Issue #13, opened 2016-01-22
Reorganize project infrastructure


ebkalderon:
There are some major changes coming to our core infrastructure. Currently, the overall Amethyst project consists of three repositories: amethyst, amethyst_cli, and amethyst_opengl.

Amethyst is set for future expansion, and the current approach to organizing our code is not maintainable. I propose the following:

- [x] Consolidate the entire Amethyst engine into the main "amethyst" repository. All future engine code and text for the Amethyst Book will reside **only** in this repository.
  - [x] Split the Amethyst engine into separate sub-crates.
     - [x] Engine
     - [x] Renderer
  - [x] Have the "amethyst" crate re-export these crates as one API.

This might cause some breakage, but this must be minimized as much as possible, especially in other branches like [ecs](https://github.com/ebkalderon/amethyst/tree/ecs) and [gpures](https://github.com/ebkalderon/amethyst/tree/gpures). Let's use [Piston's approach for managing multiple crates under one repository](https://github.com/PistonDevelopers/piston) as prior art.

Progress has begun on the [reorg](https://github.com/ebkalderon/amethyst/tree/reorg) branch.

This is the engine-side counterpart of [Amethyst Tools issue #4](https://github.com/ebkalderon/amethyst_tools/issues/4).

ebkalderon:
It seems that splitting the engine into separate sub-crates causes issues with unit testing (see the annotation in pull request #15). If anyone has better ideas on the reorganization process, please discuss them here.

kvark:
We don't have this problem in [GFX](https://github.com/gfx-rs/gfx). Perhaps, you could follow the same structure?

We have most of the crates inside the same repository, and they are in sub-folders (`src/core`, `src/render`, `src/backend/gl`, etc). Each has its own tests. Then there is a "meta" crate, which is just an empty `src/lib.rs`, it depends on all the sub-crates via [Cargo.toml](https://github.com/gfx-rs/gfx/blob/master/Cargo.toml#L17). Thus, calling `cargo build` builds everything, calling `cargo test` tests everything. This meta crate doesn't need to be uploaded to `crates.io`, it's just for developer convenience.

ebkalderon:
@kvark After playing around with my local copy of the repository, I can conclude that while `cargo build` does indeed build everything and `cargo clean` does clean everything, this behavior does not appear to extend to `cargo test`. Can you confirm?

ebkalderon:
[...]

PR #15, opened 2016-01-24
Unit tests rearrangement

Oflor:
Reorganisation of Amethyst into several crates will require 'cargo test' command to be run for each sub-crate separately (e.g. 'cargo test -p amethyst_engine'). I couldn't find a way to make cargo run tests for each of Amethyst's subcrates via one command other than change them into integration tests, which might not be the best way.

The most notable disadvantage of that way is that tests will have to follow Rust's usual rules of visibility. Tests for private structs, fields and methods will have to be a submodule. There are no such tests for now though.


ebkalderon:
Thanks for the pull request! Though I am certain that the engine and tools should each be consolidated into their own repositories, the rest is in flux. If you can think of better ways to organize the crates before 0.2.0 is released, please discuss it in issue #13.

PR #25, opened 2016-02-23
Change error handling in Simulation, SimBuilder

White-Oak:
Changed error handling to more 'Rusty' way with Results:
- Introduced `pub type ProccessorResult = Result<(), String>;` instead of ProcessorError
- SimBuilder collects all errors, while building
- Simbuilder's `done` returns `Result<Simulation, Vec<String>`, where
  - `Ok` is returned if no errors met
  - `Err` is returned if at least one error happened

Should be noted (and is noted in processor.rs), that original note for ProccessorError specified it was an error type reported by a **processor**, although, currently, Processor doesn't have any way to return an error, thus, only SimBuilder can return Result as of now.

### DIFF
--- a/src/ecs/examples/proposed_usage.rs
+++ b/src/ecs/examples/proposed_usage.rs
@@ -24,9 +24,13 @@ struct Position {
 fn main () {
     let mut world = ecs::World::new();
 
-    let (mut sim, _errs) = ecs::Simulation::build()
+    let sim_builder_result = ecs::Simulation::build()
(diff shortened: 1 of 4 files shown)


ebkalderon:
@White-Oak This is great. Thanks for the pull request!

I am painfully aware that `Processor` has no mechanism in place to return errors anyway, and I'm working on that. But this is a full step forward.

I'll merge this one.

PR #26, opened 2016-03-11
Move Blog to GitHub-pages and cobalt.rs

LucioFranco:
This PR adds support for blogging through [cobalt.rs](https://github.com/cobalt-org/cobalt.rs). This is still very early. It is using bootstrap 3.0. I am not the best designer but I will try and work some more on the styling, nothing here is done. :smile:  

The This Week in Amethyst 6 is currently completed in the markdown format.

To run:
`./build_web.sh`

Then you just need to serve the web directory. To do this you will need an HTTP server to run the site locally. I personally use [http-server](https://github.com/indexzero/http-server) to host files from any directory.

I am open to any suggestions!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 # Unwanted files
 book/html
-blog/build
 web
 prototypes
 target
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 # Unwanted files
 book/html
+blog/build
+web
 prototypes
 target
 Cargo.lock
--- a/blog/_layouts/default.tpl
+++ b/blog/_layouts/default.tpl
@@ -1,0 +1,66 @@
+<!DOCTYPE html>
+<html>
+    <head>
+	<meta charset="utf-8">
+    	<meta http-equiv="X-UA-Compatible" content="IE=edge">
+    	<meta name="viewport" content="width=device-width, initial-scale=1">
+        <title>Amethyst</title>
+
+	<!-- Latest compiled and minified CSS -->
+     	<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css" integrity="sha384-1q8mTJOASx8j1Au+a5WDVnPi2lkFfwwEAa8hDDdjZlpLegxhjVME1fgjWPGmkzs7" crossorigin="anonymous">
+
+      	<!-- Optional theme -->
+     	<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap-theme.min.css" integrity="sha384-fLW2N01lMqjakBkx3l/M9EahuwpSfeNvV63J5ezn3uZzapT0u7EYsXMjQV+0En5r" crossorigin="anonymous">
+      <link rel="stylesheet" href="/cover.css" />
+      <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js"></script>
+      	<!-- Latest compiled and minified JavaScript -->
+     	<script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/js/bootstrap.min.js" integrity="sha384-0mSbJDEHialfmuBBQP6A4Qrprq5OVfW37PRR3j5ELqxss1yVqOtnepnHVP9aJ7xS" crossorigin="anonymous"></script>
+    </head>
+    <body>
+
+    <div class="site-wrapper">
+
+      <div class="site-wrapper-inner">
+
+        <div class="cover-container">
+
+          <div class="masthead clearfix">
+            <div class="inner">
+              <h3 class="masthead-brand">Amethyst</h3>
+              <nav>
+                <ul class="nav masthead-nav">
+                  <li class="active"><a href="#">Home</a></li>
+                  <li><a href="/book">Book</a></li>
+                  <li><a href="/doc/amethyst">Doc</a></li>
+                  <li><a href="https://github.com/ebkalderon/amethyst">Github</a></li>
+                </ul>
+              </nav>
+            </div>
+          </div>
+
+          <div class="inner cover">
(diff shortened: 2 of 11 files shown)


ebkalderon:
@LucioFranco I'll be trying to build it on my local machine. I want to see how it looks like. :smile:

ebkalderon:
The website builds for me! I know you said someone should help improve the styling, but could you confirm that this is what you see?

screenshot from 2016-03-11 02-40-23

LucioFranco:
@ebkalderon I did a quick fix for the CSS. That is not what it is supposed to look like. It should work now but the links to the other parts of the site do not work yet. I would recommend using an HTTP server locally to view the site instead of opening the file from finder/windows explorer.

ebkalderon:
@LucioFranco Thanks, will do.

ebkalderon:
I think I've uncovered a major bug in markdown.rs while trying to remake the *TWIA* articles in Markdown. :confused: See johannhof/markdown.rs#13.

ebkalderon:
@LucioFranco The more I play around with Cobalt, the more I prefer keeping *TWIA* on Wordpress for now. Cobalt lacks many basic blogging features like comments, surveys, archiving, site search, etc. Its stability also leaves something to be desired, haha.

I am still intent on setting up the website, book, and docs on Cobalt, but I think I'll stick with Wordpress for *TWIA* unless Cobalt improves its blogging functionality significantly. Perhaps I could point the static website to "www.amethyst.rs" and point the "blog.amethyst.rs" subdomain to Wordpress. Does this seem like a reasonable course of action?

LucioFranco:
@ebkalderon I understand. One thing I will say is that I believe that you do not need comments or even search on a simple blog like TWIA. Both This Week in Servo and This Week in Rust both run off of statically generated websites similar to `cobalt.rs`. I believe there are many benefits to using one of these. One it makes the project more open source because every part of the project is available to see. It also allows people to submit spelling fixes or anything via a PR. Second something like cobalt takes away the need for databases. This means that it is 100% free, whereas wordpress.com charges you to even connect a domain. As for the comments I did some research and there is [disqus](https://disqus.com/) which can be embedded on static sites.

For `cobalt.rs` to me it seems like it supports all the features that you use on WordPress, minus all the nice javascript stuff. I ran into a few issues when I copied directly from the web page but besides that everything worked for me. I will look more into the issue you opened on `markdown.rs`. There are other options that are way more mature. This Week in Rust uses `Pelican` which is python but is way more mature than `cobalt.rs`. This Week in Servo uses `Jekyll` which is the blog engine that gh-pages uses. This means that there is no need to compile the site on Travis you would just upload the source and gh-pages does the rest. Jekyll is very mature and the go to static blogging platform but it is Ruby and I did not want to bring in another language. These are too very mature options we could look at.

If you are still keen on keeping it on WordPress you would just point blog.amethyst.rs to the word press and the *.amethyst.rs to GitHub. This could work and we could just have a very simple home page with basic information about amethyst and links to the docs, book, and blog.

ebkalderon:
@LucioFranco Thanks for the response! I agree that not everything Wordpress offers is necessary for *TWIA*. The only two features I'm really interested in are the site search and public surveys. Perhaps we can leave out the commenting system if we direct people to Gitter instead. I also agree that it would be great to have the blog hosted in the open on GitHub, hence my enthusiasm for the idea. :smile:

Unfortunately, `markdown.rs` simply isn't stable enough to parse all of the blog posts I've rewritten in Markdown. From here, we can either:

1. Craft our Markdown files very, very carefully so the parser doesn't panic.
2. Keep *TWIA* on Wordpress for now until the issues get fixed and make only the static pages on `cobalt.rs`
3. Look elsewhere for another static site generator.

I'll mess around with the Markdown files some more and let you know if I get it working. Should nothing come of it, we may have to consider options 2 or 3. If you have other ideas, I'd be glad to hear them.

ebkalderon:
@LucioFranco I've reposted the panicking issue on the `cobalt.rs` issue tracker as well (https://github.com/cobalt-org/cobalt.rs/issues/58), in case that the bug might be in the site generator itself rather than in the Markdown parser.

LucioFranco:
@ebkalderon Sorry for the late response. I think we should go forth with both option 1 and 2. What I mean by this is that is continue to figure out how to get the blog working with markdown.rs or [pulldown-cmark](https://github.com/google/pulldown-cmark). We could do this in another branch/PR. We should then also  merge the basic cobalt.rs site with the links the book/docs/homepage into master and setup the .rs domain to point to everything. There are some more styling issues that need to be cleaned up but it could be ready very soon. We can then have the blog at blog.amethyst.rs point to the wordpress then once cobalt is up we can move that to amethyst.rs/blog and have blog.amethyst.rs point to that. To me that seems like the best road ahead for this.

ebkalderon:
@LucioFranco Sounds good to me. I was thinking about moving the site content to its own separate repository, since we are preparing to collect everything under a single "amethyst" organization (see the definition of [organization pages](https://help.github.com/articles/user-organization-and-project-pages/) here). That way we can avoid clogging up the engine issue tracker with website-related issues, and we can more easily get pull requests adding content to the blog much like *This Week in Rust*. What do you think?

LucioFranco:
@ebkalderon Sounds good to me. I think having PR's for spell checking is going to be awesome! When are you planning on moving everything over?

ebkalderon:
As soon as possible! I'm aiming for possibly today or tomorrow. I just need to make sure that the points on #27 are put to rest before we move on.

LucioFranco:
@ebkalderon okay I will take a look at that, I have sometime tonight if you need help moving the cobalt stuff over. I can submit a PR on the web/blog repo.

ebkalderon:
@LucioFranco Beautiful! Thanks.

ebkalderon:
The bulk of the changes in this pull request has been moved to a separate [website repository](https://github.com/amethyst/amethyst.github.io) as part of the ongoing transition to a GitHub organization (issue #27), so I'm closing this. Please check there for the latest changes. Thanks for the help in getting this kick-started, @LucioFranco!

PR #45, opened 2016-05-04
[WIP] Gfx Renderer

csherratt:
Basically from a high level command are abstracted away from the data in the rendering pipeline. Commands are matched to a Method that can work with it's data. Right now that is just based on the render target and the Command itself. I think this can expand in the future.

Users can add their own targets and their own methods to the Renderer, and in addition replace methods. This replacing could be neat for doing something like a shader toy.

Right now I have a basic flat shading pipeline, and a terribly unoptimized deferred lighting. But it should be easy to to keep adding new passes and build a library of useful ones.

Demo of it working: https://www.youtube.com/watch?v=m3Q_3behpPA

Problem areas:
 - No texture support yet
 - The mesh format is static, we will need something like `gfx_mesh` to allow for multiple formats without static types
 - Resizing of buffers is incomplete
 - Single threaded


src/renderer/examples/main.rs
```diff
@@ -0,0 +1,234 @@
...
+        .with_title("Amethyst Renderer Demo".to_string())
+        .with_dimensions(800, 600)
+        .with_vsync();
+
+    let (window, mut device, mut factory, main_color, main_depth) =
+        gfx_window_glutin::init::<ColorFormat, DepthFormat>(builder);
+    let combuf = factory.create_command_buffer();
+
+    let sphere = build_sphere();
+    let buffer = factory.create_vertex_buffer(&sphere);
```
kvark: you can use `let (buffer, slice) = factory.create_vertex_buffer_with_slice(&sphere, ());`

src/renderer/examples/main.rs
```diff
@@ -0,0 +1,234 @@
...
+        scenes: std::collections::HashMap::new(),
+        cameras: std::collections::HashMap::new()
+    };
+
+    frame.scenes.insert(format!("main"), scene);
+    frame.passes.push(amethyst_renderer::RenderPasses::new(format!("gbuffer")));
+    frame.passes[0].operations = vec![
+        Box::new(amethyst_renderer::Clear{color: [0., 0., 0., 1.]}),
+        Box::new(amethyst_renderer::Draw{
+            camera: format!("main"),
```
kvark: There is too much `format!` noise. I understand that we need to receive `String` there, but maybe we can help it by adding some constructors like `amethyst_render::Draw::new("main", "main")`?

src/renderer/examples/main.rs
```diff
@@ -0,0 +1,234 @@
...
+    let mut frame = amethyst_renderer::Frame{
+        passes: vec![],
+        targets: HashMap::new(),
+        scenes: std::collections::HashMap::new(),
+        cameras: std::collections::HashMap::new()
+    };
+
+    frame.scenes.insert(format!("main"), scene);
+    frame.passes.push(amethyst_renderer::RenderPasses::new(format!("gbuffer")));
+    frame.passes[0].operations = vec![
```
kvark: I'd much prefer the passes to be constructed in advance, and then you'd have `passes: vec![pass1, pass2]` in the `frame` construction code. Modifying `passes[i]` directly can be error-prone during refactoring.

src/renderer/examples/main.rs
```diff
@@ -0,0 +1,234 @@
...
+        }),
+    ];
+    frame.targets.insert(
+        format!("main"),
+        Box::new(amethyst_renderer::ScreenOutput{
+            output: main_color,
+            output_depth: main_depth
+        }
+    ));
+    frame.targets.insert(
```
kvark: we could have something like this as a helper:
```rust
frame.add_target("gbuffer", amethyst_renderer::GBuffer::new(..));
```

src/renderer/examples/main.rs
```diff
@@ -0,0 +1,234 @@
...
+        format!("gbuffer"),
+        Box::new(amethyst_renderer::GBuffer::new(&mut factory, (800, 600)))
```
kvark: should probably extract those dimensions from the `main_color`

src/renderer/examples/main.rs
```diff
@@ -0,0 +1,234 @@
...
+    frame.scenes.insert(format!("main"), scene);
+    frame.passes.push(amethyst_renderer::RenderPasses::new(format!("gbuffer")));
+    frame.passes[0].operations = vec![
+        Box::new(amethyst_renderer::Clear{color: [0., 0., 0., 1.]}),
+        Box::new(amethyst_renderer::Draw{
+            camera: format!("main"),
+            scene: format!("main")
+        })
+    ];
+    frame.passes.push(amethyst_renderer::RenderPasses::new(format!("main")));
```
kvark: the default passes could be something simpler for this example since you already have this code bound on `Key1`

csherratt: Yeah, need to pull out the description of the passes into a function.

src/renderer/src/lib.rs
```diff
@@ -1,16 +1,204 @@
...
+use mopa::Any;
+
+pub use gbuffer::{GBuffer, Draw, BlitAmbiant, Lighting};
+
+pub type ColorFormat = gfx::format::Rgba8;
+pub type DepthFormat = gfx::format::DepthStencil;
+
+pub struct Renderer<R: gfx::Resources, C: gfx::CommandBuffer<R>> {
+    command_buffer: gfx::Encoder<R, C>,
+    methods: HashMap<(TypeId, TypeId), Box<Fn(&Box<Operation>, &Target, &Frame<R>, &mut gfx::Encoder<R, C>)>>
```
kvark: uh oh, ok
this is apparently what forces you to do all these `R::* : Any` constraints when implementing `Method`

csherratt: I think I can add a magic trait that will make this prettier, one of the todos.

src/renderer/src/method/mod.rs
```diff
@@ -0,0 +1,14 @@
+pub mod forward;
+pub mod deferred;
+
+use gfx;
+
+/// A `Method` is an implemnatnion of a Pass
+pub trait Method<A, T, R, C>
```
kvark: Shouldn't `Pass` and `Framebuffer` be associated types? This would reduce the mental overhead of reading the trait and improve clarity. Unless there are cases where the same struct implements methods for different passes and framebuffers.
Nitpic: `framebuffer` is a very GL-specific name. I'd prefer `Target`, which happens to match `T`.

src/renderer/examples/main.rs
```diff
@@ -0,0 +1,211 @@
...
+        let proj = cgmath::perspective(cgmath::deg(60.0f32), w / h, 1.0, 100.0);
+        frame.cameras.insert(
+            format!("main"),
+            amethyst_renderer::Camera{projection: proj.into(), view: view.mat.into()}
+        );
+
+        renderer.submit(&frame, &mut device);
+        window.swap_buffers().unwrap();
+    }
+}
```
kvark: you don't seem to have newlines at the end of the files

src/renderer/src/lib.rs
```diff
@@ -1,16 +1,151 @@
...
+use std::any::TypeId;
+use std::collections::HashMap;
+
+pub use pass::Pass;
+pub use framebuffer::Framebuffer;
+pub use method::Method;
+
+pub struct Renderer<R: gfx::Resources, C: gfx::CommandBuffer<R>> {
+    command_buffer: gfx::Encoder<R, C>,
+    methods: HashMap<(TypeId, TypeId), Box<Fn(&Box<Pass>, &Framebuffer, &Frame<R>, &mut gfx::Encoder<R, C>)>>
```
kvark: I wonder if making a trait instead of this function would allow you to remove the `Pass` boxing, thus stripping the need to derive from `Any` for passes.

kvark:
Reminder: this PR is not to be merged, since it is targeting the wrong branch.

PR #48, opened 2016-05-04
specs::System as Processor

lschmierer:


lschmierer:
The context is used to pass elapsed time to the processors.

src/ecs/src/lib.rs
```diff
@@ -1,9 +1,10 @@
...
 extern crate specs;
 
-mod processor;
 mod sim;
 
-pub use specs::*;
-pub use self::processor::{Processor, ProcessorResult};
+use time::Duration;
+
+pub use specs::{World, Component, System as Processor, RunArg, Entity, EntityBuilder, Entities,
```
kvark: I wonder if it should just re-export `specs` directly instead.

src/ecs/src/sim.rs
```diff
@@ -39,44 +38,35 @@ impl Simulation {
 
     /// Computes the next state of the world using the given processors.
     pub fn step(&mut self, dt: Duration) {
-        for p in self.procs.iter_mut() {
-            p.process(&mut self.planner, dt);
-        }
+        self.planner.dispatch(dt);
```
kvark: I think `dispatch` calls the wait for you

src/ecs/src/lib.rs
csherratt: I have been thinking about doing this in a few places. The main reason being that users don't need to specify a version of your dependent library, which means less versioning problems when you upgrade a dependency.

kvark: if you just do `pub use specs;`, the users would neither need to specify a version

kvark:
Thanks!
I think this is a good step forward, although not sure if we should really expose all the stuff inside `specs` instead of just exposing `specs` itself directly.

PR #49, opened 2016-05-05
GFX based Renderer

csherratt:
Basically from a high level command are abstracted away from the data in the rendering pipeline. Commands are matched to a Method that can work with it's data. Right now that is just based on the render target and the Command itself. I think this can expand in the future.

Users can add their own targets and their own methods to the Renderer, and in addition replace methods. This replacing could be neat for doing something like a shader toy.

Right now I have a basic flat shading pipeline, and a terribly unoptimized deferred lighting. But it should be easy to to keep adding new passes and build a library of useful ones.

Demo of it working: https://www.youtube.com/watch?v=m3Q_3behpPA

Problem areas:
 - No texture support yet
 - The mesh format is static
 - Single threaded


src/renderer/Cargo.toml
```diff
@@ -13,4 +13,14 @@ readme = "README.md"
...
+gfx_device_gl = "0.10"
+gfx_window_glutin = "0.11"
+glutin = "0.5"
+cgmath="0.7"
+mopa ="0.2"
+
+
+[dev-dependencies]
+genmesh="0.4"
+rand="*"
```
kvark: new line?

src/renderer/examples/main.rs
```diff
@@ -0,0 +1,226 @@
...
+
+
+    let mut renderer = amethyst_renderer::Renderer::new(combuf);
+    renderer.load_all(&mut factory);
+
+    window.set_title("Amethyst Renderer [Deferred]");
+    frame.layers = pipeline_deferred();
+
+    let start = SystemTime::now();
+    let (mut w, mut h) = (800., 600.);
```
kvark: need to get this from the Glutin window

csherratt:
Cleaned up! Unless there are new problems I'm done for now.

kvark:
Thanks! :shipit: :rocket:

palodequeso:
Are you planning on putting your example code into the examples?

PR #51, opened 2016-05-10
Improve texturing & foward pipeline

csherratt:
This adds a forward lighting pass
This also adds the ability to texture things.

### DIFF
--- a/src/renderer/src/lib.rs
+++ b/src/renderer/src/lib.rs
@@ -101,7 +101,7 @@ impl<R, C> Renderer<R, C>
     }
 }
 
-/// holds a 1x1 texture that can be used to store constnat colours
+/// holds a 1x1 texture that can be used to store constant colors
 pub struct ConstantColorTexture<R: gfx::Resources> {
     texture: gfx::handle::Texture<R, gfx::format::R8_G8_B8_A8>,
     view: gfx::handle::ShaderResourceView<R, [f32; 4]>
@@ -113,7 +113,15 @@ impl<R: gfx::Resources> ConstantColorTexture<R> {
         where F: gfx::Factory<R>
     {
         let kind = gfx::tex::Kind::D2(1, 1, gfx::tex::AaMode::Single);
-        let (text, view) = factory.create_texture_const_u8::<gfx::format::Rgba8>(kind, &[&[0, 0, 0, 0]]).unwrap();
+        let text = factory.create_texture::<gfx::format::R8_G8_B8_A8>(
+            kind,
+            1,
+            gfx::SHADER_RESOURCE,
+            gfx::Usage::Dynamic,
+            Some(gfx::format::ChannelType::Unorm)
+        ).unwrap();
+        let levels = (0, text.get_info().levels - 1);
+        let view = factory.view_texture_as_shader_resource::<gfx::format::Rgba8>(&text, levels, gfx::format::Swizzle::new()).unwrap();
         ConstantColorTexture{
             texture: text,
             view: view
@@ -141,11 +149,7 @@ impl<R: gfx::Resources> Texture<R> {
                 encoder.update_texture::<_, gfx::format::Rgba8>(
                     &texture.texture,
                     None,
-                    gfx::tex::ImageInfoCommon{
-                        xoffset: 0, yoffset: 0, zoffset:0,
-                        width: 1, height: 1, depth: 0,
-                        format: (), mipmap: 0
-                    },
+                    texture.texture.get_info().to_image_info(0),
                     &color[..]
                 ).unwrap();
                 texture.view.clone()
--- a/src/renderer/src/lib.rs
+++ b/src/renderer/src/lib.rs
@@ -57,7 +57,7 @@ impl<R, C> Renderer<R, C>
         where F: gfx::Factory<R>
     {
         self.add_pass(pass::forward::Clear);
-        self.add_pass(pass::forward::DrawNoShading::new(factory));
+        self.add_pass(pass::forward::DrawFlat::new(factory));
         self.add_pass(pass::forward::DrawShaded::new(factory));
         self.add_pass(pass::forward::Wireframe::new(factory));
 
--- a/src/renderer/src/lib.rs
+++ b/src/renderer/src/lib.rs
@@ -37,6 +37,7 @@ pub struct Renderer<R: gfx::Resources, C: gfx::CommandBuffer<R>> {
 gfx_vertex_struct!( VertexPosNormal {
     pos: [f32; 3] = "a_Pos",
     normal: [f32; 3] = "a_Normal",
+    tex_coord: [f32; 2] = "a_TexCoord",
 });
 
 impl<R, C> Renderer<R, C>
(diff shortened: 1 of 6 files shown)


src/renderer/src/pass/forward.rs
```diff
@@ -14,8 +14,14 @@ pub static VERTEX_SRC: &'static [u8] = b"
...
     in vec3 a_Pos;
+    in vec3 a_Normal;
+
+    out vec4 v_Position;
+    out vec3 v_Normal;
 
     void main() {
+        v_Position = u_Model * vec4(a_Pos, 1.0);
+        v_Normal = mat3(u_Model) * a_Normal;
         gl_Position = u_Proj * u_View * u_Model * vec4(a_Pos, 1.0);
```
kvark: could use `v_Position` here

src/renderer/src/pass/forward.rs
```diff
@@ -25,10 +31,47 @@ pub static FLAT_FRAGMENT_SRC: &'static [u8] = b"
...
+    in vec3 v_Normal;
+    out vec4 o_Color;
+
+    void main() {
+        vec4 color = u_Ka;
+        for (int i = 0; i < u_LightCount; i++) {
+            vec4 delta = light[i].center - v_Position;
+            vec4 light_to_point_normal = normalize(delta);
+
+            float dist = length(delta);
```
kvark: could compute `light_to_point_normal` via `dist` to avoid normalization

src/renderer/src/pass/forward.rs
```diff
@@ -125,13 +168,110 @@ impl<R> Pass<R> for DrawNoShading<R>
...
+        center: [f32; 4] = "center",
+        color: [f32; 4] = "color",
+    }
+
+    pipeline shaded {
+        vbuf: gfx::VertexBuffer<VertexPosNormal> = (),
+        ka: gfx::Global<[f32; 4]> = "u_Ka",
+        kd: gfx::Global<[f32; 4]> = "u_Kd",
+        lights: gfx::ConstantBuffer<PointLight> = "u_Lights",
+        light_count: gfx::Global<i32> = "u_LightCount",
```
kvark: I think this should be refactored to avoid globals. It's not like they are here for GL2 compatibility, since you are using a constant buffer for the lights anyway.

src/renderer/src/pass/forward.rs
```diff
@@ -125,13 +168,110 @@ impl<R> Pass<R> for DrawNoShading<R>
...
+        model: gfx::Global<[[f32; 4]; 4]> = "u_Model",
+        view: gfx::Global<[[f32; 4]; 4]> = "u_View",
+        proj: gfx::Global<[[f32; 4]; 4]> = "u_Proj",
+        out_ka: gfx::RenderTarget<gfx::format::Rgba8> = "o_Color",
+        out_depth: gfx::DepthTarget<gfx::format::DepthStencil> = gfx::preset::depth::LESS_EQUAL_WRITE,
+    }
+);
+
+pub struct DrawShaded<R: gfx::Resources>{
```
kvark: I think `DrawNoShading` should be renamed to `DrawFlat` for consistency

src/renderer/src/lib.rs
```diff
@@ -99,6 +101,60 @@ impl<R, C> Renderer<R, C>
     }
 }
 
+/// holds a 1x1 texture that can be used to store constnat colours
```
kvark: "constnat colours" -> "constant color"

src/renderer/src/lib.rs
```diff
@@ -99,6 +101,60 @@ impl<R, C> Renderer<R, C>
...
+    view: gfx::handle::ShaderResourceView<R, [f32; 4]>
+}
+
+impl<R: gfx::Resources> ConstantColorTexture<R> {
+    /// Create a texture buffer
+    pub fn new<F>(factory: &mut F) -> ConstantColorTexture<R>
+        where F: gfx::Factory<R>
+    {
+        let kind = gfx::tex::Kind::D2(1, 1, gfx::tex::AaMode::Single);
+        let (text, view) = factory.create_texture_const_u8::<gfx::format::Rgba8>(kind, &[&[0, 0, 0, 0]]).unwrap();
```
kvark: You can't use `create_texture_const` here, it's only for immutable textures. You'd need to explicitly say that this is a dynamic texture by creating the texture object manually with `create_texture` and passing `Usage::Dynamic` in there.

src/renderer/src/lib.rs
```diff
@@ -99,6 +101,60 @@ impl<R, C> Renderer<R, C>
...
+                let color: [[u8; 4]; 1] = [[
+                    (color[0] * 255.) as u8,
+                    (color[1] * 255.) as u8,
+                    (color[2] * 255.) as u8,
+                    (color[3] * 255.) as u8,
+                ]];
+                encoder.update_texture::<_, gfx::format::Rgba8>(
+                    &texture.texture,
+                    None,
+                    gfx::tex::ImageInfoCommon{
```
kvark: FYI, could use `texture.get_descriptor().to_image_info(0)`

src/renderer/src/pass/deferred.rs
```diff
@@ -32,16 +32,21 @@ impl<R> ::Pass<R> for Clear
 pub static DRAW_VERTEX_SRC: &'static [u8] = b"
     #version 150 core
 
-    uniform mat4 u_Proj;
-    uniform mat4 u_View;
-    uniform mat4 u_Model;
+    layout (std140) uniform u_VertexArgs {
```
kvark: Nothing wrong with this, but we need to discuss the naming of stuff at some point. With DX11 on board, it kinda went off the original plan. For the constant buffers, we may consider `cb_Something` scheme.

src/renderer/src/pass/deferred.rs
```diff
@@ -50,83 +55,210 @@ pub static DRAW_VERTEX_SRC: &'static [u8] = b"
...
     }
 ";
 
+pub static LIGHT_FRAGMENT_SRC: &'static [u8] = b"
+    #version 150 core
+    #define MAX_NUM_TOTAL_LIGHTS 128
+
+    layout (std140) uniform u_FragmentLightArgs {
+        mat4 u_Proj;
+        mat4 u_InvProj;
```
kvark: would be simpler to combine into `u_InvViewProj`

src/renderer/src/pass/forward.rs
```diff
@@ -69,72 +170,213 @@ impl<R> Pass<R> for Clear
...
+        let lights = factory.create_constant_buffer(512);
+        let vertex = factory.create_constant_buffer(1);
+        let fragment = factory.create_constant_buffer(1);
+        let pso = factory.create_pipeline_simple(
+            VERTEX_SRC,
+            FRAGMENT_SRC,
+            shaded::new()
+        ).unwrap();
+
+        let sampler = factory.create_sampler(
```
kvark: seeing this repeated a lot in our code (not just in Amethyst), we should make this more ergonomic

src/renderer/src/lib.rs
csherratt: oops, good catch.

kvark:
Awesome stuff!

PR #58, opened 2016-05-12
Fix forward renderer on more strict graphics drivers such as Intel Mesa

palodequeso:
I just chose some names for the in and out structures, but if we want to change them, or develop a standard, we should do it now.

### DIFF
--- a/src/renderer/src/pass/forward.rs
+++ b/src/renderer/src/pass/forward.rs
@@ -20,16 +20,16 @@ pub static VERTEX_SRC: &'static [u8] = b"
     in vec2 a_TexCoord;
 
     out VertexData {
-        vec4 v_Position;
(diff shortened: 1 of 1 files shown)


kvark:
I don't think it makes sense to keep `v_` prefixes for stuff inside your interface structs.

palodequeso:
> I don't think it makes sense to keep v_ prefixes for stuff inside your interface structs.

I struggled with that, and landed on shooting in the dark. but we can't just use ```in``` and ```out``` because those are reserved... suggestions?

kvark:
No, I mean for stuff **inside** the structs, like `v_In.position` instead of `v_In.v_Position`.

palodequeso:
That should do it.

PR #59, opened 2016-05-12
Added Renderer into Travis script

kvark:


csherratt:
Not sure why this doesn't work. I think it might be a cargo build. An easy work around might be just doing the cargo test from the `src/amethyst_renderer` folder.

csherratt:
https://github.com/rust-lang/cargo/issues/860

PR #61, opened 2016-05-16
Configuration and loading from yaml files

Aceeri:
Allows using macros to create a structure along with loading a yaml file into that struct.

It isn't entirely finished as I'd like to add a couple more loading types and clean it up a bit, but it would be very helpful if anyone can spot something I missed or give suggestions.

- [x] Basic enum loading? Complex enums would take a rather large feat of macro-ness (tokenizer hackiness?) to get working and I don't see much benefit so it might be better to only allow really basic enums e.g.
 `enum Example { Option1, Option2, Option3 }`
- [x] Hashmap loading
- [x] Vec loading
- [x] Writing to files?
- [x] Option loading? null return None, else Some(T)?

I'm going to be a bit busy until next wednesday, so I probably won't be making any changes until then.


src/config.rs
```diff
@@ -0,0 +1,313 @@
...
+}
+
+macro_rules! config {
+    ($root:ident {
+        $( $field:ident: $ty:ty = $name:expr, )*
+    }) => {
+
+        #[derive(Clone, Debug)]
+        pub struct $root {
+            _meta: ConfigMeta,
```
Aceeri: Should I move this to be completely invisible/in the background to the user or keep it exposed? It may be useful if we want to do something like writing to the configs.

src/config.rs
```diff
@@ -0,0 +1,313 @@
...
+
+                for (index, element) in meta.fields.iter().enumerate() {
+                    if index != 0 {
+                        path = path + "->";
+                    }
+
+                    path = path + element;
+                }
+
+                format!("{}: Failed to parse YAML: {}: expect {}", meta.path.display(), path, meta.ty)
```
Aceeri: Should probably re-add in a check if the key is a "BadValue", if it then say "Could not find YAML" instead of "Failed to parse YAML".

src/config.rs
```diff
@@ -0,0 +1,452 @@
...
+
+impl fmt::Display for ConfigError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.to_string())
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct ConfigMeta {
+    path: PathBuf, // Where the file is located, "" if not from a file.
```
lschmierer: Maybe use Option<PathBuf>. here ;)
A little bit cleaner.

lschmierer:
>  Writing to files? Not sure how necessary this would be.
I think existing Yaml files should be automatically updated to match the layout of the `config!{...}` defined Configs. So, when the config struct is changed, the developer does not need to take care of the Yaml config files.

src/config.rs
Aceeri: Decided to make it default to "config\\config.yml" instead as a default directory for amethyst.

Aceeri:
Looks like this is almost done, couple of rather minor bugs to fix.

lschmierer:
Does writing keep comments in the yaml files?

Aceeri:
@lschmierer Not at the moment, hadn't thought of that.

thiolliere:
I think that from_file_raw must result in an error if there is there is both display.yml and display/config.yml

thiolliere:
it can also be cool in case display.yml have a key not in its structure definition to warn unexpected key.

Aceeri:
Alright!

- Added a warning for unexpected keys in the yaml.
- `from_file()` now outputs a warning/error if there are multiple files and defaults instead of loading.
- Documentation for configuration structures and enums work properly now in the macro, along with custom derives so there is less need for comments inside of the .yml/.yaml files.

Anything else that could be a problem?

Cargo.toml
```diff
@@ -23,3 +23,9 @@ version = "0.1.0"
 [dependencies.amethyst_renderer]
 path = "src/renderer/"
 version = "0.3.0"
+
+[dependencies.amethyst_config]
+path = "src/config/"
+version = "0.1.0"
+
+[dependencies]
```
kvark: missing new line

src/config/examples/main.rs
```diff
@@ -0,0 +1,21 @@
+#[macro_use]
+extern crate amethyst_config;
+extern crate yaml_rust;
+
+use amethyst_config::Element;
+use std::path::Path;
+
+fn main() {
+    let config = amethyst_config::Config::from_file(Path::new("../../config/config.yml"));
```
kvark: If you use `AsRef<Path>`, like `File::open` [does](http://doc.rust-lang.org/std/fs/struct.File.html#method.open), you could avoid `Path::new()` here.

src/config/src/definitions.rs
```diff
@@ -0,0 +1,384 @@
...
+impl ConfigMeta {
+    /// Displays the meta's fields in order e.g. Config->nested_config->field
+    pub fn tree(&self) -> String {
+        let mut tree = "".to_string();
+
+        for (index, element) in self.fields.iter().enumerate() {
+            if index != 0 {
+                tree = tree + "->";
+            }
+
```
kvark: You do seem to like empty lines a lot :)

src/config/src/yaml.rs
```diff
@@ -0,0 +1,524 @@
...
+                match element {
+                    &Yaml::Array(_) |
+                    &Yaml::Hash(_) => {
+                        complex = true;
+                    },
+                    _ => { },
+                }
+            }
+
+            if array.len() > 10 {
```
kvark: declare `10` as some sort of a constant for clarity?

src/config/src/yaml.rs
```diff
@@ -0,0 +1,524 @@
...
+yaml_array!(1 => 0);
+yaml_array!(2 => 0 1);
+yaml_array!(3 => 0 1 2);
+yaml_array!(4 => 0 1 2 3);
+yaml_array!(5 => 0 1 2 3 4);
+yaml_array!(6 => 0 1 2 3 4 5);
+yaml_array!(7 => 0 1 2 3 4 5 6);
+yaml_array!(8 => 0 1 2 3 4 5 6 7);
+yaml_array!(9 => 0 1 2 3 4 5 6 7 8);
+yaml_array!(10 => 0 1 2 3 4 5 6 7 8 9);
```
kvark: ah, this is where `10` came from, I see

src/config/src/definitions.rs
Aceeri: Just a habit I have, I usually separate things that modify scope.

kvark:
Thanks @Aceeri !

PR #62, opened 2016-06-17
Add depth pass to renderer

### LINKED ISSUE #56: [Renderer] Add support for a Z-Prepass
A simple pass that renders the scene into a depth buffer leaving the other frame attachments unaffected. This can improve performance in some pipelines so it is worth having it as an optional pass.

nchashch:
Closes #56


kvark:
Looks good, thanks!
One note though. If one uses this depth pre-pass, shouldn't they disable the depth writes for the main drawing pass (and possibly use depth test = equal instead of less-equal)?
I suppose we can do it later.

PR #63, opened 2016-06-23
Add amethyst_context crate

### LINKED ISSUE #54: Windowing Support
This should be part of the engine. Probably just a way to configure `glutin` from a config file. This ties in the renderer and the event system.

nchashch:
Add `Context` struct which owns all global resources e.g. `Window`, `RenderingContext`, `AudioContext`, `Logger`, `AssetManager`, `NetworkManager`, etc.

`Context` is created from `Config` (it is an amethyst_config sturct which holds subconfigs for particular resources) and `Rc<RefCell<Context>>` is passed to every party requiring access to the context (see examples/window.rs).

Currently `Context` only contains a `Window` and a `RenderingContext` (struct containing device, factory, render targets and an `amethyst_renderer::Renderer`).

Potentially it can be used for event handling (passing an `Rc<RefCell<Context>>` to `Application` and polling events from it in `Applcation::advance_frame`).

It should close #54


kvark:
Looks good to me. Summoning @ebkalderon , @csherratt , @Aceeri and whoever else interested!

src/context/src/rendering_context.rs
```diff
@@ -0,0 +1,90 @@
...
+        let multisampling = display_config.multisampling;
+        let visibility = display_config.visibility;
+
+        let mut builder = WindowBuilder::new()
+            .with_title(title)
+            .with_multisampling(multisampling)
+            .with_visibility(visibility);
+
+        match display_config.min_dimensions {
+            Some ((w, h)) => builder = builder.with_dimensions(w, h),
```
palodequeso: Why not combine line 56's some block into this one?

palodequeso:
I had one small nit-pick, but I still consider myself a rust beginner, so there may be a really good reason for not doing what I suggested.

src/context/src/rendering_context.rs
nchashch: It is an error, on line 51 `display_config.min_dimensions` should be `display_config.dimensions`, thanks for pointing it out!

palodequeso: Wee, I was useful!

palodequeso:
+1

White-Oak:
Seems good, but should be kept in mind that there are other than GL contexts :)

palodequeso:
> Seems good, but should be kept in mind that there are other than GL contexts :)

Agreed.

Aceeri:
Looks good to me, the GL context problem should probably be handled separately in the `amethyst_renderer` crate with some sort of `Window` trait over glutin/glfw/sdl/dx11/metal stuff.

src/context/src/rendering_context.rs
```diff
@@ -0,0 +1,90 @@
...
+        let title = display_config.title;
+        let multisampling = display_config.multisampling;
+        let visibility = display_config.visibility;
+
+        let mut builder = WindowBuilder::new()
+            .with_title(title)
+            .with_multisampling(multisampling)
+            .with_visibility(visibility);
+
+        match display_config.dimensions {
```
Aceeri: Since the None never actually does something, I think it might be better to just make this a 
```rust
if let Some((w, h)) = display_config.dimensions {
    builder = builder.with_dimensions(w, h);
}
```

Aceeri: Actually, just remembered that `display_config.dimensions` shouldn't be an `Option`. So that only applies to the other ones.

src/context/src/rendering_context.rs
```diff
@@ -0,0 +1,90 @@
...
+use self::gfx_device_gl::{Device, Factory, Resources, CommandBuffer};
+use self::gfx::handle::{RenderTargetView, DepthStencilView};
+
+config!(
+    /// Contains display config,
+    /// it is required to create a `(Window, RenderingContext)` pair
+    struct DisplayConfig {
+        pub title: String = "Amethyst game".to_string(),
+        pub fullscreen: bool = false,
+        pub dimensions: Option<(u32, u32)> = None,
```
Aceeri: May want to change this to just `(u32, u32)` instead of an `Option`, since otherwise it would be defaulted to `None` and the window won't know what the default resolution is.

nchashch: I think it should be an `Option` for ergonomics reasons. If `display_config.dimensions` is an `Option` then it is possible to get a `Window` in fullscreen with native resolution just by not passing anything to `display_config.dimensions` field, and if it is not in fullscreen then it will just default to some arbituary dimensions.

If `display_config.dimensions` is not an `Option` then in fullscreen resolution would be set to some hard coded value that doesn't necessarily match the native resolution in case if the proper resolution is not set in the config. Which might be unexpected.

Aceeri: True, forgot that setting dimensions when in exclusive mode might create some problems.

kvark:
Looks like everyone is happy. Merging.

PR #64, opened 2016-07-09
Add support for non GL backends (amethyst_context)

nchashch:
Changes:
- Rename `RenderingContext` to `VideoContext`.
- `VideoContext` now holds the `window` field.
- Make `VideoContext` an enum, variants of which represent different backends, the backend to use is specified in `backend` field of `DisplayConfig`.
- Add a stub implementation for `VideoContext::Direct3D`.
- Do some minor fixes.


kvark:
Looks good! Is this done, or WIP?
Btw, don't forget to gate the DX stuff with `#[cfg(windows)]` or something.

nchashch:
In general it is done, but the DX implementation is just a stub (although it is gated with `#[cfg(windows)]`). I think it would be better to unstub the DX support separately.

kvark:
Thanks!

PR #65, opened 2016-07-12
Replace time with std::time

serprex:
The changes from time to std::time are straightforward. Also included is a 3rd commit reworking Stopwatch to be an enum
```
enum Stopwatch {
    Waiting,
    Started(Instant),
    Ended(Duration),
}
```
A change in stopwatch is that Started(start).elapsed() returns start.elapsed() instead of Duration::new(0, 0)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "amethyst"
-version = "0.3.0"
+version = "0.4.0"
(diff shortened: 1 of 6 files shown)


Aceeri:
Looks good, however, this should go to the `develop` branch instead of the master and probably doesn't warrant a whole new version.

serprex:
#66

PR #66, opened 2016-07-12
Replace time with std::time, refactor Stopwatch

serprex:
As requested, this is #65 applied to develop

### DIFF
--- a/src/engine/app.rs
+++ b/src/engine/app.rs
@@ -7,7 +7,6 @@ use context::Context;
 use ecs::{Planner, World, Processor, Priority};
 use std::sync::{Arc, Mutex};
 use std::ops::DerefMut;
-use std::time::Instant;
 
 /// User-friendly facade for building games. Manages main loop.
 pub struct Application {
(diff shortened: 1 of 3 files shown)


ebkalderon:
@serprex I like these changes overall, especially the enum'ed Stopwatch. However, the reason why I used the `time` crate originally was because `Duration` can only return full seconds + trailing nanoseconds. To me, this seems very inconvenient for game development since it's difficult to convert that in/out of floating-point formats, especially if time values are meant to be passed into graphical shaders which often expect floats/doubles in millisecond format.

That said, I am all for using the standard library as much as possible. I would merge this in a heartbeat if `Duration` were to gain a `fn as_millis() -> f64` method. I would like to hear other members' thoughts on this. Perhaps my concerns are overblown?

serprex:
You can always expose some basic dur_to_f32/dur_to_f64 utility functions or create a trait which when imported would add such methods to std::time::Duration

ebkalderon:
That sounds like a good solution, yeah. I suppose those could be added to the `timing` module as well.

serprex:
I'd prefer a discussion of such an API to be a separate PR. Unless the API is literally only ``as_millis``

ebkalderon:
@serprex Don't know why this PR hasn't been merged yet, it looks pretty good! Unfortunately, it looks like some merge conflicts have arisen due to the removal of the `amethyst_engine` crate. Are you still working on this?

serprex:
I'll rebase this evening

ebkalderon:
Awesome! Thank you, and I'm sorry the PR fell through. Your help is definitely appreciated.

serprex:
Done

src/engine/app.rs
```diff
@@ -67,7 +68,7 @@ impl Application {
         let fixed_step = self.context.lock().unwrap().fixed_step.clone();
         let last_fixed_update = self.context.lock().unwrap().last_fixed_update.clone();
 
-        if SteadyTime::now() - last_fixed_update > fixed_step {
+        if Instant::now() - last_fixed_update > fixed_step {
             self.states.fixed_update(self.context.lock().unwrap().deref_mut());
             self.context.lock().unwrap().last_fixed_update = last_fixed_update + fixed_step;
```
ebkalderon: Minor tweak: with the standard library's `Duration` type, I think we can actually use the `AddAssign` operator here. I don't have access to a computer at the moment, so I can't check whether this is the case.

ebkalderon:
Everything checks out! I'll merge this now. Thanks for the help and your patience, @serprex! :tada:

PR #67, opened 2016-07-16
Add support for event handling

### LINKED ISSUE #46: Event Handling
We need a way to handle events in a generic way. This also has to be expandable, meaning users can add their own events as they see fit without having to recompile the base of the engine.

The expandable requirement means we can't just have a `Vec<Event>` where `Event` is an enum. This is because the enum has to be defined somewhere in `amethyst` which means the users can't define their own types.

This will also have to work with `specs` meaning that system can write events for other systems to pick up. This needs to be sequenced correctly so that downstream systems get the all the events they are expected.

A good example is that button presses could be abstracted away from the intent. Think of it like a RTS ui, there is always shortcuts on the keyboard and GUI elements for the beginners. This also allows for mobile development, as most devices have zero buttons and it is all just figure taps.

nchashch:
Changes:
- Add `Broadcaster` struct which allows publishing specs entities. An instance of `Broadcaster` is owned by `Context`.

- Add `EngineEvent` component. At the begining of every `Application::advance_frame()` events of this type are polled from the `Context` and published using `Broadcaster`.
Note: `EngineEvent` is a wraper around `glutin::Event`. And `amethyst_context::event` module reexports all event structs and enums from glutin.

- Add `Context::poll_engine_events()` method. It polls platform specific events, translates them to `EngneEvent`s and returns them.

- Bump version of specs used by amethyst_ecs to 0.7.0

- Reexport all specs structs, enums and traits in amethyst_ecs.

- Add `VideoContext::Null` backend.

It should close #46


src/context/src/broadcaster.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// Build and publish an entity,
+    /// using `EntityBuilder` syntax
+    pub fn publish(&mut self) -> EntityBuilder {
+        self.world.create_now()
+    }
+
+    /// Return a vector containing clones of all published entities
+    pub fn poll(&self) -> Vec<Entity> {
+        let entities = self.world.entities();
+        let _entities: Vec<Entity> = entities.iter().map(|e| e.clone()).collect();
```
kvark: should probably keep that vector owned, re-use it, and return `&[Entity]` instead, to avoid re-allocations

kvark:
Looks good to me. @ebkalderon ?

ebkalderon:
Solid work, @nchashch! Thank you. I think this can be merged.

src/context/src/lib.rs
```diff
@@ -29,33 +29,56 @@ extern crate glutin;
...
+        match self.video_context {
+            VideoContext::OpenGL { ref window, .. } =>
+                for event in window.poll_events() {
+                    let event = EngineEvent::new(event);
+                    events.push(event);
+                },
+            #[cfg(windows)]
+            VideoContext::Direct3D {  } => {
+                // stub
+                let event = EngineEvent::new(Event::Closed);
```
Emilgardis: `Event` is not defined. New to amethyst, but I imagine it should be glutin::Event.

Emilgardis:
Used githubs provided web-interface for making pull-requests, bad mistake. Anyway, e04fbc6 should be fixed by either appending `glutin::` to `Event::Closed` or makeing `glutin::Event` in scope via `use glutin::Event`.

src/context/src/lib.rs
nchashch: It is guarded with #[cfg(windows)], so that is probably why it is accepted by Travis, and why I missed it (I built it only on linux).

nchashch:
It would be better to import `event::Event` instead of `glutin::Event`. Because `glutin::Event` is reexported in `amethyst_context::event` module, also `extern crate glutin;` in `amethyst_context` module is unnecessary.

PR #68, opened 2016-07-18
Patch e04fbc6

Emilgardis:
See #67. Travis did not pick this up, this PR can be rejected and instead rebased into/squashed with e04fbc6


Emilgardis:
Ok, I'm done with git for today,,,

Issue #42, opened 2016-04-02
Crate structure question


ebkalderon:
Currently, the crate dependency map of Amethyst looks like this:

```
+----------------------------------------------------+
|                      amethyst                      |
+----------------------------------------------------+
         ^                 ^             ^
         |                 |             |
+--------+--------+        |             |
| amethyst_engine |        |             |
+-----------------+        |             |
     ^      ^              |             |
     |      |              |             |
     |      +------------------------+   |
     |                     |         |   |
     |          +----------+---+ +---+---+-----------+
     +----------| amethyst_ecs | | amethyst_renderer |
                +--------------+ +-------------------+
```

Is there any reason to have the `amethyst_engine` crate at all? Can we not push its functionality upward to the `amethyst` crate? Doing so would simplify things quite a bit with no difference in functionality (see below).

```
+------------------------------------+
|              amethyst              |
+------------------------------------+
        ^                  ^
        |                  |
+-------+------+ +---------+---------+
| amethyst_ecs | | amethyst_renderer |
+--------------+ +-------------------+
```

White-Oak:
Yes

ebkalderon:
Resolved by #70.

PR #70, opened 2016-07-19
Merge amethyst_engine into root crate

ebkalderon:
There is little reason for this crate to exist, as we discussed a while back. This pull request merges it into the root `amethyst` crate, thereby resolving issue #42. On the side, it adopts the newer `pub extern crate <name> as <new-name>` convention.


ebkalderon:
@amethyst/engine-devs Any thoughts?

kvark:
thumbs up all the way!

PR #71, opened 2016-07-19
Updates to book

Emilgardis:
Some changes to the book. 
Updated the way of getting the git version since the way that was described is unnecessary if the user doesn't want to make changes to amethyst.

Also note:

    amethyst = {git = "https://github.com/amethyst/amethyst.git", branch = "develop" } 

This will make cargo use the develop branch.

Changed method of switching between toolchains to rustup.

### DIFF
--- a/book/src/getting_started.md
+++ b/book/src/getting_started.md
@@ -19,8 +19,11 @@ Rust compiler. Here are the system requirements (they're pretty modest):
 ## Setting Up
 
 > Note: This guide assumes you have nightly Rust and Cargo installed, and also
-> have a working Internet connection. Please take care of these prerequisites
-> first before proceeding.
+> have a working Internet connection. 
+> Please take care of these prerequisites first before proceeding.
+> See [rustup][ru] for an easy way to have two installations of rust.
+
+[ru]: https://rustup.rs
 
 There are two ways to get started working with Amethyst:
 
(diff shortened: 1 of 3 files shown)


Emilgardis:
Also I noted that on the first page there are references like . 

> I've wanted to play around [...]
> Most of those I did find stuck [...]
> In short, I am writing Amethyst to scratch [...]

These should probably be taken out as this is not a single-man project anymore. However, it could be nice to somehow keep the _charm_ it gives to readers.

book/src/getting_started.md
```diff
@@ -19,8 +19,11 @@ Rust compiler. Here are the system requirements (they're pretty modest):
 ## Setting Up
 
 > Note: This guide assumes you have nightly Rust and Cargo installed, and also
-> have a working Internet connection. Please take care of these prerequisites
-> first before proceeding.
+> have a working Internet connection. 
+> Please take care of these prerequisites first before proceeding.
+> See [rustup][ru] for an easy way to have two installations of rust.
```
Emilgardis: Now when I read this out loud, it doesn't sound that good, any suggestions for an edit?

LucioFranco:
👍 this seems like well-needed update to the book.

Emilgardis:
Took the liberty of amending and then force pushing the previous commit, I had missed one link.

LucioFranco:
LGTM @ebkalderon what do you think?

ebkalderon:
Looks good to me! Thanks for the updates, @Emilgardis. I think we can merge this.

PR #72, opened 2016-07-19
Make Renderer owned by VideoContext::OpenGL usable

nchashch:
Changes:
- Now `VideoContext` owns `frame` instead of `main_color` and `main_depth`
- Call `Renderer::load_all()` in `new_gl()`
- Insert `ColorBuffer` target into `video_context.frame.targets` in `new_gl()`
- Clear the screen in `examples/window.rs` example
- Refactor event handling code in `Application::advance_frame()`


kvark:
Looks great, thanks!
Given that it's not going to disrupt anyone's workflow, I'm going ahead with merge.

PR #78, opened 2016-07-20
Pass &mut Context and &mut World to State methods, run Processors in Application::advance_frame()

nchashch:
Changes:
- Pass `&mut Context` to `State` methods, so `State`s no longer need to own `Rc<RefCell<Context>>`
- Move `timing` module from `engine` into `amethyst_context`
- Move `delta_time`, `fixed_step`, `last_fixed_update` fields from `Application` into `Context`
- Don't pass `_delta` to `State` methods
- Store a single `Planner` in `StateMachine`
- Pass `&mut World` to `State` methods
- Add `Applicatoin::build()` function
- Run `Processor`s in `Application::advance_frame()`


src/engine/app.rs
```diff
@@ -1,38 +1,41 @@
...
-            last_fixed_update: SteadyTime::now(),
-            states: StateMachine::new(initial_state),
+            states: StateMachine::new(initial_state, planner),
             timer: Stopwatch::new(),
             context: context,
         }
     }
 
+    /// Build a new Application using builder pattern.
+    pub fn build<T: 'static>(initial_state: T, config: Config) -> ApplicationBuilder<T>
```
Emilgardis: Instead of using `where T:` and `T: 'static` use `T: State + 'static`

src/engine/app.rs
```diff
@@ -41,38 +44,78 @@ impl Application {
...
         }
-        let events = self.context.borrow_mut().broadcaster.poll();
-        self.states.handle_events(events);
-        while SteadyTime::now() - self.last_fixed_update > self.fixed_step {
-            self.states.fixed_update(self.fixed_step);
+        let events = self.context.lock().unwrap().broadcaster.poll();
+        self.states.handle_events(events, self.context.lock().unwrap().deref_mut());
+        let fixed_step = self.context.lock().unwrap().fixed_step.clone();
+        let last_fixed_update = self.context.lock().unwrap().last_fixed_update.clone();
+        while SteadyTime::now() - self.context.lock().unwrap().last_fixed_update > fixed_step {
```
Emilgardis: This doesn't seem to work, we get stuck here after some frames.

running cargo --example window with some added println!s.
```
app.rs states is running
window.rs Handling events!
app.rs advance frame after states.handle_events
app.rs states is running
app.rs advance frame after states.handle_events
app.rs states is running
app.rs advance frame after states.handle_events
app.rs in tick loop.
app.rs in tick loop.
app.rs in tick loop.
app.rs in tick loop.
app.rs in tick loop.
app.rs in tick loop.
app.rs in tick loop.
app.rs in tick loop.
app.rs in tick loop.
app.rs in tick loop.
app.rs in tick loop.
app.rs in tick loop.
```

src/context/src/lib.rs
```diff
@@ -30,10 +30,12 @@ use std::path::Path;
 
 pub mod video_context;
 pub mod broadcaster;
+pub mod timing;
 pub mod event;
 use video_context::{VideoContext, DisplayConfig};
 use broadcaster::Broadcaster;
 use event::EngineEvent;
+use timing::{Duration, SteadyTime};
```
LucioFranco: You forgot to re-expose `Duration` it is needed for the `Processor` trait.

nchashch: It can be accessed like `amethyst_context::timing::Duration`.

LucioFranco: You are right. It was a bit hidden. I am going to test out the rest tonight and this weekend. Good work.

kvark:
@emilgardis thanks for testing this! Could you confirm that the latest version of the code works?

Emilgardis:
Already done, works as it should now :)

kvark:
I suppose I can merge then. We had plenty of time for objections :)

Emilgardis:
Well, this was discussed in Gitter. There should soon also be a wiki page describing how context, planners (worlds) and systems all interact inside contexts and states.

Issue #79, opened 2016-07-24
Need a simple 3d rendering example


ahmedaliadeel:
Need a simple 3d rendering example, like a cube or sphere, nothing fancy.  Deferred or Forward defined in yaml file.

Emilgardis:
There already is one. On the develop branch: https://github.com/amethyst/amethyst/blob/develop/src/renderer/examples/main.rs

ahmedaliadeel:
Got that working but was wondering how to fit in the main window example with ecs. Anyway thanks I will see into this tonight.

![untitled](https://cloud.githubusercontent.com/assets/6824526/17083057/a7ff9318-519b-11e6-85fb-044f3bb16404.png)

ahmedaliadeel:
I mean i need an example without directly using glutin or gfx through amethyst interface! is something is WIP?

nchashch:
There is no example with rendering yet, but you might want to take a look at #80.

nchashch:
see `examples/sphere.rs` in #80.

ebkalderon:
Given that we have a working rendering example, I think we can mark this issue resolved

Issue #52, opened 2016-05-11
Integrate Asset management with Renderer


ghost:
It would be nice if we could abstract away the Textures/Vertex Buffers so that the ECS does not need to retain direct links to the data. The asset manager would be responsible to for loading and unloading assets and provide handles to said assets. That way user code never has to touch `gfx-rs` unless they are directly working on extending the renderer for their game.

msiglreith:
I toyed around with a possible asset pipeline and want to share my quick/ugly prototype (unfortunatley requiring nightly features!) https://gist.github.com/anonymous/176941be70ac1747c557ba68c47b56eb

Possible pipeline for an asset load request:
  ```
AssetStores -> (raw data) -> [AssetSubLoader(raw) -> (data) -> AssetSubLoader(asset)] -> Asset
```

AssetStore: A storage (similiar to a package manager), which returns a `[u8]` byte stream for the specific asset requested. Possible storage types would be a file archive(zip), directory or an http-server.

It should be possible to run the different stages parallel and asynchronous. Interesting would be using the new `future` library.

If the corresponding loader and some asset stores are registered to the asset manager, an assert can be loaded directly by AssetType + an identifier(e.g. file extension) or preloaded by using an identifier only.

Example usage (only implemented the actual loading from data to asset):
``` Rust
asset_manager.load_asset_from_data::<Mesh, Vec<VertexNormals>>(&vertex_data); // doesn't require storage
asset_manager.load_asset::<Mesh>("obj", &[0; 24])); // raw [u8] data from storage
asset_manager.preload_asset("obj", &[0; 24]));
```

Writing a loader for an combination of Asset + AssetSource:
https://gist.github.com/anonymous/176941be70ac1747c557ba68c47b56eb#file-playground-rs-L157-L172

I hope this was helpful to some extent!

fHachenberg:
Is the amethyst roadmap aiming for the possibility of hot swapping assets? I consider this an essential feature to speedup iteration cycles and thus increase productivity.

bjadamson:
Has the asset manager become too big? It seems like this kind of code **should** be represented via ecs components, no?

https://github.com/amethyst/amethyst/blob/develop/examples/03_renderable/main.rs#L44

jFransham:
[...]

PR #80, opened 2016-07-24
Add platform independent renderer frontend

### LINKED ISSUE #77: Add platform independent renderer frontend
Currently in order to access renderer user has to write code like this:
```rust
let mut context = self.context.borrow_mut();
match context.video_context {
VideoContext::OpenGL { ref window,
                       ref mut renderer,
                       ref frame,
                       ref mut device,
                        .. } => {
    renderer.submit(frame, device);
    window.swap_buffers().unwrap();
}
#[cfg(windows)]
VideoContext::Direct3D {  } => {
    // Direct3D code
},
VideoContext::Null => (),
}
```
It would make sense to add a renderer frontend, that would own `VideoContext` and deal with all this boilerplate internally.

Also platform independent `amethyst_renderer::Fragment` and `amethyst_renderer::Scene` would be required, since they take `gfx::Resource` type parameter, related to #52.

It should simplify implementation of the rendering system.

nchashch:
Changes:
- Add `renderer` and `asset_manger` modules to `amethyst_context`
- Add `Renderer`, `Fragment`, `FragmentImpl`, `Texture`, and `TextureImpl` types to `context::renderer`
- Add `AssetManager` and `FactoryImpl` types to `context::asset_manager`
- Move `Factory` from `VideoContext` to `FactoryImpl`
- Now `Context` owns `context::renderer::Renderer` and `context::asset_manager::AssetManager` instead of `VideoContext`
- Change some doc comments
- Update `examples/window.rs` example
- Add `examples/sphere.rs` example
- Mark all `Direct3D` match branches with `unimplemented!()` macro

`context::renderer::Renderer` allows:
- setting the rendering pipeline
- adding, deleting, and accesing `Fragment`s, `Light`s, and `Camera`s
- adding and deleting `Target`s and `Scene`s
- submitting the rendering job to `amethyst_renderer::Renderer`

`context::asset_manager::AssetManager` allows:
- loading `Texture`s from pixel data
- creating solid color constant `Texture`s
- loading `Mesh`es from vertex data
- creating `Fragment`s from loaded `Texture`s and `Mesh`es

It should close #77, #79 and #52


src/context/src/renderer.rs
```diff
@@ -0,0 +1,407 @@
...
+            VideoContext::OpenGL {
+                ref mut factory,
+                ..
+            } => {
+                let (buffer, slice) = factory.create_vertex_buffer_with_slice(&data, ());
+
+                let ka = match ka.texture_impl {
+                    TextureImpl::OpenGL { texture } => texture,
+                    #[cfg(windows)]
+                    TextureImpl::Direct3D {  } => return None,
```
kvark: should be `unimplemented!()` I suppose

src/context/src/renderer.rs
```diff
@@ -0,0 +1,407 @@
...
+        let texture_impl = TextureImpl::OpenGL {
+            texture: texture,
+        };
+        Texture {
+            texture_impl: texture_impl,
+        }
+    }
+
+    /// Create a `Texture` from pixel data.
+    pub fn create_texture(&mut self, kind: Kind, data: &[&[<<ColorFormat as Formatted>::Surface as SurfaceTyped>::DataType]]) -> Option<Texture> {
```
kvark: so, just to be clear, this only supports a single format (`ColorFormat`), right?

src/context/src/renderer.rs
```diff
@@ -0,0 +1,407 @@
...
+                                   ref mut renderer,
+                                   ref frame,
+                                   ref mut device,
+                                   .. } => {
+                renderer.submit(frame, device);
+                window.swap_buffers().unwrap();
+            }
+            #[cfg(windows)]
+            VideoContext::Direct3D {  } => {
+                // stub
```
kvark: unimplemented!()

kvark:
I looked over it and liked what I saw so far. The only thing I wish was specified here is the list of the current capabilities/limitations of the renderer, for clarity.

src/context/src/renderer.rs
nchashch: Yes, but it probably won't be hard to make it generic.

kvark:
One thing that worries me is the number of matches over GL/D3D/etc.
If they are really the same but just parametrized by different `gfx::Resources`, perhaps we could have a helper like
```rust
self.video_context.with_factory(|fac| { do something with fac });
self.video_context.with_frame(|fr| { do something with frame });
```
Here, the closures accepted are generic over the `R: gfx::Resources`, so we can call them on the appropriate match arm under the hood.

kvark:
Great work! Why did you go with a macro instead of a helper function as I initially suggested? Were there any troubles?

nchashch:
Yes, I tried implementing it using helper functions, but I could not find a way to pass a `VideoContext` field with a concrete type to a generic `with_` function.

So for example this code won't compile because of mismatched types error, since `frame` has concrete `Frame<gfx_device_gl::Resources>` type and a `Frame<R>` type is expected.
```rust
impl Renderer {
    // ...
    fn with_frame<R, F>(&mut self, f: F)
        where R: gfx::Resources,
              F: FnMut(&mut amethyst_renderer::Frame<R>),
    {
        match self.video_context {
            VideoContext::OpenGL {
                ref mut frame,
                ..
            } => {
                f(frame)
            },
            #[cfg(windows)]
            VideoContext::Direct3D {
            } => {
                unimplemented!();
            },
            VideoContext::Null => (),
        }
    }
}
```
Also if it was possible, I am not sure if rust compiler would be able to infer the types for `R` and `F`, so a match over `VideoContext` would be required anyway to pass the proper types to the method.

kvark:
Ah, I see. So what we need is a generic function of `R` to be passed in, but that means higher-order types... It seems.

csherratt:
LGTM

PR #81, opened 2016-08-04
Add support for sound using rodio

nchashch:
Changes:
- Define `Sound` type in `context::asset_manager` mod
- Add `AssetManager::load_sound()` and `AssetManager::get_sound()` methods
- Add `examples/sound.rs` example
- Now `Context` owns a `rodio::Sink`


kvark:
Wow nice, you sir are doing awesome stuff!

src/context/src/lib.rs
```diff
@@ -52,10 +53,10 @@ config!(
 /// An `Arc<Mutex<Context>>` is passed to every `Processor` run by the engine and a `&mut Context` is passed to every `State`
 /// method.
 pub struct Context {
-    // pub video_context: VideoContext,
     pub renderer: Renderer,
     pub asset_manager: AssetManager,
     pub broadcaster: Broadcaster,
+    pub audio_sink: Option<rodio::Sink>,
```
Emilgardis: Do we want to make context depend on rodio? I believe this is one of those use cases where we should really use the (unimplemented) shard system. Does "all" audio libs use background threads for making sound? If that is the case, maybe we could make this generic by providing a wrapper for rodio where `Sink` is `AudioSink` (which has the functions needed)? Also, my limited understanding of rodio gives me that each sink can only play one sound, and there is no way to modify the sound output other than volume.

nchashch: Audio implementation can be feature gated, so it can be disabled using the shard system. And I think `rodio::Sink` doesn't need a wrapper unless we want support for multiple audio libraries.

> Also, my limited understanding of rodio gives me that each sink can only play one sound, and there is no way to modify the sound output other than volume.

You can play multiple sounds by mixing sources with `Source::mix` and there are other methods which allow modifying audio output like `Source::fade_in` or `Source::speed` see [docs](http://tomaka.github.io/rodio/rodio/source/trait.Source.html) for `Source` trait. But there is no audio spatialization yet.

ebkalderon:
@nchashch Wow, this PR looks great! Shame to see it going stale, though. Do you plan to keep working on this?

nchashch:
Yes, it looks like the problem is with the ALSA backend in rodio, see tomaka/rodio#73. The behaviour has changed after tomaka/rodio#72, it doesn't panic anymore, but it still doesn't seem to work properly with ALSA. I will update the issue in rodio repo sometime soon.

ebkalderon:
Sounds good, thank you! I'll be following the progress eagerly.

Aceeri:
Is this still being working on at all?

nchashch:
Since #133 was merged, sound can't be added the way it is added in this PR (there is no `Context` struct anymore). Also there are issues with rodio/cpal on ALSA. Sound support will have to be added in another PR.

PR #83, opened 2016-08-06
Make examples smarter by knowing where config lives absolutely.

Emilgardis:
Makes it so users don't have to run examples inside a subfolder to the manifest dir

### DIFF
--- a/examples/sphere.rs
+++ b/examples/sphere.rs
@@ -82,7 +82,10 @@ impl State for Example {
 
 fn main() {
     use amethyst::context::Config;
-    let config = Config::from_file("../config/window_example_config.yml").unwrap();
+    let config = Config::from_file(
+        format!("{}/config/window_example_config.yml",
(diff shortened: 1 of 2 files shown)


ebkalderon:
Perhaps you might want to take a look at how amethyst/tools#33 currently handles this issue? In the build tool, paths are assumed to be relative to the top level directory containing the Cargo manifest. On startup, this directory is identified in a manner very similar to Cargo itself.

Emilgardis:
Nce pr! :+1: Hovewer, I don't think that is needed as this is only the examples. We could change `Config::from_file` to be relative to the manifest, but I don't see how anyone would benefit from that. I think I understand what you are saying, but I just don't see how one could benefit from it.

ebkalderon:
I think I'll merge this. Right now, the examples are somewhat broken without this. Besides, if we ever want to revise the approach later on, we can always make another pull request. Thanks for the improvement @Emilgardis!

PR #82, opened 2016-08-05
Pass Context to Application::build instead of Config

nchashch:
It allows passing a `&mut Context` to a processor's init function before calling `Application::build`.

PR #84, opened 2016-08-16
Pass Context to Application::build instead of Config

nchashch:
It allows passing a `&mut Context` to a processor's init function before calling `Application::build`.
(#82 was closed to move changes from develop to develop-fix branch)

### DIFF
--- a/src/engine/app.rs
+++ b/src/engine/app.rs
@@ -3,7 +3,7 @@
 use super::state::{State, StateMachine};
 use context::timing::{SteadyTime, Stopwatch};
 use context::event::EngineEvent;
-use context::{Config, Context};
+use context::Context;
 use ecs::{Planner, World, Processor, Priority};
 use std::sync::{Arc, Mutex};
 use std::ops::DerefMut;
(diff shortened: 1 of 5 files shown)


kvark:
Seems alright. @ebkalderon ?

src/engine/app.rs
```diff
@@ -16,11 +16,10 @@ pub struct Application {
 }
 
 impl Application {
-    /// Creates a new Application with the given initial game state, planner, and config.
-    pub fn new<T>(initial_state: T, planner: Planner<Arc<Mutex<Context>>>, config: Config) -> Application
+    /// Creates a new Application with the given initial game state, planner, and context.
+    pub fn new<T>(initial_state: T, planner: Planner<Arc<Mutex<Context>>>, context: Context) -> Application
```
ebkalderon: Didn't notice this line before. Why do we pass in two Contexts into this method exactly, one enclosed in a Planner and one alone? Can't the `new()` method build a Planner itself, insert the given State and Context inside, and pass it into the StateMachine?

nchashch: `Planner` doesn't own an `Arc<Mutex<Context>>` it just denotes the type of `Planner::dispatch` argument and the type of `Processor::run` argument.

Passing `Planner` to `Application::new` allows adding processors to planner and registering components before it is consumed by `Application::new`, `ApplicationBuilder::done` calls it.

ebkalderon: Thanks for the explanation. I was wondering what it was for.

ebkalderon:
Everything looks good! I think we can merge this.

PR #85, opened 2016-08-23
Add default rendering processor

nchashch:
Changes:
- Add `processors::rendering` module containing `RenderingProcessor` type, component types such as `Renderable`, `Light`, `Camera`, and a `RendererConfig` type
- Add `examples/renderable.rs` and `examples/pong.rs` examples
- Add `ApplicationBuilder::register` method to allow registering components with the application using the following syntax:
```rust
Application::build(StateType::new(), context)
    .with::<ProcessorType>(processor, "processor_name", 0)
    .register::<ComponentType>()
    .done();
```
- Add `amethyst_renderer::Camera::orthographic` method, which returns an orthographic projection matrix with the given parameters
- Rename `amethyst_context::Config` to `amethyst_context::ContextConfig`
- Add `amethyst::engine::Config`, which contains `ContextConfig` and `RendererConfig`


examples/pong.rs
```diff
@@ -0,0 +1,385 @@
...
+}
+
+// Pong game processor
+impl Processor<Arc<Mutex<Context>>> for PongProcessor {
+    fn run(&mut self, arg: RunArg, context: Arc<Mutex<Context>>) {
+        use amethyst::context::event::{EngineEvent, Event, VirtualKeyCode, ElementState};
+        use std::ops::Deref;
+
+        // Get all needed component storages and resources
+        let context = context.lock().unwrap();
```
kvark: I wonder... technically, why can't this context just be an ECS resource?

examples/pong.rs
```diff
@@ -0,0 +1,385 @@
...
+    use amethyst::engine::Config;
+    let config = Config::from_file(
+        format!("{}/config/pong_example_config.yml",
+                env!("CARGO_MANIFEST_DIR"))
+        ).unwrap();
+    let mut context = Context::new(config.context_config);
+    let rendering_processor = RenderingProcessor::new(config.renderer_config, &mut context);
+    let mut game = Application::build(Pong, context)
+                   .with::<RenderingProcessor>(rendering_processor, "rendering_processor", 0)
+
```
kvark: so many empty lines

src/processors/rendering.rs
```diff
@@ -0,0 +1,442 @@
...
+                   vec![
+                       Clear::new(clear_color),
+                   ])]
+}
+
+impl RenderingProcessor {
+    pub fn new(renderer_config: RendererConfig, context: &mut Context) -> RenderingProcessor {
+        let clear_color = renderer_config.clear_color;
+        let pipeline =
+            match renderer_config.pipeline.as_str() {
```
kvark: why is the `match` on the new line?

src/processors/rendering.rs
```diff
@@ -0,0 +1,442 @@
...
+                "Forward" => {
+                    match renderer_config.shading.as_str() {
+                        "Flat" => {
+                            forward_flat(clear_color)
+                        },
+                        "Shaded" => {
+                            forward_shaded(clear_color)
+                        },
+                        _ => clear(clear_color),
```
kvark: shouldn't there be a warning/error if the requested type is not supported?

src/processors/rendering.rs
```diff
@@ -0,0 +1,442 @@
...
+            let mut light_indices = HashSet::<usize>::new();
+            for (entity, light) in (&entities, &mut lights).iter() {
+                match light.idx {
+                    Some(idx) => {
+                        // If this Light is already in frame then update it.
+                        light_indices.insert(idx);
+                        if let Some(frame_light) = context.renderer.mut_light(ACTIVE_SCENE_NAME, idx) {
+                            *frame_light = light.light.clone();
+                        } else {
+                            println!("Error: entity with id = {0} is deleted, \
```
kvark: maybe `error!` instead

src/processors/rendering.rs
```diff
@@ -0,0 +1,442 @@
...
+                renderable.update_transform_matrix();
+                match renderable.idx {
+                    // If this Renderable is already in frame then update the transform field
+                    // of the corresponding Fragment.
+                    Some(idx) => {
+                        renderable_indices.insert(idx);
+                        if let Some(transform) = context.renderer.mut_fragment_transform(ACTIVE_SCENE_NAME, idx) {
+                            *transform = renderable.transform;
+                        } else {
+                            println!("Error: entity with id = {0} is deleted, \
```
kvark: same here, `error!`

kvark:
Got a few notes on the PR, plus the cgmath dependency seems to be screwed up (according to Travis).
Other than that, here is some impressive work, thanks @nchashch !
Would be great to have @csherratt and @ebkalderon eyes here as well.

examples/pong.rs
nchashch: I think it could work, but I haven't tried it yet. Probably it would make `Application` code less ugly, since it would allow removing all `context.lock.unwrap()` calls in `Application` impl and only `&mut World` would need to be passed to `State` methods instead of `&mut Context` and `&mut World`. Also it would simplify `Processor` implementation code, since you won't have to deal with `Arc<Mutex<Context>>` argument in `Processor::run` anymore.

But I think it would be better to put this change in a separate PR.

kvark: Sure thing. What about `println!` calls though - would you be able to fix them in this PR?

Emilgardis:
[hello_world.md](https://github.com/nchashch/amethyst/blob/bd897b85795d9f95ded07d12a6d4a092d9243ec4/book/src/getting_started/hello_world.md) needs to get updated. Otherwise everything looks good! :+1:

src/processors/rendering.rs
```diff
@@ -0,0 +1,427 @@
...
+                Lighting::new(ACTIVE_CAMERA_NAME, "gbuffer", ACTIVE_SCENE_NAME)
+            ]
+        ),
+    ]
+}
+
+impl RenderingProcessor {
+    pub fn new(renderer_config: RendererConfig, context: &mut Context) -> RenderingProcessor {
+        let clear_color = renderer_config.clear_color;
+        let pipeline = match renderer_config.pipeline.as_str() {
```
csherratt: Ni: this might be tidier if you use a single match.

```rust
match (renderer_config.pipeline.as_str(), renderer_config.shading.as_str()) {
    ("Forward", "Flat") => forward_flat(clear_color),
    ("Forward", "Shaded") => forward_shaded(clear_color),
    ("Deferred", "Flat") => deferred_flat(clear_color),
    ("Deferred", "Shaded") => deferred_shaded(clear_color),
    _ => panic!("Error: Can't provide rendering pipeline requested in renderer_config, \
                 renderer_config.shading field is invalid.")
}

File: src/processors/rendering.rs [hunk 8]
```diff
@@ -0,0 +1,427 @@
...
+/// A projection enum which is required to create a `Camera` component.
+pub enum Projection {
+    Perspective {
+        fov: f32,
+        aspect: f32,
+        near: f32,
+        far: f32,
+    },
+    Orthographic {
+        left:f32,
```
csherratt: nit: spacing

src/renderer/src/lib.rs
```diff
@@ -178,6 +178,7 @@ pub struct Fragment<R: gfx::Resources> {
 }
 
 /// A basic light
+#[derive(Clone)]
```
csherratt: no reason this can't be `Copy` too.

csherratt:
Some minor nits, but otherwise it LGTM

examples/pong.rs
nchashch: There is no propper logging yet, so it is either `println!` or `panic!`. And I think logging also should be added in a separate PR.

kvark:
@nchashch ok, considering the logging is left out for another PR, could you please rebase the code (and may be fix a few @csherratt -noted nits) for the merge?

ebkalderon:
@nchashch This looks pretty good to me! Hopefully we can get this in a mergeable state soon.

kvark:
Thanks @nchashch, merging now!

Issue #88, opened 2016-09-06
Remove usage of String in the asset manager


bjadamson:
Looking up resources by a string is convenient but slow thanks to dynamic allocation. I believe this is why OpenGL uses integer handles, even if it they are less convenient to work with.

These string should be replaced with type-safe integer wrappers, or some other design to improve the performance of the asset manager.
https://github.com/amethyst/amethyst/blob/develop/src/context/src/asset_manager.rs#L41-L42

Consideration must be given to the current convenience though, especially if it is removed.

Maybe offering both versions make sense? What does everyone think?

OvermindDL1:
Shouldn't most asset strings have the 'key' only allocated once per app?  If not then a flyweight string container could do that.  That way you get a unique id comparison via taking the pointer of the string or the appropriate method in rust.

bjadamson:
> Shouldn't most asset strings have the 'key' only allocated once per app?

Perhaps, but it's forcing the user to use a string when they otherwise would not. That's really what I'm trying to point out.

bjadamson:
#90

Geemili:
Could we make it a generic so that the user can decide?

Another thought I had is making a build script that takes a resource file and bakes the strings into integers, similar to how android does it. The build script could generate a bunch of environment variables like `AMETHYST_RESOURCE_ID_sphere_mesh=123`. After the environment is generated, the user can just include using:

```
let sphere_resource = env!("AMETHYST_RESOURCE_ID_sphere_mesh");
```

Or even a custom macro so we can omit the prefix:

```
let sphere_resource = resource!("sphere_mesh");
```

Anyway, those are my thoughts. Don't know if we want to mess with build scripts.

bjadamson:
I like this, I think I can put a builder in front of this and make it general.

Geemili:
Another thought: If we're going to deal with build scripts, we could just generate a "resources.rs" that contains all the resource constants. It would mean that all the `env!()` macros wouldn't need a `u32::from_str().unwrap()` surrounding them.

Emilgardis:
One way to have both i64's and String is to with #90 have each function take id: T where T: Into<*Id>
The one problem I can see are collisions. One way to fix this is to feature-flag i64 to only have one enabled, but I'm playing around trying to find the best solution.

bjadamson:
Another thought, maybe it makes sense to always support static str& as a key too.

ebkalderon:
[...]

PR #90, opened 2016-09-07
#88. Remove String from Asset Manager.

bjadamson:
Here's an example of what I'm thinking, I've simply replaced the String types with integer wrappers.

I'm guessing there are quite a few problems with the pull request as is, but I'd love feedback on this line of changes.

### DIFF
--- a/examples/02_sphere/main.rs
+++ b/examples/02_sphere/main.rs
@@ -1,10 +1,14 @@
 //! Displays a multicolored sphere to the user.
 
 extern crate amethyst;
+extern crate amethyst_context;
 extern crate cgmath;
 
 use amethyst::engine::{Application, State, Trans};
 use amethyst::context::{ContextConfig, Context};
+use amethyst::context::asset_manager::AssetManager;
+use amethyst::context::device::DeviceManager;
+use amethyst::context::prefab::{PrefabGenerator, PrefabIndex};
 use amethyst::config::Element;
 use amethyst::ecs::{World, Join};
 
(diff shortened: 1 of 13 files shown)


kvark:
I'd rather find a way to avoid `str -> String` conversion on lookups (while still storing String) than moving to pure IDs.

Emilgardis:
@kvark a small change would be 

    fn get_mesh<T: Into<String>>(mesh: T)

That way the user chooses if a conversion is done.

kvark:
Maybe 'T: Borrow<&str> + Hash'?

kvark:
This approach wouldn't get rid of `String` conversions:
```rust
fn get_mesh<T: Into<String>>(mesh: T)
```
In the best case, it would be equivalent to `clone()` but still be allocating...

nchashch:
Maybe instead of taking ID's as arguments return them? Like:
```rust
    fn load_texture(&mut self, texture_data: TextureData) -> Option<TextureID> {
        ...
    }
```
Then `Vec`s can be used to store assets instead of `HashMap`s.
It probably wouldn't introduce much inconvenience, for example instead of writing:
```rust
context.asset_manager.create_constant_texture("dark_blue", [0.0, 0.0, 0.01, 1.]);
context.asset_manager.create_constant_texture("green", [0.0, 1.0, 0.0, 1.]);
context.asset_manager.gen_sphere("sphere", 32, 32);

let fragment = context.asset_manager.get_fragment("sphere", "dark_blue", "green", transform).unwrap();

```
you would write:
```rust
let dark_blue = context.asset_manager.create_constant_texture([0.0, 0.0, 0.01, 1.]).unwrap();
let green = context.asset_manager.create_constant_texture([0.0, 1.0, 0.0, 1.]).unwrap();
let sphere = context.asset_manager.gen_sphere(32, 32).unwrap();

let fragment = context.asset_manager.get_fragment(sphere, dark_blue, green, transform).unwrap();
```
IDs are type checked, so you can't query a `Texture` with a `MeshID`, and instead of failing silently asset manager returns `None` if the asset can't be loaded.

Although there is still a problem with "asset leaks" when an ID has gone out of scope and the user didn't free the asset.

bjadamson:
I think suggestion from @nchashch makes sense to me, but it's basically the same thing as using integer wrappers (handles). This is what OpenGL does.

> I'd rather find a way to avoid str -> String conversion on lookups (while still storing String) than moving to pure IDs. @kvark 

Why store a String? Would an abstraction that stores a String built on-top of this asset manager be equivalent for your uses?

Aceeri:
@bjadamson Are you still interested in this?

ebkalderon:
@bjadamson Hey, are you still working on this? This PR is getting quite stale. If no one is willing to pick it up and finish, we may have to close it.

ebkalderon:
Given that the companion issue #88 has been closed, I will close this PR as well.

PR #91, opened 2016-09-07
[RFC] Slight code cleanup

ebkalderon:
* Use the more idiomatic `&[Entity]` instead of `Vec<Entity>` in function signatures.
* Use `where` clause to shorten function signatures.
* Remove `// Unit tests` comment requirement from `CONTRIBUTING.md` section on testing.
* Add `cargo run --example {example-name}` to `CONTRIBUTING.md` section on testing.
* Update unit test example from `timing.rs` (no longer in the top-level `amethyst` crate) to `state.rs`.
* Remove unused `tasks` module.
* Miscellaneous formatting changes.
* Use HTTPS where possible in API docs.

PR #98, opened 2016-09-08
Issue89

bjadamson:
Remove all graphics backend logic, delegating all responsibility to Gfx.
    
I'm not sure if this code was written before GFX was added to this project, or if GFX doesn't provide enough controls, but GFX is written to be backend-agnostic.
    
See this:
https://github.com/gfx-rs/gfx/blob/master/src/lib.rs#L23-L36
    
and this:
https://github.com/gfx-rs/gfx/tree/master/src/backend
    
I'm not sure the current status of any of the backends, but it seems like a bad decision to layer backend support ontop of GFX like the code is. This commit removes all opengl/direct3d/null cfg() directives (fun fact, it also removes all windows only code-paths in this package) and instead assumes GFX can handle the platform or backend specific details of OpenGL/Direct3D/NULL.
    
It might make sense to add a "NULL" backend to the GFX project directly.


ebkalderon:
@bjadamson This is exactly what I've wanted to see! It's always bothered me how "interested" in GFX's internals the renderer has become. I am excited to see the fruits of these changes.

src/context/src/asset_manager.rs
```diff
@@ -20,60 +20,32 @@ use self::genmesh::{MapToVertices, Triangulate, Vertices};
...
-    OpenGL {
-        factory: gfx_device_gl::Factory,
-    },
-    #[cfg(windows)]
-    Direct3D {
-        // stub
-    },
-    Null,
-}
+use renderer::{Fragment};
```
ebkalderon: Minor nitpick, the braces here are unnecessary and can be omitted.

ebkalderon:
Uh oh, looks like some of the merge conflicts might be from the changes I made recently in PR #91. :confused:

kvark:
Unfortunately, this is not the way to go. You are using `gfx_backend_gl` stuff, which is not backend-agnostic by default. Problem is - GFX backends are embodied into the type system, while Amethyst wants to be able to jiggle them at run-time, hence the enum you are trying to remove.

bjadamson:
> while Amethyst wants to be able to jiggle them at run-time

Is this desirable, or just what the code is currently trying to do? It seems weird (to me) to not lock this down at compile time. What's the use case?

kvark:
I'd see 2 reasons for Amethyst, but @ebkalderon may get you some more:
  1. to avoid complicating the API with abstract generics. All the backend-abstract user code for GFX needs to use `R: Resources` bound as well as other traits occasionally.
  2. to allow selecting the backend in the configuration file, independent of the code base

bjadamson:
2 reasons is plenty!

ebkalderon:
I didn't even think about the ramifications of having the GFX backends reside solely at compile time, haha. Thanks for catching that amidst my excitement, @kvark. Still, while I do support the toggling of backends at runtime, we definitely need to improve it since I don't see the current enum approach being maintainable in the long run.

As such, I will be rejecting this PR. However, it does point out some flaws in the current design which will have to be addressed by another suitable PR sometime in the future. Thanks anyway for submitting it, @bjadamson! :+1:

PR #99, opened 2016-09-08
Remove Broadcaster::poll, State::handle_events, and StateMachine::handle_events methods

nchashch:
Events can be polled like this:

``` rust
let events = ctx.broadcaster.read::<Event>();
for event in events.iter() {
...
}
```

So there is no need to provide access to a `Vec<Entity>` or a `&[Entity]` with `Broadcaster::poll`. Also `State::handle_events`, `StateMachine::handle_events` methods are not needed, since events can be handled in `State::update`.


ebkalderon:
Thanks for tweaking the Broadcaster, this is is a huge improvement over the current state! :smile: However, I personally feel that we should retain the `handle_events()` method, since I would like to differentiate reacting to events from unrelated gameplay code, which the `update()` method is intended for.

Perhaps we could make the `State::handle_events()` method simpler by passing in `&[EngineEvent]` into `StateMachine::handle_events()`, thereby hiding the `ctx.broadcaster.read::<EngineEvent>()` from the user? What are your thoughts on this?

ebkalderon:
Any ongoing progress on this?

nchashch:
I think the problem with having `State::handle_events` method which does the same thing as `State::update` and also has a `&[EngineEvent]` as an argument is mostly that it is less elegant than having just `State::update`. Most of gameplay code probably woud be in `Processor` implementations or maybe in script files, so it isn't necessary to have 2 methods doing the same thing just to separate gameplay code from event handling code.

Also if we keep `State::handle_events` I think passing `&[EngineEvent]` to `State::handle_events` wouldn't be very elegant, since there are other possible types of events which might need to be processed in `State::handle_events`. So if for example you would want to process `EngineEvent`s and `UIEvent`s or just `UIEvents`, you would still have to call `ctx.broadcaster.read::<UIEvent>()`, and in the later case you also would have a dummy `&[EngineEvent]`argument. I think it would be better to keep it uniform, so you have to manually call `Broadcaster::read` for every event type you want to process.

nchashch:
Event handling has been changed in #133, so this PR is outdated now.

PR #100, opened 2016-09-08
Reorganize examples

ebkalderon:
Examples are now put in separate folders, with each one containing a `resources/config.yml` much like a real Amethyst project would. That is, all except for the "hello, world" example which is just a lone `main.rs` file. This demonstrates how to use the engine without a `resources` folder. Each example is now sorted by increasing level of complexity, making it straightforward for newcomers to follow. Does everyone agree with this change?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,14 +12,6 @@ repository = "https://github.com/amethyst/amethyst"
 readme = "README.md"
 license = "MIT"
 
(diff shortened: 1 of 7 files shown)


ebkalderon:
Since there appear to be no objections, I guess I'll merge this.

LucioFranco:
@ebkalderon is back!! Woo! Great work!

PR #107, opened 2016-09-12
[RFC] Refactor event handling to better match mockup

ebkalderon:
This pull request should hopefully reduce the verbosity of event handling when implementing the `State` trait. The goal is to eventually make it better resemble the mockup seen in the [API documentation](https://www.amethyst.rs/doc/develop/amethyst/index.html).

This refactor is not perfect, as it reveals some blemishes in our current `Context` design, but it does make it event handling a bit prettier from the user's point of view.

* Apply some changes proposed by @nchashch in pull request #99
  * Remove `Broadcaster.poll()` entirely
  * Update examples and doc comments to match
* Move event collection out of `update()/handle_events()` and into game loop
* Change method signature of all `handle_events()` methods to accept `&[Event]` instead of `&[Entity]`, thereby removing some boilerplate

### Before:
```rust
fn handle_events(&mut self, events: &[Entity], ctx: &mut Context, _: &mut World) -> Trans {
    use amethyst::context::event::{EngineEvent, Event};
    let storage = ctx.broadcaster.read::<EngineEvent>();
    for e in events {
        let event = storage.get(*e).unwrap();
        match event.payload {
            // React to events...
        }
    }
}
```

### After:
```rust
fn handle_events(&mut self, events: &[Event], _: &mut Context, _: &mut World) -> Trans {
    for e in events {
        match *e {
            // React to events...
        }
    }
}
```

### DIFF
--- a/examples/01_window/main.rs
+++ b/examples/01_window/main.rs
@@ -4,25 +4,22 @@ extern crate amethyst;
 
 use amethyst::engine::{Application, State, Trans};
 use amethyst::context::{Context, ContextConfig};
+use amethyst::context::event::{Event, VirtualKeyCode};
 use amethyst::config::Element;
-use amethyst::ecs::{World, Entity};
+use amethyst::ecs::World;
 
 struct Example;
 
 impl State for Example {
-    fn handle_events(&mut self, events: &[Entity], ctx: &mut Context, _: &mut World) -> Trans {
-        use amethyst::context::event::{EngineEvent, Event, VirtualKeyCode};
-        let mut trans = Trans::None;
-        let storage = ctx.broadcaster.read::<EngineEvent>();
+    fn handle_events(&mut self, events: &[Event], _: &mut Context, _: &mut World) -> Trans {
         for e in events {
-            let event = storage.get(*e).unwrap();
-            match event.payload {
-                Event::KeyboardInput(_, _, Some(VirtualKeyCode::Escape)) => trans = Trans::Quit,
-                Event::Closed => trans = Trans::Quit,
+            match *e {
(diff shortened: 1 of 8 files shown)


ebkalderon:
Given that there appear to be no major objections, is it okay if I merge this in preparation for 0.4.0?

csherratt:
LGTM

ebkalderon:
@nchashch What do you think?

nchashch:
I have written about it in https://github.com/amethyst/amethyst/pull/99#issuecomment-246743129, also I think we should keep the name `EngineEvent`, because it is not the only type of event and other types of events may be used like `UIEvent`, `PhysicsEvent`, etc.

ebkalderon:
I see what you mean, and you have some valid points. I do have two things I would like to compromise on:

1. What if we pass in the `Broadcaster` itself into `handle_events`? How about renaming it something like `Events` or `EventQueue` or something more descriptive? That would solve the redundancy problem in this pull request and provide a uniform place to react to events of any type.
2. `EngineEvent` seems too generic a name to me, personally. What about `WindowEvent`? It would make more sense, since it holds payloads like `Closed `, `Resized `,  etc. Also, what if we made it a tuple struct instead of a normal struct?  That way we could destructure events like:

```rust
fn handle_events(&mut self, events: Events, _: &mut Context) -> Trans {
    for WindowEvent(e) in events.read::<WindowEvent>().iter() {
        match e {
            // Do stuff...
        }
    }
}
```
I think I'll reject my own pull request, but I'm still not happy with the current ergonomics and I think they should be improved. If these points seem like a step in the right direction, I'll make a new pull request in time for 0.4.0.

ebkalderon:
Whoops, I forgot to close this pull request!

PR #108, opened 2016-09-13
[WIP] ECS style event processing

kvark:
Only the renderable example is ported at the moment.

### DIFF
--- a/src/engine/app.rs
+++ b/src/engine/app.rs
@@ -57,25 +57,26 @@ impl Application {
         self.states.start(self.context.lock().unwrap().deref_mut());
     }
 
+    fn update_states(states: &mut StateMachine, context: &mut Context) {
+        let events = context.poll_engine_events();
+        for e in events {
+            context.broadcaster.publish().with::<EngineEvent>(e);
+        }
+
+        let entities = context.broadcaster.poll();
+        states.handle_events(&entities, context);
+
+        if context.last_fixed_update.elapsed() >= context.fixed_step {
+            states.fixed_update(context);
+            context.last_fixed_update += context.fixed_step;
+        }
+
+        states.update(context);
+    }
+
     /// Advances the game world by one tick.
     fn advance_frame(&mut self) {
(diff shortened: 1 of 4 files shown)


kvark:
Just realized using `run1w0r` is going to make the `update` body way prettier

src/engine/app.rs
```diff
@@ -57,25 +57,26 @@ impl Application {
         self.states.start(self.context.lock().unwrap().deref_mut());
     }
 
-    /// Advances the game world by one tick.
-    fn advance_frame(&mut self) {
-        let events = self.context.lock().unwrap().poll_engine_events();
+    fn update_states(states: &mut StateMachine, context: &mut Context) {
```
nchashch: Maybe it would be better to do it in a single function like this:
```rust
fn advance_frame(&mut self) {
    {
        let mut ctx = self.context.lock().unwrap();
        let events = ctx.poll_engine_events();
        for e in events {
            ctx.broadcaster.publish().with::<EngineEvent>(e);
        }

        let entities = ctx.broadcaster.poll();
        self.states.handle_events(&entities, ctx.deref_mut());

        let fixed_step = ctx.fixed_step;
        let last_fixed_update = ctx.last_fixed_update;

        if last_fixed_update.elapsed() >= fixed_step {
            self.states.fixed_update(ctx.deref_mut());
            ctx.last_fixed_update += fixed_step;
        }

        self.states.update(ctx.deref_mut());
    }
    self.states.run_processors(self.context.clone());
    {
        let mut ctx = self.context.lock().unwrap();
        ctx.broadcaster.clean();
        ctx.renderer.submit();
    }
}
```

ebkalderon: I prefer this. Simplicity is good.

ebkalderon:
@kvark Is this still being worked on?

kvark:
Tentatively. Go ahead if you want to finish it, otherwise I'll get back to fix it eventually.

kvark:
Closing as outdated

PR #111, opened 2016-09-27
Refactor Application::advance_frame and refactor examples

nchashch:
Changes:
- No more excessive locking in `Application::advance_frame`
- Rename `context` parameter to `ctx` in some examples
- Uniformly handle events in `State::update` method in all examples


ebkalderon:
Looks like this is set to merge now. Thanks for the good work, @nchashch!

Issue #53, opened 2016-05-11
Transform Processor


csherratt:
A transform system is a pretty common way to handle the scene graph. Basically it takes a `Local Transform` and a `Parent` entity and calculates the `Global Transform`. This allows objects that are attacked to the parent entity to `follow` it around. An object without a parent global transform is it's local transform, but it is stored as a separate value.

The Parent systems might be ok to handle separately as they can be used to do neat things like handle deletes in a clean way. Deleting the parent entity should delete all child entities.

kvark:
it also can be done by the system that updates transformations: if it finds an entity parent alive, it updates the current entity model matrix. If the parent is dead, it removes the current entity.

kvark:
Possible approaches to making sure the parent transform is up to date when processing a child:
  1. ignore that. Worst thing is 1 frame latency within the transformation
  2. enforce the children entities always having higher indices from the parents, so just iterating them as usual gives you proper results
  3. do recursive walk over the tree - this may be slow, and not canonical to ECS

lukebitts:
Is the engine going to use some kind of spatial structure? I ask this because in my own entity/component project I wanted to do frustrum culling and it was really annoying to keep the spatial tree in sync with creation and destruction of entities.

kvark:
@lukebitts we can start with simple parent-child relationships and culling on per-entity level. If we get a spatial structure, it's going to be its own component storage, so part of the syncing headache is going to be addressed automatically by the ECS. It's not completely clear as to how exactly this will work in the long-term, but short-term we are all set.

OvermindDL1:
The Spatial Structures in mine were just other Systems that listened to and worked with position and other components.  Many often did add another specific component to the entities for meta-data (octree and such).

fHachenberg:
Can such a "transform processor" be understood as a very simple kinematic simulator? In which cases does one use the transform processor and in which cases does one for example let the physics engine solve rigid body joint constraints instead? What kind of scene graph functionality do you have in mind for amethyst?

kvark:
I see the Transform Processor role to simply resolve hierarchical parent-child links and keep the world transformation per entity up to date.

Doing any sort of physics integration should be left out to a different processor. That, in theory. Problem would be that the physics would need to work on those world positions, may be modify them, and then we'd need to propagate the changes back to the local positions...

So, I guess we can't have both at the same time, and it's fine.

PR #113, opened 2016-10-02
Transform Processor

Aceeri:
Adds `LocalTransform`, `Transform,` and `Parent` components as a scene graph.

`LocalTransform` can be modified as a separate vector for position, a quaternion for rotation, and another vector for scale. `Parent` defines another entity to be the "parent" of the current entity, meaning its transformation will be the parent's global/world transform \* the local transformation. `Transform` is the component to hold the global/world transformation of entities for use in the renderer and such. The `Transform` can also be modified using a separate system for things like physics instead of being forced to use this scene graph system for all positioning.

**Benchmarks:**
No parents:

```
bench_1000_flagged        ... bench:       6,051 ns/iter (+/- 195)
bench_1000_half_flagged   ... bench:       6,083 ns/iter (+/- 199)
bench_1000_third_flagged  ... bench:       6,228 ns/iter (+/- 1,319)
bench_1000_unflagged      ... bench:       6,157 ns/iter (+/- 454)
bench_5000_flagged        ... bench:      21,858 ns/iter (+/- 1,170)
bench_5000_half_flagged   ... bench:      21,774 ns/iter (+/- 1,370)
bench_5000_third_flagged  ... bench:      21,836 ns/iter (+/- 1,270)
bench_5000_unflagged      ... bench:      21,880 ns/iter (+/- 1,303)
bench_10000_flagged       ... bench:      40,032 ns/iter (+/- 2,726)
bench_10000_half_flagged  ... bench:      39,307 ns/iter (+/- 3,275)
bench_10000_third_flagged ... bench:      39,793 ns/iter (+/- 2,391)
bench_10000_unflagged     ... bench:      39,338 ns/iter (+/- 2,892)
bench_50000_flagged       ... bench:     267,418 ns/iter (+/- 64,067)
bench_50000_half_flagged  ... bench:     247,280 ns/iter (+/- 51,733)
bench_50000_third_flagged ... bench:     211,564 ns/iter (+/- 64,728)
bench_50000_unflagged     ... bench:     176,401 ns/iter (+/- 22,942)
```

All entities have parents: (Worst case?)

```
bench_1000_flagged        ... bench:     117,141 ns/iter (+/- 6,358)    
bench_1000_half_flagged   ... bench:     117,152 ns/iter (+/- 5,740)    
bench_1000_third_flagged  ... bench:     117,461 ns/iter (+/- 7,064)    
bench_1000_unflagged      ... bench:     119,360 ns/iter (+/- 9,913)    
bench_5000_flagged        ... bench:     586,708 ns/iter (+/- 57,881)   
bench_5000_half_flagged   ... bench:     595,847 ns/iter (+/- 45,915)   
bench_5000_third_flagged  ... bench:     595,789 ns/iter (+/- 53,860)   
bench_5000_unflagged      ... bench:     590,739 ns/iter (+/- 40,652)   
bench_10000_flagged       ... bench:   1,234,447 ns/iter (+/- 107,232)  
bench_10000_half_flagged  ... bench:   1,239,474 ns/iter (+/- 100,179)  
bench_10000_third_flagged ... bench:   1,216,821 ns/iter (+/- 106,645)  
bench_10000_unflagged     ... bench:   1,147,391 ns/iter (+/- 89,994)   
bench_50000_flagged       ... bench:   7,287,340 ns/iter (+/- 1,114,611)
bench_50000_half_flagged  ... bench:   7,353,714 ns/iter (+/- 379,711)  
bench_50000_third_flagged ... bench:   7,810,480 ns/iter (+/- 2,125,404)
bench_50000_unflagged     ... bench:   6,162,740 ns/iter (+/- 579,224)  
```

Should solve #53.

TODO:
- [x] Remove child entity if transform's parent is Some(_), but the parent entity is not alive.
- [x] Verify that the math of children with parents is correct.
- [x] Add tests to make sure everything is working correctly.


ebkalderon:
Did you format the code with `rustfmt` before submitting? :wink:

Emilgardis:
134eb74 is the same as 65fea53, should be cherry-picked out.

src/processors/transform.rs
```diff
@@ -0,0 +1,293 @@
...
+        for entity in new {
+            init.insert(entity, Init);
+        }
+
+        // Compute transforms (global) from local transforms and parents.
+        let mut swap_candidates: Vec<(Entity, Entity)> = Vec::new();
+        for &(entity, parent) in self.sorted.iter() {
+            let mut local = locals.get_mut(entity).unwrap();
+
+            if local.dirty {
```
msiglreith: The `Transform` component should also be updated when the parent has been changed.

src/processors/transform.rs
```diff
@@ -0,0 +1,293 @@
...
+
+use ecs::{Join, Component, NullStorage, VecStorage, Entity, RunArg, Processor};
+use context::Context;
+use std::sync::{Mutex, Arc};
+use std::collections::{HashMap, VecDeque};
+
+/// Local position and rotation from parent.
+#[derive(Debug, Clone, Copy)]
+pub struct LocalTransform {
+    pub pos: [f32; 3], // translation vector
```
msiglreith: These should all be private as we have get/set.

Aceeri: Ah whoops, thanks for that catch. Forgot to take those out when I was adding the getter/setters.

src/processors/transform.rs
Aceeri: `local.dirty` is also toggled when the parent is changed.
https://github.com/amethyst/amethyst/pull/113/files#diff-43e56d65e25ce9508669140a039f7497R48

msiglreith: Sorry, I meant when the `Transform` component of the parent changes.

Aceeri: Hmm, this is actually a bit difficult of a fix. Since we could just use the parent's `LocalTransform` as a check if its dirty, but then any changes made by other systems to the global transform are ignored.

src/processors/mod.rs
```diff
@@ -1,3 +1,4 @@
 //! Default processors.
 
 pub mod rendering;
+pub mod transform;
```
kvark: missing end line?

src/processors/transform.rs
```diff
@@ -0,0 +1,293 @@
...
+}
+
+impl Default for LocalTransform {
+    fn default() -> Self {
+        LocalTransform {
+            pos: [0.0, 0.0, 0.0],
+            rot: [0.0, 0.0, 0.0, 1.0],
+            scale: [1.0, 1.0, 1.0],
+            parent: None,
+            dirty: true,
```
kvark: so you figured that having a dirty component separately is not a good way to go?

src/processors/transform.rs
```diff
@@ -0,0 +1,293 @@
...
+            indices: HashMap::new(),
+            sorted: VecDeque::new(),
+        }
+    }
+}
+
+impl Processor<Arc<Mutex<Context>>> for TransformProcessor {
+    fn run(&mut self, arg: RunArg, _: Arc<Mutex<Context>>) {
+        let (entities, mut locals, mut globals, mut init) = arg.fetch(|w| {
+            (w.entities(), w.write::<LocalTransform>(), w.write::<Transform>(), w.write::<Init>())
```
kvark: I don't think locking the `LocalTransform` for write is a good idea. I assume that's in order to change the `dirty` flag? If that so, `dirty` should just be wrapped into a `Cell`.

src/processors/transform.rs
```diff
@@ -0,0 +1,293 @@
...
+        let transform_processor = TransformProcessor::new();
+        planner.add_system::<TransformProcessor>(transform_processor, "transform_processor", 0);
+
+        b.iter(|| {
+            let config = Config::default();
+            let ctx = Arc::new(Mutex::new(Context::new(config.context_config)));
+            planner.dispatch(ctx);
+        });
+    }*/
+}
```
kvark: new line

src/processors/transform.rs
```diff
@@ -0,0 +1,293 @@
...
+            self.sorted.push_back((entity, local.parent));
+            new.push(entity.clone());
+        }
+
+        for entity in new {
+            init.insert(entity, Init);
+        }
+
+        // Compute transforms (global) from local transforms and parents.
+        let mut swap_candidates: Vec<(Entity, Entity)> = Vec::new();
```
kvark: should have it as a member of the processor, so that the Vec storage is re-used between invocations

src/processors/transform.rs
```diff
@@ -0,0 +1,293 @@
...
+
+                if let Some(global) = globals.get_mut(entity) {
+                    global.0 = combined_transform.into();
+                }
+
+                local.dirty = false;
+            }
+        }
+
+        for (entity, parent) in swap_candidates {
```
kvark: Need to take a closer look at your logic here... don't have time ATM

Aceeri: Right now `swap_candidates` contains a list of entities with parents, then later it iterates over those to check if the child is before the parent. If so then it swaps them in the sorted vec.

src/processors/transform.rs
Aceeri: I'm just not sure how it would work correctly. Should we have a vec of `LocalTransform`s/`Transform`s of the previous frame and if they are different then add a dirty component or something?

src/processors/mod.rs
Aceeri: Odd, my text editor says there are new lines in my file, I'll try re-pushing that.

src/processors/transform.rs
kvark: It looks more complicated than I'd expect

src/processors/transform.rs
Aceeri: Problem is `Component` requires `std::marker::Sync` which `Cell` doesn't have.

src/processors/transform.rs
```diff
@@ -0,0 +1,314 @@
...
+use ecs::{Join, Component, NullStorage, VecStorage, Entity, RunArg, Processor};
+use context::Context;
+use std::sync::{Mutex, Arc};
+use std::collections::HashMap;
+
+/// Local position and rotation from parent.
+#[derive(Debug, Clone)]
+pub struct LocalTransform {
+    pos: [f32; 3], // translation vector
+    rot: [f32; 4], // rotation (radians) vector
```
msiglreith: This comment is slightly confusing. Could you change it to `rotation quaternion` (or sth similiar) and specify the component layout (vec - scalar)?

src/processors/transform.rs
```diff
@@ -0,0 +1,311 @@
...
+}
+
+impl Processor<Arc<Mutex<Context>>> for TransformProcessor {
+    fn run(&mut self, arg: RunArg, _: Arc<Mutex<Context>>) {
+        let (entities, mut locals, mut globals, mut init) = arg.fetch(|w| {
+            (w.entities(), w.write::<LocalTransform>(), w.write::<Transform>(), w.write::<Init>())
+        });
+
+        for (entity, local, _) in (&entities, &locals, !&init).iter() {
+            self.indices.insert(entity, self.sorted.len());
```
kvark: When does `self.indices` ever get cleared?

src/processors/transform.rs
```diff
@@ -0,0 +1,311 @@
...
+            }
+        }
+
+        // Checks whether the child is before the parent.
+        // If so, it swaps their positions.
+        for (entity, parent) in self.swap_candidates.drain(..) {
+            let parent_index: usize = self.indices.get(&parent).unwrap().clone();
+            let index: usize = self.indices.get(&entity).unwrap().clone();
+            if parent_index > index {
+                self.sorted.swap(parent_index.clone(), index);
```
kvark: This is done after the world transform calculus. Does that mean we'd have a one frame of invalid world transforms in case such a reverse dependency is injected?

Aceeri: Yeah, but I'm not sure if adding a another loop to check for parents is a good idea considering transforms parents are only likely going to be set once, while then the loop would be executed for every frame.

src/processors/transform.rs
```diff
@@ -0,0 +1,430 @@
...
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::collections::{HashMap, HashSet};
+
+/// Local position and rotation from parent.
+#[derive(Debug)]
+pub struct LocalTransform {
+    pos: [f32; 3], // translation vector
+    rot: [f32; 4], // quaternion [w (scalar), x, y, z]
+    scale: [f32; 3], // scale vector
+    dirty: AtomicBool,
```
msiglreith: Does this need to be an atomic?

Aceeri: Yeah, Cell and Refell don't implement sync which Component requires and using an atomic let's us not lock the storage mutably.

appveyor.yml
```diff
@@ -0,0 +1,27 @@
+# Set up the build environment.
+environment:
+    matrix:
+        - TARGET: x86_64-pc-windows-msvc
+        - TARGET: i686-pc-windows-msvc
+        - TARGET: i686-pc-windows-gnu
+
+# Set up the Rust toolchain.
+install:
+    - ps: Start-FileDownload "https://static.rust-lang.org/dist/rust-nightly-${env:TARGET}.exe"
```
nchashch: It looks like appveyor build fails, because it uses nightly. Changing toolchain to stable should fix it.

Aceeri:
This should be ready to pull. I would eventually like to replace the current Parent-Child component relationship with its own storage so that other systems can use it on the sorted over with parent before children.

There does exist a *very* small edge case where if an entity is added with a reference to a parent entity that is before it in the list and then the parent is immediately deleted and leaves the child entity to be deleted the following frame instead of the current one. This is really insignificant so I don't want to do the easy path of simply moving all the sorting code into the fetch (which would then block all other systems). A secondary fix is just moving all the sorting to the storage since then it would be properly sorted and would only check to be deleted after the parent is confirmed dead.

src/processors/transform.rs
```diff
@@ -0,0 +1,498 @@
...
+            dirty: HashSet::new(),
+            swapped: HashSet::new(),
+        }
+    }
+}
+
+impl Processor<Arc<Mutex<Context>>> for TransformProcessor {
+    fn run(&mut self, arg: RunArg, _: Arc<Mutex<Context>>) {
+        // Fetch world and gets entities/components
+        let (locals, mut globals, mut init, parents) = arg.fetch(|w| {
```
nchashch: I think variable names also should be changed, so `parents` should be `children` (since you are iterating over all `Child` components), `parent.parent()` should be `child.parent()` (since you are accessing the parent of a `Child`), and so on.

PR #114, opened 2016-10-02
More functionality to InputHandler

Emilgardis:
Changes:
- Changed from HashSet to HashMap, also renamed `keys_down` to
  `pressed_keys`.
- Added fn `keys_down`. Used to query multiple keys.
- Added fn `key_once`. Used to check if a key was pressed once, and once
  only. Note this function does not care about previous calls to
  `key_down`, only other `key_once` calls.
- Added fn `keys_once`. same as `keys_down` but with `key_once` functionality.

### DIFF
--- a/src/context/src/input.rs
+++ b/src/context/src/input.rs
@@ -1,17 +1,17 @@
 //! The input handler for the game engine
 
-use std::collections::HashSet;
+use std::collections::hash_map::{Entry, HashMap};
 use event::{ElementState, EngineEvent, Event, VirtualKeyCode};
 
 pub struct InputHandler {
-    keys_down: HashSet<VirtualKeyCode>,
+    pressed_keys: HashMap<VirtualKeyCode, bool>,
 }
 
 impl InputHandler {
     /// Create a new InputHandler
     pub fn new() -> InputHandler {
         InputHandler {
-            keys_down: HashSet::new(),
+            pressed_keys: HashMap::new(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/context/src/input.rs
```diff
@@ -1,34 +1,71 @@
...
     pub fn update(&mut self, events: &[EngineEvent]) {
         for event in events {
             match event.payload {
-                Event::KeyboardInput(ElementState::Pressed, _, Some(key_code)) => { self.keys_down.insert(key_code); }
-                Event::KeyboardInput(ElementState::Released, _, Some(key_code)) => { self.keys_down.remove(&key_code); }
-                Event::Focused(false) => self.keys_down.clear(),
+                Event::KeyboardInput(ElementState::Pressed, _, Some(key_code)) => {
+                    match self.pressed_keys.entry(key_code) {
+                        Entry::Occupied(mut entry) => {
+                            entry.insert(false);
```
msiglreith: I think this should actually be a no-op. If there is a consecutive `Pressed` event (e,g. windows' repeat events) but `key_once` hasn't been called yet, the entry should stay as `true`.

src/context/src/input.rs
```diff
@@ -1,34 +1,71 @@
 //! The input handler for the game engine
 
-use std::collections::HashSet;
+use std::collections::hash_map::{Entry, HashMap};
 use event::{ElementState, EngineEvent, Event, VirtualKeyCode};
 
 pub struct InputHandler {
-    keys_down: HashSet<VirtualKeyCode>,
+    pressed_keys: HashMap<VirtualKeyCode, bool>,
```
msiglreith: (Minor: I would prefer an enum over `bool` here)

src/context/src/input.rs
Emilgardis: Seems good to me. I'll change it.

Emilgardis:
I think there could be an array of size TOTAL_KEYS, and then map `i` (`KeyboardInput(ElementState::Pressed,i,_)`) as an enum, ie
```rust
// Needs better name(s)
enum KeyPressedState { NotPressed, Pressed, Queried }
type KeyMap = [KeyPressedState; TOTAL_KEYS]
// Do stuff
```

However, I'm not sure what the u8 in KeyboardInput actually is.

ebkalderon:
@Emilgardis Hey, what's your status? You still working on this?

Emilgardis:
Yes.

Emilgardis:
So, I've been thinking, should I switch to use a `BTreeMap` and use the scancode, the `u8` in the `KeyBoardInput`?

ebkalderon:
@Emilgardis I do not think that using a `BTreeMap` is the best idea since it is optimized for ordered data and cases where one needs to find the greatest/least values of an overall set, which I don't think we really need here.

Emilgardis:
@ebkalderon fair enough, didn't think of that.

msiglreith:
Something like this should work
```Rust
enum KeyState { Released, Pressed, Queried }
type KeyMap = [KeyState; 0xFF]; // 255 should be enough, setting the `repr(u8)` attribute for the `VirtualKeycode` enum would prevent overflows
...
Event::KeyboardInput(ElementState::Pressed, _, Some(key_code)) => { self.keys_down[key_code as usize] = Pressed; }
..
```

msiglreith:
Can we move the array mapping of the keycodes into a separate issue to unblock the PR? It's not totally related to this PR and everything seems to be good (approved) beside that.

PR #115, opened 2016-10-08
[WIP] Add FXAA anti-aliasing pass to amethyst_renderer

nchashch:
Changes:
- Add support for render targets (`AssetManager::get_color_buffer` and `AssetManager::get_geometry_buffer` methods are used to create them)
- Add `Null` filter pass and `FXAA` filter pass

Problems:
When rendering with a filter pass (`Null` or `FXAA`) there is color bending, like this:
`FXAA` filter pass:
sphere_color_bending

No `FXAA` filter pass:
sphere

### DIFF
--- a/src/context/src/asset_manager.rs
+++ b/src/context/src/asset_manager.rs
@@ -176,6 +176,50 @@ impl AssetManager {
             None => None,
         }
     }
+    /// Create and return a color buffer
+    pub fn get_color_buffer(&mut self, (width, height): (u16, u16)) -> Option<ColorBuffer> {
+        match self.factory_impl {
+            FactoryImpl::OpenGL {
+                ref mut factory,
+            } => {
+                let color_buffer = Box::new(amethyst_renderer::target::ColorBuffer::new(factory, (width, height)));
+                let color_buffer_impl = ColorBufferImpl::OpenGL {
+                    color_buffer: color_buffer,
+                };
+                let color_buffer = ColorBuffer {
+                    color_buffer_impl: color_buffer_impl,
+                };
+                Some(color_buffer)
+            },
+            #[cfg(windows)]
+            FactoryImpl::Direct3D => {
+                unimplemented!();
+            },
+            FactoryImpl::Null => None,
+        }
+    }
+    /// Create and return a geometry buffer
+    pub fn get_geometry_buffer(&mut self, (width, height): (u16, u16)) -> Option<GeometryBuffer> {
+        match self.factory_impl {
+            FactoryImpl::OpenGL {
+                ref mut factory,
+            } => {
+                let geometry_buffer = Box::new(amethyst_renderer::target::GeometryBuffer::new(factory, (width, height)));
+                let geometry_buffer_impl = GeometryBufferImpl::OpenGL {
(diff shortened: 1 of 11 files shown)


src/renderer/src/pass/filters.rs
```diff
@@ -0,0 +1,267 @@
...
+
+    void main() {
+        // o_Color = texture(t_Source, v_TexCoord);
+        float FXAA_SPAN_MAX = 8.0;
+        float FXAA_REDUCE_MIN = 1.0/128.0;
+        float FXAA_REDUCE_MUL = 1.0/8.0;
+
+        vec3 luma = vec3(0.2126, 0.7152, 0.0722);
+        float lumaM = dot(texture(t_Source, v_TexCoord).xyz, luma);
+        float lumaTL = dot(texture(t_Source, v_TexCoord + (vec2(-1.0, -1.0) * u_InverseTextureSize)).xyz, luma);
```
kvark: If you use `textureOffset` instead, you'd not need any `u_InverseTextureSize` at all, and it would even be faster

src/context/src/asset_manager.rs
```diff
@@ -176,6 +176,50 @@ impl AssetManager {
...
+            },
+            #[cfg(windows)]
+            FactoryImpl::Direct3D => {
+                unimplemented!();
+            },
+            FactoryImpl::Null => None,
+        }
+    }
+    /// Create and return a geometry buffer
+    pub fn get_geometry_buffer(&mut self, (width, height): (u16, u16)) -> Option<GeometryBuffer> {
```
kvark: I know this stuff was here before this PR, but... what's the difference between `GeometryBuffer` and `ColorBuffer`?

src/context/src/asset_manager.rs
```diff
@@ -272,3 +316,33 @@ pub enum TextureImpl {
 pub struct Texture {
     texture_impl: TextureImpl,
 }
+
+pub enum ColorBufferImpl {
```
kvark: Ouch, it sucks to have this kind of enums multiplying. I was hoping for it to be localized and well hidden from anything. It would be great to solve this, but I don't see how...

src/renderer/src/pass/filters.rs
nchashch: I replaced `texture` with `textureOffset` for `lumaTL`, `lumaTR`, `lumaBL`, and `lumaBR`. But since `textureOffset` can only take constant offset `result1` and `result2` can be computed only by using `texture`. So `u_InverseTextureSize` is still required to compute `result1` and `result2`.

src/context/src/asset_manager.rs
ebkalderon: Do you think it might be possible to use trait objects for this somehow?

ebkalderon:
@nchashch Is this still being worked on? It appears to be growing stale.

nchashch:
Rendering hookup code was changed significantly (context rewrite), so I am closing this. I guess I will get back to this eventually and salvage this code, it would be better to improve renderer design first and then add more features.

ebkalderon:
Sounds good! I really love this addition, it's just a shame it didn't get merged earlier.

Issue #28, opened 2016-03-13
Relicense under dual-license Apache2 and MIT


thiolliere:
as you may know rust libraries are mostly dual-licensed in apache-MIT

The Apache license has protections from patent trolls and an explicit contribution licensing clause.
And the MIT is for GPLv2 compatibility (because Apache is not compatible with GPLv2)

[here](https://github.com/l1048576/fbx_direct/issues/1) a message from a crate.io bot for explanation of how to set dual-license in your project.

feel free to close it if your intension is to keep MIT alone.

Anyway great project, keep going :-)

ebkalderon:
@thiolliere Thanks for pointing this out! I didn't really think about patent trolls and explicit contribution licensing when I first started the project, since I wasn't really sure if people were interested in helping or not. This seems like a good idea to me. Hopefully everyone else agrees.

# Contributor checkoff

To agree to relicensing, comment with :

```
I license past and future contributions under the dual MIT/Apache-2.0 license, allowing licensees to chose either at their option.
```

Or, if you're a contributor (looks like GitHub only shows contributors to `master`), you can check the box in this repo next to your name.

- [ ] @cooperra 
- [x] @ebkalderon
- [ ] @Oflor

ebkalderon:
The source code is changing at quite a rate, and we need everyone on this list to chime in whether you support this transition. This has been sitting here for far too long now, we should arrive at a consensus ASAP.

Aceeri:
I license past and future contributions under the dual MIT/Apache-2.0 license, allowing licensees to chose either at their option.

nchashch:
I license past and future contributions under the dual MIT/Apache-2.0 license, allowing licensees to chose either at their option.

csherratt:
I license past and future contributions under the dual MIT/Apache-2.0 license, allowing licensees to chose either at their option.

cooperra:
I license past and future contributions under the dual MIT/Apache-2.0 license, allowing licensees to chose either at their option.

ebkalderon:
@Emilgardis @lschmierer @Oflor @palodequeso @White-Oak Please respond to this thread whether you support the licensing change or not. As contributors to this repository, your input is very valuable, and this thread cannot be closed until you chime in. Thanks!

White-Oak:
My contribution to amethyst was pretty minor and I hope i can contribute more :), but, nevertheless, I checked myself in in that list.

lschmierer:
[...]

PR #126, opened 2016-10-28
Dual-license Apache and MIT

Aceeri:
#28


Emilgardis:
https://ci.appveyor.com/project/amethyst/amethyst/build/1.0.22/job/6p21sjce0bm7xm3b#L236, seems like we have to wait for stable to be fixed. See gfx-rs/gfx#1070

@kvark where are we on this matter?

kvark:
@Emilgardis I've yanked the version that was failing on stable, now it's just the nightly that keeps failing. Haven't looked into a proper solution yet, got busy. Any help is appreciated!

PR #128, opened 2016-11-16
Fix appveyor CI

nchashch:
Changes:
* Add stable and beta channels to appveyor.yml
* Add x86_64-pc-windows-gnu target to appveyor.yml
* Allow failure for nightly channel build for appveyor CI

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -34,7 +34,7 @@ matrix:
 
 # Set up the Rust toolchain.
 install:
-    - curl -fsS "https://win.rustup.rs/" -o rust-init.exe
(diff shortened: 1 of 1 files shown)


selaux:
What about using rustup?
```
-cmd: rustup-init.exe -y --default-toolchain ${env:channel}-${env:target}
```

Aceeri:
I think we might be able to just update gfx to 0.13.0 once it is published since that seems to fix any problems with `use $crate`.

Aceeri:
Solved by updating the dependencies https://github.com/amethyst/amethyst/pull/136.

kvark:
Note: it's a good habit to squash small changes like this. Helps bisection of any problems later on.

PR #129, opened 2016-11-22
Added basic mouse functionality to InputHandler.

bytebuddha:
I tried to stick to the current format of InputHandler, if their are any changes i can make let me know.

Changes: 
- Added HashSet to keep track of button press's, and tuple to keep track of 
  pointer location.
- Added fn button_down to check whether a button is currently pressed
  and fn pointer_location to get mouse pointer location.

### DIFF
--- a/src/context/src/input.rs
+++ b/src/context/src/input.rs
@@ -1,17 +1,23 @@
 //! The input handler for the game engine
 
 use std::collections::HashSet;
-use event::{ElementState, EngineEvent, Event, VirtualKeyCode};
+use event::{ElementState, EngineEvent, Event, VirtualKeyCode, MouseButton };
 
 pub struct InputHandler {
+    pub focus: bool,
     keys_down: HashSet<VirtualKeyCode>,
+    buttons_down: HashSet<MouseButton>,
+    pointer_location: (i32,i32)
 }
 
 impl InputHandler {
     /// Create a new InputHandler
     pub fn new() -> InputHandler {
         InputHandler {
+            focus: false,
             keys_down: HashSet::new(),
+            buttons_down: HashSet::new(),
+            pointer_location: (0,0)
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/context/src/input.rs
```diff
@@ -31,4 +38,14 @@ impl InputHandler {
     pub fn key_down(&self, key: VirtualKeyCode) -> bool {
         self.keys_down.contains(&key)
     }
-}
\ No newline at end of file
+
+    ///Check if `button` is pressed.
+    pub fn button_down(&self, button: MouseButton) -> bool {
```
Emilgardis: Should this be called `button_down`? It is obvious from the type `button` has, but I feel like button_down is a suboptimal name.

src/context/src/input.rs
```diff
@@ -1,17 +1,21 @@
 //! The input handler for the game engine
 
 use std::collections::HashSet;
-use event::{ElementState, EngineEvent, Event, VirtualKeyCode};
+use event::{ElementState, EngineEvent, Event, VirtualKeyCode, MouseButton };
 
 pub struct InputHandler {
     keys_down: HashSet<VirtualKeyCode>,
+    buttons_down: HashSet<MouseButton>,
+    pointer_location: (i32,i32)
```
bytebuddha: Should pointer_location use a [i32 ; 2] instead of a tuple?

src/context/src/input.rs
bytebuddha: How would 'clicked'?

Emilgardis: I'm not sure. clicked implies that the button wasn't clicked in the previous logic frame, fn `mouse_key_down` would work.

Maybe we should rethink fn `key_down`. It could handle keyboard and mouse buttons. That would mean that it would take something like `Either<T>` where Either<T>::get() returns Self::Type. I'm not certain though, generics are hard to do right.
I think we can start with `fn mouse_key_down` and `fn keyboard_key_down` and then extend it later.

bytebuddha:
i agree mouse_key_down is better, but keyboard_key_down seems a little clunky so i left it at key_down. I decided to add a focus field: bool, and not turn pointer location into an Option, because this way it would still be possible to find the last pointer location even if the window is unfocused.

src/context/src/input.rs
```diff
@@ -40,7 +47,7 @@ impl InputHandler {
     }
 
     ///Check if `button` is pressed.
```
Emilgardis: Formatting: space after ``///` in docs.

bytebuddha:
I think i'm going to close this pull request and work on it some more with some changes now that #114 is merged.

PR #131, opened 2016-11-28
Made api reference easier to access

Emilgardis:
Linking to doc/ because the default branch is not master anymore.

I'm not sure of formatting, I was thinking of adding a docs.rs badge badge but since that one isn't really accurate anymore I think that can be done another time.


Aceeri:
Would you mind rebasing/formatting? Builds for failing due to stuff outside of this PR.

Issue #116, opened 2016-10-11
Renderer tests sometimes fail


ebkalderon:
Navigating directly into the `src/renderer/` directory and calling `cargo test` works fine, but for whatever reason, staying in the top-level directory of the repository and calling `cargo test -p amethyst_renderer` or `cargo local-pkgs test` fails from Cargo not finding the `genmesh` crate.

yeliknewo:
It seems to be because dev dependencies aren't propagated even when testing. Should we add a feature for testing and use that instead of dev dependencies?

Emilgardis:
This is a known issue, however it should have been fixed, see rust-lang/cargo#860

sekhat:
I've added a pull request that should fix this issue. From my understanding of calling from a higher level cargo project, that project needs to be a workspace, and that at least needs to be pointed out explicitly, by having a [workspace] entry in the top level projects Cargo.toml.

The members of the workspace can be worked out implicitly. Mostly by path dependencies, and/or walking up the tree to a Cargo.toml with a [workspace].

I could be wrong on this, but adding the [workspace] to the Cargo.toml seems to have solved the issue.

sekhat:
@ebkalderon Did the workspace stuff not fix this issue?

Xaeroxe:
Issue no longer occurring on latest master, I'm closing this.

PR #132, opened 2016-12-01
add [workspace] to top level Cargo.toml

sekhat:
This signals to cargo that the top level Cargo.toml project should
be treated as a workspace.

The projects added as dependencies are implicit members of the
workspace.

This ends up fixing issue #116 by causing dev-dependencies to be
downloaded into the projects shared target folder.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,3 +58,5 @@ gfx = "0.11"
 
 [dependencies]
 cgmath="0.11"
+
+[workspace]


sekhat:
Nightly and Beta seem to be failing for but as far as I can tell, with issues unrelated to this change. Is that the case, or am I just be silly?

Aceeri:
It isn't you, we are still trying to fix this on our end beca use of gfx failing with changes to $crate.

Aceeri:
Do you know if this allows you to import crates without reimporting through Cargo.toml in the subcrates? As in only needing one import of a crate in the workspace to be used in all subcrates. Take src/context/ for example.

sekhat:
As far as I'm aware, each crate needs to add it's own dependencies in it's cargo.toml, but, if you `cargo build` from the top level, the crate is only downloaded and compiled once for all the sub-crates.

PR #133, opened 2016-12-01
Context rewrite

nchashch:
# Goals:
- Get rid of `Arc<Mutex<Context>>` context parameter, which is passed to all `Processor`s, and which must be locked in order to be accessed (so it is a problem for concurrency).
- Get rid of multiple levels of indirection in `RenderingProcessor` (along with `RenderingProcessor`), `Renderable`s now instead of refering to `Texture`s, `Mesh`es, and `Fragment`s inside of `AssetManager`/`amethyst_renderer::Frame` hold actual `Texture`s and `Mesh`es. `Fragment`s are recreated at every iteration of the main loop from `Renderable`, `LocalTransform` pairs.

# Design:

## `Application::advance_frame`
In `Application::advance_frame` first `StateMachine::handle_events`, `StateMachine::fixed_update`, `StateMachine::update` methods are called. Then `Planner::dispatch` and `Planner::wait` methods are calld (here `Processor`s run in parallel). Then `GfxDevice::render_world` is called, this method fulfills the same purpuse as `RenderingProcessor` did.

## `GfxDevice::render_world(&mut self, world: &World, pipeline: &Pipeline)`
In this method `Renderable`s and `LocalTransform`s are read from the passed `&World`, and `Scene` (first part of `amethyst_renderer` job submission) is constructed. Then the constructed `Scene` along with the passed `Pipeline` (second part of `amethyst_renderer` job submission) is passed to `amethyst_renderer::Renderer::submit` method.

## `Camera`
`Camera` now is a resource in the `World`.

## `Renderable`
`Renderable` component now consists of two `Texture`s (which hold actual `amethyst_renderer::Texture`s, not IDs) and one `Mesh` (which holds an actual `gfx::handle::Buffer`, `gfx::Slice` pair)

## `State` methods
`State` methods now accept as parameters: `&mut World`, `&mut AssetManager`, `&mut Pipeline`. Resources are loaded/unloaded and `Pipeline` is set in these state methods. `Processor`s can't access `AssetManager` and `Pipeline` directly (currently?).

## `AssetManager`
- `GfxLoader` (which holds `Factory`) is consumed by `AssetManager::add_loader`.
- ~~`Assets::load_asset_from_data` now returns an actual asset wrapped in `Option` instead of `AssetID`~~.
- Add `AssetManager::create_renderable` method, which takes loaded mesh name, ka texture name, kd texture name and returns an `Option<Renderable>`

# Changes:
- Add `GfxDevice` struct, containing `amethyst::Renderer`, `gfx::Device`, and `Window`
- Add `GfxLoader` struct, containing `Factory`, which is consumed by `AssetManager::add_loader`
- Add `MainTarget` struct, containing `main_color` and `main_depth` render targets, returned by `gfx_device_gl::init` function
- Move functionality from `VideoContext` and `context::Renderer` to `GfxDevice`
- Add `GfxDevice::render_world` method, which takes a `&specs::World` and renders it
- Split `amethyst_renderer::Frame` into `Scene` and `Pipeline` structs
- Now `Application` owns `GfxDevice` and `Pipeline`

# Todo:

## Design
- [x] - ~~Figure out how to properly manage `main_color`, `main_depth` render targets~~ <- Should be fixed in another PR

## Examples
- [x] - Rewrite 00_hello_world example
- [x] - Rewrite 01_window example
- [x] - Rewrite 02_sphere example
- [x] - Rewrite 03_renderable example
- [x] - Rewrite 04_pong example
- [x] - Rewrite 06_assets example

## Misc
- [x] - Polish the code
- [x] - Add comments/doc-comments


Aceeri:
> `Renderable` component now consists of two `Texture`s (which hold actual `amethyst_renderer::Texture`s, not IDs) and one `Mesh` (which holds an actual `gfx::handle::Buffer`, `gfx::Slice` pair)

Are we sure we don't want it to hold IDs? A lot of the OpenGL stuff in gfx doesn't implement `Send`/`Sync` anymore due to OpenGL being single threaded, so I'm not sure storing them inside a component will work if we want to update to 0.13.

nchashch:
AFAIK `gfx::handle::ShaderResourceView` and `gfx::handle::Buffer` just hold `GLuint`s from `glGenTextures` or `glGenBuffers`. And since they just hold integers, merely refering to actual textures/buffers somewhere inside GL context, it should be safe and cheap to clone them and to pass them betwen threads.

examples/02_sphere/main.rs
```diff
@@ -1,45 +1,64 @@
...
+        let green = asset_manager.load_asset_from_data::<Texture, [f32; 4]>([0.0, 1.0, 0.0, 1.0]).unwrap();
+        let sphere = Renderable {
+            ka: dark_blue,
+            kd: green,
+            mesh: sphere_mesh,
+        };
+        let transform = LocalTransform::default();
+        world.create_now()
+            .with::<Renderable>(sphere)
+            .with::<LocalTransform>(transform)
```
Aceeri: I think this needs to also add a `.with::<Transform>(Transform::default())` since the `LocalTransform` only works by putting the calculated position inside of the `Transform` component.

src/gfx_device/gfx_device.rs
```diff
@@ -0,0 +1,134 @@
...
+                        };
+                    let ka = match renderable.ka.texture_inner {
+                        TextureInner::OpenGL { ref texture } => texture.clone(),
+                        _ => continue,
+                    };
+                    let kd = match renderable.kd.texture_inner {
+                        TextureInner::OpenGL { ref texture } => texture.clone(),
+                        _ => continue,
+                    };
+                    let transform = local_transform.matrix();
```
Aceeri: These should just use the `Transform` component instead of the `LocalTransform` as well. Then there is no need for recalculation of the matrix every frame if isn't changed and there is also the possibility of static lights by just adding a `Transform` without a `LocalTransform`.

nchashch:
I think this PR is ready now. Probably will merge after another review.

Cargo.toml
```diff
@@ -58,5 +54,8 @@ gfx = "0.12"
 
 [dependencies]
 cgmath="0.11"
-
-[workspace]
+glutin = "0.6"
+gfx = "0.12"
+gfx_device_gl = "0.11"
+gfx_window_glutin = "0.12"
+genmesh="0.4"
```
msiglreith: Nitpick: Add newline

examples/04_pong/main.rs
```diff
@@ -127,19 +133,19 @@ impl Processor<Arc<Mutex<Context>>> for PongProcessor {
...
                     // If `S` is pressed and plank is in screen boundaries then move down
-                    if ctx.input_handler.key_down(VirtualKeyCode::S) {
+                    if input_handler.key_down(VirtualKeyCode::S) {
                         if plank.position - plank.dimensions[1]/2. > -1. {
                             plank.position -= plank.velocity * delta_time;
                         }
                     }
                     // Set translation[0] of renderable corresponding to this plank
-                    local.translation[0] = left_boundary + plank.dimensions[0]/2.0
+                    local.set_translation_index(0 , left_boundary + plank.dimensions[0]/2.0);
```
msiglreith: Is there a reason to switch back to the `set_index(index, ...)` style, or is it just conflicting with https://github.com/amethyst/amethyst/pull/127?

nchashch: `LocalTransform` fields are now private and  `LocalTransform::set_translation_index` aside from changing `translation` field also calls `LocalTransform::flag` method (which sets dirty flag).

Aceeri: I actually got it to work through the use of DerefMut so you can use it exactly how you would modify the inner wrapped part. Eventually this will be changed to use a TrackedStorage instead so there is no deref abuse and you can use the components without getters and setters.

kvark:
Ok, I think it's some very impressive work here by @nchashch , and the PR got quite big.
I don't want to delay it any longer, any subsequent work can follow separately.

PR #134, opened 2016-12-03
[WIP] Update gfx to master branch and use IDs for textures and meshes

msiglreith:
Gpu resource creation is not possible in multiple threads for some gfx backends like OpenGL. These need to be allocated on the render thread, requiring to refactor the asset manager.
Now we establish a channel between the asset manager and the renderer to send resource creation commands. Depending on the backend the resource creation will be on done on the render thread or on loading the asset.

TODO:
- [x] Upstream `specs` changes to master or fork (`allocate` as pub)
- [x] Use newly publish `gfx` 0.12.2 Edit: Using master branch as 0.12.2 depends `gfx_core 0.4` (not 0.5) 
- [ ] Add documentation
- [ ] Imrove error handling at some points
- [ ] Implement texture loading

### DIFF
--- a/src/context/Cargo.toml
+++ b/src/context/Cargo.toml
@@ -24,11 +24,11 @@ path = "../ecs/"
 version = "0.1.0"
 
 [dependencies]
-gfx = "0.11"
-gfx_core = "0.3"
-gfx_device_gl = "0.10"
-gfx_window_glutin = "0.11"
-glutin = "0.5"
+gfx = { git = "https://github.com/gfx-rs/gfx" }
+gfx_core = { git = "https://github.com/gfx-rs/gfx" }
+gfx_device_gl = { git = "https://github.com/gfx-rs/gfx" }
+gfx_window_glutin = { git = "https://github.com/gfx-rs/gfx" }
+glutin = "0.7"
 time = "0.1"
 genmesh="0.4"
 cgmath="0.11"
--- a/src/context/Cargo.toml
+++ b/src/context/Cargo.toml
@@ -24,10 +24,10 @@ path = "../ecs/"
 version = "0.1.0"
 
 [dependencies]
-gfx = { git = "https://github.com/gfx-rs/gfx" }
-gfx_core = { git = "https://github.com/gfx-rs/gfx" }
-gfx_device_gl = { git = "https://github.com/gfx-rs/gfx" }
-gfx_window_glutin = { git = "https://github.com/gfx-rs/gfx" }
+gfx = { git = "https://github.com/gfx-rs/gfx", rev = "4cc8d52"}
+gfx_core = { git = "https://github.com/gfx-rs/gfx", rev = "4cc8d52"}
+gfx_device_gl = { git = "https://github.com/gfx-rs/gfx", rev = "4cc8d52"}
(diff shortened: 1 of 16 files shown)


src/context/Cargo.toml
```diff
@@ -24,11 +24,11 @@ path = "../ecs/"
...
 [dependencies]
-gfx = "0.11"
-gfx_core = "0.3"
-gfx_device_gl = "0.10"
-gfx_window_glutin = "0.11"
-glutin = "0.5"
+gfx = { git = "https://github.com/gfx-rs/gfx" }
+gfx_core = { git = "https://github.com/gfx-rs/gfx" }
+gfx_device_gl = { git = "https://github.com/gfx-rs/gfx" }
+gfx_window_glutin = { git = "https://github.com/gfx-rs/gfx" }
```
Aceeri: May want to set rev to the current commit so it doesn't break once gfx is updated.

Aceeri:
I don't think we need to go in the direction of IDs anymore for textures and meshes and stuff since `handle::Buffer`, `gfx::Slice`, and `gfx::ShaderResourceView` are all technically sendable/syncable (just needs to be unsafely implemented because the compiler doesn't understand phantom data. May be worth waiting for the context rewrite PR and then updating the gfx version.

src/context/Cargo.toml
msiglreith: Thanks, I wanted to use the newly published 0.12.2 but I noticed that it depends on the outdated 0.4 version of `gfx_core`.

src/context/src/asset_manager.rs
```diff
@@ -221,35 +220,21 @@ impl DerefMut for AssetManager {
...
-    Direct3D {
-        // stub
-    },
-    Null,
-}
-
 impl AssetLoader<Mesh> for Vec<VertexPosNormal> {
     /// # Panics
     /// Panics if factory isn't registered as loader.
     fn from_data(assets: &mut Assets, data: Vec<VertexPosNormal>) -> Option<Mesh> {
```
Aceeri: I'm not sure if the `AssetLoader` trait was added in this or the original context, but I feel that it should return a `Result<T, LoaderError>` or something similar. Given that data is unreliable (corrupt, parsed wrongly, etc.).

Emilgardis:
#136 fixed gfx dep. Do [this](http://stackoverflow.com/questions/1186535/how-to-modify-a-specified-commit-in-git) and squash the two commits. Or maybe make a new pr only containing the changes for IDs.

Aceeri:
IIRC #136 only updated to 0.12.2 for gfx, while this uses a recent revision that is 0.13.0 which was released recently. So it is still worth having on the PR.

Emilgardis:
Yes you are correct, new version (0.13.0) landed. Does this pr need master or is 0.13.0 fine?

msiglreith:
0.13.0 should be fine, I'm gonna look into finishing it in the next days

Aceeri:
We might want to wait until 0.14.0, since that is supposedly landing soon.

kvark:
@Aceeri apparently, we got stacked up with plans for 0.14 now, so it's going to be in development for at least a few weeks from now.

msiglreith:
Hm, rebasing turns out to be really tricky after the context rewrite. Going to close this and make a new PR. Also trying to split it up abit.

PR #136, opened 2016-12-09
Bump crate deps

vvuk:
I couldn't build as-is with nightly rustc; issues with `use $crate;` and some other problems.  Bumping the dep versions fixes things; everything's at the latest except for `glutin` -- `gfx_window_glutin` has not been updated yet for glutin 0.7.


Aceeri:
I think this is already being fixed by https://github.com/amethyst/amethyst/pull/134?

Emilgardis:
Yes, but it is being blocked by being incorporated with the other things that that pr aims to do. I think we should merge this and rebase it onto #134.
cc @msiglreith

msiglreith:
We should merge this as it seems to compile and allow people to keep on working without issues. I don't know when I have time to finish #134 atm.

PR #137, opened 2016-12-12
Fix texture coordinates in gen_rectangle

jFransham:
I noticed this when adding texture support in a fork of amethyst that I'm building a little test game on. The texture coordinates are wrong and cause the texture to be split down the diagonal. None of the examples show this off since they all use flat textures. This commit fixes that issue.


Emilgardis:
f59672d is the same as the one in #140. This PR shouldn't include it.

jFransham:
Oh sorry, I PR'd from my `develop` branch instead of from a feature branch. I'll fix that now.

PR #140, opened 2016-12-15
Mark children as dirty when their parents are dirty

jFransham:
The `Transform` processor currently doesn't update the position of children when the roots (i.e. objects without parents) are updated. This commit fixes this. This is working but I haven't written tests for it, I don't want to put that effort in if it turns out the current behaviour is desired (or if it turns out that locking the entities iterator for the whole body of the `Transform` processor is unacceptable).

### DIFF
--- a/src/processors/transform.rs
+++ b/src/processors/transform.rs
@@ -221,7 +221,7 @@ impl TransformProcessor {
 impl Processor<Arc<Mutex<Context>>> for TransformProcessor {
     fn run(&mut self, arg: RunArg, _: Arc<Mutex<Context>>) {
         // Fetch world and gets entities/components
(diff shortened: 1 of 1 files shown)


Aceeri:
I think this can just be fixed by moving the local.flag() to the end of the processor.

jFransham:
I don't think I follow, that seems to be what I'm doing here already.

Aceeri:
Oh it just looked like you were putting another loop inside the other one from the diff. Why'd you put the new iterator stuff in a variable?

jFransham:
Just a style thing, I prefer to pull the argument out when it's getting long. I can change that though no problem.

PR #141, opened 2016-12-16
Fix texture coordinates in gen_rectangle

jFransham:
Original PR: #137

I noticed this when adding texture support in a fork of amethyst that I'm building a little test game on. The texture coordinates are wrong and cause the texture to be split down the diagonal. None of the examples show this off since they all use flat textures. This commit fixes that issue.

### DIFF
--- a/src/context/src/asset_manager.rs
+++ b/src/context/src/asset_manager.rs
@@ -323,25 +323,25 @@ impl AssetManager {
     }
     /// Generate and load a rectangle mesh in XY plane with given `width` and `height`.
     pub fn gen_rectangle(&mut self, name: &str, width: f32, height: f32) {
-        let norm = [0., 0., 1.];
+        const NORM: [f32; 3] = [0., 0., 1.];
         let t_left = VertexPosNormal {
             pos: [-width/2., -height/2., 0.],
-            normal: norm.clone(),
+            normal: NORM,
             tex_coord: [0., 0.],
         };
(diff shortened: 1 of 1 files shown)


kvark:
Thanks @jFransham ! Please rebase.

jFransham:
@ebkalderon The function that this is a fix for seems to no longer exist on `develop`

ebkalderon:
It still exists, it's just been moved to [line 358 of our Pong example](https://github.com/amethyst/amethyst/blob/develop/examples/04_pong/main.rs#L358). 😄

jFransham:
Yeah, I noticed that it was in an example. Sorry, I should say that it's not exposed in application code, so does it matter? I'd be happy for this PR to be closed, since the texture coords don't need to be correct in pong (everything's solid-colour)

Emilgardis:
I think a fix is still needed,as the examples are supposed to give an
outline how to model your code, and habing something that looks correct but
doesn't actually do what it is supposed to do can be hurtful to the
end-user.

ebkalderon:
@jFransham Regardless of whether the bug is present in the core library or in a code example, your fix is just as welcome! ❤️ Just rebase on the develop branch so we can merge it.

nchashch:
`AssetManager::gen_rectangle` method was removed, so this PR is out of date. Nevertheless thank you for your effort. Closing it.

Issue #60, opened 2016-05-14
Default Amethyst Processors and Components for ECS


LucioFranco:
This issue is to keep track of all the systems that will need to be implemented for amethyst engine.
### Systems
- [x] Transform | Issue: (https://github.com/amethyst/amethyst/issues/53), PR: (https://github.com/amethyst/amethyst/pull/113)
- [x] Rendering | PR: (https://github.com/amethyst/amethyst/pull/85)
- [ ] Scripting
- [x] Audio
### Components
- [x] Transform
- [ ] Velocity?
- [x] Renderable

Feel free to leave a comment and/or create a PR. Let us know if you are working on one by leaving a comment.

JDoneth:
Should networking be a target? 

For example: Some sort of automatic serialization and transmission of structs over the std library networking functions.

ebkalderon:
@JDoneth That is correct, but off-topic. Networking is an engine-level feature and not an ECS (entity-component-system) feature. This issue deals with implementing [systems in the ECS sense](https://www.amethyst.rs/book/glossary.html#Entity-component-system%20%28ECS%29%20model). Personally, I prefer the term "processors" to eliminate confusion.

JDoneth:
Ah sorry, my bad, I saw systems and thought engine layers.

ahmedaliadeel:
Position and velocity shouldnt be separate components. If u use 3rd party physics like nphysics or bullet bindings u get rigidbody access. We need to provide a rigidbody component.

Emilgardis:
I think that they should be separated, I don't see any reason why they shouldn't. Having them together means more work for the engine, i.e computing if a object is static (not moving) or dynamic (moving).

ahmedaliadeel:
I haven't explorered the nphysics yet but based on my experience on unity 3d i was suggesting this idea. 
Look, its not just velocity or angular velocity there is much more about a physical body. Like i want to know velocity at certain point? 
Let me explain in detail for a different perspective. I think we should have separate components and respective subsytem for both 2d and 3d. 
Talking about the granulity of components incase of 3d physics we should provide components the way most physics engine api is natural to map. Physix, bullet both have to concept of rigidbodies, colliders etc 
For kinemetic physics we can just add a property to rigidbody component that this object is not dynamic. (As of know nphysics doest support kinemetic bodies) or if its an unmovable static body etc.
Now that i have discovered this project i will also try to implement a reference set of components and subsystem.
Thanks for consideration.

[...]

PR #144, opened 2016-12-18
Adding nphysics integration

mechaxl:


Cargo.toml
```diff
@@ -52,11 +52,19 @@ path = "examples/04_pong/main.rs"
...
+name = "physics"
+path = "examples/07_physics/main.rs"
+
 [dev-dependencies]
 obj-rs = "0.4.16"
 gfx = "0.12"
 
 [dependencies]
 cgmath="0.11"
+nalgebra = "0.9.*"
```
mechaxl: I think this pulls in the physics demo dependencies for everything else as well, which isn't great, but I'm not sure if there's a better way of doing it.

Emilgardis: use `[dev-dependencies]` instead, should work fine :)

examples/07_physics/main.rs
```diff
@@ -0,0 +1,276 @@
...
+use ncollide::shape::{Ball as PhysicsBall, Cuboid};
+use nphysics3d::object::{RigidBody, RigidBodyHandle};
+use nphysics3d::world as nphysics_world;
+
+
+struct PhysicsExample;
+
+// Wrap the nphysics3d `World` struct in our own anonymous struct
+// so that we can impl `Send`/`Sync` for it.
+struct PhysicsWorld(nphysics_world::World<f32>);
```
mechaxl: Wondering if there's any interest in re-exporting this from within Amethyst, to avoid the `Send`/`Sync` and `Deref` hassle. Does it make sense to add it as a built-in `Processor`, to achieve something like `use amethyst::processors::physics::World;`? At any rate, probably something for another PR.

Emilgardis: We've actually had some talk about nphysics and nalgebra. We definitely want to make it easy to use with amethyst. See #60

msiglreith:
Thanks for PR, looks fine to me, but haven't tested it so far!
Out of interest, does nphysics support to dynamically resize the shape of collider, as we currently have a size parameter in the transform component?

mechaxl:
@msiglreith: I don't believe so, unfortunately. I opened up a ticket here to track that issue:

https://github.com/sebcrozet/ncollide/issues/139

Emilgardis:
This is no longer only adding an integration example. Change the title :)

Also, this looks promising :+1:

jFransham:
I also had problems with having to use `unsafe impl Sync`, is there a better way to do that? Obviously it's fine as long as the `Rc`s don't escape but it'd be better to statically enforce that, I'd say. I've forked `nphysics` already, but I don't want to just change `Rc<RefCell<T>>` to `Arc<RwLock<T>>` because it would have too much performance penalty for the 99% of it that really is single-threaded.

mechaxl:
@jFransham: Which `unsafe impl Sync` are you referring to? I just removed the ones for the `*Processor`s, as they're apparently unnecessary.

@Emilgardis: I think this is finally ready for review when you get the chance. Note that you'll probably want to use the `--release` flag, as the falling objects tend to go right through the static mesh quite often in debug mode.

Cargo.toml
mechaxl: I've moved these back to being straight dependencies instead of dev dependencies, with the introduction of the physics processor. Let me know if I should switch them back.

jFransham:
@mechaxl Well for one, you implement `Send` and `Sync` for a clearly non-send, non-sync struct [here](https://github.com/amethyst/amethyst/pull/144/files#diff-a451a59f53f192de82d0817090f835e2R29). In my integration I have a version of `nphysics` that exposes the integer body IDs, which really are safe to send and sync, but to make it truly safe I should only redeem them on a single thread, which I don't.

To be clear, it might be ok to strip away the safety in an application (or even an example, although it might not be advisable to have `unsafe` sample code), but it's simply not good enough to have library code that puports to expose a safe facade but actually hides possible data races. The handles that you store in the `PhysicsComponent`s have inner mutability, so you could have two systems that operate on immutable `PhysicsComponent`s and have them running concurrently. That way they both could get a mutable reference to the `RefCell`'s innards at the same time (for both of them the "is already borrowed" check could return false, since it is non-atomic), and oh look at that demons are flying out of my nose now.

mechaxl:
@jFransham: Thanks, I've updated those comments. I do like the idea of simply having an integer id for the `RigidBody`s, I will update this PR if that gets merged into `nphysics` before this does. Otherwise, if it's ok with @Emilgardis I can open a ticket to track that work separately from this PR.

jFransham:
I think until there is a way to do this safely (not necessarily using 100% safe rust, but safe from a consumer perspective) this should be moved out-of-tree. Would be nice to have an example but this is simply not production-ready.

examples/07_physics/main.rs
```diff
@@ -0,0 +1,400 @@
...
+                let points = vec![
+                    [ii, jj, mat[(i, j)]],
+                    [ii + 1.0, jj, mat[(i + 1, j)]],
+                    [ii, jj + 1.0, mat[(i, j + 1)]],
+                    [ii - 1.0, jj + 1.0, mat[(max(i, 1) - 1, j + 1)]],
+                ];
+
+                let u = [points[1][0] - points[0][0], points[1][1] - points[0][1], points[1][2] - points[0][2]];
+                let v = [points[2][0] - points[0][0], points[2][1] - points[0][1], points[2][2] - points[0][2]];
+                let w = [points[3][0] - points[0][0], points[3][1] - points[0][1], points[3][2] - points[0][2]];
```
Emilgardis: Format these, they are a bit long.

examples/07_physics/main.rs
```diff
@@ -0,0 +1,400 @@
...
+                let points = vec![
+                    [ii, jj, mat[(i, j)]],
+                    [ii, jj + 1.0, mat[(i, j + 1)]],
+                    [ii - 1.0, jj + 1.0, mat[(i - 1, j + 1)]],
+                    [ii + 1.0, jj, mat[(i + 1, j)]],
+                ];
+
+                let u = [points[1][0] - points[0][0], points[1][1] - points[0][1], points[1][2] - points[0][2]];
+                let v = [points[2][0] - points[0][0], points[2][1] - points[0][1], points[2][2] - points[0][2]];
+                let w = [points[3][0] - points[0][0], points[3][1] - points[0][1], points[3][2] - points[0][2]];
```
Emilgardis: Same thing, a bit long

src/processors/physics.rs
```diff
@@ -0,0 +1,160 @@
...
+                                Segment, Triangle, Torus, CompositeShape, SupportMap};
+
+// Also reëxport some `nphysics3d` objects
+pub use self::nphysics3d::object::{RigidBody, RigidBodyHandle, Sensor, SensorHandle};
+use self::nphysics3d::world::World as PhysicsWorld;
+
+
+/// Wrapper around `RigidBodyHandle` that impls some convenience traits.
+/// NOTE: It also impls `Send`/`Sync`, which makes it unsafe to use in
+/// multiple `World`'s.
```
Emilgardis: Here you are referencing the specs world, correct?

src/processors/physics.rs
```diff
@@ -0,0 +1,160 @@
...
+        &mut self.0
+    }
+}
+
+
+/// Wrapper around `nphysics`'s `World` that impls some convenience traits.
+/// NOTE: It also impls `Send`/`Sync`, which makes it unsafe to use in
+/// multiple threads.
+/// TODO: Investigate moving to a thread-safe version.
+pub struct World(PhysicsWorld<f32>);
```
Emilgardis: I would change this `World` to be `PhysicsWorld` and just use `nphysics3d::world::World` inside the new-type.

Emilgardis: Or use another name for the nphysics world.

anderejd:
I'm confused by this comment: "/// NOTE: It also impls `Send`/`Sync`, which makes it unsafe to use in
 +/// multiple threads."

As far as I know (I have only been using rust for a couple of weeks), the point of those traits is to tell the type system that a type is safe to share (Sync) or to move (Send) to other threads . Right?

jFransham:
Yes, that's absolutely right. It's normally automatically implemented (that's almost always what you want), but it's unsafe to `impl` them explicitly, because if you `impl` them spuriously it means that Rust will happily allow you to send them between threads even though that can cause data races, which in turn can cause memory unsafety.

mechaxl:
@Emilgardis: Thanks for the review, I've fixed your comments and rebased off of the latest context rewrite.

@jFransham: The `PhysicsProcessor` should be the only thing modifying the `PhysicsComponent`s and `PhysicsWorld`, so this might be a situation where it's ok to use `unsafe` as the compiler doesn't have enough information to infer that. However, @Aceeri probably knows more about how `specs` handles that.

mechaxl:
@jFransham: It looks like locking the physics world/components as writable/mutable will prevent data races from happening:

https://gitter.im/amethyst/general?at=5865b2c49e6f00e74ad7f221

Can you point out a specific example of how this will introduce memory/concurrency safety bugs?

examples/07_physics/resources/config.yml
```diff
@@ -0,0 +1,9 @@
+dimensions: null
+fullscreen: false
+max_dimensions: null
+min_dimensions: null
+multisampling: 1
+title: "Pong example"
```
Emilgardis: You probably want to change this. Noticed this now that I have ran it on a computer.

mechaxl: Whoops :P updated

mechaxl:
@anderejd: I've updated the comments for `PhysicsWorld` and `PhysicsComponent`, hopefully that makes things clearer.

src/processors/physics.rs
```diff
@@ -0,0 +1,127 @@
...
+impl Processor<()> for PhysicsProcessor {
+    fn run(&mut self, arg: RunArg, _: ()) {
+        let (
+            mut world,
+            mut handles,
+            mut locals,
+            time,
+        ) = arg.fetch(|w| (
+            w.write_resource::<PhysicsWorld>(),
+            w.write::<PhysicsComponent>(),
```
kvark: Where is it actually writing to the physics components?

src/processors/physics.rs
```diff
@@ -0,0 +1,127 @@
...
+        for (handle, local) in (&mut handles, &mut locals).iter() {
+            let body = handle.borrow();
+            let position = body.position();
+
+            // Update translation
+            local.translation[0] = position.translation[0];
+            local.translation[1] = position.translation[1];
+            local.translation[2] = position.translation[2];
+
+            // Update rotation. `nalgebra` stores rotation as a rotation matrix,
```
kvark: This is pretty unfortunate to include this code. Perhaps, we could ping @sebcrozet to publish a minor version bump of `nalgebra` with this function?

src/processors/physics.rs
mechaxl: Mostly just because it otherwise gives me a compiler error :P

```
error: no method named `iter` found for type `(&mut ecs::Storage<components::physics::PhysicsComponent, std::sync::RwLockReadGuard<'_, ecs::Allocator>, std::sync::RwLockReadGuard<'_, ecs::MaskedStorage<components::physics::PhysicsComponent>>>, &mut ecs::Storage<components::transform::local_transform::LocalTransform, std::sync::RwLockReadGuard<'_, ecs::Allocator>, std::sync::RwLockWriteGuard<'_, ecs::MaskedStorage<components::transform::local_transform::LocalTransform>>>)` in the current scope
  --> src/processors/physics.rs:73:60
   |
73 |         for (handle, local) in (&mut handles, &mut locals).iter() {
   |                                                            ^^^^
   |
   = note: the method `iter` exists but the following trait bounds were not satisfied: `&mut ecs::Storage<components::physics::PhysicsComponent, std::sync::RwLockReadGuard<'_, ecs::Allocator>, std::sync::RwLockReadGuard<'_, ecs::MaskedStorage<components::physics::PhysicsComponent>>> : ecs::Join`, `std::sync::RwLockReadGuard<'_, ecs::MaskedStorage<components::physics::PhysicsComponent>> : std::ops::DerefMut`
```

Also, the `RigidBody`s (which `PhysicsComponent`s are a wrapper around) are modified in the call to `world.step`, and I believe we need to use `w.write` to make sure that there are no data races while `world.step` is processing.

src/processors/physics.rs
mechaxl: See his reply here:

http://users.nphysics.org/t/converting-rotation3-to-quaternion/48/2

Looks like it'll happen in the next release, which is cool. It's already been written, just needs to be released:

https://github.com/sebcrozet/nalgebra/blob/nexgen/src/geometry/quaternion_construction.rs#L169

anderejd:
Just an idea: Could we start a dedicated thread that creates and owns the physics world to avoid unsafe impl Sync + Send? The physics thread could wait for PhysicsProcessor to send a message using a channel or a semaphore.

jFransham:
@mechaxl No, because other systems can read a `PhysicsComponent`, store/clone it and then write to it later (because it has internal mutability). This is unlikely, but the point is that _it's possible to use this API in a memory-unsafe way_, and therefore we should not expose it. The dedicated physics thread idea is great though, I am 100% behind doing that. Ideally we wouldn't spin up an OS thread, can we run a greenthread on whatever threadpool ECS uses?

src/processors/physics.rs
kvark: Hmm, that makes sense, but needs to be clearly expressed in the comments.

src/processors/physics.rs
kvark: Well, I don't see why he can't publish a patch *right now*. But since he has not, we might as well let this in, temporarily.

jFransham:
@kvark It should be a seperate crate with big giant warning exclamation marks around it saying "SERIOUSLY, THIS IS MEMORY UNSAFE AND MAY MAKE DEMONS FLY OUT YOUR NOSTRILS" or something similarly scary. No reason to taint this whole library for the sake of integrating a physics engine that is, let's be honest, pretty practically limited (although it's an impressive technical acheivement).

kvark:
@jFransham ok, I think you got a point that is hard to argue with. We can't make `nphysics` a part of the engine until it's safe to do.
@mechaxl do you think we can move the nphysics logic into a separate crate?

Emilgardis:
This is now a blocker for 0.3.2, something should be done about it.
We should also make a blog post about what changes has been done, as the site hasn't seen one in a long time. 

Regarding the `Send`/`Sync` stuff, I'd say we feature gate nphysics for the time being.

mechaxl:
Closing in favor of investigating alternative methods of integrating a physics engine.

erlend-sh:
@mechaxl @msiglreith @Emilgardis @jFransham @anderejd @kvark 
Perhaps this should be revisited now that nphysics is compatible with ECS-based designs, as per the latest update from @sebcrozet:

> https://www.patreon.com/posts/whats-next-on-0-16264116
> 
> **Compatibility with ECS-based designs.** The physics world is now Send+Sync so it can be used safely on ECS systems. (Previous versions of nphysics used Rc<_> internally, breaking those properties.) However, nphysics itself is not implemented as an ECS (because I think it does not fit very well).

sebcrozet:
@erlend-sh Please note that, as mentioned at the end of my post, the code will be released on Github only at the end of this month (and released to `crates.io` at the end of march). I probably should have written this paragraph in the future tense…

erlend-sh:
Yeah I did notice that after posting this. I figured it'd still be good to bring this to other developers' attention now already so they can pop by your repo and start asking questions if they have any.

erlend-sh:
That 0.8.0 release is now done:

https://www.patreon.com/posts/nphysics-0-8-0-18689882

Issue #112, opened 2016-09-28
Load Entities from RON


ebkalderon:
We should be able to construct entities from YAML files (*.yml) stored in the `resources/entities` directory. Files can be nested arbitrarily in subfolders for organizational purposes, if the user wants. The markup format should be as follows:

``` yaml

---
component1:
    field1: true
    field2: 'value'
    field3: [1.0, 2.1]

component2:
    field1:
        - thing1
        - thing2
        - thing3
```

The string name of the entity is the filename minus the `.yml` extension. The user should be able to create an instance of such an entity in Rust code.

msiglreith:
Remark: Relevant specs issues https://github.com/slide-rs/specs/issues/18

mechaxl:
After taking a look at the ability to load a world/map in #145, I'm wondering how this will work for that use case. Using this approach for loading individual entities seems fine, but I think doesn't address the overall question of how to load a game world. It seems like we can proxy off most/all of the "real work" to `specs-rs`, so if this looks right from Amethyst's POV, I can go propose this over there to see if there's any interest. Basically, it would look something like this:

```rust
impl State for SomeExample {
    fn on_start(&mut self, ctx: &mut Context, world: &mut World) {
        world.register_component(Component1, "component1");
        world.register_component(Component2, "component2");
        
        world.load("example");
    }
}
```

With a `resources/worlds/example.yml` file that looks like this:

```yaml
---
entities:
  - component1:
      field1: true
      field2: 'value'
      field3: [1.0, 2.1]

  - component1:
      field1: false
      field2: 'value'
      field3: [1.0, 2.1]
    component2:
      field1: []
```

Xaeroxe:
Same as https://github.com/amethyst/amethyst/issues/122#issuecomment-316724835

Xaeroxe:
Ron is ready to use at this point and we've decided in other threads that's what we're going to be using, so I changed the issue title.

omni-viral:
@Xaeroxe We should abstract away from any particular data formats. Loading entities from any `serde::Deserializer` is better than just `.ron`

Xaeroxe:
I'm downgrading this to "Important" because while valuable, this isn't something we can take immediate action on.

torkleyy:
I don't see how this is much different from #122, but I think it can be closed, too.

PR #145, opened 2016-12-20
[WIP] Adding maps example

mechaxl:
I got annoyed by having to recompile every time I changed some world object, so I started in on a map loader. Wanted to get some feedback about whether this is generally the right approach first before polishing it up, though.


examples/08_maps/main.rs
```diff
@@ -0,0 +1,111 @@
...
+        camera.activate();
+        world.create_now()
+            .with(camera)
+            .build();
+
+        // Set up world
+        ctx.asset_manager.register_asset::<Mesh>();
+        ctx.asset_manager.register_asset::<Texture>();
+
+        let result = load_map::<Renderable, Light, LocalTransform, Transform>(ctx, world, "examples/08_maps/resources/maps/demo.yml");
```
mechaxl: I got this to a basic working example, but I imagine it would work better as something like this:

```rust
ctx.map_manager.register_asset::<Renderable>();
ctx.map_manager.register_asset::<Light>();
ctx.map_manager.register_asset::<LocalTransform>();
ctx.map_manager.register_asset::<Transform>();
let result = ctx.map_manager.load_file("examples/08_maps/resources/maps/demo.yml", ctx, world);
```

msiglreith:
Interesting work! Our current plans are on adding a serialization API to specs based on serde, which would store /load all entities with their corresponding components. (https://github.com/amethyst/amethyst/issues/112)

mechaxl:
@msiglreith: Thanks for the response! I've got some questions about that approach in general, but I'll close this and bring the conversation over there.

PR #149, opened 2016-12-26
Making AssetManager more "batteries-included" by default

mechaxl:


examples/06_assets/main.rs
```diff
@@ -1,66 +1,32 @@
...
-use amethyst::engine::{Application, State, Trans};
-use amethyst::context::{ContextConfig, Context};
 use amethyst::config::Element;
-use amethyst::ecs::{World, Join};
-use amethyst::context::asset_manager::{Assets, AssetLoader, AssetLoaderRaw, DirectoryStore, Mesh, Texture};
-use amethyst::renderer::VertexPosNormal;
-use cgmath::{InnerSpace, Vector3};
-use std::io::BufReader;
-
-struct Obj(obj::Obj);
```
mechaxl: I'm wondering if there's any rationale behind keeping this here. Loading `*.obj` files seems like a a pretty common use case, so it made sense to me to move this directly to the asset manager module, but it could certainly be moved back to here.

examples/06_assets/main.rs
```diff
@@ -117,6 +94,10 @@ impl State for Example {
 }
 
 fn main() {
+    let assets_path = format!("{}/examples/06_assets/resources/assets",
+                   env!("CARGO_MANIFEST_DIR"));
+    set_var("AMETHYST_ASSET_DIRS", assets_path);
```
mechaxl: Also added some handling for setting up default asset directories. Do we want to demo that here, or revert this to setting up a `DirectoryStore` itself? It does seem kind of odd to have this example setting up asset loading differently from the others.

examples/06_assets/main.rs
Emilgardis: I feel like there should still be an example showing how to implement a custom asset loader. Or maybe it's too trivial

examples/06_assets/main.rs
Emilgardis: I think this should be done by the user, also the "static" set env isn't that great. But I feel like this behaviour should be kept somehow optionally.

examples/06_assets/main.rs
anderejd: I think the goal of the pull request is excellent, more standard formats supported by the asset manger would be fantastic. I also agree with @Emilgardis that there shoud be an example demonstrating custom asset loading. Maybe do both in 06_assets? Move obj loader code to asset mananger and implement some obviously custom format loader in the example code and load one obj asset and one custom asset?

diegor8: could give your thoughts about this please

diegor8:
my issue may be of interest to the people of this thread

https://github.com/amethyst/amethyst/issues/150

mechaxl:
@Emilgardis, @anderejd, @msiglreith: I've rebased this PR off of the latest changes to `develop`, and added a custom object loader example. It should be ready for another review when you get the chance. Thanks!

Cargo.toml
```diff
@@ -49,13 +49,14 @@ name = "assets"
...
-glutin = "0.6"
+cgmath = "0.11"
+genmesh = "0.4"
 gfx = "0.12"
 gfx_device_gl = "0.11"
 gfx_window_glutin = "0.12"
-genmesh="0.4"
+glutin = "0.6"
+imagefmt = "*"
+wavefront_obj = "*"
```
Aceeri: These shouldn't be wildcarded, could break the example in the future if the dependent crates publish a major change.

examples/06_assets/main.rs
```diff
@@ -59,62 +87,105 @@ impl State for Example {
...
             .with(renderable)
+            .with(transform)
             .with(Transform::default())
             .build();
 
+        // Add sphere to scene
+        let renderable = asset_manager.create_renderable("sphere", "grass", "green").unwrap();
+        let mut transform = LocalTransform::default();
+        transform.translation = [-5.0, 0.0, 7.5];
+        transform.rotation = [0.70711, 0.70711, 0.0, 0.0];
```
Aceeri: Is this correct? The rotation is a quaternion [w, x, y, z] so I'm thinking this might need to be `[1.0, 0.70711, 0.70711, 0.0]` if you wanted a 45 degree angle on x and y.

Cargo.toml
mechaxl: :+1: Updated

mechaxl:
@kvark: Looks like you're involved in the `gfx` project. Any idea what's happening with this?

https://github.com/gfx-rs/gfx/issues/1115

It's affecting the ability to run the example with just `cargo run`. `cargo run --release` works just fine, however.

kvark:
@mechaxl we'll investigate, hold on

src/asset_manager/asset_manager.rs
```diff
@@ -280,6 +320,101 @@ impl AssetStore for DirectoryStore {
...
+impl AssetLoaderRaw for imagefmt::Image<u8> {
+    fn from_raw(_: &Assets, data: &[u8]) -> Option<imagefmt::Image<u8>> {
+        imagefmt::read_from(&mut Cursor::new(data), imagefmt::ColFmt::RGBA).ok()
+    }
+}
+
+impl AssetLoader<Texture> for imagefmt::Image<u8> {
+    fn from_data(assets: &mut Assets, image: imagefmt::Image<u8>) -> Option<Texture> {
+        let pixels = image.buf.chunks(4).map(|p| [p[0], p[1], p[2], p[3]]).collect::<Vec<_>>();
+        let chunked = pixels.as_slice().chunks(image.w).collect::<Vec<_>>();
```
msiglreith: `raw` in `TextureLoadData` are slices on images/textures. This is not a 2d array of image data!
For data loaded from an image this means a single slice of pixel data.

This should also fix the mipmap issue:
```Rust
let pixels = image.buf.chunks(4).map(|p| [p[0], p[1], p[2], p[3]]).collect::<Vec<_>>();
AssetLoader::from_data(assets, TextureLoadData {
            kind: Kind::D2(image.w as u16, image.h as u16, AaMode::Single),
            raw: &[pixels.as_slice()],
       })

mechaxl: 👍 Thanks! I've fixed the texture loading, and added some comments to the `TextureLoadData` struct to help prevent anyone else from making the same mistake

examples/06_assets/main.rs
Emilgardis: Is this resolved?

mechaxl: I think so. @aceeri?

Aceeri: I suppose, just seems weird to me.

mechaxl: @Aceeri: With that rotation, I'm getting this result, which I'm guessing is not correct?

rotation

I was mostly just trying to point the top of the sphere straight up and away from the camera, so I ended up rotating it 90 degrees around the X axis.

msiglreith: Would prefer using cgmath functions to build the quaternions from axis rotations or euler angles, as they are more intuitive when reading the code.
Note: The quaternion is valid (45° around x-axis), but shouldn't make any difference for a sphere compared to not rotating it at all.

mechaxl:
@msiglreith: I've updated the rotations to use `cgmath`

kvark:
@mechaxl for the future, please consider writing a good PR description, setting the context for future reviews/discussions.

kvark:
I looked at the code, seems fine to me. @Aceeri approved, so if @msiglreith is also happy, we'll merge

Issue #122, opened 2016-10-12
Load Prefabs from RON


ebkalderon:
We should be able to construct prefabs from YAML files (*.yml) stored in the `resources/prefabs` directory. Files can be nested arbitrarily in subfolders for organizational purposes, if the user wants. The markup format should be identical to that of [entities](https://github.com/amethyst/amethyst/issues/112):

``` yaml

---
component1:
    field1: true
    field2: 'value'
    field3: [1.0, 2.1]

component2:
    field1:
        - thing1
        - thing2
        - thing3
```

The string name of the prefab is the filename minus the `.yml` extension. The user should be able to create entities deriving from such prefabs in Rust code. However, this should also be doable from within YAML as well, following the format below:

``` yaml
extends: name_of_prefab

---
component3:
    field1: [0.1, 2.6]
```

yeliknewo:
How do you go from string into field of a struct or a name of a struct?

Aceeri:
This is seeming to turn out to be *extremely* difficult to do. Mainly because first you would need to have a `&'static str` towards either the `TypeId` and/or the trait for serialization (which simply isn't possible because serde's `Deserialize` requires `Sized` and returns `Self` meaning the compiler will have no idea what type it is.

Xaeroxe:
I'm going to suggest we rename this to "Load Prefabs from [Ron](https://github.com/kvark/ron) since we seem to be in favor of using that for our files from here forward.

Xaeroxe:
Ron is ready to use at this point and we've decided in other threads that's what we're going to be using, so I changed the issue title.

Xaeroxe:
I'm downgrading this to "Important" because while valuable, this isn't something we can take immediate action on.

torkleyy:
:tada: Can be closed!

PR #152, opened 2016-12-29
Project module (Configuration rework)

Aceeri:
Note: should only be merged once macros 1.1 is stable (Rust 1.15).
Deprecates the `amethyst_config` and merges the new contents into the `amethyst/src/project`.
Migrates the configuration structures to use serde instead of a custom loader.


msiglreith:
> Note: should only be merged once macros 1.1 is stable.

Is there already a date, when this will happen?

Aceeri:
@msiglreith It should be 1.15 I believe, so in about 5 weeks.

src/lib.rs
```diff
@@ -69,7 +79,11 @@ pub mod gfx_device;
 pub mod asset_manager;
 pub mod event;
 
-#[macro_use]
-pub extern crate amethyst_config as config;
+config!(
+    struct Test {
+        pub field: u32 = 0,
+    }
+);
```
Emilgardis: ?

src/project/directory.rs
```diff
@@ -0,0 +1,102 @@
...
+use ::project::ProjectError;
+
+/// Tracking root folder for files as well as a list of files in destination.
+#[derive(Debug)]
+pub struct Directory {
+    /// Root directory of the file system
+    pub root: PathBuf,
+
+    /// Whether recursively check subfolders
+    ///
```
Emilgardis: Char

src/project/directory.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// non-recursive:
+    /// ```
+    /// resources/
+    ///   config/
+    ///     config.yml
+    ///     display.yml
+    ///     logging.yml
+    /// ```
+    ///
```
Emilgardis: Char

src/lib.rs
Aceeri: Just an example configuration, this isn't a final pass at all.

Aceeri:
Should the prefabs (https://github.com/amethyst/amethyst/issues/112, https://github.com/amethyst/amethyst/issues/122), be located in the project folder, asset manager, or the `amethyst_ecs` crate?

Aceeri:
I'm going to rework this to only include the migration to serde from the custom serialization stuff I had before.

Aceeri:
Should be working now. Error handling could be worked on a bit, but that can wait for now.

src/project/config.rs
```diff
@@ -0,0 +1,248 @@
...
+                    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+                        where D: ::serde::de::Deserializer
+                    {
+                        struct FieldVisitor;
+
+                        impl ::serde::de::Visitor for FieldVisitor {
+                            type Value = Field;
+
+                            fn expecting(&self, formatter: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+                                try!(write!(formatter, "one of "));
```
Emilgardis: Is there a reason for using `try!` vs `?`.

src/project/config.rs
```diff
@@ -0,0 +1,248 @@
...
+                let content = try!($crate::project::directory::Directory::load(path));
+                let parsed = ::serde_yaml::from_str::<$identifier>(&content);
+                parsed.map_err(|e| $crate::project::ProjectError::Parser(e.to_string()) )
+            }
+        }
+    }
+);
+
+#[macro_export]
+macro_rules! config(
```
Emilgardis: Can we add som documentation to this?

src/project/config.rs
Aceeri: Not much, I just prefer `try!(...)` over `?` when it isn't chained.

src/project/config.rs
```diff
@@ -0,0 +1,302 @@
...
+                parsed.map_err(|e| $crate::project::ProjectError::Parser(e.to_string()) )
+            }
+        }
+    }
+);
+
+/// The `config!` macro allows defining configuration structures that can load in
+/// `.yml`/`.yaml` files.
+///
+/// It implements the [`serde::Serialize`], [`serde::Deserialize`], and [`Config`] traits.
```
Emilgardis: And `Default`?

src/project/config.rs
```diff
@@ -0,0 +1,302 @@
...
+                                $field: {
+                                    match visitor.visit() {
+                                        Ok(opt) => match opt {
+                                            Some(v) => v,
+                                            None => {
+                                                $default
+                                            },
+                                        },
+                                        Err(err) => {
+                                            println!("{}: {}, expecting: `{}`", CONFIG_NAME, err, stringify!($ty));
```
Emilgardis: TODO: Logging.

Aceeri: I was waiting for the logging stuff with `slog` to be implemented, not sure what to do until then.

Aceeri:
I was looking at getting the project stuff together but I'll just do that in a later PR. I'll see if I can get this fixed up and rebased with the documentatin and such before Wednesday.

src/project/config.rs
```diff
@@ -0,0 +1,301 @@
+
+/// Trait implemented by the `config!` macro.
+pub trait Config where Self: Sized + ::serde::Deserialize + ::serde::Serialize {
+    /// Loads a configuration structure from a file.
+    /// Defaults if the file fails in any way.
+    fn load<P: AsRef<::std::path::Path>>(P) -> Self;
+
+    /// Loads a configuration structure from a file.
+    fn direct_load<P: AsRef<::std::path::Path>>(P) -> Result<Self, ::project::ProjectError>;
```
torkleyy: `load_no_fallback` may be clearer

src/project/directory.rs
```diff
@@ -0,0 +1,104 @@
...
+        for entry in list {
+            self.list.push(entry.clone());
+        }
+
+        Ok(())
+    }
+
+    /// Return contents of a file.
+    pub fn load<P: AsRef<Path>>(path: P) -> Result<String, ProjectError> {
+        let mut file = try!(fs::File::open(&path));
```
torkleyy: Would be nice to support Android, too.

src/project/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+    pub fn description(&self) -> String {
+        match self {
+            &ProjectError::File(ref err) => err.to_string(),
+            &ProjectError::Parser(ref msg) => msg.clone(),
+        }
+    }
+}
+
+/// Project structure, holds information related to configurations and meta information.
+pub struct Project {
```
torkleyy: Probably missing a `#[derive(Debug)]`? Stupid question: Sorry if I missed some discussion about this, but how is this supposed to be used? Also, why isn't it using `Directory`? I think it would make sense to have it load all the configs together.

Aceeri: I was hoping to eventually use it as some sort of meta-ish structure about the project, such as defining paths to resources, configurations, assets, etc. It isn't using `Directory` right now because I'm looking at using the walkdir crate instead, but I'm not 100% sure either way.

`#[derive(Debug)]` can be added I suppose, this is just a little mockup for what I want to do with this though, the main part of this PR is just dealing with configurations.

src/project/directory.rs
Aceeri: Do you have any examples on how to support both at the same time? I'm not entirely familiar with how the Android file system works.

torkleyy: @Aceeri I first thought about android assets, but since this is will also be for mutated - if I understand correctly - Android assets are not suitable, so the current solution works just fine (because we have to use file access on Android then, anyways) :tada:

src/project/mod.rs
torkleyy: @Aceeri I see, thanks for explaining. Do you plan to have some `Project::load`, which loads all configs (maybe addressed in a future PR)?

Aceeri: Yeah, I'm thinking of having it so the method would take in the `specs::World` and add them as resources. The user would still need to register with the project however, similar to how it works with the specs world, since the project would need to retain the configuration type in a closure.
e.g.
```rust
impl Project {
    fn load(world: &mut World) {
        for config_closure in closures {
            config_closure(world);
        }
    }
    fn register<T: Config>() {
        let closure = |world: specs::World| {
            // find file, load into rust structure with serde
            world.add_resources::<T>(config);
        };
        self.closures.push_back(closure);
    }
}
```

Could also have methods for custom named files of configurations in the config folder.

src/project/config.rs
```diff
@@ -0,0 +1,301 @@
...
+                let serialized = try!(result.map_err(|e| $crate::project::ProjectError::Parser(e.to_string()) ));
+                let mut file = try!(::std::fs::File::create(path));
+                try!(file.write(&serialized.into_bytes()));
+
+                Ok(())
+            }
+
+            fn direct_load<P: AsRef<::std::path::Path>>(path: P) -> Result<Self, $crate::project::ProjectError> {
+                let content = try!($crate::project::directory::Directory::load(path));
+                let parsed = ::serde_yaml::from_str::<$identifier>(&content);
```
torkleyy: There was already some discussion on Gitter about supporting multiple formats and also #171 addresses this issue. Should we do it? It would probably be a simple

```rust
match path.file_extension() {
    "yml" => serde_yaml::from_str(&content),
    "toml" => toml::from_str(&content),
}
```

OvermindDL1: I'd take toml over yaml any day!

ebkalderon:
@Aceeri Hey, are you still working on this PR? It's been a while, and I don't want to see it going stale.

Aceeri:
I'm probably going to wait for the project module stuff (mainly after we have the asset manager stuff done because it will mainly deal with that probably), but the configuration stuff should work fine. I'm still kind of hesitant about multiple configuration formats though since it could just get confusing.

torkleyy:
@Aceeri

Oh, I didn't know you're waiting for my PR...
There is actually a whole chain of changes that have to land before:
1) The `Engine` struct, which depends on

2) the rayon `ThreadPool` which was already integrated into specs, but there was no new release, because shred should be integrated before that

3) In shred @msiglreith would like to have closures replace tasks

4) In addition to my `Engine` struct I would like to have a headless OpenGL context, because otherwise we cannot have tests

So.. as soon as shred got closures, I'll get a PR for specs ready as fast as possible. In the meantime I'll try to get support for headless context creation for gfx. And finally, there is also the event system my asset management depends on...

Aceeri:
I don't mean that this PR has to wait for it, the project module I could just do in a later PR.

src/project/mod.rs
```diff
@@ -23,39 +22,32 @@ pub enum ProjectError {
...
-    fn from(e: io::Error) -> ProjectError {
-        ProjectError::File(e)
-    }
-}
-
 impl ProjectError {
-    /// Returns a human friendly error message for the `ProjectError`.
-    pub fn description(&self) -> String {
+    /// Displays the type of error and relevant information.
+    pub fn to_string(&self) -> &str {
```
torkleyy: Should probably implement `ToString`.

Aceeri: Yeah, I'm adding an impl for std::error::Error, std::fmt::Display, etc.

examples/03_renderable/main.rs
```diff
@@ -6,13 +6,14 @@ extern crate cgmath;
 
 use amethyst::{Application, ElementState, Event, State, Trans, VirtualKeyCode, WindowEvent};
 use amethyst::asset_manager::{AssetManager, DirectoryStore};
-use amethyst::config::Element;
```
nchashch: I like this change, `project::Config` is somewhat more self-descriptive than `config::Element`.

nchashch:
Appveyor build passes on stable but fails on beta and nightly because it can't build `dxguid` crate. This problem isn't introduced in this PR, appveyor failed on nightly for the same reason when testing 0.4.3 see ([appveyor log](https://ci.appveyor.com/project/amethyst/amethyst/build/1.0.335/job/j2j1oulfsvjll89v#L217)). So I think we can merge this despite appveyor build failure.

torkleyy:
@nchashch Could you leave it open for another day please? I'd like to review it once more. In case I forget it, feel free to ping me.

nchashch:
@torkleyy  Ok.

src/project/config.rs
```diff
@@ -0,0 +1,334 @@
...
+pub trait Config where Self: Sized {
+    /// Loads a configuration structure from a file.
+    /// Defaults if the file fails in any way.
+    fn load<P: AsRef<::std::path::Path>>(P) -> Self;
+
+    /// Loads a configuration structure from a file.
+    fn load_no_fallback<P: AsRef<::std::path::Path>>(P) -> Result<Self, ::project::ProjectError>;
+
+    /// Writes a configuration structure to a file.
+    fn write<P: AsRef<::std::path::Path>>(&self, P) -> Result<(), ::project::ProjectError>;
```
torkleyy: Let's add an import here.

Aceeri: Not entirely sure what you mean here?

src/project/config.rs
```diff
@@ -0,0 +1,337 @@
...
+}
+
+/// The `config!` macro allows defining configuration structures that can load in
+/// `.yml`/`.yaml` files.
+///
+/// It implements the [`serde::Serialize`], [`serde::Deserialize`], and [`Config`] traits.
+/// As well as the standard libraries `std::default::Default` trait for the defaulting fields.
+///
+/// It follows Rust's syntax for defining structures with one addition, defaulting values.
+/// If the file fails to specific fields of a configuration, then it will print out
```
torkleyy: Maybe "does not contain a specific field" or something like that?

src/project/config.rs
```diff
@@ -0,0 +1,337 @@
...
+/// an error describing the issue and then load in this defaulting value.
```
torkleyy: Do we really want it to print an error? If I see it correctly, having defaults is useful so you don't have to specify it. Otherwise, they're fallback values.

Aceeri: Personally, I'd say yes, if a user wants to change some configuration of a file and it isn't specified then they aren't going to know what the name of the field is nor the field type.

torkleyy: That's the same for every configuration format with optional keys; you just have to document it (which isn't a bad idea anyways, because you can explain what they're doing...). But let's here someone else's opinion.

PR #153, opened 2017-01-04
Render entities which only have Renderable component in `GfxDevice::render_world`

nchashch:
Now entities with only Renderable component attached to them (not only Renderable, Transform pair) are also rendered.

### DIFF
--- a/src/gfx_device/gfx_device.rs
+++ b/src/gfx_device/gfx_device.rs
@@ -42,11 +42,12 @@ impl GfxDevice {
         }
     }
 
-    /// Render all `Renderable`, `Transform` pairs in `World`.
+    /// Render all entities with `Renderable` components in `World`.
     pub fn render_world(&mut self, world: &mut ecs::World, pipeline: &renderer::Pipeline) {
         use components::rendering::{MeshInner, TextureInner, Renderable};
         use components::transform::Transform;
         use world_resources::camera::Projection;
+        use renderer::Fragment;
         match self.gfx_device_inner {
             GfxDeviceInner::OpenGL { ref mut renderer,
                                      ref mut device,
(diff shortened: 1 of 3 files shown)


src/gfx_device/gfx_device.rs
```diff
@@ -79,35 +80,50 @@ impl GfxDevice {
...
                 let global_transforms = world.read::<Transform>();
+                // Add all entities which only have Renderable component attached to them to the scene.
+                for (renderable, _) in (&renderables, !&global_transforms).iter() {
+                    // If Transform is not specified use the identity transform.
+                    if let Some(fragment) = unwrap_renderable(renderable, &Transform::default()) {
+                        scene.fragments.push(fragment);
+                    }
+                }
+                // Add all entities which have a Renderable, Transform pair attached to them to the scene.
                 for (renderable, global_transform) in (&renderables, &global_transforms).iter() {
```
kvark: I think it would be cleaner to just iterate over entities and `renderables`, trying to access the relevant global transform for each entity.

src/gfx_device/gfx_device.rs
```diff
@@ -79,35 +80,50 @@ impl GfxDevice {
...
+                    if let Some(fragment) = unwrap_renderable(renderable, global_transform) {
+                        scene.fragments.push(fragment);
+                    }
+                }
+                let lights = world.read::<renderer::Light>();
+                // Add all Lights to the scene.
+                for light in lights.iter() {
+                    scene.lights.push(light.clone());
```
kvark: nit: you can rewrite this with using `extend` and iterators

src/gfx_device/gfx_device.rs
```diff
@@ -79,35 +80,50 @@ impl GfxDevice {
...
+                }
+                // Render the scene.
+                renderer.submit(pipeline, &scene, device);
```
kvark: do we ever call `device.cleanup()`?

kvark:
Awesome, thanks @nchashch !

PR #154, opened 2017-01-05
Lighting updates

mechaxl:
Added specular lighting via Blinn-Phong shading

Switched from light propagation to light attenuation

Updated `03_renderable` example to show off new lighting


src/components/rendering/renderable.rs
```diff
@@ -3,21 +3,25 @@ use components::rendering::Texture;
...
-/// This struct is a `Component`, which combines geometry and ka, kd textures.
+/// This struct is a `Component`, which combines geometry and ka/kd/ks textures.
 /// Every `Renderable`, `Transform` pair attached to an entity inside the `World`
 /// is rendered by `GfxDevice::render_world` method.
 pub struct Renderable {
     pub mesh: Mesh,
     pub ka: Texture,
     pub kd: Texture,
+    pub ks: Texture,
+    pub ns: f32,
```
msiglreith: These names are kinda confusing, maybe something longer would be better and documentation.

src/renderer/src/lib.rs
```diff
@@ -184,31 +184,43 @@ pub struct Fragment<R: gfx::Resources> {
...
-    pub propagation_constant: f32,
-    /// linear level drops
-    pub propagation_linear: f32,
-    /// cubic light level drop
-    pub propagation_r_square: f32,
+
+    /// Represents constant, linear, and quadratic (kc, kl, and kq)
+    /// factors for light attenuation calculations. See these links
+    /// for some more information on how to choose good values for
+    /// these factors:
```
msiglreith: Should also provide a function to generate such useful values for the user e.g from a light radius

src/renderer/src/pass/forward.rs
```diff
@@ -82,16 +83,36 @@ pub static FRAGMENT_SRC: &'static [u8] = b"
...
-            float inv_dist = 1. / dist;
-            vec4 light_to_point_normal = delta * inv_dist;
-            float intensity = dot(light[i].propagation.xyz, vec3(1., inv_dist, inv_dist * inv_dist));
-            color += kd * light[i].color * intensity * max(0, dot(light_to_point_normal, vec4(v_In.Normal, 0.)));
+            // Calculate diffuse light
+            vec4 lightDir = normalize(light[i].center - v_In.Position);
+            float diff = max(dot(lightDir, normal), 0.0);
+            vec4 diffuse = diff * light[i].color;
+
+            // Calculate specular light
```
msiglreith: Add documentation that we are calculating the specular reflection via Blinn-Phong.

kvark:
It would be fine to combine the new example with the old one if the result features both lit and non-lit rendering.

src/components/rendering/renderable.rs
Emilgardis:
```rust
/// Ambient texture
ka: Texture,
/// Diffuse texture
kd: Texture
/// Specular texture
ks: Texture,
```
and/or add information to struct doc.

Aceeri: Why not just call them kambient, or just ambient?

kvark: `k_ambient` would work too

examples/08_lighting/main.rs
```diff
@@ -0,0 +1,215 @@
+extern crate amethyst;
```
kvark: I'd really like to either merge the example with the other one, or move the common parts out somewhere to reduce the maintenance burden

src/renderer/src/pass/deferred.rs
```diff
@@ -150,7 +150,8 @@ pub static LIGHT_FRAGMENT_SRC: &'static [u8] = b"
             float dist = length(delta);
             float inv_dist = 1. / dist;
             vec4 light_to_point_normal = delta * inv_dist;
-            float intensity = dot(light[i].propagation.xyz, vec3(1., inv_dist, inv_dist * inv_dist));
+            // float intensity = dot(light[i].propagation.xyz, vec3(1., inv_dist, inv_dist * inv_dist));
```
kvark: what's the reason for that?

src/renderer/src/pass/forward.rs
```diff
@@ -82,16 +83,36 @@ pub static FRAGMENT_SRC: &'static [u8] = b"
...
+            vec4 specular = vec4(0.0);
+            if (diff > 0.0) {
+                vec4 viewDir = normalize(-v_In.Position);
+                vec4 reflectDir = reflect(-lightDir, normal);
+                vec4 halfwayDir = normalize(lightDir + viewDir);
```
kvark: the `halfway` should really be interpolated from the vertex shader

src/renderer/src/pass/forward.rs
```diff
@@ -82,16 +83,36 @@ pub static FRAGMENT_SRC: &'static [u8] = b"
...
-            float intensity = dot(light[i].propagation.xyz, vec3(1., inv_dist, inv_dist * inv_dist));
-            color += kd * light[i].color * intensity * max(0, dot(light_to_point_normal, vec4(v_In.Normal, 0.)));
+            // Calculate diffuse light
+            vec4 lightDir = normalize(light[i].center - v_In.Position);
+            float diff = max(dot(lightDir, normal), 0.0);
+            vec4 diffuse = diff * light[i].color;
+
+            // Calculate specular light
+            vec4 specular = vec4(0.0);
+            if (diff > 0.0) {
```
kvark: I don't think this is correct. While physically obvious, it doesn't quite work like that for blinn-phong model, which can have a slight specular effect on areas unaffected by the diffuse (because it's an approximation, after all). So I suggest just computing the diffuse and specular terms independently and combining them. Otherwise, you may see a sharp cut-off of the specular for some cases.

src/renderer/src/pass/deferred.rs
mechaxl: Sorry, that was just me playing around with the deferred rendering. I've updated this line in the latest commit

mechaxl:
@kvark: If you run the current assets example in this PR with `cargo run --example assets` and then press space to show the deferred rendering pipeline, it shows all fragments as white, even though I'm attempting to set their color equal to their normal. Any idea what I'm doing wrong here?

kvark:
@mechaxl it works for me - hitting space I see the normals of the objects. What's your OS/HW/Driver? Perhaps, you could run with `RUST_LOG=gfx=debug`, assuming Amethyst enames the logging, and provide the output.

mechaxl:
@kvark: Thanks for the reply. That was on Mac OSX with OpenGL. Not sure what was happening, but switching to my Linux machine fixed it. ¯\\\_(ツ)\_/¯

src/renderer/src/pass/forward.rs
mechaxl: I'm a little confused as to how to interpolate `halfway`. Do you have some pointers to where I can read about how to do that?

kvark: @mechaxl you have both the light vector and the view vector in the vertex shader already. So instead of interpolating them to FS and then averaging there, you could just average them in the VS and interpolate the result (which would just need to be re-normalized in PS).

kvark:
@mechaxl thanks for the changes!
Please update the half-vector computation and make sure the CI passes.

src/renderer/src/lib.rs
```diff
@@ -184,31 +190,99 @@ pub struct Fragment<R: gfx::Resources> {
...
+    /// http://ogldev.atspace.co.uk/www/tutorial20/tutorial20.html
+    pub attenuation: [f32; 3],
+}
+
+impl PointLight {
+    /// Creates a new PointLight with the attenuation factors
+    /// set to create a light that falls below a cutoff at `radius`.
+    /// Since most monitors won't display anything less than an RGB
+    /// value of 1/255, this function uses that value as a cutoff.
+    pub fn from_radius(radius: f32) -> PointLight {
```
msiglreith: Good point regarding cutoff! Useful for possible light culling.
(I'm not really convinced by the display range argumentation with respect to HDR rendering.)

src/renderer/src/lib.rs
```diff
@@ -184,31 +190,99 @@ pub struct Fragment<R: gfx::Resources> {
...
+        let cutoff = 1.0 / 255.0;
+        let quadratic = ((1.0 / cutoff - 1.0) / radius - 0.1) / radius;
```
msiglreith: How did you derive this formula?
The engineers from the Frostbite engine gave a very nice talk about PBR at SIGGRAPH 2014 (http://www.frostbite.com/wp-content/uploads/2014/11/course_notes_moving_frostbite_to_pbr.pdf). It also includes a description (p.31) about their attenuation formula, which might be interesting for implementing a nice cutoff function (needs to be done in the shader):

Frostbite: `attenuation = 1.0 / max(distance, 0.1).powi(2) * cutoff`
UE4 [Karis13]: `attenuation = 1.0 / (distance.powi(2) + 1.0) * cutoff`

We might could implement something like: `attenuation = 1.0 / (distance + 0.1).powi(2) * cutoff`
This would fit better into our attenuation model and would also take care of `distance=0`.
Suggestions?

kvark:
@mechaxl so basically we need to fix the half-vector computation and the attenuation equation. Would you be able to do this fixes for the PR to be merged?

mechaxl:
Sure, sorry about the wait. Will get that pushed up soon when I get a chance, though

kvark:
It appears that if we compute the half vectors in VS, we can't pass all of them to PS. So the way it's computed now is fine.
As for @msiglreith's concern on the cut-off equation, this can be fixed in a follow-up.
I consider the PR ready for merging, just needs a rebase.

mechaxl:
Closing in favor of #166, as I screwed something up and can no longer edit this PR

Issue #160, opened 2017-01-26
RFC: Roadmap


msiglreith:
# Summary
[summary]: #summary

Establish a new vision and work out a long term roadmap for the project.

# Motivation
[motivation]: #motivation

Sparked from kvark's gitter posts: https://gitter.im/amethyst/engine/archives/2017/01/23

# Detailed design
[design]: #detailed-design

## Vision

> Amethyst aims to be a modular, parallel and data-driven game engine

**Modularity** The game engine should be easily extendable to fit the needs of the game developers for their game idea. Therefore we expose an entity-component system, which allows to define new components and processors without needing to change the game engine code. This should emphasize the 'game engine as library' aspect.

**Parallelism** Current and future CPUs are multi-core processors, which can run multiple tasks in parallel. Rust provides us useful language features to implement parallelism, which should be used to provide an easy and safe API for the engine.

**Data-Driven** In order to achieve high performance, we will also have to take care of data-orientation next to parallelism. This is closely coupled with the ECS system: processors will perform certain operations over large data sets.

## Goals
These goals are mainly based on priority and will run through several iterations over the whole development process. So, short-term goals won't be finished when starting with mid-term content but still be worked on.
### Short term
These are the core functionalities for the game engine required to build a game. Mainly focuses on using the engine as library to build upon.
* Entity-Component System
* Rendering
* Input Handling
* Audio
* Serialization/Reflection
* Streaming/Asset Handling
* Documentation/Tutorial

### Mid-term

* Physics
* Tooling
* Animation

### Long-term

* Scripting
* Graphical UI
* Networking
* AI

### Develop A Game
A game engine is built in order to build games! The best tutorial for potential users are working examples, and fancier examples are even better.
Under this assumption we should also try to spend some time on translating our implementations into a working game, showcasing the technical aspects of our engine.

## Systems/Aspects
Discussing the scopes and features of the different subsystems in more detail.

### Entity-Component System
Our current system is based on `specs` and should probably stick to it in near future (in case some magically new library pops up..).
As the ECS is the core part of the architecture, this needs be carefully designed and should be clear under our 3 main vision aspects:

**Modularity** We expose the API from `specs` to the developers, in order to allow them to implement new components, processors and resources (larger singleton objects). The engine will have built-in processors for common operations like `Transform` to ensure spatial hierarchy.

**Parallelism** Processor run in parallel with respect to task- and data-dependencies with low overhead.

**Data-Driven** TODO

**Next steps**
[...]

Issue #34, opened 2016-03-25
RFC: Extensions system "Shards"


White-Oak:
Each Amethyst game project can contain several 'shards', which are essentialy Amethyst fucnctionality packages with code and/or resources.
Examples of such shards can be:
- Renderer
- ECS
- Audio
- Physics
- AI

Shards can be connected together in the core `amethyst_engine` crate to have additional functionality such as
- ECS & Renderer will provide aditional functionality such as Rendering systems/components.
- ECS & Audio will provide additional systems/components.

Following tasks are considered completed when discussed and implemented. If the team comes to a conclussion on a task, it is marked in a certain way:
- [ ] **Part 1: Features**
  Technically, adding shards to projects will be done through 'features' -
  conditional compilation supported by Rust and Cargo. Thus, adding, for example 'ecs' shard will result in adding 'ecs' feature to Cargo.toml of a project. Feature `ecs` will enable ecs crate in `amethyst_engine`.
  Some features, as mentioned above, can be glued together in `amethyst_engine` crate using combinned features (i.e. `cfg(all(ecs, renderer))`).
  Some shards may not be included in base `amethyst_engine` crate and CLI will need to inject an additional dependency, when ading such a shard.
  - [x] CLI (**agreed on**) (impl PR: [subcommands for CLI to add/remove shards](https://github.com/amethyst/tools/pull/28))  
    Proposed:  
  
  ```
  amethyst add <name>  
  amethyst remove <name>
  ```
  
  Alternatives:  
  
  ```
  amethyst shard add <name>  
  amethyst shard remove <name>
  ```
  - [ ] Shards list  
    There may be inconsistence between shard name and its crate name. As well, some shards may not be included in the base crate, so an additonal dependency needs to be injected. A list of such mapping ('exotic_shard' -> 'exotic') and a list of necessary dependencies needs to exist.
    - [ ] A format of a list. Those are basically two maps. What format to use?
    - [ ] A place to keep it. Options are:
      - inside `tools` crate
      - in some (existing or separate) repository on GH, being synced when using CLI to add shards.
    - [ ] ~~User-specific shards. Users may share their own shards for a specific purpose. How to handle them?~~ These are called third-party libraries or extensions and are not shards.
- [ ] **Part 2: Shards' resources**
  Each shard may bring its own resources, and/or configuration files, and/or modify existing. I have no idea of this part. It should be thouroughfully discussed and, maybe, even separated to its own RFC.

White-Oak:
[...]

PR #163, opened 2017-01-28
RFC: Roadmap

msiglreith:
Followup to #160

[Rendered](https://github.com/msiglreith/amethyst/blob/9315bd09a7284ee9e6b2a97b095890c230b6d5f9/ROADMAP.md)

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,0 +1,154 @@
+## Vision
+
+> Amethyst aims to be a modular, parallel and data-driven game engine
+
+**Modularity** The game engine should be easily extendable to fit the needs of the game developers for their game idea. Therefore we expose an entity-component system, which allows to define new components and processors without needing to change the game engine code. This should emphasize the 'game engine as library' aspect.
+
+**Parallelism** Current and future CPUs are multi-core processors, which can run multiple tasks in parallel. Rust provides us useful language features to implement parallelism, which should be used to provide an easy and safe API for the engine.
+
+**Data-Driven** In order to achieve high performance, we will also have to take care of data-orientation next to parallelism. This is closely coupled with the ECS system: processors will perform certain operations over large data sets.
+
+## Goals
+These goals are mainly based on priority and will run through several iterations over the whole development process. So, short-term goals won't be finished when starting with mid-term content but still be worked on.
+### Short term
+These are the core functionalities for the game engine required to build a game. Mainly focuses on using the engine as library to build upon.
+* Entity-Component System
+* Rendering
+* Input Handling
+* Audio
+* Serialization/Reflection
+* Streaming/Asset Handling
+* Documentation/Tutorial
+
+### Mid-term
+
+* Physics
+* Tooling
+* Animation
+
+### Long-term
+
+* Scripting
+* Graphical UI
+* Networking
+* AI
+
+### Develop A Game
+A game engine is built in order to build games! The best tutorial for potential users are working examples, and fancier examples are even better.
+Under this assumption we should also try to spend some time on translating our implementations into a working game, showcasing the technical aspects of our engine.
+
+## Systems/Aspects
+Discussing the scopes and features of the different subsystems in more detail.
+
+### Entity-Component System
+Our current system is based on `specs` and should probably stick to it in near future (in case some magically new library pops up..).
+As the ECS is the core part of the architecture, this needs be carefully designed and should be clear under our 3 main vision aspects:
+
+**Modularity** We expose the API from `specs` to the developers, in order to allow them to implement new components, processors and resources (larger singleton objects). The engine will have built-in processors for common operations like `Transform` to ensure spatial hierarchy.
+
+**Parallelism** Processor run in parallel with respect to task- and data-dependencies with low overhead.
+
+**Data-Driven** TODO
+
+**Next steps**
+ * Evaluating different approaches for defining task-dependencies between the processors, including support to run a processor multiple times
+ * Research ways to handle single-threaded processors (e.g nphysics) if possible
+
+### Rendering
+The trend in industry is to achieve realistic rendering, therefore most engines implement physically based rendering.
+Using `gfx-rs` for backend abstraction allows us to provide compatibility with multiple platforms without caring that much about the different graphic APIs
+
+**Modularity** TODO: easily integrate own rendering approaches
+**Parallelism** TODO: parallel command list creation
+**Data-Driven** TODO: efficient rendering -> minimizing draw calls and overhead
+
+**Next steps**
+ * Transfer to physically based rendering (pretty large tasks..)
+ * Evaluate current system in terms of extendability for developers to implement own rendering approaches
+ * Feedback practical experience into improvements for `gfx-rs`
+
+### Input Handling
+**Modularity** The API should be flexible enough to support multiple different input devices including gamepads and maybe VR devices.
+**Data-Driven** TODO: raw input
+
+**Next steps**
+ * Raw input support (winit?)
+ * Gamepad support (winit?)
+ * Input Mapping
+
+### Audio
+TODO: Unfortunately I'm not that deep into the audio components of game engines (state of art?)
+I imagine some kind of OpenAL like API, supporting spatial sound source objects, audio streaming, attenuation, etc.
+Provide low-level access?
+
+** Next Steps**
+
+### Serialization/Reflection
+Type reflection offers additional information about the type, which is available at compile- or run-time. In comparison to C++, to can be achieved easier with Rusts macro system.
+In our engine we want to couple the ECS with serialization/reflection to easily load/save game scenes from various formats. Additionally, it helps us in to build safe network protocols, writing script and GUI bindings.
+
+** Next Steps**
+ * Integrate serde based reflection
+ * Serialize/Deserialize ecs worlds
+
+### Streaming/Asset Handling
+TODO
+
+**Parallelism** TODO: Load assets in parallel
+**Data-Driven** TODO: low loading overhead
+
+** Next Steps**
+ * Improve current approach regarding robustness (error handling) and flexibility with scene formats
+ * Implement loaders for common file formats and archives
+
+### Documentation/Tutorial
+TODO
+
+** Next Steps**
+
+### Physics
+TODO: I would consider physics pretty difficult part to do *right*, therefore not part of the short term goals.
+Maybe rely on external libraries first.
+Should still provide ways for collision detection, allowing to do collision checks, raycasts, etc. (-> ncollide?).
+
+**Parallelism** TODO: internal physic parallelism, possible GPU implementation (hard), difficult with external libraries
+**Data-Driven** TODO: integration issue: physics world <-> ECS world, having 2 different representations? if so how to keep them in sync and integrate with other conflicting processors (e.g spatial hierarchy)
+
+** Next Steps**
+ * Integrate an existing library or write from scratch (requires further planning then)
+ * Provide collision testing API (triggers, raycasts, etc.)
+
+### Tooling
+We want to focus on developing a game engine, even though a rich toolset is required to simplify different task for users. If possible, we can integrate existing tools like profilers and 3D editors.
+'In-house' tools would be implemented on demand, e.g. engine specific data formats, own editor (Qt based?).
+
+** Next Steps**
+ * Integrate better logging (high priority -> short-term)
+
+### Scripting
+Possible scripting languages: mruby, lua, moonscript, dyon, ..
+
+Supporting a scripting language would support rapid-prototyping and faster iteration cycles. Also would be more attractive for developers who are not so much familiar with Rust.
+Scripting could be use for several subsystems like animation, particle system, AI, etc.
+
+** Next Steps**
+ * Evaluate the different scripting languages with respect to interoperability, performance, usability, ..
+ 
+### Graphical UI
+** Next Steps**
+ * Compatibility of GUIs with entity component systems (UI elements as entities <-> only root UI elements)
+ * Integrate an existing UI library (e.g conrod) or develop a new library from scratch (requires further planning then)
+ 
+Future:
+A common trend for large engines is embedding webbrowser frameworks like CEF for UI. Allowing to re-use existing JS/HTML/CSS frameworks to create complex UIs in a short time.
+Unfortunately CEF is pretty large, therefore a Servo based approach would be desirable for performance and easier integration.
+
+### Networking
+TODO: same as physics, hard to do correctly, interesting would be something like `libyojimbo` but well..
+
+** Next Steps**
+
+### AI
+TODO: also no real experience with this..
+
+** Next Steps**
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -74,12 +74,13 @@ Using `gfx-rs` for backend abstraction allows us to provide compatibility with m
 **Next steps**
  * Raw input support (winit?)
  * Gamepad support (winit?)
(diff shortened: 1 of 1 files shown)


ROADMAP.md
```diff
@@ -0,0 +1,154 @@
...
+A common trend for large engines is embedding webbrowser frameworks like CEF for UI. Allowing to re-use existing JS/HTML/CSS frameworks to create complex UIs in a short time.
+Unfortunately CEF is pretty large, therefore a Servo based approach would be desirable for performance and easier integration.
+
+### Networking
+TODO: same as physics, hard to do correctly, interesting would be something like `libyojimbo` but well..
+
+** Next Steps**
+
+### AI
+TODO: also no real experience with this..
```
Emilgardis: AI support should be pretty minimal, I'm thinking something like A* implementations and the likes. How does this integrate with the physics engine?

ROADMAP.md
```diff
@@ -0,0 +1,154 @@
...
+**Parallelism** TODO: internal physic parallelism, possible GPU implementation (hard), difficult with external libraries
+**Data-Driven** TODO: integration issue: physics world <-> ECS world, having 2 different representations? if so how to keep them in sync and integrate with other conflicting processors (e.g spatial hierarchy)
+
+** Next Steps**
+ * Integrate an existing library or write from scratch (requires further planning then)
+ * Provide collision testing API (triggers, raycasts, etc.)
+
+### Tooling
+We want to focus on developing a game engine, even though a rich toolset is required to simplify different task for users. If possible, we can integrate existing tools like profilers and 3D editors.
+'In-house' tools would be implemented on demand, e.g. engine specific data formats, own editor (Qt based?).
```
Emilgardis: As was discussed earlier, making plugins for tools like Blender, 3DS max etc would make this process much easier for the team and the user.

ROADMAP.md
```diff
@@ -0,0 +1,154 @@
...
+ * Integrate an existing UI library (e.g conrod) or develop a new library from scratch (requires further planning then)
+ 
+Future:
+A common trend for large engines is embedding webbrowser frameworks like CEF for UI. Allowing to re-use existing JS/HTML/CSS frameworks to create complex UIs in a short time.
+Unfortunately CEF is pretty large, therefore a Servo based approach would be desirable for performance and easier integration.
+
+### Networking
+TODO: same as physics, hard to do correctly, interesting would be something like `libyojimbo` but well..
+
+** Next Steps**
```
Emilgardis: We should look into if we need to make support for protocols like [Quake](http://www.gamers.org/dEngine/quake/QDP/qnp.html), protobuf, etc easier (not the actual implementations, just making the interactions as smooth as possible). Nom would be an excellent thing for the user to use, but should we provide ways to integrate it with the networking shard?

ROADMAP.md
```diff
@@ -0,0 +1,154 @@
...
+### Audio
+TODO: Unfortunately I'm not that deep into the audio components of game engines (state of art?)
+I imagine some kind of OpenAL like API, supporting spatial sound source objects, audio streaming, attenuation, etc.
+Provide low-level access?
+
+** Next Steps**
+
+### Serialization/Reflection
+Type reflection offers additional information about the type, which is available at compile- or run-time. In comparison to C++, to can be achieved easier with Rusts macro system.
+In our engine we want to couple the ECS with serialization/reflection to easily load/save game scenes from various formats. Additionally, it helps us in to build safe network protocols, writing script and GUI bindings.
```
Emilgardis: Related to networking, add a note inside that goal.

ROADMAP.md
```diff
@@ -0,0 +1,154 @@
...
+ * Feedback practical experience into improvements for `gfx-rs`
+
+### Input Handling
+**Modularity** The API should be flexible enough to support multiple different input devices including gamepads and maybe VR devices.
+**Data-Driven** TODO: raw input
+
+**Next steps**
+ * Raw input support (winit?)
+ * Gamepad support (winit?)
+ * Input Mapping
```
Emilgardis: What does this mean?

ROADMAP.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Audio
+TODO: Unfortunately I'm not that deep into the audio components of game engines (state of art?)
+I imagine some kind of OpenAL like API, supporting spatial sound source objects, audio streaming, attenuation, etc.
+Provide low-level access?
```
Emilgardis: [dsp-chains](https://github.com/RustAudio/dsp-chain)

ROADMAP.md
```diff
@@ -0,0 +1,154 @@
...
+## Systems/Aspects
+Discussing the scopes and features of the different subsystems in more detail.
+
+### Entity-Component System
+Our current system is based on `specs` and should probably stick to it in near future (in case some magically new library pops up..).
+As the ECS is the core part of the architecture, this needs be carefully designed and should be clear under our 3 main vision aspects:
+
+**Modularity** We expose the API from `specs` to the developers, in order to allow them to implement new components, processors and resources (larger singleton objects). The engine will have built-in processors for common operations like `Transform` to ensure spatial hierarchy.
+
+**Parallelism** Processor run in parallel with respect to task- and data-dependencies with low overhead.
```
Emilgardis: I'm not sure if this is possible on AMD cards though. This may be a major drawback.

ROADMAP.md
msiglreith: Allow the users to define custom actions like `Jump`, `Forward`, etc.
Instead of asking for `is_key_pressed(A)`, we can use the newly defined actions. Players can then remap these actions to actual input events e.g keyboard press oder gamepad events.

Will try to define this point better in the roadmap!

ROADMAP.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Streaming/Asset Handling
+TODO
+
+**Parallelism** TODO: Load assets in parallel
+**Data-Driven** TODO: low loading overhead
+
+** Next Steps**
+ * Improve current approach regarding robustness (error handling) and flexibility with scene formats
+ * Implement loaders for common file formats and archives
```
msiglreith: Yes, we can use extisting loaders for formats like `obj`. If they don't exist and we have need for supporting a certain format these libraries then should live in another crate.
The loading code glue for our assetsloader still have to be provided in the engine.

ROADMAP.md
```diff
@@ -0,0 +1,154 @@
...
+Supporting a scripting language would support rapid-prototyping and faster iteration cycles. Also would be more attractive for developers who are not so much familiar with Rust.
+Scripting could be use for several subsystems like animation, particle system, AI, etc.
+
+** Next Steps**
+ * Evaluate the different scripting languages with respect to interoperability, performance, usability, ..
+ 
+### Graphical UI
+** Next Steps**
+ * Compatibility of GUIs with entity component systems (UI elements as entities <-> only root UI elements)
+ * Integrate an existing UI library (e.g conrod) or develop a new library from scratch (requires further planning then)
```
msiglreith: Something rendered by our gfx backend, not using the native UI toolkits of the OS.

ROADMAP.md
```diff
@@ -0,0 +1,154 @@
+## Vision
+
+> Amethyst aims to be a modular, parallel and data-driven game engine
+
+**Modularity** The game engine should be easily extendable to fit the needs of the game developers for their game idea. Therefore we expose an entity-component system, which allows to define new components and processors without needing to change the game engine code. This should emphasize the 'game engine as library' aspect.
```
mechaxl: How opinionated do we want to be for each subsystem? For example, when we integrate scripting, do we want to pick a particular option as default while still leaving the option of switching it out for something else, or do we leave that decision completely up to the user?

I'm in favor of being opinionated, as that makes getting started for new users a lot easier.

ROADMAP.md
```diff
@@ -0,0 +1,154 @@
...
+Under this assumption we should also try to spend some time on translating our implementations into a working game, showcasing the technical aspects of our engine.
+
+## Systems/Aspects
+Discussing the scopes and features of the different subsystems in more detail.
+
+### Entity-Component System
+Our current system is based on `specs` and should probably stick to it in near future (in case some magically new library pops up..).
+As the ECS is the core part of the architecture, this needs be carefully designed and should be clear under our 3 main vision aspects:
+
+**Modularity** We expose the API from `specs` to the developers, in order to allow them to implement new components, processors and resources (larger singleton objects). The engine will have built-in processors for common operations like `Transform` to ensure spatial hierarchy.
```
mechaxl: What counts as a common operation, e.g. will there be built-in processors for everything in the goals list above, like physics/scripting/AI/etc? Seems like we'd want to be batteries-included, but that does bring up the issue of what happens when someone wants to replace a built-in processor.

ROADMAP.md
Emilgardis: > (...) do we want to pick a particular option as default while still leaving the option of switching it out for something else (...)

Do you mean that we pick what language to use or an actual script doing somehing on default?

ROADMAP.md
Emilgardis: I believe this is what the shard system (#34) is supposed to make possible. Should we also make that into a pr?

ROADMAP.md
Emilgardis: So a keymap, where a config file is used for different input variants.
Should we use yaml for this or something else, I feel like XML is the most used format for keymaps, but Rust is not XSLT.

ROADMAP.md
Aceeri: I'm thinking we could have some basic `specs` systems with a macro to automatically register components and systems when passed in a world.

Emilgardis: So we provide some of these macros defining a world, but the user can also define their own world. That seems like a good way to do it.

ROADMAP.md
diegor8: the op commented in order levereage web tecnology for gui it would be better to use servo , as it is more ligthweight than chormiun , however the projects that would have allowed to utilize it as a framework or library , instead of an embedded browser havent materialized  , as i understand it , the  way to interact with it 3for the near future,   will be by using the cef  api ; finally , it is now expected to run in multiple processes ( again , just as chorme does ) , which could hurt performance 

All

ROADMAP.md
mechaxl: I was talking about picking which language is used by default, it doesn't seem like we'd want to implement the actual scripts themselves by default.

ROADMAP.md
```diff
@@ -74,12 +74,13 @@ Using `gfx-rs` for backend abstraction allows us to provide compatibility with m
 **Next steps**
  * Raw input support (winit?)
  * Gamepad support (winit?)
- * Input Mapping
+ * Input Mapping: Map input events (e.g key press) to abstract action events (`Jump`, `Forward`, ..)
```
nchashch: This sort of can be done now. You can catch keyboard events in `State::handle_events` method and then republish them as abstracts events using `Broadcaster`. Like this:
```rust
fn handle_events(&mut self, events: &[WindowEvent], _: &mut World, _: &mut AssetManager, _: &mut Pipeline) -> Trans {
    use amethyst::event::*;
    let broadcaster = world.write_resource::<Broadcaster>();
    for event in events {
        let movement_event = match event.payload {
            Event::KeyboardInput(_, _, Some(VirtualKeyCode::W)) => Some(MovementEvent::Up),
            Event::KeyboardInput(_, _, Some(VirtualKeyCode::A)) => Some(MovementEvent::Left),
            Event::KeyboardInput(_, _, Some(VirtualKeyCode::S)) => Some(MovementEvent::Down),
            Event::KeyboardInput(_, _, Some(VirtualKeyCode::D)) => Some(MovementEvent::Right),
            _ => None,
        }
        if let Some(movement_event) = movement_event {
            broadcaster.publish(movement_event);
        }
        match event.payload {
            Event::KeyboardInput(_, _, Some(VirtualKeyCode::Escape)) => return Trans::Quit,
            Event::Closed => return Trans::Quit,
            _ => (),
        }
    }
    Trans::None
}
```
`MovementEvent` enum must have a `Component` trait impl and must be registered with `Broadcaster`, also `Broadcaster` must be added to the `world`. Registering `MovementEvent`/adding `Broadcaster` can be done in `State::on_start`.

ebkalderon: This is a good point! 👍 We obviously need to improve the ergonomics of this in the long term by building this functionality into the engine proper. From a UX perspective, matching over simple strings comes to mind, or perhaps an `Action ` tuple struct which contains the string and also the keycode which triggered the action.

nchashch: I don't really like using strings. It would be better to use types instead, so everything can be typechecked and more errors can be cought at compile time.

ebkalderon: Remember the goals of the project, however. Data-driven behavior and real-time hot reloading of logic through text files and scripts is a major goal.

nchashch: I guess we can use types and have text based configs. Maybe by using `serde`. Then for example we could serialize/deserialize a `HashMap<event::Event, MovementEvent>` to/from YAML (or something else `serde` supports).

ebkalderon: That sounds like a decent approach, I think. We can optimize later if better options come our way!

Emilgardis: Serializing with serde seems like the correct approach. In an file we specify something like this
```toml
# TOML syntax in config.toml, I am not familiar with yaml syntax, but should be the same.
[key-map]
up = "w" # all of these should be parsed as lowercase, so case shouldn't matter
down = "s"
left = "a"
right = "d"
run = "shift"
config = "^m" # `^` is CTRL, parsing should support "ctrl+m" aswell.
jump = "space" # is ` ` same as spacebar?
```
I agree we should open an issue specifying the specifics

ROADMAP.md
```diff
@@ -99,7 +100,7 @@ TODO
 
 ** Next Steps**
  * Improve current approach regarding robustness (error handling) and flexibility with scene formats
- * Implement loaders for common file formats and archives
+ * Implement loaders for common file formats and archives. Write `AssetLoader` implements based on external loaders or develope new libraries for other formats.
```
Emilgardis: "develope"

ROADMAP.md
```diff
@@ -111,15 +112,15 @@ TODO: I would consider physics pretty difficult part to do *right*, therefore no
 Maybe rely on external libraries first.
 Should still provide ways for collision detection, allowing to do collision checks, raycasts, etc. (-> ncollide?).
 
-**Parallelism** TODO: internal physic parallelism, possible GPU implementation (hard), difficult with external libraries
+**Parallelism** TODO: internal physic parallelism, possible GPU implementation (hard) (Rust to GPU compilation in the future?), difficult with external libraries
```
Emilgardis: Should probably be changed to:

> Rust - GPU interop

ROADMAP.md
ebkalderon: @Emilgardis Sounds good! I think I'll try prototyping this using `TypeId`s later today or tomorrow, maybe put up a PR if anything comes out of this.

nchashch:
This PR looks abandoned. I think it would be nice to do some strategic planning, but it would probably be better to put it on the wiki or into a separate repo. Anyway, thank you for your work.

msiglreith:
Unfortunatley it didn't spark enough discussions I fear, thanks nonetheless! (I should have pushed it more aggressively!)
Maybe a smaller proposal would be a better way and splitting the different components into sub-issues to keep the discussion more focused at a certain aspect.

kvark:
@msiglreith I really hope nothing from this RFC will be lost at the end of the day.

diegor8:
Further this discussion in the original thread should have taken place before opening this

PR #166, opened 2017-01-31
Adding specular lighting, switching propagation -> attenuation

mechaxl:
Resubmission of #154, as I screwed something up with that PR and can't edit it any longer.

Includes all suggested changes from #154, such as improving lighting calculations based on http://www.frostbite.com/wp-content/uploads/2014/11/course_notes_moving_frostbite_to_pbr.pdf


kvark:
Thanks!

PR #180, opened 2017-02-09
Fix bug in InputHandler

derekdreery:
I think this is a bug. If I'm mistaken feel free to close the pull request. :)

### DIFF
--- a/src/ecs/resources/input.rs
+++ b/src/ecs/resources/input.rs
@@ -45,7 +45,7 @@ impl InputHandler {
                             // second `Pressed` event.
                         }
                         Entry::Vacant(entry) => {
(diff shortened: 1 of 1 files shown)


src/ecs/resources/input.rs
```diff
@@ -45,7 +45,7 @@ impl InputHandler {
                             // second `Pressed` event.
                         }
                         Entry::Vacant(entry) => {
-                            entry.insert(KeyQueryState::Queried);
+                            entry.insert(KeyQueryState::NotQueried);
```
derekdreery: My reasoning here is: when the key is pressed it has not been queried, so key_once (etc.) will return true.

derekdreery:
Closing as the change is in my other request.

PR #181, opened 2017-02-09
Add mouse button events to InputHandler

derekdreery:
Added mouse buttons to the InputHandler (exactly like keys) and run through rustfmt.

I have implemented something locally for mouse movements, but I'm not sure whether there's a better API than the one I'm using so I'm submitting it as a separate PR.

Blocks #186

PR #186, opened 2017-02-10
Implement InputHandler for mouse movement

derekdreery:
This is my state implementation for mouse movement. It adds to PR #181 .

It suits my needs nicely as I can respond to any change in mouse position since the last frame, so I want to see if the behaviour I want is universal.

One particular aspect I invite comments on is defaulting to (0, 0) rather than using an Option. In my model, a movement of (0, 0) is basically a no-op, where as a position of (0, 0) is not a no-op (it's top left window). Does this seem reasonable?

### DIFF
--- a/src/ecs/resources/input.rs
+++ b/src/ecs/resources/input.rs
@@ -64,6 +64,9 @@ impl<'a> Iterator for PressedMouseButtons<'a> {
 pub struct InputHandler {
     pressed_keys: HashMap<VirtualKeyCode, KeyQueryState>,
     pressed_mouse_buttons: HashMap<MouseButton, KeyQueryState>,
+    // defined to match glium
+    mouse_position: Option<(i32, i32)>,
+    previous_mouse_position: Option<(i32, i32)>,
 }
 
 impl InputHandler {
@@ -72,6 +75,8 @@ impl InputHandler {
         InputHandler {
             pressed_keys: HashMap::default(),
             pressed_mouse_buttons: HashMap::default(),
+            mouse_position: None,
+            previous_mouse_position: None,
         }
     }
 
@@ -97,9 +102,14 @@ impl InputHandler {
                 Event::MouseInput(ElementState::Released, button) => {
                     self.pressed_mouse_buttons.remove(&button);
                 }
+                Event::MouseMoved(x, y) => {
+                    self.mouse_position = Some((x, y));
+                }
                 Event::Focused(false) => {
                     self.pressed_keys.clear();
                     self.pressed_mouse_buttons.clear();
+                    self.mouse_position = None;
+                    self.previous_mouse_position = None;
                 }
                 _ => {}
             }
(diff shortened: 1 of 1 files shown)


src/ecs/resources/input.rs
```diff
@@ -13,26 +13,68 @@ enum KeyQueryState {
...
+///
+/// ```ignore
+/// fn handle_events(&mut self,
+///                  events: &[WindowEvent],
+///                  _: &mut World,
+///                  _: &mut AssetManager,
+///                  _: &mut Pipeline)
+///                  -> Trans {
+///     // ...
+///     input_handler.update(events);
```
Emilgardis: This wouldn't work. Show how to get `input_handler` aswell.

src/ecs/resources/input.rs
```diff
@@ -13,26 +13,68 @@ enum KeyQueryState {
...
+///     // ...
+/// }
+/// ```
 pub struct InputHandler {
     pressed_keys: HashMap<VirtualKeyCode, KeyQueryState>,
+    // use KeyQueryState for mouse button query state as well
+    pressed_mouse_buttons: HashMap<MouseButton, KeyQueryState>,
+    // defined to match glium
+    mouse_position: Option<(i32, i32)>,
+    previous_mouse_position: Option<(i32, i32)>,
```
Emilgardis: Why does `InputHandler` store the previous mouse position?

src/ecs/resources/input.rs
```diff
@@ -48,30 +90,46 @@ impl InputHandler {
                             // second `Pressed` event.
                         }
                         Entry::Vacant(entry) => {
-                            entry.insert(KeyQueryState::Queried);
+                            entry.insert(KeyQueryState::NotQueried);
```
Emilgardis: This is wierd, I remember that this was working, but following my logic in the code of key_once it shouldn't work.
Note also that you don't reflect this change in the commit.

Emilgardis:
Also, should #181 be closed inplace of this PR? Seems like they are identical except for `b54a4dc`

derekdreery:
I split the PR up because I thought that the mouse stuff might be more controversial.

src/ecs/resources/input.rs
derekdreery: previous_mouse_position is the mouse position the previous time mouse_change was called. It allows us to compare to the current mouse position and get the amount the mouse has moved. Using this information,  you can work out for example how much to rotate the camera in the ecs etc.

src/ecs/resources/input.rs
derekdreery: Good point!

derekdreery:
I just rebased this on my other PR

torkleyy:
@derekdreery It's not blocked anymore, please rebase onto the current master. I think we can do #225 after this PR is merged.

derekdreery:
This is rebased on current upstream/develop.

PR #192, opened 2017-02-16
Replace constant paddle width with actual value

Hammster:
Just stumbled about this when I changed the dimensions for the ball and paddle. If the value is constant the ball gets positioned inside the paddle. 

Cheers 😃


ebkalderon:
No objections, so I'm merging this now. Thank you again for the fix! ❤️

PR #193, opened 2017-02-20
Make rendering parallel using a scoped_threadpool::Pool

nchashch:
How it works:
1. `Fragment`s vector is chunked up into num_cpus equal parts, the remainder is added to the 0th chunk.
2. Each chunk is passed to a threadpool thread along with a `&mut Encoder` (one for each thread) and `&Renderer`, `&Pipeline`, `&Scene` (shared for all threads).
3. `Renderer::submit` is called from each thread.
4. All `Encoder`s are flushed at the end of `GfxDevice::render_world` call.

Problems:
- 2*num_cpus threads are run, for `specs::Planner` and for a rendering threadpool. Fixing this would require adding some sort of data parallelism to `specs` (slide-rs/specs#54), which would be useful for other things besides rendering.

Changes:
- `Application` now owns a `scoped_threadpool::Pool`.
- There is no more `MainTarget` struct.
- `GfxDevice` owns num_cpus `Encoder`s one for each thread.
- `Renderer` no longer owns an `Encoder`.
- There is no more `Vec<Fragment>` in `renderer::Scene`.
- `Fragment`s are now passed as `&[Fragment]` to `Renderer::submit` method.
- There is no more `Clear` pass.
- `ecs::resources::ClearColor` was added (holding `clear_color` and `clear_depth`). So the screen is cleared once by calling `self.encoder[0].clear(...)` in `GfxDevice::render_world` before rendering anything.


torkleyy:
@nchashch Could you somehow change your PR to use just `threadpool`? Or would cloning be impossible / too much cost? Because I found a way to reuse the `threadpool` from specs for #195.

nchashch:
I am currently trying to replace `scoped_threadpool` with `futures-cpupool`. It looks like it could work, although I will have to use `Arc`s instead of plain references to pass `Renderer`, `Pipeline`, `Scene`, and `Vec<Fragment>`.

I was thinking about replacing `threadpool` in `specs` with `futures-cpupool` and then exposing something like `Planner::spawn` method, to allow spawning custom `Future`s unrelated to ECS. Then there won't be a need for `AssetLoader` or for `scoped_threadpool::Pool` in `Application`.

torkleyy:
@nchashch I think we are working in the opposite direction right now. I think just using `threadpool` would be better (it's used in specs, I've restructured my project and using futures-cpupool would be more expensive for asset loading than threadpool and I can still make it a `Future` with threadpool).

ebkalderon:
Given that both Amethyst and Specs appear to be settling mostly on Rayon for parallelism, I think we can close this pull request for now. Thanks anyways for sparking some interesting debate, @nchashch!

PR #194, opened 2017-02-23
Add a `from_string` method to `Element` of the config crate

torkleyy:
This way, the code of a config file can be embedded into the source file.

I'm trying to get Android support for this engine and this would be one step which is necessary.

### DIFF
--- a/src/config/src/yaml.rs
+++ b/src/config/src/yaml.rs
@@ -182,7 +182,7 @@ pub trait Element: Sized {
         let found_path = found[0].clone();
         next_meta.path = found_path.clone();
 
-        let mut file = File::open(found_path.as_path()).map_err(|e| ConfigError::FileError(found_path.clone(), e));
+        let mut file = File::open(found_path.as_path()).map_err(|e| ConfigError::FileError(found_path.clone(), e))?;
         let mut buffer = String::new();
 
         file.read_to_string(&mut buffer)
--- a/src/config/src/yaml.rs
+++ b/src/config/src/yaml.rs
@@ -117,7 +117,7 @@ pub trait Element: Sized {
     fn from_string<'a>(src: &'a str) -> Result<Self, ConfigError> {
         let mut meta = ConfigMeta::default();
         meta.path = PathBuf::from("");
-        let yaml = try!(YamlLoader::load_from_str(src).map_err(|e| ConfigError::YamlScan(e)));
+        let yaml = YamlLoader::load_from_str(src).map_err(|e| ConfigError::YamlScan(e))?;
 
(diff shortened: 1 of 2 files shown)


agmcleod:
Not a contributor, but I do wonder if the example should be updated. Encourage those to use files instead of strings. I like the capability though :)

ebkalderon:
@agmcleod I think we're mostly waiting on @Aceeri's project module to be finished. 😛

src/config/src/yaml.rs
```diff
@@ -113,6 +113,21 @@ pub trait Element: Sized {
         None
     }
 
+    /// From some string (should be used for top level elements if you want to embed the code)
+    fn from_string<'a>(src: &'a str) -> Result<Self, ConfigError> {
```
ebkalderon: Does using `&'static str` not work here?

src/config/src/yaml.rs
```diff
@@ -113,6 +113,21 @@ pub trait Element: Sized {
...
+        let mut meta = ConfigMeta::default();
+        meta.path = PathBuf::from("");
+        let yaml = try!(YamlLoader::load_from_str(src).map_err(|e| ConfigError::YamlScan(e)));
```
ebkalderon: I think you should stick to the more idiomatic `?` notation instead of `try!`, if possible.

src/config/src/yaml.rs
torkleyy: @ebkalderon One use case of `from_string` would be to load the config from an android asset, which one can only read as bytes. Therefore the lifetime is necessary.

ebkalderon: That makes sense. I'll let it slide. 😄

ebkalderon:
@amethyst/engine-devs Everyone okay with these changes? If so, I'm merging it.

src/config/src/yaml.rs
nabijaczleweli: Do you need the explicit lifetime, tho? You're not returning a reference, so there seems to not be any use for it

nabijaczleweli:
I added a line comment on there so hold off until a response from PRer maybe?

Emilgardis:
This is good. We should tell the user that a build step is recommended where a config file is copied into the out folder and imported using `include_str!()` macro.

torkleyy:
@nabijaczleweli I removed the explicit lifetime and the checks have passed. Are there any other changes needed?

nabijaczleweli:
You're golden, fam :+1:

ebkalderon:
Alright, let's merge this sucker. Thanks for the contribution, @agmcleod!

PR #195, opened 2017-02-26
Restructuring of the asset management

torkleyy:
This PR tries to rework and improve the current asset management, which in my opinion is already good but has three negative points:

* not flexible enough: it decides too much for the user (loads assets into a hash map, has multiple stores to load from, ..)
* not fast enough: no parallelization, hash map of closures (-> dynamic dispatch)
* not too easy to comprehend and somehow "dirty" because of `Any` and casting 

I have been thinking and working on this design for many, many hours and I hope you like it!

Note that this PR is not complete yet, see the To do list at the end for the current progress.

## Contents

This (overly long) PR comment is structured as shown here:

* **What should asset management look like in amethyst?**
* **How can we solve these tasks?**
* **Further improvements**
* **Drawbacks**
* **Unresolved questions**
* **What has to be done?**

## What should asset management look like in amethyst?

In my opinion, the following criteria are important:

The core properties should be

* Flexibility
* Performance
* Extensibility

### Flexibility

* Location of the assets (managed with `AssetStore`s)
    * Assets are allowed to be in a directory, on a web server, in a zip file or any other arbitrary source you can read byte-wise
    * However, the user should explicitly state where the asset is (you usually do not want to look up on the web server if you have the file locally, but if you still want you can create a custom asset store which manages this for you)
* Allow different formats to be used, but provide a recommended default binary format which can be imported (decision needed: should the amethyst cli do this or a `build.rs` automatically?)
* Don't assume requirements, but create a general purpose asset management
* Let the user decide which error is recoverable and which not -> use `Result` in more places
* Provide a `Default` asset store which loads the asset from a directory on desktop platforms and to the embedded assets on other platforms, in order to make another step towards a cross platform engine
* I think the assets should decide what to do with the asset (-> not store it automatically, but give the option to do so); the user should decide what to do (most likely he wants to use it as a component and he may want to manage it himself (without a `HashMap`))

These features allow great flexibility, by letting the user choose what he needs, following one of Rust's principles: Pay only for what you need, in order to gain performance.

### Performance

* As already mentioned in the previous section, only pay what you want for
* Do not use `HashMap`s and `Any` everywhere
* Parallelize asset management
* Avoid too much dynamic dispatch
* Implement custom binary formats which exactly match the internals of the engine

### Extensibility

* Provide the user with a general purpose asset loader, which can load
    * any format
    * any asset
    * from any store
* Also, the engine does not cache assets, but we may provide a `CachedAssetManager`, so the user can choose; again, the user should choose what he needs

All of these can also be implemented by the engine user.

## How can we solve these tasks?

Instead of registering loaders and formats and stores, let the compiler search for the right asset loader:

Every type you want to load from a store, has to implement the `Asset` trait, which only contains an associated type:

```rust
pub trait Asset: Sized {
    type Data;
    type Error: Debug;
    // This may also be the place to provide some other metadata if that's necessary

    // Context contains the gfx factory
    fn from_data(data: Self::Data, context: &mut Context) -> Result<Self, Self::Error>;
}
```

Why do we want this? The `Data` type should be some pure data struct which contains the imported stuff without any modification. For example, think about a texture. First we want to load colors and dimensions into this `Data` format, which can then be uploaded to the graphics card. This way, we can easily implement a way to convert from one format to another and the implementor of the file format does not need to know the internals of the engine.

#### How does import and export work?

With two very simple traits:

```rust
pub trait Import<T> {
    fn import(&self, bytes: Box<[u8]>) -> Result<T, Error>;
}

// May be added in the future, not by this PR
pub trait Export<T> {
    fn export<W: Write>(&self, stream: W, data: T) -> Result<(), Error>;
}
```

#### How does the compiler know how to load an asset?

This is pretty handy, considering the performance gains and the cleaner structure:

Previously, an asset was loaded like this:

```rust
assets.load_asset::<Mesh>("monkey", "obj");
```

---

With the proposed design, it will work this way:

```rust
// load_default will load with the `Default` store
let monkey = loader.load_default("monkey", Obj);
```

`Obj` here is ZST which implements `AssetFormat`. And this `Obj` also implements `Import<MeshData>` where `Mesh` can be created from a `MeshData` (currently called `VertexPosNormal`). This way the compiler now replaces our hash map, meaning everything can be statically dispatched.

So this design is more data-oriented than the previous one; it is as simple as

`Type`, `Name`, `Store`, `Format` -> `Type`, `Name`, `Bytes`, `Format` -> `Type`, `Data` -> `Asset`
| Parameters of `AssetLoader` | After reading from the store | After `Import`ing | Output of `from_data` |

The user may then display some loading screen and wait for the assets to finish loading.

There will be an `AssetLoader` which just returns a [`Future`](https://github.com/alexcrichton/futures-rs). So parallelization is solved with a `CpuPool`. Additionally, if you want more convenience and a better runtime performance, there'll be a way to read that, at compile time, from a scene file.

## Further improvements

* Uses `Result` instead of `Option`
* Gets rid of `has_asset` which has a bug (it will return true if there is a directory called like the asset)
* Provides better documentation
* By storing assets in variables, unused assets can be spotted easier (the "pink" texture in the renderable example)
* Another step towards android support
* The user can decide himself which asset loader he wants to use because he creates it (`AssetLoader` or a custom one)

## Drawbacks

* This is a breaking change
* At the moment, only `'static` stores and formats can be used for parallel loading (single threaded is still supported) because of a limitation in the `future` crate
* A bit uglier syntax when working with `AssetLoader` manually (fixed by scene files)

## Unresolved questions

* Are there missing features with this approach which can not be fixed without changing APIs (is this PR forward-compatible enough)?
* How to change this design to allow for parallel asset creation (non blocking `from_data`) in the future without changing `Data` to be more than just data?

## What has to be done?

- [x] Create core design for this system
- [x] Finish core traits
- [x] Implement `DefaultStore`
- [x] Implement `DirectoryStore`
- [x] Fix import implementations
- [x] Provide good documentation with implementation examples
- [x] Implement image and obj file formats
- [x] Add `Engine` struct and do refactoring
- [x] Fix examples
- [x] Fix and write unit tests
- [x] Implement `load_now`
- [x] Edit the book if necessary
- [x] Delete `asset_manager.rs` (previously, the `mod.rs` only contained a reexport)
- [x] Answer unresolved questions

I think the following tasks can be implemented by other PRs in the future:

* Scene format (I'll do that after this PR is merged)
* Implement custom binary format for meshes (I'll do that after this PR is merged)
* Implement a binary format for packaging assets
* Implement a web asset store
* Provide easy conversion (`build.rs` / amethyst cli)

That's it! I'm inviting everybody to make further suggestions and to tell me what you think about this!


src/asset_manager/asset.rs
```diff
@@ -0,0 +1,184 @@
...
+    /// The data type, an intermediate format.
+    /// This may also be `Self` if this asset does
+    /// not depend on `Context`.
+    type Data;
+    /// The error type that may be returned if
+    /// `from_data` fails.
+    type Error: Debug;
+
+    /// Create the asset from the data and the context (used to create buffers for the gpu).
+    fn from_data(data: Self::Data, context: &mut Context) -> Result<Self, Self::Error>;
```
msiglreith: Does this need to take `Context` as it limits the users a bit as they can't extend it without changing the library?

src/asset_manager/formats.rs
```diff
@@ -0,0 +1,228 @@
...
+// Image formats
+// -----------------------------------------------------------------------------------------
+
+use imagefmt;
+use dds::{Header, DDS};
+
+use ecs::components::{TextureData, TextureLoadData};
+use gfx::texture::{Kind, AaMode};
+
+macro_rules! replace_expr {
```
msiglreith: Maybe move these macros up, out of the image format section.

src/ecs/components/rendering/texture.rs
```diff
@@ -1,48 +1,53 @@
...
 }
 
-impl AssetLoader<Texture> for [f32; 4] {
-    fn from_data(_: &mut Assets, color: [f32; 4]) -> Option<Texture> {
-        Some(renderer::Texture::Constant(color))
+impl Asset for Texture {
+    type Data = TextureData;
+    type Error = GfxError;
+
+    fn from_data(data: TextureData, context: &mut Context) -> Result<Self, GfxError> {
```
msiglreith: Could there be a way to do the texture creation part `import` for backends with parallel factory support? So OpenGL would fallback to this way, for other backends like vulkan or d3d12 this would only a pass-through and the actual creation would happen in the `import()`.

torkleyy: Yes, with conditional compilation we could define `<Texture as Asset>::Data` as `Self` and do everything in `import`. But I thought glx wouldn't support Vulkan yet?

torkleyy:
Please wait for your review, I'll push some changes in a few minutes. Sorry, I needed longer than planned.

torkleyy:
So the current asset loading isn't really syntactically beautiful, however, by adding compile-time generated scenes (with some external file), this isn't a problem anymore. I still have to fix the test, but the examples should work now.

src/ecs/components/rendering/texture.rs
msiglreith: > But I thought glx wouldn't support Vulkan yet?

Not yet, but working on it!

Anyway, I assume in the future we could split the `Factory` into a single-threaded and a parallel part. The first one would be used in `from_data` and the parallel in `import`. This would be pretty neat imo.

src/asset_manager/asset.rs
msiglreith: I understand your point concerning passing `context` directly, which simplifies handling (e.g for the AssetFuture). Personally, I would consider making `Context` an associated type also an option to consider. This avoids blocking the context and allows the user to pass own related data for loading, which might not be contained in the context provided by the engine. But that's not a blocker from my side, other opinions on regarding this topic would be appreciated!

src/ecs/components/rendering/texture.rs
torkleyy: I've thought about this again, and what I wrote above also has some drawbacks:

I wanted the `Data` type to be just plain data, in order to allow

* Reexporting
* Compile-time importing

Which is only possible if `import` only imports the data. If we already had specialization, I could specialize for `T: Send`, in which case `from_data` should be executed async.

So the only solution I'd knew is to have two import steps from the data (`from_data_async` and `from_data`) which return `Async` and `Asset` (-> another associated type). I think that would complicate it again, but it would of course allow for great flexibility

src/asset_manager/asset.rs
torkleyy: What I meant by my second point is that I cannot implement `AssetManager` anymore then.

```rust
pub struct AssetManager {
    loader: AssetLoader,
    submissions: Vec<AssetFuture>,
    assets: HashMap<TypeId, Box<Any>>,
}

impl AssetManager {
    pub fn submit<F>(&mut self, name: &str, format: F) {
        self.submissions.push(self.loader.load_default(name, format));
    }

    pub fn wait(&mut self, context: &mut Context) {
        // If context is an associated type,
        // I cannot pass it here

        // I'd have to use something
        // like `HashMap<TypeId, Box<Any>>
        // which provides the parameters,
        // which is ugly

        while let Some(future) = self.submissions.pop() {
            let asset = future.finish( /* What to pass here? */ );
            // Store asset in hash map
        }
    }
}
```

torkleyy:
Unfortunately, I had to remove the test in `engine/state.rs` because it would now require an `Engine` to be created, which is not supported on Travis.

torkleyy:
@msiglreith I see that your approach (with the associated type instead of `Context`) is better, I also found a solution to uphold flexibility, however there's one big problem:

You cannot use lifetimes in associated types, meaning we cannot pass a `&mut Context` to the Texture / Mesh which is a problem I think. Does anybody know how we could work around that until we have Associated Type Constructors?

torkleyy:
I've worked on an implementation [here](https://github.com/torkleyy/amethyst/tree/assets-async) which would allow what @msiglreith wrote:

> Could there be a way to do the texture creation part import for backends with parallel factory support? So OpenGL would fallback to this way, for other backends like vulkan or d3d12 this would only a pass-through and the actual creation would happen in the import()

while still allowing to reexport / compile-time import data. I think it would work somehow.

However, it seems that it isn't possible to make the parameter for `from_data` / `from_asset` generic atm (see my previous comment).
The only workaround I'd knew was to make them type parameters, which would add two type parameters more everywhere in the asset system.

Could you please review my changes?
Because I think at the current stage the design of this PR should be pretty clear and I want to hear what extensions you'd like or what you suggest?

@nchashch 
@ebkalderon 
@msiglreith

examples/05_assets/main.rs
```diff
@@ -4,60 +4,51 @@
...
 
 // Implement custom asset loader that reads files with a simple format of
 // 1 vertex and 1 normal per line, with coordinates separated by whitespace.
-struct CustomObj {
-    vertices: Vec<[f32; 3]>,
-    normals: Vec<[f32; 3]>,
+struct CustomObj;
+
+impl AssetFormat for CustomObj {
+    fn file_extensions(&self) -> &[&str] {
```
msiglreith: Could you replace this with the `string_array!()` macro?

src/asset_manager/asset.rs
```diff
@@ -0,0 +1,200 @@
...
+///         Ok(data) // Ommitted here: Check for size
+///     }
+/// }
+///
+/// impl AssetFormat for WhitespaceTable {
+///     fn file_extensions(&self) -> &[&str] {
+///         const FE: [&'static str; 1] = ["wst"];
+///         const FE_REF: &'static [&'static str; 1] = &FE;
+///
+///         FE_REF
```
msiglreith: Should also encourage the user to use the `string_array!()` macro if possible.

src/asset_manager/asset.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// Return the typical file extensions a file
+    /// with this format has, without the preceding `"."`.
+    /// If there is no asset with the first extension, the
+    /// next ones will be tested fo existence.
+    ///
+    /// # Panics
+    ///
+    /// May cause a panic if the length of
+    /// file extensions is 0! Do not do this.
+    fn file_extensions(&self) -> &[&str];
```
msiglreith: Can we remove the `&self` borrow, or are there use-cases in which this might be useful?

src/asset_manager/asset.rs
```diff
@@ -0,0 +1,200 @@
...
+/// the asset format is not the same as the type.
+/// There may be multiple formats for the same
+/// asset. A format should also implement the
+/// `Import` trait for the target `Asset`s data
+/// type.
+pub trait AssetFormat {
+    /// Return the typical file extensions a file
+    /// with this format has, without the preceding `"."`.
+    /// If there is no asset with the first extension, the
+    /// next ones will be tested fo existence.
```
msiglreith: typo: `fo`/ `for`

src/asset_manager/asset.rs
```diff
@@ -0,0 +1,200 @@
...
+    ///
+    /// # Panics
+    ///
+    /// May cause a panic if the length of
+    /// file extensions is 0! Do not do this.
+    fn file_extensions(&self) -> &[&str];
+}
+
+/// An asset store may be a ".zip" file, a server,
+/// a custom binary format conataining levels, or just
```
msiglreith: typo: `conataining`

src/asset_manager/common.rs
```diff
@@ -0,0 +1,243 @@
...
+              F: AssetFormat
+    {
+        let mut last_error = None;
+
+        let mut path = PathBuf::from(path.as_ref());
+
+        for extension in format.file_extensions() {
+            path.set_extension(extension);
+
+            println!("Trying path {:?}", &path);
```
msiglreith: Is this debug log? Hm, not sure if we support the log macros (`debug!`, `info!`, etc.)

src/asset_manager/common.rs
```diff
@@ -0,0 +1,243 @@
...
+
+#[cfg(not(android))]
+impl AssetStore for DirectoryStore {
+    fn read_asset<T, F>(&self, name: &str, format: &F) -> Result<Box<[u8]>, AssetStoreError>
+        where F: AssetFormat,
+              T: Asset
+    {
+        let path = self.path.join(T::category()).join(name);
+
+        println!("path: {:?}", path);
```
msiglreith: Also, debugging only?

src/asset_manager/formats.rs
```diff
@@ -0,0 +1,253 @@
...
+/// loading a texture from it.
+pub struct Dds;
+
+fn read_image<F>(bytes: Box<[u8]>, f: F) -> Result<TextureData, ImportError>
+    where F: FnOnce(&mut Cursor<&[u8]>, imagefmt::ColFmt) -> imagefmt::Result<imagefmt::Image<u8>>
+{
+    let mut bytes_ref = Cursor::new(bytes.as_ref());
+
+    let data = f(&mut bytes_ref, imagefmt::ColFmt::RGBA)
+            .map_err(|x| ImportError::FormatError(format!("Failed to load png: {}", x)))?;
```
msiglreith: The error message might be a bit confusing for loading non-png files.

src/asset_manager/mod.rs
```diff
@@ -1,4 +1,221 @@
...
+    /// out. Note that this does not necessarily mean
+    /// that the data is invalid.
+    DataError(T::Error),
+}
+
+impl AssetLoader {
+    /// Creates a new asset loader with a cpu pool
+    /// using the number of cpu cores for the number of threads.
+    pub fn new() -> Self {
+        AssetLoader { cpupool: CpuPool::new_num_cpus() }
```
msiglreith: I hope we can re-use the thread pool from our ecs for this in near future!

src/engine/app.rs
```diff
@@ -3,25 +3,55 @@
...
 use engine::state::{State, StateMachine};
 use engine::timing::Stopwatch;
 use gfx_device;
-use gfx_device::{DisplayConfig, GfxDevice, gfx_types};
+use gfx_device::{DisplayConfig, GfxDevice};
+use gfx_device::gfx_types::Factory;
 use renderer::{AmbientLight, DirectionalLight, Pipeline, PointLight, target};
 
+/// A context, which stores structs
+/// that are required for asset instantioation,
```
msiglreith: typo: `instantioation`

src/asset_manager/common.rs
```diff
@@ -0,0 +1,243 @@
...
+impl AssetStore for ZipStore {
+    fn read_asset<T, F>(&self, name: &str, format: &F) -> Result<Box<[u8]>, AssetStoreError>
+        where F: AssetFormat,
+              T: Asset
+    {
+        let mut pathbuf = PathBuf::from(T::category()).join(name);
+
+        for extension in format.file_extensions() {
+            pathbuf.set_extension(extension);
+            let s = pathbuf.to_str().unwrap();
```
msiglreith: Can we return an error instead of just unwrapping here?

src/asset_manager/common.rs
```diff
@@ -0,0 +1,243 @@
...
+
+        Self::from(stream)
+    }
+
+    /// Open a zip file and read its contents
+    #[cfg(android)]
+    pub fn open<P: AsRef<Path>>(p: P) -> Result<Self, IoError> {
+        use std::io::Cursor;
+
+        let stream = read_bytes(p.to_str().unwrap())?;
```
msiglreith: (Same as below)

examples/05_assets/main.rs
torkleyy: Do we really want to export the macro? I just thought I'd leave it internal. Oh I found another way to return a `&'static [&'static str]`:

```rust
const EXTENSIONS: &'static [&'static str] = &["jpg"];

EXTENSIONS
```

Would this one be okay or should I still export it?

src/asset_manager/asset.rs
torkleyy: It was an associated function, but I thought this way it's more flexible if someone wants to create a more dynamic format. I'm unsure if there are any practical use cases though.

src/asset_manager/common.rs
torkleyy: Ah! I was debugging it (because I used `.join()` instead of `set_extension` and forgot to remove it

src/asset_manager/common.rs
torkleyy: The unwrap only panics if there is invalid Unicode which is so rare that I wouldn't handle it

src/asset_manager/mod.rs
torkleyy: For that to work `specs` has to use `futures` though. I'll try to get that working, but it will take me quite some time because I don't know how it's working exactly atm (with this `Pulse` and the channel)

Aceeri:
> Scene format (I'll do that after this PR is merged)

If you are talking about serializing/deserializing the ecs world and such it may be a bit more difficult than it looks due to trait object limitations. Or at least there isn't a very clean way to do it as far as I've seen.

torkleyy:
@Aceeri I planned to do it at compile time.

src/asset_manager/common.rs
```diff
@@ -0,0 +1,239 @@
...
+              T: Asset
+    {
+        let mut pathbuf = PathBuf::from(T::category()).join(name);
+
+        for extension in F::file_extensions() {
+            pathbuf.set_extension(extension);
+            let s = pathbuf.to_str().unwrap();
+
+            if let Some(bytes) = self.files.get(s) {
+                return Ok(bytes.clone());
```
torkleyy: (see summary)

src/asset_manager/mod.rs
```diff
@@ -1,4 +1,221 @@
...
+    /// If you already have the asset data and you just want to
+    /// import it, use `MyAsset::from_data(data, context)`.
+    pub fn load<T, S, F>(&self, store: &S, name: &str, format: F) -> AssetFuture<T>
+        where T: Asset + 'static,
+              T::Data: Send + 'static,
+              T::Error: Send + 'static,
+              S: AssetStore + Clone + Send + Sync + 'static,
+              F: AssetFormat + Import<T::Data> + Send + 'static
+    {
+        let store: S = store.clone();
```
torkleyy: (see summary)

src/asset_manager/common.rs
torkleyy: Is there a way this could be optimized out if `import` does not consume the bytes?

src/asset_manager/mod.rs
torkleyy: Here is the PR: https://github.com/slide-rs/specs/pull/102

I'm not exactly sure where I messed it up, but it doesn't pass yet.

examples/05_assets/main.rs
msiglreith: I would consider it easier to read with the macro (highly subjective), but I'm also fine with the current solution!

src/asset_manager/asset.rs
```diff
@@ -0,0 +1,200 @@
...
+    fn file_extensions() -> &'static [&'static str];
+}
+
+/// An asset store may be a ".zip" file, a server,
+/// a custom binary format conataining levels, or just
+/// a directory.
+pub trait AssetStore {
+    /// Read an asset from a given name and format and
+    /// return the bytes.
+    fn read_asset<T, F>(&self, name: &str, format: &F) -> Result<Box<[u8]>, AssetStoreError>
```
msiglreith: Minor: Can remove `format: &F` here imo, because `AssetFormat` only has `file_extensions()`, which is static.

src/asset_manager/mod.rs
```diff
@@ -1,4 +1,287 @@
...
+///
+/// #### Why isn't it available immediately?
+///
+/// The reason is that assets are loaded in
+/// separate threads, without blocking the
+/// main thread. Once they're finished,
+/// you can access it.
+///
+/// As soon as you need the asset, you call
+/// `finish` on this future which will return
```
nchashch: In order to have asset loading that doesn't block the main thread [Future::poll](https://docs.rs/futures/0.1.10/futures/future/trait.Future.html#tymethod.poll) should be used instead of [Future::wait](https://docs.rs/futures/0.1.10/futures/future/trait.Future.html#method.wait) (it is used inside `AssetFuture::finish`).

Probably the best way to achieve this would be to add a `AssetLoader` to `Engine` struct. Then `AssetFuture::wait` can be called repeatedly from `State::update` (as described above in my `03_renderable` example comment). Then this doc comment would have to be updated to reflect the proper usage.

src/asset_manager/mod.rs
nchashch: If #193 and #195 are both merged there would be `3*num_threads` threads. Which I might be a problem, because there is overhead for switching between threads. Modern desktop CPUs may have 16 virtual cores, so an `amethyst` instance then might have to spin up 48 threads.

I think the best solution would be to make `specs` API flexible enough to allow reusing the underlying threadpool for both rendering and asset loading. Maybe we could even move "checking if asset is loaded and adding it to the `World`" code into something like a `specs::System` from `State::update`.

src/engine/app.rs
```diff
@@ -3,25 +3,55 @@
...
 use gfx_device;
-use gfx_device::{DisplayConfig, GfxDevice, gfx_types};
+use gfx_device::{DisplayConfig, GfxDevice};
+use gfx_device::gfx_types::Factory;
 use renderer::{AmbientLight, DirectionalLight, Pipeline, PointLight, target};
 
+/// A context, which stores structs
+/// that are required for asset instantiation,
+/// like the gfx factory.
+pub struct Context {
```
nchashch: Maybe expose `Factory` directly? It has a more descriptive name and It is `pub` anyway. I am not sure if something like a `Factory` would be requried for loading audio files. For example `lewton` (ogg loading library) doesn't need it, you just create an [`OggStreamReader`](https://github.com/RustAudio/lewton/blob/master/examples/perf.rs#L28) from a file.

torkleyy:
The performance improvements aren't too high (most likely because of the OpenGL restriction):

Old asset management: 7,770,539 ns/iter
New, single threaded: 7,722,735 ns/iter
New, multi threaded: 5,924,114 ns/iter

Well, it is an improvement. I think we can do further optimizations in this asset management, but it should be okay for now

torkleyy:
Note of course that this benchmark isn't a real usecase and the performance gains might be higher in an actual game, espicially with the changes suggested by @nchashch

torkleyy:
@nchashch 

> I think the best solution would be to make specs API flexible enough to allow reusing the underlying threadpool for both rendering and asset loading. Maybe we could even move "checking if asset is loaded and adding it to the World" code into something like a specs::System from State::update.

Yes, it would be great if asset loading worked like a system. But this won't work that well, because of line 204 of `engine/app.rs`:

```rust
self.engine.planner.wait();
```

So the asset loader is pretty limited...

src/asset_manager/mod.rs
```diff
@@ -215,14 +234,55 @@ impl AssetLoader {
...
+    }
+
+    /// Process a number of asset datasets
+    /// as long as `as_long` returns true.
+    ///
+    /// Intended to be used from the internals
+    /// of the engine to allow loading assets
+    /// as long as the wished frame time wasn't reached
+    /// yet.
+    pub fn process<F: FnMut() -> bool>(&mut self, context: &mut Context, mut as_long: F) {
```
torkleyy: I changed asset management quite a bit, so that it now supports the features requested by @nchashch. Additionally, I think this implementation is quite nice: If there is still time left in this frame, then assets will be created as long as we didn't reach `(1/FPS)*0.95`. If the computer is too slow for the requested frames per second, it will only load one asset in order not slow it down too much

src/asset_manager/asset.rs
```diff
@@ -0,0 +1,200 @@
...
+/// ```
+pub trait Asset: Sized {
+    /// The data type, an intermediate format.
+    /// This may also be `Self` if this asset does
+    /// not depend on `Context` (if it is just
+    /// plain data).
+    type Data;
+    /// The error type that may be returned if
+    /// `from_data` fails.
+    type Error: Debug;
```
Emilgardis: This should be [`std::error::Error`](https://doc.rust-lang.org/std/error/trait.Error.html)

torkleyy: yes, feels right, but what if it cannot fail? `()` does not implement `Error`

Aceeri: Well in the case that it cannot fail couldn't the person just have a dummy type?
e.g.
```rust
struct NoFail;
impl std::error::Error for NoFail {
    fn description(&self) -> &str {
        "This should never occur"
    }
}
```
I can't think of many asset loaders that wouldn't have some possibility of failure.

torkleyy: @Aceeri yes, there aren't too much, but `Mesh` for example can not fail.

Emilgardis: We can probably define a NoFail error for the user

torkleyy:
Okay guys, thanks for the feedback so far! Because I changed the basic idea multiple times and the history is pretty much screwed up, I think I'll replace this PR with 2 or more smaller ones, the first being the `specs` upgrade.

I was pretty busy the last weeks with school, but now I have some free time. I hope I'm not blocking `0.5.0` release for too long!

src/asset_manager/asset.rs
SCareAngel: It must be
```
enum NoFail {}
```

torkleyy: @SCareAngel This PR will be replaced anyways, but you're right of course.

torkleyy:
Replaced by #244 Thanks for the great discussion everybody!

Issue #188, opened 2017-02-14
Project template fails to build


assumptionsoup:
>\> $ cargo run
>     Updating registry `https://github.com/rust-lang/crates.io-index`
>    Compiling bitflags v0.7.0
>    Compiling gfx_core v0.5.1
>    Compiling gfx_core v0.6.0
>    Compiling xml-rs v0.3.6
>    Compiling gl_generator v0.5.2
>    Compiling gfx_gl v0.3.1
>    Compiling glutin v0.7.4
>    Compiling gfx_device_dx11 v0.4.0
>    Compiling gfx v0.14.0
>    Compiling gfx_device_gl v0.13.0
>    Compiling gfx_window_dxgi v0.5.0
>    Compiling gfx_window_glutin v0.14.0
>    Compiling amethyst_renderer v0.4.1
> error[E0412]: type name `ColorFormat` is undefined or not in scope
>    --> C:\Users\jordan\.cargo\registry\src\github.com-1ecc6299db9ec823\amethyst_renderer-0.4.1\src\pass\deferred.rs:261:31
>     |
> 261 |         out: gfx::BlendTarget<ColorFormat> = ("o_Color", gfx::state::MASK_ALL, gfx::preset::blend::MULTIPLY),
>     |                               ^^^^^^^^^^^ undefined or not in scope
>     |
>     = help: you can import it into scope: `use target::ColorFormat;`.
> 
> error[E0412]: type name `ColorFormat` is undefined or not in scope
>    --> C:\Users\jordan\.cargo\registry\src\github.com-1ecc6299db9ec823\amethyst_renderer-0.4.1\src\pass\deferred.rs:261:31
>     |
> 261 |         out: gfx::BlendTarget<ColorFormat> = ("o_Color", gfx::state::MASK_ALL, gfx::preset::blend::MULTIPLY),
>     |                               ^^^^^^^^^^^ undefined or not in scope
>     |
>     = help: you can import it into scope: `use target::ColorFormat;`.
> 
> error[E0412]: type name `ColorFormat` is undefined or not in scope
>    --> C:\Users\jordan\.cargo\registry\src\github.com-1ecc6299db9ec823\amethyst_renderer-0.4.1\src\pass\deferred.rs:261:31
>     |
> 261 |         out: gfx::BlendTarget<ColorFormat> = ("o_Color", gfx::state::MASK_ALL, gfx::preset::blend::MULTIPLY),
[...]

PR #196, opened 2017-02-27
fix compile error on rust stable

mabruckner:
I have been unable to compile amethyst 0.4.* on rust stable - this fixes the issue, but I have no idea why. Issue #188 starts with the error I've been having, though the thread seems to quickly move to an unrelated bug. A bisect revealed that this problem was introduced with commit 00cd53b4010b5c5635e7eef01999fe9fa2114cd4. I have not tested this on nightly, though I can't imagine that it breaks anything.

### DIFF
--- a/src/renderer/src/pass/deferred.rs
+++ b/src/renderer/src/pass/deferred.rs
@@ -5,7 +5,7 @@ use gfx::traits::FactoryExt;
 
 use ConstantColorTexture;
 use pass;
-use target::{ColorFormat, GeometryBuffer};
+pub use target::{ColorFormat, GeometryBuffer};
 
 gfx_vertex_struct!(Vertex {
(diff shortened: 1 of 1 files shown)


torkleyy:
@mabruckner Which compiler and platform are you using? Is it still an issue with everything updated?

ebkalderon:
@mabruckner Thanks for looking into it! However, I simply cannot reproduce this issue with Rust stable on Linux. I also do not understand why this fix works. Why should publicly re-exporting another module's types have any impact? The `target` module is already re-exported at the crate root, IIRC.

To reiterate @torkleyy's questions, would you please list which compiler version and platform you are using?

mabruckner:
Well, it looks like it was a compiler error. I'm on arch linux, I was on rust 1.14 but updating to 1.15.1 fixed it for me.

PR #198, opened 2017-03-03
Stopwatch fixes

scorbeil:
Added a Duration to the Started variant of the Stopwatch enum to accumulate time from previous start/stop cycles to match behavior with documentation.

Also, adjusted tests to allow for +/- 1% in elapsed time comparison, as they were periodically failing on my machine (Windows 7 64 bit) due to very small differences in expected vs actual elapsed time.

### DIFF
--- a/src/engine/timing.rs
+++ b/src/engine/timing.rs
@@ -8,7 +8,7 @@ pub enum Stopwatch {
     /// Initial state with an elapsed time value of 0 seconds.
     Waiting,
     /// Stopwatch has started counting the elapsed time since this `Instant`.
-    Started(Instant),
+    Started(Duration, Instant),
     /// Stopwatch has been stopped and reports the elapsed time `Duration`.
     Ended(Duration),
 }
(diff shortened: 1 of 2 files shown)


src/engine/timing.rs
```diff
@@ -8,7 +8,7 @@ pub enum Stopwatch {
     /// Initial state with an elapsed time value of 0 seconds.
     Waiting,
     /// Stopwatch has started counting the elapsed time since this `Instant`.
```
torkleyy: The documentation should be edited here

scorbeil:
I'll fix that shortly.  In the meantime, how do you feel about changing Started, rather than adding a Restarted variant?  Do you think a distinction will ever be important? This technically changes the public API, as well, but I assumed that was less important given the unstable state and the fact that it seems the intention is to use the methods rather than destructure the enum value.

src/engine/timing.rs
```diff
@@ -7,8 +7,9 @@ use std::time::{Duration, Instant};
 pub enum Stopwatch {
     /// Initial state with an elapsed time value of 0 seconds.
     Waiting,
-    /// Stopwatch has started counting the elapsed time since this `Instant`.
-    Started(Instant),
+    /// Stopwatch has started counting the elapsed time since this `Instant`
+    /// and accumuluated time from previous start/stop cycles `Duration`.
+    Started(Duration, Instant),
```
torkleyy: I think `Started` is okay.

ebkalderon:
Sweet! I guess the `Stopwatch ` subtly broke when we switched it from a struct to an enum. Thanks for the fix, @scorbeil!

torkleyy:
@ebkalderon Why did you only merge into the release branch? I think develop would be good, too.

ebkalderon:
@torkleyy Sorry, I was busy with school so I didn't have a chance to merge `release-0.4` into `develop`. 😛 I'll do that today.

PR #201, opened 2017-03-06
Use fnv hasher for HashMap and HashSet.

Object905:
With ecs, keys for hashmaps and sets are usually some kind of id, hence is very small. So fnv hash function will give significant performance in this case. [Here is compassion](http://cglab.ca/~abeinges/blah/hash-rs/). Also default hasher resistant to HashDoS attacks, but we don't need it anyway.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -16,3 +16,4 @@
 * Robbie Cooper
 * Scott Corbeil
 * White-Oak
+* Konstantin Zverev
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -8,6 +8,7 @@
 * Emil Gardström
 * Eyal Kalderon
 * Ilya Bogdanov
+* Konstantin Zverev
 * Lucio Franco
 * Lukas Schmierer
 * msiglreith
@@ -16,4 +17,3 @@
 * Robbie Cooper
 * Scott Corbeil
 * White-Oak
-* Konstantin Zverev
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ glutin = "0.7"
 imagefmt = "4.0"
 specs = "0.7"
 wavefront_obj = "5.0"
+fnv = "1.0.5"
 
 [target.'cfg(windows)'.dependencies]
 gfx_device_dx11 = "0.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ amethyst_config = { path = "src/config/", version = "0.2.1" }
 amethyst_renderer = { path = "src/renderer", version = "0.4.1" }
 cgmath = "0.12"
 dds-rs = "0.4"
+fnv = "1.0"
 num_cpus = "1.2"
 genmesh = "0.4"
 gfx = "0.14"
@@ -36,7 +37,6 @@ glutin = "0.7"
 imagefmt = "4.0"
 specs = "0.7"
 wavefront_obj = "5.0"
-fnv = "1.0.5"
 
 [target.'cfg(windows)'.dependencies]
 gfx_device_dx11 = "0.4"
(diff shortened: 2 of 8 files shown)


src/ecs/resources/input.rs
```diff
@@ -1,6 +1,7 @@
 //! World resource that handles all user input.
 
-use std::collections::hash_map::{Entry, HashMap, Keys};
+use std::collections::hash_map::{Entry, Keys};
+use fnv::FnvHashMap;
```
ebkalderon: Please try to keep the `use` statements in alphabetical order.

src/asset_manager/asset_manager.rs
```diff
@@ -6,7 +6,7 @@ use dds::DDS;
 use gfx::texture::{AaMode, Kind};
 use imagefmt::{ColFmt, Image, read_from};
 use std::any::{Any, TypeId};
-use std::collections::HashMap;
+use fnv::FnvHashMap;
```
ebkalderon: Please try to keep the `use` statements in alphabetical order.

Cargo.toml
```diff
@@ -36,6 +36,7 @@ glutin = "0.7"
 imagefmt = "4.0"
 specs = "0.7"
 wavefront_obj = "5.0"
+fnv = "1.0.5"
```
ebkalderon: Please try to keep the dependencies in alphabetical order. Also, I would recommend omitting the `.5` at the end so Cargo can automatically update patch releases.

src/lib.rs
```diff
@@ -77,6 +77,7 @@ extern crate imagefmt;
 extern crate num_cpus;
 extern crate specs;
 extern crate wavefront_obj;
+extern crate fnv;
```
ebkalderon: Please try to keep the `extern crate` statements in alphabetical order.

src/renderer/Cargo.toml
```diff
@@ -25,6 +25,7 @@ gfx_window_glutin = "0.14"
 glutin = "0.7"
 mopa = "0.2"
 specs = "0.7"
+fnv = "1.0.5"
```
ebkalderon: Please try to keep the dependencies in alphabetical order. Also, I would recommend omitting the `.5` at the end so Cargo can automatically update patch releases.

AUTHORS.md
```diff
@@ -16,3 +16,4 @@
 * Robbie Cooper
 * Scott Corbeil
 * White-Oak
+* Konstantin Zverev
```
ebkalderon: Please try to keep the authors list in alphabetical order.

src/renderer/src/lib.rs
```diff
@@ -17,7 +18,7 @@ pub mod target;
 
 use specs::{Component, VecStorage};
 use std::any::TypeId;
-use std::collections::HashMap;
+use fnv::FnvHashMap;
```
ebkalderon: Please try to keep the `use` statements in alphabetical order.

Aceeri:
Are we sure we don't want to keep the hashmap/sets described as the `FnvHashMap/Set` types? It might be a little confusing to modify in the future since you'd normally guess the `HashMap` type is the std one.

vitvakatu:
@Aceeri but you can always take a look at the first lines of file and find the truth.

Aceeri:
Yeah but it feels like an unnecessary abstraction and is harder to understand by reading only those portions of the code. Do we gain anything from abstracting it like this?

vitvakatu:
Well, it makes sense. I agree with you

ebkalderon:
@Aceeri Remember that the `FnvHashMap` and `FnvHashSet` types are simply type aliases to the `std::collections::HashMap` and `HashSet` types. Using `as HashMap` is perfectly fine in this case.

Aceeri:
Ah I see, I was thinking it was another implementation of a hashmap/set.

PR #212, opened 2017-03-13
Profile Application::advance_frame using thread_profiler

### LINKED ISSUE #208: Add profiling mechanism
It would be nice to have a mechanism for measuring performance of various bits of the engine. For example we could measure the average time it takes to run `GfxDevice::render_world`/`State::update`/`Planner::dispatch` methods inside `Application::advance_frame`.

Having this info would allow us to make performance related design decisions based on measurement. Also it would be useful for engine users optimizing their games.

nchashch:
* Update `CONTRIBUTING.md` to include a section about profiling.
* Save profiling info to `thread_profile.json` file living in the same
  directory as `Cargo.toml`.

It should resolve #208.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,9 +18,10 @@ appveyor = { repository = "amethyst/amethyst" }
 travis-ci = { repository = "amethyst/amethyst" }
 
 [features]
-default = ["opengl"]
+default = ["opengl", "profiler"]
 opengl = []
 direct3d = []
+profiler = ["thread_profiler/thread_profiler"]
 
 [dependencies]
 amethyst_config = { path = "src/config/", version = "0.2.2" }
@@ -37,6 +38,7 @@ glutin = "0.7"
 imagefmt = "4.0"
 specs = "0.7"
 wavefront_obj = "5.0"
+thread_profiler = "0.1"
 
 [target.'cfg(windows)'.dependencies]
 gfx_device_dx11 = "0.4"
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -37,6 +38,7 @@ glutin = "0.7"
 imagefmt = "4.0"
 specs = "0.7"
 wavefront_obj = "5.0"
+thread_profiler = "0.1"
```
kvark: should probably be `{ version = "0.1", optional = true }`

src/engine/app.rs
```diff
@@ -1,5 +1,6 @@
 //! The core engine framework.
 
+use thread_profiler::{register_thread_with_profiler, write_profile};
```
kvark: missing feature check

src/lib.rs
```diff
@@ -64,6 +64,8 @@
 #![doc(html_logo_url = "https://tinyurl.com/jtmm43a")]
 
 #[macro_use]
+pub extern crate thread_profiler;
```
kvark: missing feature check

Cargo.toml
```diff
@@ -18,9 +18,10 @@ appveyor = { repository = "amethyst/amethyst" }
 travis-ci = { repository = "amethyst/amethyst" }
 
 [features]
-default = ["opengl"]
+default = ["opengl", "profiler"]
```
kvark: I don't think we want to ship with the profiler. At the same time, enabling it for the debug builds doesn't make much sense either, since we want to profile the release. Thus, it should probably be off by default (but still be checked by CI).

kvark:
Thanks! Last thing - please add testing to travis.

kvark:
Thanks!

PR #218, opened 2017-04-06
Ignore Intellij IDEA project files

torkleyy:
This PR adds "*.iml" and ".idea" to the git ignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,7 @@ thumbs.db
 
 # IDE / Editor files
 *.iml
-/.idea
+.idea
 
 # Thread profiler output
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -11,5 +11,9 @@ thumbs.db
 *.rs.bk
 *.swp
 
+# IDE / Editor files
+*.iml
+/.idea
```
ebkalderon: Is the forward slash here necessary?

torkleyy: Not really, just thought this folder would only appear in the top level directory. But you're right, I'll fix that.

PR #219, opened 2017-04-06
Upgrade specs to 0.8

torkleyy:


src/engine/app.rs
```diff
@@ -230,10 +235,12 @@ impl<T> ApplicationBuilder<T>
     /// Creates a new ApplicationBuilder with the given initial game state and
     /// display configuration.
     pub fn new(initial_state: T, cfg: DisplayConfig) -> ApplicationBuilder<T> {
+        let pool = Arc::new(ThreadPool::new(num_cpus::get()));
```
torkleyy: I'm going to use this pool in the next PR, which will allow a shared thread pool.

kvark:
Good stuff!
Usage of `join` for non-tuples is weird, there could be a better way, but it's not significant.

torkleyy:
@kvark Yeah, was thinking the same.

kvark:
Also, a squash would be appropriate.

torkleyy:
@kvark Isn't there a way to squash when merging?

kvark:
@torkleyy yep, I'll do that.
Typically, you'd have some sort of a homu/bors bot that doesn't do squashing, so I'm used to request that from time to time ;)

PR #221, opened 2017-04-07
Share thread pool using `Engine` struct

torkleyy:
`Engine` contains the pipeline, asset manager, planner and a thread pool. It is passed to all state methods.

### DIFF
--- a/src/engine/app.rs
+++ b/src/engine/app.rs
@@ -19,14 +19,46 @@ use gfx_device;
 use gfx_device::{DisplayConfig, GfxDevice, gfx_types};
 use renderer::{AmbientLight, DirectionalLight, Pipeline, PointLight, target};
 
+/// The engine type, which holds
+/// several useful structs which can be accessed
+/// by engine and game.
+/// It allows for sharing a thread pool,
+/// and it holds the asset manager.
+pub struct Engine {
+    /// The graphics pipeline
+    pub pipe: Pipeline,
+    /// The ecs planner
+    ///
+    /// To get the world, use `world_mut`.
+    pub planner: Planner<()>,
+    /// The asset manager used to submit
+    /// assets to be loaded in parallel
+    pub manager: AssetManager,
+    /// A thread pool which is used
+    /// for asset loading and the ecs.
+    pub pool: Arc<ThreadPool>,
+}
+
+impl Engine {
+    fn new(pipe: Pipeline, planner: Planner<()>,
+           manager: AssetManager, pool: Arc<ThreadPool>) -> Self {
+        Engine {
+            pipe: pipe,
+            planner: planner,
+            manager: manager,
+            pool: pool,
+        }
+    }
+}
+
 /// User-friendly facade for building games. Manages main loop.
 pub struct Application {
+    /// Holds "global" resources per engine.
+    pub engine: Engine,
+
     // Graphics and asset management structs.
     // TODO: Refactor so `pipe` and `gfx_device` are moved into the renderer.
-    assets: AssetManager,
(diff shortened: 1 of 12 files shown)


torkleyy:
Is the Appveyor build disabled?

torkleyy:
Please do not merge yet, I'm unsure which thread pool we should be using. Seems @nchashch needs a scoped threadpool...

src/engine/app.rs
```diff
@@ -19,14 +19,46 @@ use gfx_device;
...
+
 /// User-friendly facade for building games. Manages main loop.
 pub struct Application {
+    /// Holds "global" resources per engine.
+    pub engine: Engine,
+
     // Graphics and asset management structs.
     // TODO: Refactor so `pipe` and `gfx_device` are moved into the renderer.
-    assets: AssetManager,
     gfx_device: GfxDevice,
```
nchashch: Maybe move `GfxDevice` inside `Engine`? Then `Application` would just hold an `Engine` struct and main loop related things (time variables and state machine), so `Application` would just handle the main loop and not care about engine internals (they all would live inside `Engine` struct).

Since `Engine` is exposed to `State` methods I think `GfxDevice` should be private (inside `Engine`), because making `GfxDevice` public doesn't add value and the user would be able to shoot himself in the foot in a couple weird ways:
1. call `GfxDevice::render_world` manually from `State::update`.
2. call `GfxDevice::poll_events` manually from `State::update`.

It might make sense to add an `Engine::advance_frame` method, which would be called from `Application::advance_frame` and which would handle all the `Planner::dispatch`/`GfxDevice::render_world` buisness (to make `Engine` more opaque to `Application`), then if we wanted to change how engine internals work we would have to modify `Engine::advance_frame` method, which is cleanly separated from all the main loop logic.

src/engine/state.rs
```diff
@@ -234,9 +234,9 @@ mod tests {
         let assets = AssetManager::new();
         let pipe = Pipeline::new();
         let pool = Arc::new(ThreadPool::new(2));
-        let planner = Planner::from_pool(World::new(), pool.clone());
+        let planner: Planner<()> = Planner::from_pool(World::new(), pool.clone());
 
-        let mut engine = Engine { manager: assets, pipe: pipe, planner: planner, pool: pool };
+        let mut engine = unsafe { ::std::mem::uninitialized() };
```
vitvakatu: Is it necessary?

torkleyy:
@vitvakatuYes, if you want to keep the test. It's because travis cannot create a window, which is required by `GfxDevice`, which is required by `Engine`.

src/engine/state.rs
torkleyy: @vitvakatu Yes, if you want to keep the test. It's because travis cannot create a window, which is required by `GfxDevice`, which is required by `Engine`.

vitvakatu: Oh, I see

torkleyy:
Reminder: Please don't merge yet, still waiting for https://github.com/slide-rs/specs/pull/125, so I can use `rayon` then.

kvark:
@torkleyy sorry for the delay, it's merged now. I suppose you need a new published version?

torkleyy:
@kvark Thanks for merging. A new version would be nice, although making `ticket` additive should probably be included in `0.9`. If you want anything else merged before the version bump, I could try to implement it.

kvark:
@torkleyy your change doesn't require 0.9, we can roll it out as a patch release.

torkleyy:
Closing because @ebkalderon implemented this in #233.

PR #226, opened 2017-04-12
Add link to FAQ

### LINKED ISSUE #200: Provide a clean and simple math module
I think the user shouldn't explicitly use something like `cgmath` and I also think that we are using `Vector3<f32>` and not `f64`, so I'd like a math module with typedefs for commonly used math structures. If you agree I'd start a PR.

torkleyy:
Fixes #200

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,20 @@ $ cargo doc
 
 The API reference can be found in `target/doc/amethyst/index.html`.
 
+
+## Questions / Help
+
+Please check out the [FAQ][faq] before asking.
+
+If you have an easy question, just ask on [Gitter][gt] and we'll help you and add it to the FAQ.
+
+Other places you may want to check out are [r/rust_gamedev][rg] and [#rust-gamedev IRC][irc].
+
+[faq]: https://github.com/amethyst/amethyst/wiki/Frequently-Asked-Questions
+[gt]: https://gitter.im/amethyst/general
+[rg]: https://www.reddit.com/r/rust_gamedev/
+[irc]: https://botbot.me/mozilla/rust-gamedev/
+
 ## License
 
 Amethyst is free and open source software distributed under the terms of both


vitvakatu:
I'm not sure that we should close #200

README.md
```diff
@@ -108,6 +108,19 @@ $ cargo doc
...
+
+## Questions / Help
+
+Please check out the [FAQ][faq] before asking.
+
+If you have an easy question, just ask on [Gitter][gt] and we'll help you and add it to the FAQ.
+
+For more complex questions, [r/rust_gamedev][rg] might be the right place.
+
+[faq]: https://www.reddit.com/r/rust_gamedev/
```
msiglreith: I assume this should link to our wiki?

README.md
```diff
@@ -108,6 +108,19 @@ $ cargo doc
...
 The API reference can be found in `target/doc/amethyst/index.html`.
 
+
+## Questions / Help
+
+Please check out the [FAQ][faq] before asking.
+
+If you have an easy question, just ask on [Gitter][gt] and we'll help you and add it to the FAQ.
+
+For more complex questions, [r/rust_gamedev][rg] might be the right place.
```
msiglreith: All amethyst related questions should be asked in the gitter in the first place imo.

torkleyy: Thought it would flood gitter if somebody has a bigger project.

README.md
torkleyy: right, fixed that

README.md
torkleyy: @msiglreith What do you think about 

> Another place you may want to check out is r/rust_gamedev

?

msiglreith: Sounds good to me! We could also add the #rust-gamedev irc channel (double check the channel name please).

> Thought it would flood gitter if somebody has a bigger project.

I would prefer to root all amethyst questions to the gitter as it's also feedback about the library and the current design!

PR #227, opened 2017-04-12
Fix InputHandler bug

derekdreery:
When a keypress is inserted into the hashmap, it should be marked as `NotQueried`.


derekdreery:
Blocking #181

Emilgardis:
I'm not sure how this was missed when this was implemented in #114, thank you

PR #228, opened 2017-04-16
Add pre-commit hook to automate local testing for commits

Lionex:
I started using a `pre-commit` hook on my own repositories and it's made my life much easier by automating the process of running tests; because it stashes all of the unstaged changes, it prevents bugs from cropping up because I've forgotten to add some code to a respective commit that my changes depend on.  It also means I don't have to remember every command I need to use for testing, so if I've forgotten to test with a feature flag this script will catch that for me.

I added instructions for setting up the `pre-commit` hook to CONTRIBUTING.md, but I'm not sure I'm totally satisfied with them.

As for the content of the script, I considered adding `cargo fmt` but it would fail because some of the example and source files were longer than 100 lines, and it introduced a whole host of changes to the source code that I wasn't sure how to deal with.

Finally, I'm not exactly an expert with shell scripts, so if there are changes that can make the script better please suggest away!

### DIFF
--- a/.pre-commit.sh
+++ b/.pre-commit.sh
@@ -6,10 +6,14 @@ RED='\033[1;31m'
 GREEN='\033[1;32m'
 YELLOW='\033[1;33m'
 NC='\033[0m'
+BOLD='\033[1m'
 
-# Check to make sure commit isn't emtpy
+stash=0
+# Check to make sure commit isn't emtpy, exit with status 1 if it is
 if git diff-index --quiet HEAD --; then
-    echo "${RED} Empty commit"
+    echo "${RED}You've tried to commit an empty commit${NC}"
+    echo "\tMake sure to add your changes with 'git add'"
+    exit 1
 else
     # Stash all changes in the working directory so we test only commit files
     if git stash save -u -k -q $STASH_NAME; then
@@ -17,6 +21,8 @@ else
         stash=1
     fi
 fi
+
+echo "${GREEN} Testing commit\n\n"
 
 cargo doc --no-deps &&
 cargo build &&
@@ -27,6 +33,12 @@ cargo test --all --features profiler
 
 # Capture exit code from tests
 status=$?
+
+# Inform user of build failure
+if [ "$status" -ne "0" ]
+then
+    echo "${RED}Build failed:${NC} if you still want to commit use ${BOLD}'--no-verify'${NC}"
+fi
 
 # Revert stash if changes were stashed to restor working directory files
 if [ "$stash" -eq 1 ]
--- a/.pre-commit.sh
+++ b/.pre-commit.sh
@@ -12,12 +12,15 @@ if git stash save -u -k -q $STASH_NAME; then
     stash=1
 fi
 
-cargo doc --no-deps
-cargo build
+cargo doc --no-deps &&
+cargo build &&
 # Build and test without profiler
-cargo test --all
+cargo test --all &&
 # Build and test with profiler
 cargo test --all --features profiler
+
+# Capture exit code from tests
+status=$?
 
 if [ "$stash" -eq 1 ]
 then
@@ -27,3 +30,6 @@ then
         echo "\n\n${RED}Unable to revert stash command${NC}"
     fi
 fi
+
+# Keep exit code from tests, so if they fail, prevent commit
+exit $status
--- a/.pre-commit.sh
+++ b/.pre-commit.sh
@@ -7,9 +7,15 @@ GREEN='\033[1;32m'
 YELLOW='\033[1;33m'
 NC='\033[0m'
 
-if git stash save -u -k -q $STASH_NAME; then
-    echo "${YELLOW}Stashed changes as:${NC} ${STASH_NAME}\n\n"
-    stash=1
+# Check to make sure commit isn't emtpy
+if git diff-index --quiet HEAD --; then
+    echo "${RED} Empty commit"
+else
+    # Stash all changes in the working directory so we test only commit files
+    if git stash save -u -k -q $STASH_NAME; then
+        echo "${YELLOW}Stashed changes as:${NC} ${STASH_NAME}\n\n"
+        stash=1
+    fi
 fi
 
 cargo doc --no-deps &&
@@ -22,6 +28,7 @@ cargo test --all --features profiler
 # Capture exit code from tests
 status=$?
 
+# Revert stash if changes were stashed to restor working directory files
 if [ "$stash" -eq 1 ]
 then
     if git stash pop -q; then
--- a/.pre-commit.sh
+++ b/.pre-commit.sh
@@ -2,53 +2,62 @@
 # pre-commit.sh
 
 STASH_NAME="pre-commit-$(date +%s)"
+BRANCH_NAME=$(git branch | grep '*' | sed 's/* //')
 RED='\033[1;31m'
 GREEN='\033[1;32m'
 YELLOW='\033[1;33m'
 NC='\033[0m'
 BOLD='\033[1m'
 
-stash=0
-# Check to make sure commit isn't emtpy, exit with status 1 if it is
-if git diff-index --quiet HEAD --; then
-    echo "${RED}You've tried to commit an empty commit${NC}"
-    echo "\tMake sure to add your changes with 'git add'"
-    exit 1
-else
-    # Stash all changes in the working directory so we test only commit files
-    if git stash save -u -k -q $STASH_NAME; then
-        echo "${YELLOW}Stashed changes as:${NC} ${STASH_NAME}\n\n"
-        stash=1
-    fi
-fi
-
-echo "${GREEN} Testing commit\n\n"
-
-cargo doc --no-deps &&
-cargo build &&
-# Build and test without profiler
-cargo test --all &&
-# Build and test with profiler
-cargo test --all --features profiler
-
-# Capture exit code from tests
-status=$?
-
-# Inform user of build failure
-if [ "$status" -ne "0" ]
+# Check if commit is on a rebase, if not proceed as usual
+if [ $BRANCH_NAME != '(no branch)' ]
 then
-    echo "${RED}Build failed:${NC} if you still want to commit use ${BOLD}'--no-verify'${NC}"
-fi
-
-# Revert stash if changes were stashed to restor working directory files
-if [ "$stash" -eq 1 ]
-then
-    if git stash pop -q; then
-        echo "\n\n${GREEN}Reverted stash command${NC}"
+    stash=0
+    # Check to make sure commit isn't emtpy, exit with status 1 if it is
+    if git diff-index --quiet HEAD --; then
+        echo "${RED}You've tried to commit an empty commit${NC}"
+        echo "\tMake sure to add your changes with 'git add'"
+        exit 1
     else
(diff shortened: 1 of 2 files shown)


torkleyy:
I really like this idea! Already thought myself about this, will review in four hours.

Lionex:
@torkleyy I guess we are on the same wavelength then!

I found a bug with exit codes from the tests failing to propagate, fixed it now; if you need me to squash the commits I can go ahead and do that.

I think there are also lots of other improvements to be made, so I'm open to feedback!

.pre-commit.sh
```diff
@@ -0,0 +1,35 @@
...
+
+STASH_NAME="pre-commit-$(date +%s)"
+RED='\033[1;31m'
+GREEN='\033[1;32m'
+YELLOW='\033[1;33m'
+NC='\033[0m'
+
+if git stash save -u -k -q $STASH_NAME; then
+    echo "${YELLOW}Stashed changes as:${NC} ${STASH_NAME}\n\n"
+    stash=1
```
vitvakatu: I guess you need to set `stash` to 0 in the beginning.

torkleyy: @vitvakatu Why that? Without setting the `stash` flag the changes aren't applied afterwards.

CONTRIBUTING.md
```diff
@@ -127,6 +127,28 @@ done the following things first:
...
 
+To streamline the process, the `.pre-commit.sh` script will run tests on each
+commit in a sanitary way for you; to have it run automatically on each commit
+simply create a symbolic link with
+
+```bash
+$ ln -s -f ../../.pre-commit.sh .git/hooks/pre-commit
+```
+
+This ensures that you can't commit your changes if tests fail.  If you need to
```
torkleyy: Two spaces

CONTRIBUTING.md
```diff
@@ -127,6 +127,28 @@ done the following things first:
...
+make a commit without running tests, then simply pass the `--no-verify` flag.
+We recommend the following aliases:
+
+```bash
```
torkleyy: Wouldn't necessarily recommend these in a `CONTRIBUTING.md`; users not familiar to `git` aliases won't know what they did and how to commit.

.pre-commit.sh
```diff
@@ -0,0 +1,35 @@
...
+fi
+
```
torkleyy: Not sure if we should do / ask for a rebase?

.pre-commit.sh
```diff
@@ -0,0 +1,35 @@
...
+cargo build &&
+# Build and test without profiler
+cargo test --all &&
+# Build and test with profiler
+cargo test --all --features profiler
+
+# Capture exit code from tests
+status=$?
+
+if [ "$stash" -eq 1 ]
```
torkleyy: Good idea to stash them!

.pre-commit.sh
vitvakatu: @torkleyy what if I have `stash` flag in my environment? For example after running `export stash=1`?

torkleyy: See [The Open Group - Environment Variables](http://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap08.html):

> Environment variable names [...] consist solely of uppercase letters

It's not a requirement, but by convention, environment variables are UPPERCASE_ONLY.

CONTRIBUTING.md
Lionex: Should I just mention the `--no-verify` flag as an alternative?

.pre-commit.sh
vitvakatu: I can be stupid enough or not experienced enough. Well, I will approve it, but still, imo it's not problem to init variable before use.

CONTRIBUTING.md
torkleyy: Yes, I think that is really useful.

.pre-commit.sh
Lionex: I tried setting `stash` to 0 at the beginning and it seemed to change the behavior; now, if i test the script on a commit with no staged changes and a clean working directory, the script stashes the changes and then when it tries to un-stash it fails and results in unmerged paths, requiring me to `git reset HEAD -- && git checkout .` to get back to my pre-commit state.

torkleyy: @vitvakatu sorry I missed the "in the beginning" part of your comment! You can of course do that. Still, I never saw that because if you don't have an environment `stash`.

torkleyy: @Lionex that's strange.

Lionex: @torkleyy I did some more testing and it only appears if working directory is clean; I guess it's a rare edge case, should we worry about it?

Lionex: @torkleyy I figured it out, if there are no local changes to stash, `git stash save` exits without error and sets `stash` to 1; later the script attempts to unstash something I saved from another branch causing unmerged paths.  

That doesn't explain why it doesn't happen if my commit isn't empty though.

.pre-commit.sh
Lionex: If people do make the symbolic link and switch to a branch in the project that doesn't include the script their `pre-commit` hook won't run.  I don't know if that prevents committing though and requires the `--no-verify` flag..

.pre-commit.sh
torkleyy: If not setting `stash` to zero solves it, I'd just leave it out.

.pre-commit.sh
torkleyy: @Lionex What does that have to do with rebasing? Right, the symbolic link is problematic in this case, what about `cp`ing it?

.pre-commit.sh
Lionex: Actually it appears in both cases, I had just been testing it differently when I wrote the script the first time.

.pre-commit.sh
Lionex: I guess I misunderstood the context for mentioning rebasing.

On my other repositories I'm using `cp` until the hook shows up in all of my branches.  That _does_ mean that updating it can be a pain though.

.pre-commit.sh
torkleyy: What about using `git diff-index --quiet HEAD --` (not sure if that's correct) to determine if there are are untracked changes?

CONTRIBUTING.md
```diff
@@ -127,6 +127,36 @@ done the following things first:
...
+$ ln -s -f ../../.pre-commit.sh .git/hooks/pre-commit
+```
+
+or copy it to your `.git` folder with
+
+```bash
+$ cp .pre-commit.sh .git/hooks/pre-commit
+```
+
+Note: if you use `cp` you won't get upstream changes, but if you use `ln` and
```
torkleyy: Would adding it to `.gitignore` work?

.pre-commit.sh
```diff
@@ -0,0 +1,42 @@
...
+# Build and test without profiler
+cargo test --all &&
+# Build and test with profiler
+cargo test --all --features profiler
+
+# Capture exit code from tests
+status=$?
+
+# Revert stash if changes were stashed to restor working directory files
+if [ "$stash" -eq 1 ]
```
torkleyy: stash is "" if it's an empty commit

.pre-commit.sh
```diff
@@ -0,0 +1,42 @@
...
+
+STASH_NAME="pre-commit-$(date +%s)"
+RED='\033[1;31m'
+GREEN='\033[1;32m'
+YELLOW='\033[1;33m'
+NC='\033[0m'
+
+# Check to make sure commit isn't emtpy
+if git diff-index --quiet HEAD --; then
+    echo "${RED} Empty commit"
```
torkleyy: probably exit the pre-commit hook?

.pre-commit.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env sh
+# pre-commit.sh
+
+STASH_NAME="pre-commit-$(date +%s)"
+RED='\033[1;31m'
+GREEN='\033[1;32m'
+YELLOW='\033[1;33m'
+NC='\033[0m'
+
```
torkleyy: Would be nice to have some informative message like "Testing commit"

.pre-commit.sh
```diff
@@ -0,0 +1,42 @@
...
+# Revert stash if changes were stashed to restor working directory files
+if [ "$stash" -eq 1 ]
+then
+    if git stash pop -q; then
+        echo "\n\n${GREEN}Reverted stash command${NC}"
+    else
+        echo "\n\n${RED}Unable to revert stash command${NC}"
+    fi
+fi
+
```
torkleyy: Would be nice to tell the user it didn't quite work out:

```bash
if [ "$status" -eq "0" ]
then
    echo "Build failed; if you still want to commit use '--no-verify'"
fi
```

CONTRIBUTING.md
Lionex: I don't think it would have any effect

Lionex:
@torkleyy Thanks for the suggestions, incorporating them now!

.pre-commit.sh
Lionex: They still see all of the test output but I'll add it right now, the suggestion to use `--no-verify` would be especially helpful

CONTRIBUTING.md
torkleyy: Ah you mean upstream changes to the `pre-commit` hook, now it makes sense.

Lionex:
@torkleyy Are contributors on windows not using something like `git bash` in the first place?

torkleyy:
@Lionex if there is a bash shell executed on Windows, I'm perfectly fine with this PR. It should work and if it does not, we can still fix it later when someone notices an issue.

.pre-commit.sh
```diff
@@ -0,0 +1,54 @@
...
+    echo "\tMake sure to add your changes with 'git add'"
+    exit 1
+else
+    # Stash all changes in the working directory so we test only commit files
+    if git stash save -u -k -q $STASH_NAME; then
+        echo "${YELLOW}Stashed changes as:${NC} ${STASH_NAME}\n\n"
+        stash=1
+    fi
+fi
+
```
torkleyy: Oh I forgot, what about a rebase here?

Lionex: Good catch!  
I found a [blog post](http://gmurphey.com/2013/02/02/ignoring-git-hooks-when-rebasing.html) with a solution, will test it when I get home.

Lionex:
@torkleyy I tested on a rebase, the updated script doesn't interfere.

torkleyy:
@Lionex :+1:

I thought a bit more about this and since the advantage of this is that we don't have travis fail anymore, it would probably be nice to check if a rebase is required and if so, prompt the user to do that

Lionex:
@torkleyy What would require a rebase?  I think that may be more appropriate for a pre-push hook.

Lionex:
I'm down to write a `pre-push` hook whenever I have time again.  Finals are coming up and I might be starting a new job, but if we can lay out exactly what situation calls for a `rebase` (perhaps in an issue?) I'll get to it as soon as possible and open a separate PR for it.

torkleyy:
@Lionex thank you for efforts! The pre-push hook is definitely not a high priority. I think we can merge this PR as is.

Issue #191, opened 2017-02-15
Module organization - Seperation of concerns


lschmierer:
I would advocate for organizing the modules of the `amethyst` crate by "features".

So instead of the current structure like,
```
* ecs
  * components
    * rendering
    * transform
    * ...
  * systems
    * rendering
    * transform
    * ...
  * ressources
    * ...
```
I would suggest something like
```
* rendering
  * ecs
    * components
    * systems
    * ressources
* transform
  * ecs
    * components
    * systems
    * ressources
* ...
```
where everything that directly belongs together can be found in one place.

nchashch:
If we adopted this approach it would make sense to drop `ecs` submodule, so it would look like:
```
* rendering
  * components
  * systems
  * resources
* transform
  * components
  * systems
  * resources
* ...
```

lschmierer:
To avoid confusion, the `amethyst_renderer` crate should probably moved out of  the `amethyst` crates src directory.

See #121.

ebkalderon:
Hmm, I am inclined to disagree. I see the following issues with this approach:

1. Regarding components and resources that are shared between systems: the `Transform`, `Rendering`, and `Physics` systems might all be interested in the `Transform` component. Which subfolder should it go under?
2. The `Transform` system is an essential part of the engine and is not a "feature" in the Cargo sense, if that is what you were implying. If I misinterpreted your statement, please ignore this complaint.
3. What about @Aceeri's upcoming `project` module? It sits on top of everything, whether ECS-related or not. Another example is the `engine` module. Is it really a good idea to mix these modules in with all the ECS stuff?
4. It goes against idiomatic Rust style. The norm is to have data structures be placed directly alongside the functions and methods that operate on them. I think that having dedicated modules only for data (components), mixed data and logic (resources), and almost only logic (systems) flies in the face of this idea. Besides, traversing up and down a deeply nested hierarchy which might even mislead people (see issue 1 above) looking for what they want is very annoying.

lschmierer:
> 1. Regarding components and resources that are shared between systems: the Transform, Rendering, and Physics systems might all be interested in the Transform component. Which subfolder should it go under?

Probably under `transform` together with the `Transform` system. `Rendering` shard then depends on `Transform` shard.

> 2. The Transform system is an essential part of the engine and is not a "feature" in the Cargo sense, if that is what you were implying. If I misinterpreted your statement, please ignore this complaint.

[...]

Issue #159, opened 2017-01-26
RFC: Renderer redesign


nchashch:
# Current state of renderer:
At the end of the frame all `Renderable`s and `Light`s are iterated over and packed into a `Scene`, which is then submitted to `amethyst_renderer::Renderer` along with `Pipeline` struct (it contains all the information about passes/targets), which produces an `Encoder` and then `encoder.flush(&mut device)` is called. So the work of producing an `Encoder` isn't parallelized.

# Proposal:

## Compile time backend choice (or @msiglreith's proposed way of having runtime choice without wrappers)
Dealing with rendering this way would be easier if there are no `GfxDeviceInner`, `EncoderInner`, `MainTargetInner`, etc. So code examples in this issue use `gfx_types` module defined in #155.

### _Update_: Problems with trait object `Component`s which make using them unacceptable
- It breaks data locality (not idiomatic ECS design/bad for cache friendliness)
- It introduces additional levels of indirection (a `Box`ed value/virtual methods) which have a performance cost (it is significant because `Renderable`s are iterated over/dereferenced every frame and there can be a large number of `Renderable`s in the `World`)

Define a `Renderable` as something which can fill an `Encoder`, so the most natural thing would be to make `Renderable` a trait implementing a method for filling an `Encoder`.

Maybe it would make sense to store a separate `Pipeline` for each `Renderable` (to make `Renderable`s self contained and allow for different graphic pipelines for different `Renderables`).
```rust
pub trait Renderable {
    fn render_to_encoder(&self,
                         world: &specs::World,
                         main_target: &MainTarget,
                         encoder: &mut gfx::Encoder<gfx_types::Resources, gfx_types::CommandBuffer>);
}
```
Then a component trait can be implemented for a boxed `Renderable` trait object.
```rust
impl Component for Box<Renderable + Send + Sync> {
    type Storage = VecStorage<Box<Renderable + Send + Sync>>;
}
```
Different concrete structs can now implement this trait, for example:
```rust
pub struct StaticModel {
    pub mesh: Mesh,
    pub ka: Texture,
    pub kd: Texture,
    pub pipeline: Pipeline,
}

impl Renderable for StaticModel { ... }
```
And they can be attached to entities like this:
```rust
let static_model = asset_manager.create_static_model("mesh_name", "ka_name", "kd_name");
let static_model = Box::new(static_model);
world.create_now().
         .with::<Box<Renderable + Send + Sync>>(static_model);
```

## Rendering processor
[...]

Issue #231, opened 2017-04-28
Initial Android support


torkleyy:
As mobile/tablet games become pretty popular (40% of the games market in 2016), it's important for Amethyst to support the Android platform.

This issue tries to focus on getting initial support, the overall process could be tracked as a project (probably as cross platform support in general?).

These are the first steps towards Android support:

- [ ] Support context creation on most Android devices (https://github.com/tomaka/glutin/issues/861)
- [ ] Provide touch-friendly input mechanisms
- [ ] Abstract away savestates and other files which will be modified
- [x] Merge the future asset system which provides a default cross-platform storage

derekdreery:
`glutin::WindowEvent` has a `Touch` variant.

torkleyy:
Constant buffers are not supported by OpenGL ES, we need a workaround there. 
The old renderer creates constant buffers [here](https://github.com/gfx-rs/gfx/blob/8f9b09c45/src/render/src/factory.rs#L149), not sure about the new renderer.

Thanks to @ozkriff for the research on this!

ebkalderon:
@torkleyy This definitely does need to be refactored. I'm looking into possible alternatives.

torkleyy:
@ebkalderon Great, thanks!

ebkalderon:
One possible solution to the constant buffers problem is to have the `EffectBuilder` determine whether we are using OpenGL ES, and if so, to silently split the constant buffer's fields into individual uniforms somehow, probably through some `#[cfg]` trickery. This leaves the GLSL out of date, though, and I'm not too keen on performing regex on the GLSL. Thoughts?

kvark:
Why not provide the same data in both constant buffers and uniforms on the Rust side, similarly to how some of the gfx-rs examples do it?

torkleyy:
Sounds difficult but does anybody have an idea how to handle input cross-platform? Considering that Android controls work very different from desktop, I'm not sure how one would go about that.

cc @Xaeroxe

Xaeroxe:
Android controls are so different from desktop controls I'm not sure if it's useful to abstract over them like that.  Some mobile games use emulated buttons and joysticks but in my opinion that's far from an ideal solution.  If at some point we feel it would help we can add support for that kind of emulation, but I would consider abstraction over input in this way to be a low priority.

torkleyy:
I think it would already be helpful to have something like dragging on the touch screen setting some axis value. I mean that's what all the input abstraction is about, right?

[...]

PR #233, opened 2017-05-07
[WIP] Renderer Rewrite

ebkalderon:
### Added
- [ ] Add high-level physically-based material system.
- [x] Add new render pass abstraction.
- [x] Macro-less runtime defined PSO (pipeline state object) wrapper.
  - [x] `Data`, `Init`, and `Meta` structs.
  - [ ] `Effect` abstraction around all three types.
- [ ] Wire up support for arbitrary vertex format types.
- [x] Add `prelude` module to `amethyst` crate for convenience.
- [x] Depend on `rayon::ThreadPool` (similar to #221)

### Changed
- [x] Redesign renderer API to be usable outside of engine crate. (#159, #191)
- [x] Adopt `cgmath` as renderer math library. (#200)
- [x] Reorganize workspace (BREAKING CHANGE). (#191)

### Removed
- [x] Eliminate `gfx_device` module from engine crate. (#191)


derekdreery:
I think I'll need to wait for this to land before fixing the resize problem, since the bits of code I use are disappearing.

kvark:
Glad to see so much progress lately! 👍

amethyst_renderer/Cargo.toml
```diff
@@ -25,19 +25,24 @@ opengl = ["gfx_device_gl", "gfx_window_glutin", "glutin"]
...
-cgmath = { version = "0.14", features = ["eders", "use_simd"] }
+cgmath = { version = "0.14", features = ["eders"] }
+derivative = "1.0"
 fnv = "1.0"
-gfx = { version = "0.15", features = ["cgmath-types", "serialize"], git = "https://github.com/gfx-rs/gfx" }
-gfx_core = { version = "0.7", features = ["cgmath-types", "serialize"], git = "https://github.com/gfx-rs/gfx" }
+#gfx = { version = "0.15", features = ["cgmath-types", "serialize"], git = "https://github.com/gfx-rs/gfx" }
+gfx = { version = "0.15", features = ["cgmath-types", "serialize"], path = "/home/ekalderon/gfx/src/render/" }
+#gfx_core = { version = "0.7", features = ["cgmath-types", "serialize"], git = "https://github.com/gfx-rs/gfx" }
+gfx_core = { version = "0.7", features = ["cgmath-types", "serialize"], path = "/home/ekalderon/gfx/src/core/" }
```
kvark: why do you need to include the core?

amethyst_renderer/Cargo.toml
```diff
@@ -25,19 +25,24 @@ opengl = ["gfx_device_gl", "gfx_window_glutin", "glutin"]
...
 num_cpus = "1.0"
 rayon = "0.7"
 serde = "1.0"
 serde_derive = "1.0"
 winit = "0.6"
 
-gfx_device_gl = { version = "0.14", optional = true, git = "https://github.com/gfx-rs/gfx" }
-gfx_window_glutin = { version = "0.16", optional = true, git = "https://github.com/gfx-rs/gfx" }
+#gfx_device_gl = { version = "0.14", optional = true, git = "https://github.com/gfx-rs/gfx" }
+gfx_device_gl = { version = "0.14", optional = true, path = "/home/ekalderon/gfx/src/backend/gl/" }
```
kvark: perhaps, reconsider including `gfx_device_gl` too. I believe `gfx_window_glutin` now re-exports the essentials

amethyst_renderer/src/pipe/effect/pso.rs
```diff
@@ -102,9 +112,10 @@ impl<'d> PipelineInit for Init<'d> {
     }
 }
 
-#[derive(Clone, Debug, Default, Eq, Hash, PartialEq)]
+#[derive(Clone, Debug, Default, PartialEq)]
 pub struct Data {
     pub const_bufs: Vec<<RawConstantBuffer as DataBind<Resources>>::Data>,
```
kvark: hmm, not sure it has to get all the `Data` from the component types. Couldn't you just specify the types directly, and what would be the downside of this?

amethyst_renderer/Cargo.toml
ebkalderon: @kvark There were a handful of types that I needed but weren't re-exported in the `gfx` crate, so I had to pull in `gfx_core` directly.

amethyst_renderer/Cargo.toml
ebkalderon: @kvark Really? If so, that's terrific! I'll try it out.

amethyst_renderer/src/pipe/effect/pso.rs
ebkalderon: @kvark I only did this to mimic the `gfx_pipeline!` macro included in `gfx`. Personally, I find it clearer that the `Data` struct contains `DataBind::Data` associated types.

amethyst_renderer/src/mtl/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+pub struct Material {
+    /// Diffuse map.
+    pub albedo: Texture,
+    /// Emission map.
+    pub emission: Texture,
+    /// Normal map.
+    pub normal: Texture,
+    /// Metallic map.
+    pub metallic: Texture,
+    /// Reflectance value.
```
nchashch: nit: `Reflectance value.` -> `Reflectance map.`

amethyst_renderer/src/scene.rs
```diff
@@ -0,0 +1,63 @@
...
+pub type LightIter<'l> = Values<'l, String, Light>;
+
+/// Immutable slice iterator of meshes.
+pub type MeshIter<'m> = Values<'m, String, Mesh>;
+
+/// Collection of lights and meshes to render.
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct Scene {
+    camera: Vec<Camera>,
+    lights: HashMap<String, Light>,
```
nchashch: `Scene` is read every frame and there might be a lot of lights, materials, meshes. So maybe use `Vec<Thing>` here? Because `HashMap<String, Thing>` is expensive (AFAIK it is less cache friendly than `Vec`, because elements are scattered around memory - so `lights.values()` and `light.values().next()` aren't guaranteed to come one after another in RAM, hence it is less likely that both `lights.values()` and `light.values().next()` would be in cache when read).

amethyst_renderer/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+    /// Draws a scene with the given pipeline.
+    pub fn draw(&mut self, scene: &Scene, pipe: &Pipeline, _delta: Duration) {
+        use gfx::Device;
+        use rayon::prelude::*;
+
+        {
+            let encoders = self.encoders.as_mut_slice();
+            self.pool.install(|| {
+                pipe.stages()
+                    .par_iter()
```
nchashch: If I understand how rayon works correctly this would slice up `stages` vector into chunks and process these chunks in parallel. But stages are supposed to be processed sequentially, the `Scene` should be chunked up into `num_cpus` chunks and processed in parallel.

Also the number of `Encoder`s in `encoders` `Vec` should be equal to the number of CPUs and `Encoder`s would correspond to "workers" churning through chunks of the `Scene`, so it wouldn't make sense to `.zip` `stages` and `encoders` `Vec`s.

amethyst_renderer/src/mtl/mod.rs
ebkalderon: Whoops, good catch.

amethyst_renderer/src/lib.rs
```diff
@@ -265,7 +265,8 @@ fn init_backend(wb: WindowBuilder, el: &EventsLoop) -> Result<Backend> {
 
     let wb = glutin::WindowBuilder::from_winit_builder(wb)
         .with_gl_profile(GlProfile::Core)
-        .with_gl(GlRequest::Latest);
+        .with_gl(GlRequest::Latest)
```
torkleyy: Could we have a temporary workaround for https://github.com/tomaka/glutin/issues/861 here?

Maybe something like this:

```rust
    .with_pixel_format(24, 8)
    .with_gl(glutin::GlRequest::GlThenGles {
        opengles_version: (2, 0),
        opengl_version: (2, 1),
})
```

amethyst_renderer/src/lib.rs
```diff
@@ -0,0 +1,285 @@
...
+
+/// Creates the OpenGL backend.
+#[cfg(feature = "opengl")]
+fn init_backend(wb: WindowBuilder, el: &EventsLoop) -> Result<Backend> {
+    use glutin::{GlProfile, GlRequest};
+    use gfx_window_glutin as win;
+
+    let wb = glutin::WindowBuilder::from_winit_builder(wb)
+        .with_gl_profile(GlProfile::Core)
+        .with_gl(GlRequest::Latest)
```
torkleyy: Could we have a temporary workaround for https://github.com/tomaka/glutin/issues/861 here?

Maybe something like this:

```rust
    .with_pixel_format(24, 8)
    .with_gl(glutin::GlRequest::GlThenGles {
        opengles_version: (2, 0),
        opengl_version: (2, 1),
})
```

amethyst_renderer/src/pipe/effect/pso.rs
```diff
@@ -0,0 +1,165 @@
...
+type AccessInfo = pso::AccessInfo<Resources>;
+type DepthStencilTarget = target::DepthStencilTarget<DepthFormat>;
+type Manager = handle::Manager<Resources>;
+type RenderTarget = target::RenderTarget<ColorFormat>;
+type RawDataSet = pso::RawDataSet<Resources>;
+type InitResult<'r, M> = Result<M, InitError<&'r str>>;
+
+#[derive(Clone, Debug, Default, Eq, Hash, PartialEq)]
+pub struct Meta {
+    const_bufs: Vec<RawConstantBuffer>,
```
torkleyy: Constant buffers are not supported by OpenGL ES, how should we handle that? Also see https://github.com/amethyst/amethyst/issues/231

amethyst_renderer/src/scene.rs
ebkalderon: This leads back into our discussion on Gitter a while back in regards to using string IDs as opposed to integer IDs. I agree with your suggestion to store the values inside `Vec`s, and I am looking into alternatives.

amethyst_renderer/src/lib.rs
ebkalderon: Hmm, perhaps I misunderstood the usage of `par_iter` in this case. I assumed that the parallel iterator would keep the initial ordering of the `Vec` consistent, meaning I could process each stage with an `Encoder` in parallel and later flush the results sequentially into the `Device`. Is this is not the case?

amethyst_renderer/src/lib.rs
ebkalderon: @torkleyy Will do, thanks for the suggestion.

amethyst_renderer/src/scene.rs
torkleyy: @ebkalderon Maybe you could use a `Vec` and a `FnvHashMap` which maps `String` -> `usize`? Would be more cache efficient and still ergonomic.

ebkalderon: That's a good idea, @torkleyy. At least during drawing time, the data can remain in a contiguous block. I'll try that!

amethyst_renderer/src/lib.rs
torkleyy: It doesn't change the vector, but the elements won't be dispatched in order.

That said, by looking at the methods, it seems to me that before calling `flush` no changes are uploaded. Given that, I'd say the code is correct, because it just sends scene to all the encoders in parallel, whereas their content gets uploaded sequentially. But I never used these encoders in parallel, so I'm not sure about this.

nchashch: I think if we want to use rayon then `.par_iter` should be called on a `Vec` of `Fragment`s. Here is a code example illustrating how this might work:
```rust
// Warning: Pseudo code - probably won't compile.
// encoders: Vec<Encoder>;
assert!(encoders.length() == num_cpus);
let fragments_len = fragments.length();
let repeating_encoders = encoders.iter().cycle().take(fragments_len);
for stage in pipe.stages() {
    if stage.is_enabled() {
        self.pool.install(|| {
            fragments
                .par_iter()
                .zip(repeating_encoders)
                .for_each(|fragment, enc| stage.apply(fragment, enc, scene));
            // scene contains lights and other things that are the same for all fragments
        }
    }
}
```

amethyst_renderer/src/lib.rs
```diff
@@ -0,0 +1,285 @@
...
+mod tex;
+mod types;
+
+/// Generic renderer.
+pub struct Renderer {
+    device: types::Device,
+    encoders: Vec<Encoder>,
+    factory: Factory,
+    main_target: Arc<Target>,
+    pool: rayon::ThreadPool,
```
torkleyy: This should be `Arc<ThreadPool>`.

amethyst_renderer/src/mesh.rs
```diff
@@ -0,0 +1,142 @@
...
+    attrs: Vec<Attribute>,
+    prim: Primitive,
+    slice: Slice,
+    transform: Matrix4<f32>,
+    vbuf: RawBuffer,
+}
+
+impl Mesh {
+    /// Builds a new mesh from the given vertices.
+    pub fn new<'v, V: VertexFormat + 'v>(verts: &'v [V]) -> MeshBuilder {
```
torkleyy: Maybe remove this method? From the name I'd assume it gives me a mesh.

amethyst_renderer/src/mesh.rs
```diff
@@ -0,0 +1,142 @@
...
+    /// This four-by-four matrix applies translation, rotation, and scaling to
+    /// the mesh. It is often referred to in the computer graphics industry as
+    /// the "model matrix".
+    pub fn with_transform<M: Into<Matrix4<f32>>>(mut self, mat: M) -> Self {
+        self.transform = mat.into();
+        self
+    }
+
+    /// Builds and returns the new mesh.
+    #[doc(hidden)]
```
torkleyy: Why do you hide this method?

amethyst_renderer/src/pipe/effect/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+pub struct Effect {
+    pso: PipelineState<Meta>,
+    #[derivative(Hash = "ignore")]
+    pso_data: Data,
+    #[derivative(Hash = "ignore")]
+    samplers: HashMap<String, Sampler>,
+}
+
+impl Effect {
+    pub fn new_simple_prog<S>(vs: S, ps: S) -> EffectBuilder
```
torkleyy: Same as for `Mesh`. Please move these functions to the builder type, I think that's clearer.

amethyst_renderer/src/pipe/target.rs
```diff
@@ -0,0 +1,178 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub struct Target {
+    color_bufs: Vec<ColorBuffer>,
+    depth_buf: Option<DepthBuffer>,
+    size: (u32, u32),
+}
+
+impl Target {
+    /// Creates a new TargetBuilder with the given name.
+    pub fn new<N: Into<String>>(name: N) -> TargetBuilder {
```
torkleyy: Same here.

amethyst_renderer/src/pipe/effect/pso.rs
ebkalderon: I'm looking into possible alternatives. I am currently not in front of my home computer, so I do not have access to my uncommitted work ATM.

amethyst_renderer/src/mesh.rs
ebkalderon: Well, the `Mesh` needs a constructor of some kind due to the complexity involved. Maybe I could rename it to `from_verts`?

amethyst_renderer/src/lib.rs
ebkalderon: Whoops, good catch!

amethyst_renderer/src/mesh.rs
ebkalderon: I hid it because it's meant to be called by the `Renderer` and not directly by the user. Would you prefer I left it visible?

amethyst_renderer/src/mesh.rs
torkleyy: Why not just use the `MeshBuilder::new()`?

amethyst_renderer/src/lib.rs
```diff
@@ -0,0 +1,285 @@
...
+            encoders: encoders,
+            factory: fac,
+            main_target: Arc::new(main),
+            pool: pool,
+            window: win,
+        })
+    }
+
+    /// Builds a new mesh from the given vertices.
+    pub fn create_mesh(&mut self, mb: MeshBuilder) -> Result<Mesh> {
```
torkleyy: Ah oh I see, you're passing it in here. Maybe you could use `pub(crate)` for the builder then?

amethyst_renderer/src/mesh.rs
omni-viral: IMHO `SomeType::build().with(...).with(...).done()` looks clearer than `SomeTypeBuilder::new().with(...).with(...).done()`

torkleyy: @omni-viral While I thin `Mesh::build` is fine, here we have `Mesh::new` returning a `MeshBuilder` which seems a bit confusing to me.

amethyst_renderer/src/lib.rs
omni-viral: You can't zip `ParallelIterator` with `Cycle`

This is what (I think it) should look like

```
for stage in pipe.stages() {
    if stage.is_enabled() {
        let encoders = self.encoders.as_mut_slice();
        let encoders_count = encoders.len();
        self.pool.install(|| {
            let mut enc_par_iter = encoders.into_par_iter();
            let mut mod_par_iter = scene.par_chunks_models(encoders_count);
            assert_eq!(enc_par_iter.len(), mod_par_iter.len());                        
            enc_par_iter.zip(mod_par_iter)
                .for_each(|(enc, models)| for model in models { stage.apply(enc, model, scene) });
        });
        for enc in self.encoders.iter_mut() {
            enc.flush(&mut self.device);
        }
    }
}
```

amethyst_renderer/src/lib.rs
```diff
@@ -0,0 +1,321 @@
...
+        use gfx::Device;
+        use rayon::prelude::*;
+
+        for stage in pipe.stages() {
+            assert_eq!(self.encoders.len(), num_cpus::get());
+            let encoders = self.encoders.as_mut_slice();
+            if stage.is_enabled() {
+                self.pool.install(|| {
+                    scene.par_iter_models()
+                        .zip(encoders)
```
omni-viral: This code draws only `num_cpus::get()` models.
This is what (I think) it should look like.
```rust
for stage in pipe.stages() {
    if stage.is_enabled() {
        let encoders = self.encoders.as_mut_slice();
        let encoders_count = encoders.len();
        self.pool.install(|| {
            let mut enc_par_iter = encoders.into_par_iter();
            let mut mod_par_iter = scene.par_chunks_models(encoders_count);
            assert_eq!(enc_par_iter.len(), mod_par_iter.len());                        
            enc_par_iter.zip(mod_par_iter)
                .for_each(|(enc, models)| for model in models { stage.apply(enc, model, scene) });
        });
        for enc in self.encoders.iter_mut() {
            enc.flush(&mut self.device);
        }
    }
}

// in impl Scene
pub fn par_chunks_models(&self, count: usize) -> rayon::slice::Chunks<Model> {
    use rayon::prelude::*;
    let size = self.models.len();
    self.models.par_chunks(((size - 1) / count) + 1)
}
```

amethyst_renderer/src/pipe/effect/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+    pub fn new_tess_prog<S>(vs: S, hs: S, ds: S, ps: S) -> EffectBuilder
+        where S: Into<&'static [u8]>
+    {
+        EffectBuilder::new_tess_prog(vs, hs, ds, ps)
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct EffectBuilder {
+    init: Init<'static>,
```
omni-viral: `'static` here is not required. `EffectBuilder` can have lifetime parameter.

amethyst_renderer/src/vertex.rs
```diff
@@ -0,0 +1,74 @@
...
+use gfx::pso::buffer::Structure;
+use gfx::traits::Pod;
+
+/// Handle to a vertex attribute.
+pub type Attribute = gfx::pso::buffer::Element<Format>;
+
+/// Trait implemented by all valid vertex formats.
+pub trait VertexFormat: Pod + Structure<Format> + Sized {
+    /// Returns a list of all attributes specified in the vertex.
+    fn attributes() -> Vec<Attribute>;
```
omni-viral: `Attribute`s must be paired with names to make `<RawVertexBuffer as DataLink>::Init`
But user don't know which attribute corresponds to which field without looking into code.
It he looks into code then he can create static array to link RawVertexBuffer to attributes in shader without this method.

amethyst_renderer/src/scene.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// Iterates through all stored models in parallel.
+    pub fn par_iter_models(&self) -> Models {
+        use rayon::prelude::*;
+        self.models.par_iter()
+    }
+}
+
+#[allow(missing_docs)]
+#[derive(Clone, Debug, PartialEq)]
+pub struct Model {
```
omni-viral: How one can add more data into `Model` to access it in custom shader?

amethyst_renderer/src/pipe/pass.rs
```diff
@@ -0,0 +1,93 @@
...
+use error::Result;
+use pipe::{Effect, EffectBuilder, Target, Targets};
+use scene::{Model, Scene};
+use std::fmt::{Debug, Formatter, Result as FmtResult};
+use std::sync::Arc;
+use types::{Encoder, Factory};
+
+pub type FunctionFn = Arc<Fn(&mut Encoder, &Target) + Send + Sync>;
+pub type PostFn = Arc<Fn(&mut Encoder, &Target, &Effect, &Scene) + Send + Sync>;
+
```
omni-viral: Need a pass function that receive model as well

amethyst_renderer/src/pipe/stage.rs
```diff
@@ -0,0 +1,100 @@
...
+use pipe::pass::{Pass, PassBuilder};
+use scene::{Model, Scene};
+use std::sync::Arc;
+use types::{Encoder, Factory};
+
+/// A stage in the rendering pipeline.
+#[derive(Clone, Debug)]
+pub struct Stage {
+    enabled: bool,
+    passes: Vec<Pass>,
```
omni-viral: Shouldn't be passes split into three kinds?
Pre-processing passes
Main-phase passes that handles models
Post-processing passes

While pre-processing passes should be executed first post-processing passes - last
And only main-phase passes should be executed in parallel

encoder flushing should preserve the order between those kinds

amethyst_renderer/src/mesh.rs
```diff
@@ -0,0 +1,141 @@
...
+
+        let verts = self.vertices;
+        let stride = self.stride;
+        let role = Role::Vertex;
+        let bind = Bind::empty();
+
+        let vbuf = fac.create_buffer_immutable_raw(verts, stride, role, bind)?;
+        let slice = Slice {
+            start: 0,
+            end: verts.len() as u32,
```
msiglreith: This bound is incorrect as `verts` is an `u8` slice and doesn't have the same length as the original slice of vertices due to the `cast_slice` on creation.

omni-viral: Do you think it cause your "hedgehog" problem?

torkleyy:
@ebkalderon Do you think we could merge this as soon as we have the functionality of the old renderer covered and @omni-viral's PR merged? After all, merging a PR (especially at this stage of development) does not mean it's full-featured and production ready.

ebkalderon:
@torkleyy Yes, that's exactly my intention!

amethyst_renderer/src/cam.rs
```diff
@@ -0,0 +1,57 @@
...
+
+use cgmath::{Deg, Matrix4, Ortho, PerspectiveFov, Point3, Vector3};
+
+/// The projection mode of a `Camera`.
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub enum Projection {
+    /// An [orthographic projection][op].
+    ///
+    /// [op]: https://en.wikipedia.org/wiki/Orthographic_projection
+    Orthographic(Ortho<f32>),
```
Xaeroxe: Thanks!  This saves me a lot of work!

torkleyy:
@Xaeroxe why not do it after it is merged?

Xaeroxe:
@torkleyy We could do that too.  Order of operations isn't terribly important to me I just want to drop our old asset management system.

kvark:
@ebkalderon I'm happy to see progress here! I'm not able to keep up with all the changes, so please ping me when doubting about particular aspects of gfx usage.

ebkalderon:
@amethyst/engine-devs I was working all day yesterday to rebase my code on the latest `develop`, improve the `Effect` wrapper, and update all engine-wide examples and tests to match. Expect this pull request to be ready in a few more days.

amethyst_renderer/src/pipe/effect/mod.rs
```diff
@@ -70,91 +73,101 @@ impl<'a> ProgramSource<'a> {
...
+    globals: HashMap<String, usize>,
 }
 
 impl Effect {
-    pub fn new_simple_prog<'a, S>(vs: S, ps: S) -> EffectBuilder<'a>
-        where S: Into<&'a [u8]>
-    {
-        EffectBuilder::new_simple_prog(vs, ps)
+    pub fn update_global<N: AsRef<str>, T: ToUniform>(&mut self, name: N, data: T) {
+        if let Some(i) = self.globals.get(name.as_ref()) {
```
omni-viral: I'm not sure that hash-table lookup for each global/constant_buffer for each frame is a good idea.

amethyst_renderer/src/pipe/effect/mod.rs
```diff
@@ -70,91 +73,101 @@ impl<'a> ProgramSource<'a> {
...
     {
-        EffectBuilder::new_tess_prog(vs, hs, ds, ps)
+        if let Some(i) = self.const_bufs.get(name.as_ref()) {
+            let raw = &self.data.const_bufs[*i];
+            enc.update_constant_buffer::<T>(unsafe { mem::transmute(raw) }, &data);
+        }
+    }
+
+    /// FIXME: Add support for arbitrary materials and textures.
+    pub fn draw(&self, model: &Model, enc: &mut Encoder) {
```
omni-viral: There is no need in whole `Model`. `Mesh` would be sufficient.

amethyst_renderer/src/pipe/mod.rs
```diff
@@ -32,63 +33,75 @@ mod effect;
 mod stage;
 mod target;
 
+/// Immutable iterator of pipeline stages.
+#[derive(Debug)]
+pub struct Stages<'s>(Filter<Iter<'s, Stage>, fn(&&Stage) -> bool>);
```
omni-viral: This type is used with only one function `Stage::is_enabled`.
I suggest to encapsulate the function behavior in `Iterator` implementation.

```rust
pub struct Stages<'s>(Iter<'s, Stage>);

impl<'s> Iterator for Stages<'s> {
    type Item = &'s Stage;

    #[inline]    
    fn next(&mut self) -> Option<Self::Item> {
        while let Some(stage) = self.0.next() {
            if stage.is_enabled() {
                return Some(stage);
            }
        }
        None
    }

    #[inline]
    fn size_hint(&self) -> (usize, Option<usize>) {
        let (_, upper) = self.iter.size_hint();
        (0, upper) // can't know a lower bound, due to the predicate
    }
}

File: amethyst_renderer/src/pipe/mod.rs [hunk 27]
```diff
@@ -32,63 +33,75 @@ mod effect;
...
-    pub fn forward<'a>() -> PipelineBuilder<'a> {
+    pub fn forward() -> PipelineBuilder {
         use pass::*;
         PipelineBuilder::new()
             .with_stage(Stage::with_backbuffer()
-                .with_pass(ClearTarget::with_values([1.0; 4], None)))
+                .clear_target([1.0; 4], None))
     }
 
     /// Returns an immutable slice of all stages in the pipeline.
```
omni-viral: All enabled stages

amethyst_renderer/src/pipe/pass.rs
```diff
@@ -3,174 +3,60 @@
...
     }
 }
 
 #[derive(Clone)]
-pub enum PassBuilder<'a> {
-    Basic(BasicFn),
-    Simple(SimpleFn, EffectBuilder<'a>),
-    Model(ModelFn, EffectBuilder<'a>),
-    Light(LightFn, EffectBuilder<'a>),
+pub struct CompiledPass {
```
omni-viral: Deferred shading couldn't be implemented with only model passes.
I'm preparing PR to allow `Pass` to work on different kinds of components. It requires for new render to be integrated with ECS first. ECS types will be referenced only in `RenderSystem` ofc.

amethyst_renderer/src/pipe/pass.rs
```diff
@@ -3,174 +3,60 @@
...
-    pub(crate) fn finish(self, fac: &mut Factory, t: &Targets, out: &Target) -> Result<Pass> {
-        match self {
-            PassBuilder::Basic(f) => Ok(Pass::Basic(BasicPass(f))),
-            PassBuilder::Simple(f, e) => Ok(Pass::Simple(SimplePass(f, e.finish(fac, out)?))),
-            PassBuilder::Model(f, e) => Ok(Pass::Model(ModelPass(f, e.finish(fac, out)?))),
-            PassBuilder::Light(f, e) => Ok(Pass::Light(LightPass(f, e.finish(fac, out)?))),
-        }
+impl CompiledPass {
+    pub fn apply(&self, enc: &mut Encoder, scene: &Scene, model: &Model) {
+        self.inner.apply(enc, &mut self.effect.clone(), scene, model);
```
omni-viral: `Effect` cloning is far from free.
But this part could be optimized later.

amethyst_renderer/src/pipe/stage.rs
```diff
@@ -37,51 +37,78 @@ impl Stage {
...
+    /// Applies all passes within this stage.
+    pub fn apply(&self, encoders: &mut [Encoder], scene: &Scene) {
+        use rayon::prelude::*;
+
+        let mut encs = encoders;
+        self.clear_color.map(|c| self.target.clear_color(&mut encs[0], c));
+        self.clear_depth.map(|d| self.target.clear_depth_stencil(&mut encs[0], d));
+
+        for pass in self.passes.iter() {
+            let mut models = scene.par_chunks_models(4);
```
omni-viral: 4 is a good number. But constant is really bad here.
Use `encoders.len() / self.passes.len()`.
Or on machines with less then 4 cores there will be panic 3 lines below.

amethyst_renderer/src/pipe/stage.rs
```diff
@@ -37,51 +37,78 @@ impl Stage {
...
+            models.zip(enc)
+                .for_each(move |(models, enc)| for model in models {
+                              pass.apply(enc, scene, model);
+                          });
+        }
     }
 
     /// Get count of parallelable passes
     pub fn encoders_required(&self, jobs_count: usize) -> usize {
         self.passes
```
omni-viral: It can be simplified into `self.passes.len() * jobs_count + 1`

amethyst_renderer/src/pipe/stage.rs
```diff
@@ -37,51 +37,78 @@ impl Stage {
...
             .iter()
-            .map(|pass| match *pass {
-                     Pass::Basic(_) => 1,
-                     Pass::Simple(_) => 1,
-                     Pass::Model(_) => jobs_count,
-                     Pass::Light(_) => jobs_count,
-                 })
-            .sum::<usize>()
+            .map(|_| jobs_count)
+            .sum::<usize>() + 1
```
omni-viral: It can be simplified into `self.passes.len() * jobs_count + 1`

amethyst_renderer/src/pipe/stage.rs
```diff
@@ -37,51 +37,78 @@ impl Stage {
...
-        &self.target
+            let enc = {
+                let slice = encs;
+                let (count, left) = slice.split_at_mut(models.len());
+                encs = left;
+                count
+            };
+
+            models.zip(enc)
+                .for_each(move |(models, enc)| for model in models {
```
omni-viral: I actually think that all passes should be applied in parallel with only one join at the end. Instead of joining in each stage. That's why I previously moved this code into `Renderer::draw`.

amethyst_renderer/src/pipe/stage.rs
ebkalderon: Correct, I haven't fixed that yet.

amethyst_renderer/src/pipe/mod.rs
ebkalderon: Good catch, thank you.

amethyst_renderer/src/pipe/effect/mod.rs
ebkalderon: The intention is for the `Model` to be passed in such that it can apply the textures and samplers in the material.

omni-viral: But samplers and textures are applied in `Pass::apply`.
How does `Effect` could know how to apply them?

amethyst_renderer/src/pipe/stage.rs
ebkalderon: I hadn't thought of that. The deeply nested `match` statement was pretty ugly, though. Perhaps we could have `apply` return a parallel iterator to the caller, and have them chained and executed all in parallel.

amethyst_renderer/src/pipe/stage.rs
ebkalderon: Yeah, this isn't ideal. Technically, the number of encoders required is `self.passes.len() * jobs_count` if `self.passes.len() > 0`, and `1` otherwise.

amethyst_renderer/src/pipe/pass.rs
ebkalderon: Definitely noted! I need to find a nice way to avoid this clone and remain thread-safe.

amethyst_renderer/src/pipe/stage.rs
omni-viral: Technically you can reuse one encoder for each pass if you join workers as you do now.
So the number is `self.passes.len() * (max(jobs_count, 1) - 1) + 1`.
But I think it is better to avoid joining at all except one before flushing.

amethyst_renderer/src/pipe/stage.rs
omni-viral: It was ugly but efficient. I was hoping to make it pretty later without losing performance.
Returning parallel iterator is what I was thinking of :+1:
I just haven't figured out exact type :smile:

amethyst_renderer/src/pipe/effect/mod.rs
ebkalderon: @omni-viral Unfortunately, the solution is TBD. Having a single `Material` struct is nice and simple but ultimately is starting to look rather limiting. In an ideal scenario, users should be able to define their own materials with data, and passes compatible with one or more of said materials can draw them.

Maybe we could make `Material` a trait and associate each one with a compatible `Effect` variant? But that might prevent users from defining new material types at runtime, though.

omni-viral: @ebkalderon I don't think that `Effect` should be more then convenient way to bind data for shaders.
Pass defines what shader to use and so it must bind all data required to shader.
Not all passes will work with any material type.

amethyst_renderer/src/pipe/stage.rs
```diff
@@ -69,29 +69,28 @@ impl Stage {
...
-                models.zip(enc).map_with(
-                    pass,
-                    (|pass, (models, enc)| (*pass, models, enc)) as ApplyPassFn<'a>,
-                )
-            })
-            .collect::<Vec<_>>();
+        let mut encoders = encoders.iter_mut();
+        let mut update = Vec::new();
+        for pass in self.passes.iter() {
+            let mut models = scene.par_chunks_models(num_cpus::get());
```
omni-viral: I suggest to use `encoders.len() / passes.len()`.
Note that `Stage::encoders_required` doesn't use `num_cpu::get()` also.

omni-viral:
Looks really good to me 👍 
The only thing required to do before merging is integration with ECS.
I think it could be done by transforming `Scene` into trait with same methods it has now and implement it for type that holds all data in `Storage`s (and implements `SystemData` as well).

Xaeroxe:
@ebkalderon I'm getting close to where I'd like to start working on rendering 2D sprites.  How far off do you think this is from merge?

omni-viral:
@Xaeroxe rendering 2D sprites is what I'm going to do too. I think we should put our heads together on this topic.

torkleyy:
Merged this into https://github.com/amethyst/amethyst/tree/renderer for now, so @omni-viral can work on it.

torkleyy:
@omni-viral did great work, so `renderer` branch is nearly there! I'm closing this now, because these changes will be merged by merging `renderer` into `develop`. @ebkalderon Just trying out this renderer, it's fantastic!

PR #244, opened 2017-06-13
Core asset management

torkleyy:
* Adds a crate "amethyst-assets"
* Examples and `README.md`

This PR contains the core functionality for asset management. This means that e.g. formats or special syntax are to be done.

It also does not contain any tests yet. I'm creating this PR to get some feedback early on.

Replaces #195


CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 
 ## [Unreleased]
```
ebkalderon: I don't think this trailing newline should be here, strictly speaking, according to KeepAChangelog.

torkleyy:
Wow, thanks for the detailed review! I'm actually still resolving somequestions myself. And sorry for the weird API, I did a lot of refactoring so not everything has been optimized for usage yet.

assets/src/loader.rs
```diff
@@ -0,0 +1,173 @@
...
+pub struct Loader {
+    contexts: FnvHashMap<TypeId, Box<Any>>,
+    directory: Arc<Directory>,
+    pool: Arc<ThreadPool>,
+}
+
+impl Loader {
+    /// Creates a new asset loader, initializing the directory store with the
+    /// given path.
+    pub fn new(alloc: &Allocator, directory: PathBuf, pool: Arc<ThreadPool>) -> Self {
```
torkleyy: That's true; I didn't clean up yet after refactoring.

torkleyy:
> I guess amethyst_assets would end up replacing the project module eventually?

Yes, it could be made an asset then, but the `config!` macro would remain.

assets/src/store.rs
```diff
@@ -0,0 +1,95 @@
+use std::error::Error;
+use std::fs::File;
+use std::io::Error as IoError;
+use std::path::PathBuf;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+/// An `Allocator`, holding a counter for producing unique ideas for the stores.
```
torkleyy: :smile: :smile: :smile: Err, this somehow happens to me in the last time - I speak what I'm about to write in my head but write words that sound similar instead.

assets/src/store.rs
```diff
@@ -0,0 +1,95 @@
...
+
+impl Store for Directory {
+    type Error = IoError;
+    type Location = PathBuf;
+
+    fn store_id(&self) -> StoreId {
+        self.id
+    }
+
+    fn load(&self, category: &str, name: &str, ext: &str) -> Result<Vec<u8>, IoError> {
```
torkleyy: No, I planned to only allow `/` as separator. Should probably add that to the docs.

If your mesh is in `assets/meshes/a/really/long/name` then you should probably consider a flatter hierarchy :smile: I think writing something like `buildings/small01` is just fine.

assets/src/store.rs
```diff
@@ -0,0 +1,95 @@
...
+    fn store_id(&self) -> StoreId;
+}
+
+/// A unique store id, used to identify the storage in `AssetSpec`.
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct StoreId(usize);
+
+impl StoreId {
+    /// Returns a copy of the internal id.
+    pub fn id(&self) -> usize {
```
torkleyy: No, `Deref` should only be used for smart pointers.

assets/src/loader.rs
```diff
@@ -0,0 +1,173 @@
...
+              F::Error: Send + 'static,
+              I: AsRef<str> + Send + 'static,
+              S: Store + Send + Sync + 'static,
+              S::Error: Send
+    {
+        let context = self.contexts
+            .get(&TypeId::of::<A>())
+            .expect("Assets need to be registered");
+
+        let context: &Arc<A::Context> = context.downcast_ref().unwrap();
```
torkleyy: If the asset has been registered, there is no way this could fail; that's why I used `unwrap()` in the first place.

assets/src/loader.rs
```diff
@@ -0,0 +1,173 @@
...
+        load_asset_future::<A, F, I, S>(context.clone(), format, id, store, &*self.pool)
+    }
+}
+
+/// Loads an asset with a given context, format, specifier and storage right now.
+pub fn load_asset<A, F, S>(context: &A::Context,
+                           format: &F,
+                           name: &str,
+                           storage: &S)
+                           -> Result<A, LoadError<A::Error, F::Error, S::Error>>
```
torkleyy: Given that typedefs don't support bounds AFAIK, this wouldn't be much shorter: 
`AssetResult<A, A::Error, F::Error, S::Error>`. And it hides the real type, so probably more to lookup?

assets/src/asset.rs
```diff
@@ -0,0 +1,165 @@
...
+    ///
+    /// ## Examples
+    ///
+    /// * `"png"`
+    /// * `"obj"`
+    /// * `"wav"`
+    fn extension() -> &'static str;
+
+    /// Imports the given bytes and produces asset data.
+    fn import(&self, bytes: Vec<u8>) -> Result<Self::Data, Self::Error>;
```
torkleyy: > Also, why are you passing in a Vec<u8> and using it as a slice? Can't you just pass in a &[u8]?

When writing the old PR, I remember passing a slice in but then collecting it back to a `Vec` for some reason, so it was allocated twice. I don't remember what it was exactly.

assets/src/store.rs
ebkalderon: I completely agree with you about flatter hierarchies. But keep in mind that users can sometimes be astonishingly dumb, haha. Be sure to code defensively and consider all possibilities!

torkleyy:
> Speaking of, do you have any plans to include hot reloading of modified assets eventually?

That would be a great feature! But there is so much else to figure out. This is much harder than I thought from a design point of view.

assets/LICENSE-APACHE
```diff
@@ -0,0 +1,201 @@
+                            Apache License
```
torkleyy: I'm not sure; AFAIK licenses should be distributed with the source code. Now, a cargo package contains the source code, but not the top-level license. That's the reason I put it in there, just to not cause any trouble. But I personally don't care whether or not there's a license in these crates.

assets/src/asset.rs
```diff
@@ -0,0 +1,165 @@
...
+    type Error: Error;
+
+    /// Returns the extension (without `.`).
+    ///
+    /// ## Examples
+    ///
+    /// * `"png"`
+    /// * `"obj"`
+    /// * `"wav"`
+    fn extension() -> &'static str;
```
torkleyy: First of all, let me address the easier question:

> Can we and should we support files with no extension?

Well, while the concept of always having extension is somehow only a Windows one, I think we should follow it here. It makes the API simpler and potentially allows an extension -> format mapping.

---

> What happens if there are multiple extensions that refer to the same format? (e.g. mp4/m4a, yml/yaml, and cfg/conf/ini)

It's a simplification, again. I wanted to return an iterator here, but we'd need Generic Associated Types (aka ATCs) then. The other option would be to return a `Vec`, but that's one allocation per asset. So you'd just define one function for loading the asset and call it from both `Format` implementations.

assets/src/loader.rs
```diff
@@ -0,0 +1,173 @@
...
+              A::Error: Send + 'static,
+              F: Format<Data = A::Data> + Send + 'static,
+              F::Error: Send + 'static,
+              I: AsRef<str> + Send + 'static,
+              S: Store + Send + Sync + 'static,
+              S::Error: Send
+    {
+        let context = self.contexts
+            .get(&TypeId::of::<A>())
+            .expect("Assets need to be registered");
```
torkleyy: If the application didn't register the asset, it's clearly a programming error. How would you handle a programming error in code? Furthermore, I have to return a future here, so the options would be returning a result of a future (so two errors essentially) or sending the error to a thread pool :smile:

assets/src/store.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+/// A trait for asset stores, which provides
+/// methods for loading
+pub trait Store {
+    type Error: Error + 'static;
+    type Location;
+
+    /// Loads the bytes given a category, id and extension of the asset.
+    fn load(&self, category: &str, id: &str, ext: &str) -> Result<Vec<u8>, Self::Error>;
```
vitvakatu: In this case, every S (`category`, `id` and `ext`) must be the same type. I'm not sure it's okay

torkleyy: Hold on, there's no need to use a type parameter here, it's never called by the user.

assets/src/store.rs
Aceeri: `Deref` and `DerefMut` from all the discussioms Ive heard about it, are mainly for wrapper types. So that does include smart pointers, but isnt limited to.

assets/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+authors = ["torkleyy <torkleyy@gmail.com>"]
+readme = "README.md"
+description = """
+Asynchronous asset management for games.
+"""
+
+[dependencies]
+fnv = "1"
+futures = "0.1.14"
+parking_lot = "0.4.4"
```
Aceeri: Id personally prefer it if we did since there is no guarantee a dependency couldnt accidentally break something. Moreso it is just an honor code right now.

assets/src/store.rs
torkleyy: Here a reference for that: Rust API guidelines, [C-DEREF](https://github.com/brson/rust-api-guidelines#c-deref).

assets/Cargo.toml
torkleyy: I basically agree with what @Aceeri said: I prefer to just specify the latest patch version, since I don't know whether or not I'm using features of a patch, so it could work for me and on CI, but the dependency version isn't correct.

assets/src/asset.rs
SCareAngel: Can't it be `&'static [&'static str]`?

assets/src/asset.rs
SCareAngel: Maybe you can put several functions here. With `Vec`, slice or even `Read` arguments.

assets/src/loader.rs
```diff
@@ -0,0 +1,180 @@
...
+                              -> Result<A, AssetError<A::Error, F::Error, S::Error>>
+    where A: Asset,
+          F: Format<Data = A::Data>,
+          N: Into<String>,
+          S: Store
+{
+    let name = name.into();
+
+    let store_id = storage.store_id();
+    let spec = AssetSpec::new(name.clone(), F::extension(), store_id);
```
SCareAngel: I'd like to avoid allocations (`String::clone`) in case of getting cached asset

assets/src/store/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+    pub fn next_store_id(&self) -> StoreId {
+        StoreId(self.store_count.fetch_add(1, Ordering::Relaxed))
+    }
+}
+
+/// A trait for asset stores, which provides
+/// methods for loading
+pub trait Store {
+    type Error: Error + 'static;
+    type Location;
```
SCareAngel: I didn't find how it is used. Can you explain? )

assets/src/asset.rs
torkleyy: Why? You would have to implement multiple methods for every format.

assets/src/store/mod.rs
torkleyy: Refactoring... sorry

torkleyy:
Thanks for so much feedback so far! I'll continue development soon; no need to review in that time, I may have already fixed it. I'll tell you when it's ready!

assets/src/asset.rs
SCareAngel: I think it'll give more flexibility.
And you can implement one which is easier to implement. And then implement others through the first one.

torkleyy: If they're just given the bytes, the user can convert to the format they need.

SCareAngel: There are two users of interface. One which implements it and one which call methods. And they can both have data as some arbitrary type `T` that is not a `&[u8]`.

torkleyy: I'm sorry, I don't really understand your point here.

assets/LICENSE-APACHE
ebkalderon: Looking at gfx and Servo as an example, the subcrates do not each contain licenses, presumably because the license files found at the repository root apply to all source code in the tree. But then again, IANAL.

assets/src/asset.rs
```diff
@@ -0,0 +1,164 @@
...
+    pub ext: &'static str,
+    /// The name of this asset.
+    pub name: String,
+    /// The storage id, uniquely identifying the storage it's been loaded from.
+    pub store: StoreId,
+}
+
+impl AssetSpec {
+    /// Creates a new asset specifier from the given parameters.
+    pub fn new(name: String, ext: &'static str, store: StoreId) -> Self {
```
ebkalderon: Would you be able to make this an `Into<String>` as well? I think I missed this on the last review.

assets/src/asset.rs
```diff
@@ -0,0 +1,164 @@
...
+pub struct Cache<T> {
+    map: RwLock<FnvHashMap<AssetSpec, T>>,
+}
+
+impl<T> Cache<T>
+    where T: Clone
+{
+    /// Creates a new `Cache` and initializes it with the default values.
+    pub fn new() -> Self {
+        Default::default()
```
ebkalderon: Shouldn't you `#[derive(Default)]` and then replace this line with `Self::default()`? Or would that complicate things with `T` if `T` doesn't implement `Default`?

assets/src/error.rs
```diff
@@ -0,0 +1,118 @@
...
+            LoadError::StorageError(_) => "Failed to load from storage",
+        }
+    }
+}
+
+/// An error type which cannot be instantiated.
+/// Used as a placeholder for associated error types if
+/// something cannot fail.
+#[derive(Debug)]
+pub enum NoError {}
```
kvark: A phantom type has better semantics than `()`

torkleyy: In addition to what @kvar said, `()` does not implement `Error`.

assets/src/loader.rs
```diff
@@ -0,0 +1,180 @@
...
+    where A: Asset,
+          F: Format<Data = A::Data>,
+          S: Store
+{
+    let bytes = storage
+        .load(A::category(), &spec.name, spec.ext)
+        .map_err(LoadError::StorageError)?;
+    let data = format.import(bytes).map_err(LoadError::FormatError)?;
+    let a = Asset::from_data(data, context)
+        .map_err(LoadError::AssetError)?;
```
torkleyy: That's not possible; I didn't define the error type I'm getting an instance of here, it's an associated type implementing `Error`.

assets/src/asset.rs
torkleyy: It's only used in the code of this crate, so I don't think it's necessary.

assets/src/asset.rs
torkleyy: * `Self::default` and `Default::default` are the same here
* derive doesn't work here because of exactly that reason

Xaeroxe:
So it looks like some of my work might end up having a merge conflict with this, I'm specifically working on key rebinding.  I'd like to make the key rebinding PR sometime between 6/17-6/18.  I'm thinking that the "default" keybinding should probably be an asset, with the active player rebinding file being a config file, similar to our display config files. 

This PR kind of looks like it's in the thick of iteration, so I don't think it'll be merged by the time the key rebinding functionality is ready.  Should I just integrate that default file with the existing asset system?  I could also make a PR to the branch that is hosting this PR to give it the same functionality I'd be giving the old asset management system.

torkleyy:
@Xaeroxe I'm not sure because I don't know how much this PR will change. I don't expect it to change too much though, I guess the main part I'll change is the `Loader`, so I think you should be able to base your key bindings on the `Format` and `Asset` traits.

assets/src/asset.rs
torkleyy: @SCareAngel If we want that, we'd have to assume that there's only one file with an extension contained in the returned array. Is that correct?

assets/LICENSE-APACHE
torkleyy: Let's remove them then.

assets/src/asset.rs
ebkalderon: Cool! Never mind my comment here, then.

assets/src/asset.rs
ebkalderon: Cool, fair enough.

assets/src/asset.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /// Inserts an asset, locking the internal `RwLock` to get write access to the hash map.
+    pub fn insert(&self, spec: AssetSpec, asset: T) {
```
Xaeroxe: If a value exists for the given key in a HashMap the insert operation will return the value it replaced.  It might be helpful to have this function do the same, so that users of this function can identify when they did a redundant insert, and modify their code to not do those redundant inserts.  It's certainly not necessary, it's just helpful for debugging.

assets/src/asset.rs
```diff
@@ -0,0 +1,164 @@
...
+        self.map.write().insert(spec, asset);
+    }
+
+    /// Retrieves an asset, locking the internal `RwLock` to get read access to the hash map.
+    /// In case this asset has been inserted previously, it will be cloned and returned.
+    /// Otherwise, you'll `None`.
```
Xaeroxe: Doc comment is slightly bad english, should be "Otherwise, you'll receive `None`.

assets/src/asset.rs
Xaeroxe: Personally I think we shouldn't support files with no extension.  Not all formats/files have header data and extensions are a nice way to make some sane assumptions about the underlying data.

assets/src/asset.rs
Xaeroxe: Some in-memory asset data structures (I might even argue most) won't need to store the source data contained on disk.  So it'd be nice to pass in a slice here and let the implementation clone the data if it needs to.  Having to clone the vec when one was already allocated is an unfortunate downside to this but I personally think it's the correct approach.

assets/src/error.rs
```diff
@@ -0,0 +1,118 @@
...
+            LoadError::FormatError(ref e) => e,
+            LoadError::StorageError(ref e) => e,
+        };
+
+        Some(cause)
+    }
+
+    fn description(&self) -> &str {
+        match *self {
+            LoadError::AssetError(_) => "Failed to load asset",
```
Xaeroxe: `Failed to do thing` type error messages have been very frustrating for me to debug in the past, due to the lack of additional information.  Perhaps we could also append the `description()` for the inner value?  That would help this a bit.

assets/src/loader.rs
Xaeroxe: I feel like panic is a good thing to do here, mainly because if a problem causes code to be fundamentally broken it's important to have high visibility on it in debugging.

assets/src/error.rs
Aceeri: The error `description` method is extremely unhelpful. Normal use cases would just use a `Debug` implementation because here you *need* a reference to a string, which makes it impossible to create a new string other than one stored inside the structure or is a static string.

Aceeri:
No idea why, but I can't comment on:
> I feel like panic is a good thing to do here, mainly because if a problem causes code to be fundamentally broken it's important to have high visibility on it in debugging.

Anyways, `expect` does panic, it is basically a match of:
```rust
match result {
    Ok(ok) => ok,
    Err(_) => panic!(msg),
}
```

Xaeroxe:
@Aceeri yeah that comment is out of context thanks to GitHub.  I was responding to @ebkalderon.

assets/src/error.rs
Xaeroxe: Yeah you're right.  These thoughts started occurring to me after I made my comment of course :P

assets/src/asset.rs
Xaeroxe: However, since you can't slice over hard drive space that may be nonsense.  Especially since the user of this method clearly doesn't the Vec anymore.

assets/src/asset.rs
SCareAngel: @torkleyy, for arbitrary filename, yes.

assets/src/asset.rs
torkleyy: Well, why would somebody check the value? The `insert` will only be called if there hasn't been a cached version.

assets/src/asset.rs
torkleyy: Oops, I think I the verb here :joy:

assets/src/asset.rs
Xaeroxe: I guess the "somebody" would be us, in the event we ever have to modify how this function is called.

torkleyy: Agreed.

torkleyy:
Blocked on https://github.com/nikomatsakis/rayon/pull/344

assets/src/loader.rs
```diff
@@ -0,0 +1,180 @@
...
+          S: Store + Send + Sync + 'static,
+          S::Error: Send + 'static
+{
+    use futures::future::lazy;
+
+    let name = name.into();
+
+    let closure = move || load_asset::<A, F, _, S>(&*context, &format, name, &*storage);
+
+    AssetFuture(thread_pool.spawn_future_async(lazy(closure)))
```
lschmierer: Mayber return a Stream for hot-reloading support later?

For noe we can simply use
`futures::stream::futures_ordered`
http://alexcrichton.com/futures-rs/futures/stream/fn.futures_ordered.html

torkleyy: Interesting idea..I have to think about it. Just keep in mind that hot-reloading is more a feature for development, so it's not necessary for the final game.

assets/src/loader.rs
```diff
@@ -0,0 +1,338 @@
...
+
+use asset::AssetSpec;
+
+use store::AnyStore;
+use {Allocator, Asset, BoxedErr, Directory, Format, AssetError, LoadError, Store};
+
+/// Represents a future value of an asset. This future may be
+/// added to the ECS world, where the responsible system can poll it and merge
+/// it into the persistent storage once it is `Ready`.
+pub struct AssetFuture<A, E> {
```
torkleyy: So there is something wrong about `AssetFuture`. Calling `Future::wait` on it just blocks forever...

assets/src/loader.rs
torkleyy: I don't see how that is possible.

torkleyy:
Okay, as discussed with @Xaeroxe, this PR should be merged after the core functionality is implemented, which it is. Stuff like hot reloading will be implemented later.

So this PR is ready for review now.

assets/src/error.rs
```diff
@@ -0,0 +1,118 @@
...
+
+/// An error type which cannot be instantiated.
+/// Used as a placeholder for associated error types if
+/// something cannot fail.
+#[derive(Debug)]
+pub enum NoError {}
+
+impl Display for NoError {
+    fn fmt(&self, _: &mut Formatter) -> FmtResult {
+        match *self {}
```
Xaeroxe: This is a bizarre syntactic construct.  I'm kind of surprised it even compiles.  This appears to return `()` in a function which should be returning `FmtResult`.  According to https://github.com/rust-lang/rust/issues/24590 this appears to be undefined behavior.  Is there any way we can improve upon this?  Return some kind of default useless value maybe?  This might be a pain point in the future.

assets/src/error.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+impl Error for NoError {
+    fn description(&self) -> &str {
+        match *self {}
```
Xaeroxe: Same as above.

assets/src/loader.rs
```diff
@@ -0,0 +1,348 @@
...
+            match self.poll() {
+                Ok(Async::Ready(x)) => return Ok(x),
+                Ok(Async::NotReady) => {}
+                Err(x) => return Err(x),
+            }
+        }
+    }
+}
+
+struct AssetFutureInner<A, E> {
```
Xaeroxe: `AssetFutureCell` might be a better name for this struct.

assets/src/error.rs
torkleyy: I could not find the issue saying it's undefined behavior. I'd actually see it's very clearly defined and makes sense - `NoError` can never be instantiated, and this empty match perfectly reflects that.

torkleyy: The undefined behavior is still instantiating `NoError`, not the match.

Xaeroxe: I still think we should change it primarily because this could in future Rust versions lead to [E0308](https://doc.rust-lang.org/error-index.html#E0308).  This statement appears to return `()`, not `FmtResult`. I feel something like `Ok(())` is much less likely to break in the future.

Xaeroxe: Basically I'm pretty sure the only reason this compiles is because the compiler optimizes out this implementation in the final binary.  In my opinion that's a bug in rustc and we shouldn't rely on bugs in our compiler.

Aceeri: Could just put an `unreachable!()` here I believe.

Xaeroxe:
Requesting review from @ebkalderon and @nchashch.  I'd like to prioritize getting this merged as it's blocking rather a lot of work.

assets/src/asset.rs
```diff
@@ -0,0 +1,166 @@
...
+    ///
+    /// The storage may use this information, to e.g. search the identically-named
+    /// subfolder.
+    fn category() -> &'static str;
+
+    /// Provides the conversion from the data format to the actual asset.
+    fn from_data(data: Self::Data, context: &Self::Context) -> Result<Self, Self::Error>;
+
+    /// Notifies about an asset load. This is can be used to cache the asset.
+    /// To return a cached asset, see the `cached` function.
```
omni-viral: where is this `cached` function?

lschmierer:
As the crate name is `amethyst_assets`, the folder should probably called `amethyst_assets`, too.

Like `amethyst_renderer` on @ebkalderon's renderer PR.

lschmierer:
At least some tetsts would be nice.

torkleyy:
@lschmierer I know that, but I thought that at this stage we would like to proceed a bit faster because changes are rather big atm. But I agree - I'll definitely write some before we release a new version!

torkleyy:
> As the crate name is amethyst_assets, the folder should probably called amethyst_assets, too.

But why prefix everything with `amethyst_`? I mean that's already clear from the repository the folder is in, isn't it?

OTOTH, consistency would be nice. So maybe @ebkalderon could rename the folder to just `renderer`?

assets/src/error.rs
brendanzab: The reason it compiles is because `NoError` has no variants, and thus can never be created. This means that it is impossible for `fmt` to ever be called! You do have to figure out a way to return a value from the function though. To do this you use the funny looking `match` on no cases. In the future we will have the never type, `!` (under the `#[feature(never_type)]` flag). This basically gives us:

```rust
enum ! {}
```

I think it will impl a bunch of core traits for cases like this. Using a `NoError` type in the interim seems like a reasonable thing to do though though.

See https://github.com/rust-lang/rfcs/blob/master/text/1216-bang-type.md

assets/src/error.rs
brendanzab: It's not really a phantom type. It's a void or 'bottom' type - ie. an enum with no variants. The neat thing about it is that it is impossible to create at run time, unlike `()` which can be created exactly one time - ie. using `()`. If you used `()` you would have to sprinkle around `unreacable!()`s in a bunch of places, where as with an empty enum the compiler will allow you to return from an impossible-to-call function by doing `match impossible {}`.

assets/src/error.rs
brendanzab: Think about it. Say if I have:

```rust
enum Impossible {}

fn foo(impossible Impossible) -> Vec<String> {
    // ...
}
```

How do I ever call `foo`?

```
let vec = foo(Impossible::/*ugghhhhh*/);
````

Nothing you ever write, will ever satisfy the compiler. Therefore the compiler will allow you to construct a vector from nothing, because it knows it can never be called:

```rust
fn foo(impossible Impossible) -> Vec<String> {
    match impossible {}
}
```

This is a super useful tool when using other folk's API's that take type parameters, or associated types, like the futures crate. It allows you to tell the compiler at the type level that, 'this future will always succeed'. Eg.

```
fn no_unwrapping_wait<F: Future<Error = Impossible>>(f: F) -> F::Item {
    match f.wait() {
        Ok(item) => item,
    }
}
```

Xaeroxe: Yeah that was more or less the conclusion I came to below.  I'm content with this solution.

brendanzab: Yeah, sorry, missed your update! 😅 - was fun thinking about how to explain it though!

Aceeri: Was thinking for some reason that the enum could be instantiated similar to a zero type struct.

assets/src/asset.rs
nchashch: I think we will have to mention this behavior (recognizing only one extension per format) in the documentation, because it can potentially be very frustrating to have your `.m4a` file  fail to load while the only thing you have to do is rename it into a `.mp4` file if you don't know about this quirk of the engine.

lschmierer: What's about @omni-viral's `&'static [&'static str]` approach?

`Store` would then just needs to take `&'static [&'static str]` as well in its functions, `Directory` https://github.com/torkleyy/amethyst/blob/assets/assets/src/store/dir.rs#L52 for example would just look for the first file with one of the given extensions.

assets/examples/cache.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+fn main() {
+    use futures::Future;
+
+    let cfg = Configuration::new().num_threads(8);
+    let pool = Arc::new(ThreadPool::new(cfg).expect("Invalid config"));
+
+    let alloc = Allocator::new();
+    let mut loader = Loader::new(&alloc, "examples/assets", pool);
```
Aceeri: Might be better to do
```rust
format!("{}/examples/assets", env!("CARGO_MANIFEST_DIR"));
```
here so running the example doesn't require building from the crate root.

assets/src/asset.rs
```diff
@@ -0,0 +1,166 @@
...
+    fn clear(_context: &Self::Context) {}
+
+    /// Request for clearing the whole cache.
+    fn clear_all(_context: &Self::Context) {}
+}
+
+/// A specifier for an asset, uniquely identifying it by
+///
+/// * the extension (the format it was provided in)
+/// * it's name
```
Aceeri: "it's" -> "its"

assets/src/asset.rs
```diff
@@ -0,0 +1,166 @@
...
+/// * the storage it was loaded from
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct AssetSpec {
+    /// The extension of this asset
+    pub ext: &'static str,
+    /// The name of this asset.
+    pub name: String,
+    /// The storage id, uniquely identifying the storage it's been loaded from.
```
Aceeri: same as above

assets/src/asset.rs
```diff
@@ -0,0 +1,166 @@
...
+    ///
+    /// ## Examples
+    ///
+    /// * `"png"`
+    /// * `"obj"`
+    /// * `"wav"`
+    fn extension() -> &'static str;
+
+    /// Reads the given bytes and produces asset data.
+    fn parse(&self, bytes: Vec<u8>) -> Result<Self::Data, Self::Error>;
```
Aceeri: I feel like there was a discussion about this earlier, but I can't find it. Would it be better for this to take in a `&'a [u8]`? I'd assume that this wouldn't be modifying the bytes passed in at least.

assets/src/loader.rs
```diff
@@ -0,0 +1,348 @@
...
+    {
+        Loader {
+            contexts: Default::default(),
+            directory: Arc::new(Box::new(Directory::new(alloc, directory))),
+            pool: pool,
+            stores: Default::default(),
+        }
+    }
+
+    pub fn add_store<I, S>(&mut self, name: I, store: S)
```
Aceeri: Should this have docs or do we think it's self-explanatory enough?

assets/src/loader.rs
```diff
@@ -0,0 +1,348 @@
...
+    /// Registers an asset and inserts a context into the map.
+    pub fn register<A>(&mut self, context: A::Context)
+        where A: Asset + 'static,
+              A::Context: 'static
+    {
+        self.contexts
+            .insert(TypeId::of::<A>(), Box::new(Arc::new(context)));
+    }
+
+    /// Like `load_from`, but doesn't ask the cache for a
```
Aceeri: Incomplete documentation.

assets/src/loader.rs
```diff
@@ -0,0 +1,348 @@
...
+        self.load_from::<A, F, _, _>(id, format, "")
+    }
+
+    /// Loads an asset with a given id and format from a custom store.
+    /// The actual work is done on a worker thread, thus this method immediately returns
+    /// a future.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the asset wasn't registered.
```
Aceeri: Is this ideal? Are assets statically registered similar to components/resources in `specs` or can they be unregistered and such?

assets/src/store/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+        Default::default()
+    }
+
+    /// Produces a new store id.
+    pub fn next_store_id(&self) -> StoreId {
+        StoreId(self.store_count.fetch_add(1, Ordering::Relaxed))
+    }
+}
+
+pub trait AnyStore: Send + Sync + 'static {
```
Aceeri: Should probably be docs here as well or a forward to the `Store` docs.

assets/src/loader.rs
```diff
@@ -0,0 +1,348 @@
...
+    let name = name.into();
+
+    let closure = move || load_asset::<A, F, _, S>(&*context, &format, name, &*storage);
+
+    AssetFuture::spawn(thread_pool, closure)
+}
+
+/// Like `load_asset`, but loads the asset on a worker thread and returns
+/// an `AssetFuture` immediately.
+pub fn reload_asset_future<A, F, N, S>
```
Aceeri: Should these `reload` methods specify the difference between what they do and just the regular `load` methods?

assets/src/asset.rs
Xaeroxe: There was a discussion on this but Github marked it outdated due to other changes.

No it would not be better to take a slice, primarily because the bytes would be dropped after this function is called and for some very simple asset types the bytes themselves might be part of the asset, so passing in a Vec here allows us to save an allocation in some instances, and is harmless in others.

assets/src/loader.rs
Xaeroxe: I feel it's pretty self-explanatory.

assets/src/loader.rs
Xaeroxe: Assets can't be unregistered and with this system are intended to be statically registered, so I think this is fine.

assets/src/loader.rs
torkleyy: Oh yeah, probably, but I'd like every item in the crate to have docs. Should enable that lint.

assets/src/loader.rs
torkleyy: Yeah it's *fine*, not ideal, however. The problem is that registering can only work mutably, but loading an asset should be immutable. So they cannot be auto-registered or smth like that.

Aceeri: Is it just an unregistered asset format? Or do you also need to register each asset themselves for it to not panic?

Xaeroxe: You must register asset formats before loading an asset of that format.  You are not required to register the individual assets only the formats which they take.

assets/src/asset.rs
torkleyy: Are you sure? I thought I could use it as a shorthand for "it has"?

assets/src/asset.rs
torkleyy: The best would probably be a stream.

assets/src/loader.rs
torkleyy: You register every type that implements `Asset` (in case you want to load it at some point).

assets/src/loader.rs
torkleyy: Yes.

assets/src/asset.rs
Aceeri: I suppose, it just sounds odd to me.

assets/src/store/mod.rs
torkleyy: Can add a short note, but it's not exposed to the user.

torkleyy:
Okay, so I see there are still things left (like streams for the storages, hot reloading, implementation of asset formats, ...), but I think we can still merge this for now. That way @Xaeroxe can already build audio stuff, and @ebkalderon rendering assets. If there are critical issues to fix before merging, feel free to push to my branch, because I won't have access to my computer for the next few days.

assets/src/asset.rs
ebkalderon: Just checking in. Have we arrived at a reasonable conclusion to this issue yet?

assets/src/asset.rs
ebkalderon: Do you plan to make this and the other similar methods accept a stream instead of a `Vec` before merging?

assets/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "amethyst_assets"
+version = "0.1.0"
+authors = ["torkleyy <torkleyy@gmail.com>"]
+readme = "README.md"
+description = """
+Asynchronous asset management for games.
+"""
+
```
ebkalderon: To keep `amethyst_assets` consistent with other crates, please don't forget to add:

```toml
license = "MIT/Apache-2.0"
keywords = ["game", "asset", "resource", "management", "amethyst"]
categories = ["filesystem"]

documentation = "https://www.amethyst.rs/doc/master/amethyst"
homepage = "https://www.amethyst.rs/"

[badges]
appveyor = { repository = "amethyst/amethyst", branch = "develop" }
```

Cargo.toml
```diff
@@ -75,4 +75,4 @@ name = "assets"
 path = "examples/05_assets/main.rs"
 
 [workspace]
-members = ["src/renderer"]
+members = ["assets", "src/renderer"]
```
ebkalderon: I would prefer to name this folder `amethyst_assets` simply to differentiate subcrates from the `book` and `src` directories. It may look a little verbose, but at least users won't be misled into thinking there are real game assets in that folder, as @lschmierer [pointed out in the Gitter chat](https://gitter.im/amethyst/engine?at=5969031689aea4761d9b76c3) a while ago.

assets/src/asset.rs
alteous: Could returning a MIME type be a reasonable alternative?

ebkalderon: @alteous That could work, and it could potentially be more cross-platform. However, we need to ensure the given MIME types are always valid, and we can't be checking long strings all the time. We could make a big `enum` or `enum`s of all possible valid MIME types, though that seems somewhat hacky.

What does everyone think about this? Do you have an opinion on this, @torkleyy?

assets/src/asset.rs
Xaeroxe: @ebkalderon I think that was going to be a follow-up PR.

ebkalderon:
Thanks to everyone who participated and contributed! This is a pretty big win for us. 🎉 

Special thanks to @torkleyy for his hard work, and to @Xaeroxe for finalizing and merging this PR on such short notice.

Issue #225, opened 2017-04-10
Input manager


torkleyy:
I just reviewed the input manager and I think there are some problems:

- [ ] Don't store a `HashMap` of pressed keys, but of keys mapping to their state

[Link](https://github.com/amethyst/amethyst/blob/develop/src/ecs/resources/input.rs#L32)

I think this will increase performance because we don't add/remove them all the time.

- [ ] Rename `key_down` to `key_pressed`

because `key_down` is widely used for "the key was pressed this frame".

---

It's the same for #181, so I'll fix it after it's merged.

Emilgardis:
I've tried addressing your first point. What you will find is that what needs to happen is that `VirtualKeyCode` gets an `#[repr(u8)]`

https://github.com/tomaka/winit/issues/87
some discussion here: https://github.com/amethyst/amethyst/pull/114

derekdreery:
I can add the `key_down -> key_pressed` change in #181 if you like?

Emilgardis:
It should be added (by rebasing) when that pr is merged

Xaeroxe:
@torkleyy I think we determined in https://github.com/amethyst/amethyst/pull/247 that we weren't going to store the key state in the way described here due to it requiring a larger copy every frame.  I also addressed your second point in that PR.  How do you feel about closing this issue?

PR #247, opened 2017-06-17
Input rebind patch

Xaeroxe:
I'm working on enhancements to the input system and adding support for key rebinding.  I'm opening this pull request now to start gathering feedback.

### DIFF
--- a/src/ecs/resources/input.rs
+++ b/src/ecs/resources/input.rs
@@ -414,7 +414,6 @@ impl InputHandler {
     /// If any action in this list is invalid this will return the id of it in Err.
     pub fn actions_pressed(&self, actions: &[i32]) -> Result<bool, Vec<i32>> {
         let mut all_buttons_pressed = true;
-        //Prefer to skip initializing a new vec if we can.
         let mut bad_values = Vec::new();
         for action in actions {
             if let Some(button) = self.actions.get(action) {
--- a/src/ecs/resources/input.rs
+++ b/src/ecs/resources/input.rs
@@ -286,9 +286,14 @@ impl InputHandler {
         self.current_frame.key_pressed(key) && !self.previous_frame.key_pressed(key)
     }
 
+    /// Checks if a given key was released on this frame.
+    pub fn key_released(&self, key: VirtualKeyCode) -> bool {
+        !self.current_frame.key_pressed(key) && self.previous_frame.key_pressed(key)
+    }
+
     /// Checks if the all the given keys are down and at least one was pressed on this frame.
     pub fn keys_down(&self, keys: &[VirtualKeyCode]) -> bool {
-        keys.iter().any(|key| self.key_down(*key)) && self.keys_pressed(keys)
+        keys.iter().any(|&key| self.key_down(key)) && self.keys_pressed(keys)
     }
 
     /// Returns an iterator for all the pressed down mouse buttons.
@@ -310,6 +315,12 @@ impl InputHandler {
     pub fn mouse_button_down(&self, button: MouseButton) -> bool {
         self.current_frame.mouse_button_pressed(button) &&
         !self.previous_frame.mouse_button_pressed(button)
+    }
+
+    /// Checks if the given mouse button was released this frame.
+    pub fn mouse_button_released(&self, button: MouseButton) -> bool {
+        !self.current_frame.mouse_button_pressed(button) &&
+        self.previous_frame.mouse_button_pressed(button)
     }
 
     /// Checks if the all the given mouse buttons are down and at least one was pressed this frame.
@@ -354,6 +365,14 @@ impl InputHandler {
         match button {
             Button::Key(k) => self.key_down(k),
             Button::Mouse(b) => self.mouse_button_down(b),
+        }
+    }
+
+    /// Checks if the given button was released on this frame.
+    pub fn button_released(&self, button: Button) -> bool {
+        match button {
+            Button::Key(k) => self.key_released(k),
+            Button::Mouse(b) => self.mouse_button_released(b),
         }
     }
 
@@ -416,6 +435,13 @@ impl InputHandler {
     /// Checks if the given action was pressed on this frame.
     pub fn action_down(&self, action: i32) -> Option<bool> {
         self.actions.get(&action).map(|&b| self.button_down(b))
+    }
+
+    /// Checks if the given action was released on this frame.
+    pub fn action_released(&self, action: i32) -> Option<bool> {
+        self.actions
+            .get(&action)
+            .map(|&b| self.button_released(b))
     }
 
     /// Checks if the all the given actions are being pressed and at least one was pressed this frame.
--- a/src/ecs/resources/input.rs
+++ b/src/ecs/resources/input.rs
@@ -2,21 +2,34 @@
 
 use fnv::FnvHashMap as HashMap;
 
-use std::collections::hash_map::{Entry, Keys};
-use std::iter::Iterator;
+use std::iter::{Iterator, Chain, Map};
+use std::slice::Iter;
 
 use engine::{ElementState, WindowEvent, Event, VirtualKeyCode, MouseButton};
 
-/// Indicates whether a given `VirtualKeyCode` has been queried or not.
-#[derive(Eq, PartialEq)]
-enum KeyQueryState {
-    NotQueried,
-    Queried,
+/// A Button is any kind of digital input that the engine supports.
+#[derive(Eq, PartialEq, Debug, Copy, Clone)]
+pub enum Button {
+    Key(VirtualKeyCode),
+    Mouse(MouseButton),
+    //TODO: Add controller buttons here when the engine has support.
+}
+
+impl From<VirtualKeyCode> for Button {
+    fn from(keycode: VirtualKeyCode) -> Self {
+        Button::Key(keycode)
+    }
+}
+
+impl From<MouseButton> for Button {
+    fn from(mouse_button: MouseButton) -> Self {
+        Button::Mouse(mouse_button)
+    }
 }
 
 /// An iterator over the currently pressed down keys.
 pub struct PressedKeys<'a> {
-    iterator: Keys<'a, VirtualKeyCode, KeyQueryState>,
+    iterator: Iter<'a, VirtualKeyCode>,
 }
 
 impl<'a> Iterator for PressedKeys<'a> {
@@ -28,7 +41,7 @@ impl<'a> Iterator for PressedKeys<'a> {
 
 /// An iterator over the currently pressed down mouse buttons.
 pub struct PressedMouseButtons<'a> {
-    iterator: Keys<'a, MouseButton, KeyQueryState>,
+    iterator: Iter<'a, MouseButton>,
 }
 
 impl<'a> Iterator for PressedMouseButtons<'a> {
@@ -36,6 +49,23 @@ impl<'a> Iterator for PressedMouseButtons<'a> {
     fn next(&mut self) -> Option<Self::Item> {
         self.iterator.next()
     }
+}
+
+pub struct PressedButtons<'a> {
+    iterator: Chain<Map<PressedMouseButtons<'a>, fn(&MouseButton) -> Button>,
+                    Map<PressedKeys<'a>, fn(&VirtualKeyCode) -> Button>>,
+}
+
+impl<'a> Iterator for PressedButtons<'a> {
+    type Item = Button;
+    fn next(&mut self) -> Option<Self::Item> {
+        self.iterator.next()
+    }
+}
+
+pub struct Axis {
+    pub pos: Button,
+    pub neg: Button,
 }
 
 /// This struct holds state information about input devices.
@@ -62,45 +92,73 @@ impl<'a> Iterator for PressedMouseButtons<'a> {
 /// ```
 #[derive(Default)]
 pub struct InputHandler {
-    pressed_keys: HashMap<VirtualKeyCode, KeyQueryState>,
-    pressed_mouse_buttons: HashMap<MouseButton, KeyQueryState>,
+    pressed_keys: Vec<VirtualKeyCode>,
+    previous_pressed_keys: Vec<VirtualKeyCode>,
+    pressed_mouse_buttons: Vec<MouseButton>,
+    previous_pressed_mouse_buttons: Vec<MouseButton>,
     // defined to match glium
     mouse_position: Option<(i32, i32)>,
     previous_mouse_position: Option<(i32, i32)>,
+    axes: HashMap<i32, Axis>,
+    actions: HashMap<i32, Button>,
+    text_this_frame: String,
 }
 
 impl InputHandler {
     /// Creates a new input handler.
     pub fn new() -> InputHandler {
         InputHandler {
-            pressed_keys: HashMap::default(),
-            pressed_mouse_buttons: HashMap::default(),
+            pressed_keys: Vec::new(),
+            previous_pressed_keys: Vec::new(),
+            pressed_mouse_buttons: Vec::new(),
+            previous_pressed_mouse_buttons: Vec::new(),
             mouse_position: None,
             previous_mouse_position: None,
+            axes: HashMap::default(),
+            actions: HashMap::default(),
+            text_this_frame: String::new(),
         }
     }
 
     /// Updates the input handler with new engine events.
     pub fn update(&mut self, events: &[WindowEvent]) {
+        // Before processing these events store the input states of the previous frame.
+        self.previous_mouse_position = self.mouse_position;
+        self.previous_pressed_keys.clear();
+        self.previous_pressed_keys
+            .extend_from_slice(&self.pressed_keys);
+        self.previous_pressed_mouse_buttons.clear();
+        self.previous_pressed_mouse_buttons
+            .extend_from_slice(&self.pressed_mouse_buttons);
+        self.text_this_frame.clear();
         for event in events {
             match event.payload {
+                Event::ReceivedCharacter(c) => {
+                    self.text_this_frame.push(c);
+                }
                 Event::KeyboardInput(ElementState::Pressed, _, Some(key_code)) => {
-                    // The check for Entry::Vacant allows more accurate `key_once` calls,
-                    // I.e `key_once(key)` is queried after second `Pressed` event.
-                    if let Entry::Vacant(entry) = self.pressed_keys.entry(key_code) {
-                        entry.insert(KeyQueryState::NotQueried);
+                    if self.pressed_keys.iter().all(|&k| k != key_code) {
+                        self.pressed_keys.push(key_code);
                     }
                 }
                 Event::KeyboardInput(ElementState::Released, _, Some(key_code)) => {
-                    self.pressed_keys.remove(&key_code);
+                    let index = self.pressed_keys.iter().position(|&k| k == key_code);
+                    if let Some(i) = index {
+                        self.pressed_keys.swap_remove(i);
+                    }
                 }
                 Event::MouseInput(ElementState::Pressed, button) => {
-                    if let Entry::Vacant(entry) = self.pressed_mouse_buttons.entry(button) {
-                        entry.insert(KeyQueryState::NotQueried);
+                    if self.pressed_mouse_buttons.iter().all(|&b| b != button) {
+                        self.pressed_mouse_buttons.push(button);
                     }
                 }
                 Event::MouseInput(ElementState::Released, button) => {
-                    self.pressed_mouse_buttons.remove(&button);
+                    let index = self.pressed_mouse_buttons
+                        .iter()
+                        .position(|&b| b == button);
+                    if let Some(i) = index {
+                        self.pressed_mouse_buttons.swap_remove(i);
+                    }
                 }
                 Event::MouseMoved(x, y) => {
                     self.mouse_position = Some((x, y));
@@ -116,86 +174,65 @@ impl InputHandler {
         }
     }
 
+    /// Returns a string representation of all text entered this frame.
+    ///
+    /// Intended for use with text entry fields, insert this string at the cursor position
+    /// every frame.
+    pub fn text_entered(&self) -> &str {
+        self.text_this_frame.as_str()
+    }
+
     /// Returns an iterator for all the pressed down keys.
     pub fn pressed_keys(&self) -> PressedKeys {
-        PressedKeys { iterator: self.pressed_keys.keys() }
+        PressedKeys { iterator: self.pressed_keys.iter() }
     }
 
     /// Checks if the given key is being pressed.
-    pub fn key_down(&self, key: VirtualKeyCode) -> bool {
-        self.pressed_keys.contains_key(&key)
+    pub fn key_pressed(&self, key: VirtualKeyCode) -> bool {
+        self.pressed_keys.iter().any(|&k| k == key)
     }
 
     /// Checks if all the given keys are being pressed.
-    pub fn keys_down(&self, keys: &[VirtualKeyCode]) -> bool {
+    pub fn keys_pressed(&self, keys: &[VirtualKeyCode]) -> bool {
         keys.iter().all(|key| self.key_down(*key))
     }
 
-    /// Checks if the given key is being pressed and held down.
-    ///
-    /// If `key` hasn't been released since the last `key_once()` query, this
-    /// function will return false.
-    pub fn key_once(&mut self, key: VirtualKeyCode) -> bool {
-        if !self.pressed_keys.contains_key(&key) {
-            return false;
-        }
-
-        if let Some(value) = self.pressed_keys.get_mut(&key) {
-            if *value == KeyQueryState::NotQueried {
-                *value = KeyQueryState::Queried;
-                return true;
-            }
-        }
-
-        false
+    /// Checks if the given key was pressed on this frame.
+    pub fn key_down(&self, key: VirtualKeyCode) -> bool {
+        self.key_pressed(key) && self.previous_pressed_keys.iter().all(|&k| k != key)
     }
 
-    /// Checks if the all the given keys are being pressed and held down.
-    ///
-    /// If the `keys` haven't been released since the last `key_once()` query,
-    /// this function will return false.
-    pub fn keys_once(&mut self, keys: &[VirtualKeyCode]) -> bool {
-        keys.iter().any(|key| self.key_once(*key)) && self.keys_down(keys)
+    /// Checks if the all the given keys are down and at least one was pressed on this frame.
+    pub fn keys_down(&self, keys: &[VirtualKeyCode]) -> bool {
+        keys.iter().any(|key| self.key_down(*key)) && self.keys_down(keys)
     }
 
     /// Returns an iterator for all the pressed down mouse buttons.
     pub fn pressed_mouse_buttons(&self) -> PressedMouseButtons {
-        PressedMouseButtons { iterator: self.pressed_mouse_buttons.keys() }
+        PressedMouseButtons { iterator: self.pressed_mouse_buttons.iter() }
     }
 
     /// Checks if the given mouse button is being pressed.
-    pub fn mouse_button_down(&self, button: MouseButton) -> bool {
-        self.pressed_mouse_buttons.contains_key(&button)
+    pub fn mouse_button_pressed(&self, button: MouseButton) -> bool {
+        self.pressed_mouse_buttons.iter().any(|&b| b == button)
     }
 
     /// Checks if all the given mouse buttons are being pressed.
-    pub fn mouse_buttons_down(&self, buttons: &[MouseButton]) -> bool {
+    pub fn mouse_buttons_pressed(&self, buttons: &[MouseButton]) -> bool {
         buttons.iter().all(|btn| self.mouse_button_down(*btn))
     }
 
-    /// Checks if the given mouse button is being pressed and held down.
-    ///
-    /// If `button` hasn't been released since the last `mouse_button_once()` query, this
-    /// function will return false.
-    pub fn mouse_button_once(&mut self, button: MouseButton) -> bool {
-        if !self.pressed_mouse_buttons.contains_key(&button) {
-            return false;
-        }
-        if let Some(value) = self.pressed_mouse_buttons.get_mut(&button) {
-            if *value == KeyQueryState::NotQueried {
-                *value = KeyQueryState::Queried;
-                return true;
-            }
-        }
-        false
+    /// Checks if the given mouse button was pressed this frame.
+    pub fn mouse_button_down(&self, button: MouseButton) -> bool {
+        self.mouse_button_pressed(button) &&
+        self.previous_pressed_mouse_buttons
+            .iter()
+            .all(|&b| b != button)
     }
 
-    /// Checks if the all the given mouse buttons are being pressed and held down.
-    ///
-    /// If the `buttons` haven't been released since the last `mouse_button_once()` query,
-    /// this function will return false.
-    pub fn mouse_buttons_once(&mut self, buttons: &[MouseButton]) -> bool {
-        buttons.iter().any(|btn| self.mouse_button_once(*btn)) && self.mouse_buttons_down(buttons)
+    /// Checks if the all the given mouse buttons are down and at least one was pressed this frame.
+    pub fn mouse_buttons_down(&self, buttons: &[MouseButton]) -> bool {
+        buttons.iter().any(|&btn| self.mouse_button_down(btn)) && self.mouse_buttons_down(buttons)
     }
 
     /// Gets the current mouse position.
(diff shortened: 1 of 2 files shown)


Xaeroxe:
The only file modified in any meaningful way in this PR so far is src/ecs/resource/input.rs everything else is just whitespace changes.

src/ecs/resources/input.rs
```diff
@@ -62,45 +80,63 @@ impl<'a> Iterator for PressedMouseButtons<'a> {
 /// ```
 #[derive(Default)]
 pub struct InputHandler {
-    pressed_keys: HashMap<VirtualKeyCode, KeyQueryState>,
-    pressed_mouse_buttons: HashMap<MouseButton, KeyQueryState>,
+    pressed_keys: Vec<VirtualKeyCode>,
```
torkleyy: It would be great if we could instead just have a boolean `Vec` where the index is the key. Also see [this issue](https://github.com/amethyst/amethyst/issues/225).

src/ecs/resources/input.rs
```diff
@@ -206,16 +213,71 @@ impl InputHandler {
...
+            } else {
+                return Some(-1.0);
+            }
+        } else {
+            return None;
+        }
+    }
+}
+
+fn mouse_button_to_button(mb: &MouseButton) -> Button {
```
torkleyy: Could we have that as an `Into` implementation?

src/ecs/resources/input.rs
```diff
@@ -206,16 +213,71 @@ impl InputHandler {
...
+        }
+    }
+
+    /// Checks if the all the given buttons are being pressed and at least one was pressed this frame.
+    pub fn buttons_once(&self, buttons: &[Button]) -> bool {
+        buttons.iter().any(|&b| self.button_once(b)) && buttons.iter().all(|&b| self.button_down(b))
+    }
+
+    /// Returns the value of an axis by the i32 id, if the id doesn't exist this returns None.
+    pub fn axis_value(&self, id: i32) -> Option<f32> {
```
torkleyy: I see, so this is how you would refer to an axis defined in the config file?

Xaeroxe: Eventually yes, I haven't quite fleshed out the rebinding part yet

src/ecs/resources/input.rs
Xaeroxe: Sure!  I'll make one.

src/ecs/resources/input.rs
Xaeroxe: I had similar thoughts, I initially wanted to store a fixed size array of bools as part of the struct, where each `VirtualKeyCode` (or MouseButton) `as usize` is the index into the array.  This reduces the amount of times we have to follow pointers and do allocations, but determining what that fixed size should be is a bit hard as there isn't an easy way to get a count of all the variants of a `VirtualKeyCode`.  I could just hard code a value but this makes for some brittle code if new codes are ever added.  Additionally every frame the current input state needs to be copied to the previous input state as well, so if the input state structures are larger this gets more expensive.  There are also limitations on how large this Vec<VirtualKeyCode> can get, most of the time it will have < 5 elements, so iterating it to determine key state isn't that expensive, especially when compared to the cost of copying the entire keyboard structure every frame.

src/ecs/resources/input.rs
Xaeroxe: It's been added!

Xaeroxe:
This PR is as finished as it can be until https://github.com/amethyst/amethyst/pull/244 is merged.

Changelog entry:

- All _down functions have been renamed to _pressed
- All _once functions have been renamed to _down, and they all will report the correct values even if they've already been called once on this frame.
- Button type has been added, which is an enum containing the `VirtualKeyCode` and `MouseButton` types
- Functions to compliment Button have been added,
  - pub fn pressed_buttons(&self) -> PressedButtons
  - pub fn button_pressed(&self, button: Button) -> bool
  - pub fn buttons_pressed(&self, buttons: &[Button]) -> bool
  - pub fn button_down(&self, button: Button) -> bool
  - pub fn buttons_down(&self, buttons: &[Button]) -> bool
- Actions have been added, which are mappings between Buttons and i32 values, allowing for rebinding of inputs.
- Similar functions for actions were added, with the exception that there is no pressed_actions.  The function was difficult to write and I couldn't think of a use case for it.
- Axis struct added, which is a mapping of two`Buttons`, one positive and one negative.
- Axes can be mapped to i32s on the InputHandler and then queried for with the `axis_value` function.

src/ecs/resources/input.rs
```diff
@@ -2,21 +2,34 @@
...
-    Queried,
+/// A Button is any kind of digital input that the engine supports.
+#[derive(Eq, PartialEq, Debug, Copy, Clone)]
+pub enum Button {
+    Key(VirtualKeyCode),
+    Mouse(MouseButton),
+    //TODO: Add controller buttons here when the engine has support.
+}
+
+impl Into<Button> for VirtualKeyCode {
```
kvark: I think it's more idiomatic to have `impl From<VirtualKeyCode> for Button`

src/asset_manager/asset_manager.rs
```diff
@@ -410,59 +422,67 @@ impl AssetLoader<Mesh> for ObjSet {
...
-                                            None => [0.0, 0.0, 0.0],
-                                        },
-                                        tex_coord: match ti {
-                                            Some(i) => {
-                                                let tvertex = object.tex_vertices[i];
-                                                [tvertex.u as f32, tvertex.v as f32]
-                                            }
-                                            None => [0.0, 0.0],
-                                        },
+                        geometry
```
kvark: I don't think mixing your input PR with rustfmt pass is a good idea. Better have them separate. FWIW, I like the old formatting better.

src/ecs/resources/input.rs
```diff
@@ -28,7 +41,7 @@ impl<'a> Iterator for PressedKeys<'a> {
 
 /// An iterator over the currently pressed down mouse buttons.
 pub struct PressedMouseButtons<'a> {
-    iterator: Keys<'a, MouseButton, KeyQueryState>,
+    iterator: Iter<'a, MouseButton>,
```
kvark: is there any benefit over doing just `type PressedMouseButtons<'a> = Iter<'a, MouseButton>`?

src/ecs/resources/input.rs
```diff
@@ -62,45 +92,63 @@ impl<'a> Iterator for PressedMouseButtons<'a> {
...
-    pressed_keys: HashMap<VirtualKeyCode, KeyQueryState>,
-    pressed_mouse_buttons: HashMap<MouseButton, KeyQueryState>,
+    pressed_keys: Vec<VirtualKeyCode>,
+    previous_pressed_keys: Vec<VirtualKeyCode>,
+    pressed_mouse_buttons: Vec<MouseButton>,
+    previous_pressed_mouse_buttons: Vec<MouseButton>,
     // defined to match glium
     mouse_position: Option<(i32, i32)>,
     previous_mouse_position: Option<(i32, i32)>,
+    axes: HashMap<i32, Axis>,
```
kvark: I assume these `i32` are specific indices of the elements, in which case it would be a good idea to make them opaque (i.e. `struct AxeId(i32)`) so that the user is not able to change them

src/ecs/resources/input.rs
```diff
@@ -62,45 +92,63 @@ impl<'a> Iterator for PressedMouseButtons<'a> {
...
+            actions: HashMap::default(),
         }
     }
 
     /// Updates the input handler with new engine events.
     pub fn update(&mut self, events: &[WindowEvent]) {
+        // Before processing these events store the input states of the previous frame.
+        self.previous_mouse_position = self.mouse_position;
+        self.previous_pressed_keys = self.pressed_keys.clone();
+        self.previous_pressed_mouse_buttons = self.pressed_mouse_buttons.clone();
```
kvark: nit: should re-use the existing allocation instead of cloning, i.e.
```rust
vector.clear();
vector.extend_from_slice(&other_vector);
```

src/ecs/resources/input.rs
```diff
@@ -62,45 +92,63 @@ impl<'a> Iterator for PressedMouseButtons<'a> {
...
         for event in events {
             match event.payload {
                 Event::KeyboardInput(ElementState::Pressed, _, Some(key_code)) => {
-                    // The check for Entry::Vacant allows more accurate `key_once` calls,
-                    // I.e `key_once(key)` is queried after second `Pressed` event.
-                    if let Entry::Vacant(entry) = self.pressed_keys.entry(key_code) {
-                        entry.insert(KeyQueryState::NotQueried);
+                    if self.pressed_keys.iter().all(|&k| k != key_code) {
```
kvark: I wonder if a `HashSet` would be more convenient here

src/ecs/resources/input.rs
```diff
@@ -206,16 +225,187 @@ impl InputHandler {
...
+            .and_then(|&b| Some(self.button_pressed(b)))
+    }
+
+    /// Checks if all the given actions are pressed.
+    ///
+    /// If any action in this list is invalid this will return the id of it in Err.
+    pub fn actions_pressed(&self, actions: &[i32]) -> Result<bool, Vec<i32>> {
+        let mut all_buttons_pressed = true;
+        //Prefer to skip initializing a new vec if we can.
+        let mut bad_values: Option<Vec<i32>> = None;
```
kvark: `Vec::new()` costs absolutely nothing, no need to wrap it into `Option`

src/ecs/resources/input.rs
```diff
@@ -206,16 +225,187 @@ impl InputHandler {
...
+                if bad_values.is_none() {
+                    bad_values = Some(Vec::new());
+                }
+                if let Some(ref mut values) = bad_values {
+                    values.push(*action);
+                }
+            }
+        }
+        if let Some(values) = bad_values {
+            return Err(values);
```
kvark: no need for `return` statements here

src/ecs/resources/input.rs
```diff
@@ -206,16 +225,187 @@ impl InputHandler {
...
+    }
+
+    /// Checks if the all the given actions are being pressed and at least one was pressed this frame.
+    ///
+    /// If any action in this list is invalid this will return the id of it in Err.
+    pub fn actions_down(&self, actions: &[i32]) -> Result<bool, Vec<i32>> {
+        let mut all_buttons_pressed = true;
+        let mut any_button_pressed_this_frame = false;
+        //Prefer to skip initializing a new vec if we can.
+        let mut bad_values: Option<Vec<i32>> = None;
```
kvark: same here as above

src/ecs/resources/input.rs
Xaeroxe: Ah you're right, thanks.

src/ecs/resources/input.rs
Xaeroxe: This conceals the implementation details a bit better, but other than that no not really.  When I opened the file it was a wrapper struct so I left it that way.  I'm happy to change it either way.

src/ecs/resources/input.rs
Xaeroxe: Good note, I'll do that.

src/ecs/resources/input.rs
Xaeroxe: 0_o TIL a cool new trick, thanks!

src/ecs/resources/input.rs
Xaeroxe: It might be, although I think the performance of Vec might actually be better as iterating over <5 elements in contiguous memory is probably faster than any hashing function

src/ecs/resources/input.rs
Xaeroxe: Cool!  Thanks!

src/ecs/resources/input.rs
Xaeroxe: Yeah I guess it's more of a stylistic choice.  I'll remove them though.

src/asset_manager/asset_manager.rs
Xaeroxe: I asked @ebkalderon about mixing this with rustfmt and he didn't seem to mind, however in this specific instance I just realized it's very likely to have a merge conflict with #244 and I'd rather not have that.  So I'll remove the other files from this PR.

src/ecs/resources/input.rs
Xaeroxe: On second thought it'd probably be better for this to be transparent right now, as the facilities for making input rebinding part of the asset system don't exist yet.  I'll probably see if I can do this again when #244 is merged.

src/ecs/resources/input.rs
kvark: Right, but I don't think it's even worth optimizing if there are just 5 keys. `HashSet` is just more convenient.

kvark:
I don't think `HashSet` is essential here (we can change it later, it's an impl detail anyway), so I'd be ready to merge this. Unless there is something else you want to do in the PR, or other reviewers want to chime in.

Xaeroxe:
As of this moment I consider this merge ready.  If any other reviewers want to chime in I'll be happy to hear what they have to say.

Xaeroxe:
@Aceeri This API doesn't store key press ordering in any way, in order to accomplish that you'd need some kind of state machine, which could be powered by the functions given here but no this won't handle the order in which keys are pressed.

torkleyy:
@Xaeroxe How should we handle text input (e.g. for an in-game chat)?

Xaeroxe:
@torkleyy That's a great question.  This API is awful for that currently, as it has no buffered input, so if a player were to press and release a key in the span of a single frame (as is common in typing) the key wouldn't be registered.  I'd like to change that.  Please don't merge until I have that in.

Xaeroxe:
@torkleyy function `text_entered` has been added, which returns a &str of all characters sent to the window during the current frame, you can insert the output of that function wherever your cursor is in your text entry field.

I consider this ready to merge again.

Xaeroxe:
Also as an aside: This won't handle for backspace and delete.  Full text entry support is more of a UI problem and in my opinion slightly out of scope for this.

Xaeroxe:
And now that you guys got me thinking about sub-frame input speeds it's occurred to me that if a key press and key release occur in the same frame of event handling the fact that the key was pressed will never register, so for keys that were pressed and released in the same frame I'm going to wait until the next frame to process the release event, so that the key entry gets registered for at least one frame of computation.  Seems unlikely but if an amethyst game ever hits the pro gaming scene those pros will thank me.

Xaeroxe:
InputHandler will now register an input for a single frame even if the input was released in the same frame it was pressed.

src/ecs/resources/input.rs
```diff
@@ -62,45 +92,73 @@ impl<'a> Iterator for PressedMouseButtons<'a> {
...
+    text_this_frame: String,
 }
 
 impl InputHandler {
     /// Creates a new input handler.
     pub fn new() -> InputHandler {
         InputHandler {
-            pressed_keys: HashMap::default(),
-            pressed_mouse_buttons: HashMap::default(),
+            pressed_keys: Vec::new(),
```
msiglreith: Abstracting `pressed_*` and `previous_*` into a separate struct could reduce code duplication.

src/ecs/resources/input.rs
```diff
@@ -116,86 +174,65 @@ impl InputHandler {
...
     /// Checks if the given key is being pressed.
-    pub fn key_down(&self, key: VirtualKeyCode) -> bool {
-        self.pressed_keys.contains_key(&key)
+    pub fn key_pressed(&self, key: VirtualKeyCode) -> bool {
+        self.pressed_keys.iter().any(|&k| k == key)
     }
 
     /// Checks if all the given keys are being pressed.
-    pub fn keys_down(&self, keys: &[VirtualKeyCode]) -> bool {
+    pub fn keys_pressed(&self, keys: &[VirtualKeyCode]) -> bool {
```
msiglreith: `keys_pressed` and `key_pressed` seem to have a different behavior as the first one uses `key_down` internally.

src/ecs/resources/input.rs
```diff
@@ -116,86 +174,65 @@ impl InputHandler {
...
 
-    /// Checks if the all the given keys are being pressed and held down.
-    ///
-    /// If the `keys` haven't been released since the last `key_once()` query,
-    /// this function will return false.
-    pub fn keys_once(&mut self, keys: &[VirtualKeyCode]) -> bool {
-        keys.iter().any(|key| self.key_once(*key)) && self.keys_down(keys)
+    /// Checks if the all the given keys are down and at least one was pressed on this frame.
+    pub fn keys_down(&self, keys: &[VirtualKeyCode]) -> bool {
+        keys.iter().any(|key| self.key_down(*key)) && self.keys_down(keys)
```
msiglreith: This looks like an infinite recursion.

src/ecs/resources/input.rs
```diff
@@ -116,86 +174,65 @@ impl InputHandler {
...
     /// Checks if the given mouse button is being pressed.
-    pub fn mouse_button_down(&self, button: MouseButton) -> bool {
-        self.pressed_mouse_buttons.contains_key(&button)
+    pub fn mouse_button_pressed(&self, button: MouseButton) -> bool {
+        self.pressed_mouse_buttons.iter().any(|&b| b == button)
     }
 
     /// Checks if all the given mouse buttons are being pressed.
-    pub fn mouse_buttons_down(&self, buttons: &[MouseButton]) -> bool {
+    pub fn mouse_buttons_pressed(&self, buttons: &[MouseButton]) -> bool {
```
msiglreith: (Same as above)

src/ecs/resources/input.rs
```diff
@@ -116,86 +174,65 @@ impl InputHandler {
...
 
-    /// Checks if the all the given mouse buttons are being pressed and held down.
-    ///
-    /// If the `buttons` haven't been released since the last `mouse_button_once()` query,
-    /// this function will return false.
-    pub fn mouse_buttons_once(&mut self, buttons: &[MouseButton]) -> bool {
-        buttons.iter().any(|btn| self.mouse_button_once(*btn)) && self.mouse_buttons_down(buttons)
+    /// Checks if the all the given mouse buttons are down and at least one was pressed this frame.
+    pub fn mouse_buttons_down(&self, buttons: &[MouseButton]) -> bool {
+        buttons.iter().any(|&btn| self.mouse_button_down(btn)) && self.mouse_buttons_down(buttons)
```
msiglreith: (Same as above, recursion)

src/ecs/resources/input.rs
```diff
@@ -206,16 +243,174 @@ impl InputHandler {
...
+            Button::Mouse(b) => self.mouse_button_down(b),
+        }
     }
+
+    /// Checks if all the given buttons are pressed.
+    pub fn buttons_pressed(&self, buttons: &[Button]) -> bool {
+        buttons.iter().all(|b| self.button_down(*b))
+    }
+
+    /// Checks if the given mouse button was pressed on this frame.
```
msiglreith: Documentation seems misleading here as the function also includes keys

src/ecs/resources/input.rs
```diff
@@ -206,16 +243,174 @@ impl InputHandler {
...
+
+    /// Checks if the all the given buttons are being pressed and at least one was pressed this frame.
+    pub fn buttons_down(&self, buttons: &[Button]) -> bool {
+        buttons.iter().any(|&b| self.button_down(b)) &&
+        buttons.iter().all(|&b| self.button_pressed(b))
+    }
+
+    /// Returns the value of an axis by the i32 id, if the id doesn't exist this returns None.
+    pub fn axis_value(&self, id: i32) -> Option<f32> {
+        if let Some(a) = self.axes.get(&id) {
```
msiglreith: Could be written better via `get(&id).map(|a| .. )` imo

src/ecs/resources/input.rs
```diff
@@ -206,16 +243,174 @@ impl InputHandler {
...
+        for action in actions {
+            if let Some(button) = self.actions.get(action) {
+                if all_buttons_pressed && !self.button_pressed(*button) {
+                    all_buttons_pressed = false;
+                }
+            } else {
+                bad_values.push(*action);
+            }
+        }
+        if bad_values.len() > 0 {
```
msiglreith: nit: `!bad_values.is_empty()`

src/ecs/resources/input.rs
```diff
@@ -206,16 +243,174 @@ impl InputHandler {
...
+                    any_button_pressed_this_frame = true;
+                }
+                if all_buttons_pressed && !self.button_pressed(*button) {
+                    all_buttons_pressed = false;
+                }
+            } else {
+                bad_values.push(*action);
+            }
+        }
+        if bad_values.len() > 0 {
```
msiglreith: (Same as above)

src/ecs/resources/input.rs
```diff
@@ -206,16 +243,174 @@ impl InputHandler {
...
+        buttons.iter().any(|&b| self.button_down(b)) &&
+        buttons.iter().all(|&b| self.button_pressed(b))
+    }
+
+    /// Returns the value of an axis by the i32 id, if the id doesn't exist this returns None.
+    pub fn axis_value(&self, id: i32) -> Option<f32> {
+        if let Some(a) = self.axes.get(&id) {
+            let pos = self.button_down(a.pos);
+            let neg = self.button_down(a.neg);
+            if pos == neg {
```
msiglreith: `Some( if pos == neg { .. } else if ... )` might be nicer here

src/ecs/resources/input.rs
```diff
@@ -206,16 +243,174 @@ impl InputHandler {
...
+        } else {
+            Ok(all_buttons_pressed)
+        }
+    }
+
+    /// Checks if the given action was pressed on this frame.
+    pub fn action_down(&self, action: i32) -> Option<bool> {
+        self.actions
+            .get(&action)
+            .and_then(|&b| Some(self.button_down(b)))
```
msiglreith: `map(|&b| self.button_down(b)))` should do the same I think

src/ecs/resources/input.rs
```diff
@@ -206,16 +243,174 @@ impl InputHandler {
...
+        } else {
+            return None;
+        }
+    }
+
+    // Pressed actions has been omitted as the function is a little difficult to write and I
+    // can't think of a use case for it.
+
+    /// Checks if the given button is currently pressed.
+    pub fn action_pressed(&self, action: i32) -> Option<bool> {
```
msiglreith: yay, actions!

Xaeroxe:
Oh man @msiglreith found lots of problematic typos.  Thanks a ton @msiglreith I'll get to work on your comments.

src/ecs/resources/input.rs
Xaeroxe:
```Rust
Some(if pos == neg {
                     0.0
                 } else if pos {
                1.0
            } else {
                -1.0
            })
```
This is the rustfmt version of that.  In my opinion it really hurts readability.

src/ecs/resources/input.rs
Xaeroxe: Woops, copy paste error, thanks.

Xaeroxe:
Code has been rearranged to have specific InputFrameData structures, this is nice because it more clearly separates where the functionality lives and makes my code less error prone.

Xaeroxe:
@msiglreith since all your comments are outdated I'm giving you a ❤️ here, thanks for catching all the typos I made while tired.

Xaeroxe:
Just added _released functions to detect when an input is released.

Xaeroxe:
I skipped having functions like `keys_released` because the behavior of that would be pretty wonky and hard to predict.  It'd only be reasonable to check if any of the given keys were released this frame, which is not how it's opposite `keys_pressed` behaves.  You can make `keys_pressed` work because it relies on action rather than inaction, but `keys_released` would rely on inaction making it a bit confusing.

src/ecs/resources/input.rs
```diff
@@ -38,6 +51,112 @@ impl<'a> Iterator for PressedMouseButtons<'a> {
...
+impl FrameInputData {
+    pub fn new() -> FrameInputData {
+        FrameInputData {
+            pressed_keys: Vec::new(),
+            pressed_mouse_buttons: Vec::new(),
+            mouse_position: None,
+        }
+    }
+
+    fn advance_frame(&mut self, previous_frame: &mut FrameInputData) {
```
ebkalderon: Missing a doc comment here.

src/ecs/resources/input.rs
```diff
@@ -62,160 +181,335 @@ impl<'a> Iterator for PressedMouseButtons<'a> {
...
+                } else if pos {
+                    1.0
+                } else {
+                    -1.0
+                }
+            })
+    }
+
+    // Pressed actions has been omitted as the function is a little difficult to write and I
+    // can't think of a use case for it.
```
ebkalderon: Given that `action_pressed()` is present below, this comment seems somewhat misleading.

src/ecs/resources/input.rs
```diff
@@ -38,6 +51,112 @@ impl<'a> Iterator for PressedMouseButtons<'a> {
...
+    fn mouse_position(&self) -> Option<(i32, i32)> {
+        self.mouse_position
+    }
+
+    /// Returns an iterator over the buttons that are currently pressed
+    fn pressed_buttons(&self) -> PressedButtons {
+        let mouse_button_convert = mouse_button_to_button as fn(&MouseButton) -> Button;
+        let key_convert = key_to_button as fn(&VirtualKeyCode) -> Button;
+        let mouse_buttons = self.pressed_mouse_buttons().map(mouse_button_convert);
+        let keys = self.pressed_keys().map(key_convert);
```
ebkalderon: Perhaps you could use closures here instead of free-standing functions?

src/ecs/resources/input.rs
```diff
@@ -38,6 +51,112 @@ impl<'a> Iterator for PressedMouseButtons<'a> {
...
+        self.iterator.next()
+    }
+}
+
+pub struct Axis {
+    pub pos: Button,
+    pub neg: Button,
+}
+
+#[derive(Default)]
```
ebkalderon: Missing doc comment here.

src/ecs/resources/input.rs
```diff
@@ -38,6 +51,112 @@ impl<'a> Iterator for PressedMouseButtons<'a> {
...
+}
+
+impl<'a> Iterator for PressedButtons<'a> {
+    type Item = Button;
+    fn next(&mut self) -> Option<Self::Item> {
+        self.iterator.next()
+    }
+}
+
+pub struct Axis {
```
ebkalderon: Missing doc comment here.

src/ecs/resources/input.rs
Xaeroxe: I tried, it seems rustc doesn't support casting closures to fn types yet.  I spent a few hours trying to get this to work and this was the best I could do.  Storing functions in types without using trait objects or generics turns out to be a bit of a sore spot for Rust currently.

src/ecs/resources/input.rs
Xaeroxe: Maybe this is a good reason to rename these functions.  This comment is actually referring to pressed_actions, with pressed_ in front and an s on actions.  It's a different operation from action_pressed.  See key_pressed vs pressed_keys for a comparison.

src/ecs/resources/input.rs
ebkalderon: I meant passing closures into the `map()` methods directly, rather than storing them as types.

src/ecs/resources/input.rs
ebkalderon: Oh, I see what you mean now. Sorry, the terminology is a bit confusing.

src/ecs/resources/input.rs
Xaeroxe: I sincerely would love to do this, but this is the compiler error I get:

```
jacob@jacob-laptop ~/amethyst $ cargo check
   Compiling amethyst v0.4.3 (file:///home/jacob/amethyst)
error: non-capturing closure to fn coercion is experimental (see issue #39817)
   --> src/ecs/resources/input.rs:139:36
    |
139 |         let mouse_button_convert = (|&mb| Button::Mouse(mb)) as fn(&MouseButton) -> Button;
    |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^

error: non-scalar cast: `[closure@src/ecs/resources/input.rs:139:36: 139:61]` as `fn(&glutin::MouseButton) -> ecs::resources::input::Button`
   --> src/ecs/resources/input.rs:139:36
    |
139 |         let mouse_button_convert = (|&mb| Button::Mouse(mb)) as fn(&MouseButton) -> Button;
    |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: non-capturing closure to fn coercion is experimental (see issue #39817)
   --> src/ecs/resources/input.rs:140:27
    |
140 |         let key_convert = (|&k| Button::Key(k)) as fn(&VirtualKeyCode) -> Button;
    |                           ^^^^^^^^^^^^^^^^^^^^^

error: non-scalar cast: `[closure@src/ecs/resources/input.rs:140:27: 140:48]` as `fn(&glutin::VirtualKeyCode) -> ecs::resources::input::Button`
   --> src/ecs/resources/input.rs:140:27
    |
140 |         let key_convert = (|&k| Button::Key(k)) as fn(&VirtualKeyCode) -> Button;
    |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to 4 previous errors

error: Could not compile `amethyst`.

To learn more, run the command again with --verbose.
```

These `fn`s have to be stored in the PressedButtons iterator that is passed back so it is necessary to define a type for them.  I could just Box the iterator but then we're moving this structure onto the heap unnecessarily.

Basically this whole system ends up being a type system jigsaw puzzle and this was the only way I could see to solve it and maintain performance.

ebkalderon: I'm sorry, perhaps I'm not being very clear. I meant you could do:

```rust
let mouse_buttons = self.pressed_mouse_buttons().map(|mb| Button::Mouse(mb));
let keys = self.pressed_keys().map(|k| Button::Key(k));
```

Xaeroxe: Tried that too, with that you get

```
jacob@jacob-laptop ~/amethyst $ cargo check
   Compiling amethyst v0.4.3 (file:///home/jacob/amethyst)
error[E0308]: mismatched types
   --> src/ecs/resources/input.rs:141:36
    |
141 |         PressedButtons { iterator: mouse_buttons.chain(keys) }
    |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^ expected fn pointer, found closure
    |
    = note: expected type `std::iter::Chain<std::iter::Map<ecs::resources::input::PressedMouseButtons<'_>, fn(&glutin::MouseButton) -> ecs::resources::input::Button>, std::iter::Map<ecs::res
ources::input::PressedKeys<'_>, fn(&glutin::VirtualKeyCode) -> ecs::resources::input::Button>>`
               found type `std::iter::Chain<std::iter::Map<ecs::resources::input::PressedMouseButtons<'_>, [closure@src/ecs/resources/input.rs:139:62: 139:85]>, std::iter::Map<ecs::resources
::input::PressedKeys<'_>, [closure@src/ecs/resources/input.rs:140:44: 140:63]>>`

error: Could not compile `amethyst`.

To learn more, run the command again with --verbose.
```

Xaeroxe: I spent hours trying to jam a closure in there and I couldn't see a way to do it.  I'm sorry if I come off flippant, it's just the majority of the time spent on this PR was trying to get this piece of code to compile.

src/ecs/resources/input.rs
Xaeroxe: I just renamed the functions to
pressed_keys()
key_is_pressed()
keys_are_pressed()

I think that helps disambiguate it a bit better.

src/ecs/resources/input.rs
Xaeroxe: Relevant rustc issue: https://github.com/rust-lang/rust/issues/39817

It seems that we'll be able to correct this code in a very short time, as this has been stabilized in the master branch of rustc.

Xaeroxe:
@ebkalderon

Xaeroxe:
Woops, closing was a misclick, my bad.

src/ecs/resources/input.rs
ebkalderon: @Xaeroxe No problem at all; I was just curious. 😁 Thank you for all your effort!

src/ecs/resources/input.rs
ebkalderon: Sounds great! Thank you.

ebkalderon:
@Xaeroxe Thank you so much for addressing all my comments. LGTM!

Xaeroxe:
Excellent!  So what's our usual procedure for merging?  It's gotten approval from 3 collaborators, so I think it's ready now.

ebkalderon:
Aside from pinging @torkleyy that he should probably rebase #244 to the latest upstream, nothing else. Let's merge this now.

PR #252, opened 2017-07-10
Remove unnecessary unsafe Sync impl in Pong example

aochagavia:


Aceeri:
I was also wondering why it was in there when I was sifting through the code.

PR #254, opened 2017-07-12
No nightly support

Xaeroxe:
Add language to indicate we don't officially support nightly or beta branches of Rust.  I'm of the opinion that trying to support nightly or beta is an exercise in futility and a recipe for wasted time.  If others are in agreement with me I'd like to merge this.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,6 +28,10 @@ issues first, as it's possible that someone else reported the same issue before
 you. Though it helps save time, don't worry! We won't mind if you accidentally
 post a duplicate issue.
 
+Please note that if an issue can only be reproduced on a nightly compiler it will be closed and marked "Will not fix."
+If an issue can only be reproduced on a beta compiler the next action is to determine if this issue will also occur
+in the next release of stable.  If it will, we'll address it, otherwise it will be closed as "Will not fix."
+
 That's all there is to it! Thanks for posting your issue; we'll take it to heart
 and try our best to resolve it.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,9 +28,10 @@ issues first, as it's possible that someone else reported the same issue before
 you. Though it helps save time, don't worry! We won't mind if you accidentally
 post a duplicate issue.
 
-Please note that if an issue can only be reproduced on a nightly compiler it will be closed and marked "Will not fix."
-If an issue can only be reproduced on a beta compiler the next action is to determine if this issue will also occur
-in the next release of stable.  If it will, we'll address it, otherwise it will be closed as "Will not fix."
+Amethyst does not officially support beta or nightly channels of Rust.
+If an issue can only be reproduced on those channels our resolution strategy will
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -113,6 +113,8 @@ The API reference can be found in `target/doc/amethyst/index.html`.
 
 ## Questions / Help
 
+We do not support anything other than the most recent Rust stable release.  Use nightly and beta channels with this project at your own risk.
```
torkleyy: Maybe we should say "We do not explicitly support"? Also, there are two spaces between the two sentences.

torkleyy:
Is it possible you pushed to an upstream branch?

Xaeroxe:
@torkleyy This is being merged into the "develop" branch.  I don't feel the word "explicitly" is necessary as it just seems extraneous.  We're already being explicit by putting it here.

I feel it would be a good idea to add this to the readme as projects I've worked in previously did have the express goal of supporting nightly.  I feel like one of our first questions when troubleshooting issues that are difficult to reproduce should be "What version of Rust are you running?" 

This PR was inspired by a recent exchange in gitter that could have been resolved by asking that question at the beginning rather than at the end.

torkleyy:
Since CONTRIBUTING.md pops up when opening an issue, maybe we could add a note that the user should try with stable before opening it?

Xaeroxe:
Agreed, I'll add that.

CONTRIBUTING.md
```diff
@@ -28,6 +28,10 @@ issues first, as it's possible that someone else reported the same issue before
 you. Though it helps save time, don't worry! We won't mind if you accidentally
 post a duplicate issue.
 
+Please note that if an issue can only be reproduced on a nightly compiler it will be closed and marked "Will not fix."
```
torkleyy: Err, this doesn't seem very welcome / may cause more problems later. I mean, if somebody experiences an issue, it may be unknown to rust-lang/rust yet, so we should report it. I'd therefore just say that they should try stable, but still create an issue if it fails on nightly only.

Xaeroxe: I'd say it's the responsibility of the person running the nightly compiler to report that, not ours.  They took the jump into nightly, we didn't.

kvark: I agree with @torkleyy , it's in our best interest to resolve the nightly breakage as soon as it's discovered.

Xaeroxe: If we're accepting nightly issues I'd at least like to add an issue tracker label of "nightly-only"

vitvakatu: I agree with @kvark and @torkleyy, it's too unfriendly to new users. We'll miss a lot of bugs in this case.

Xaeroxe: I guess I want to set the expectation that Amethyst is not going to change its codebase to accommodate rust compiler bugs on beta/nightly that will disappear in a few weeks.  The language used specifically to accomplish that could be improved.

Xaeroxe:
Contributing.MD language has been updated, please review.

PR #256, opened 2017-07-16
Change axis behaviour from button down to pressed

stenverbois:
The `InputHandler::axis_value` function is not really useful when it only returns the value for one frame.
This PR changes the queried button function from `button_down` to `button_is_pressed`.


ebkalderon:
Since there appear to be no objections, I think this is ready for merge. Thanks for your contribution, @stenverbois! 🎉

Xaeroxe:
Yeah this was a typo during my PR.  thanks for the fix @stenverbois

PR #258, opened 2017-07-20
Don't warn when defaults are used for config

ahmedcharles:


torkleyy:
I mentioned the same when reviewing @Aceeri's PR, but they mentioned that we should always have all keys in the config.

Aceeri:
For the most part if you are dealing with a YAML file, then it isn't really a problem since writing to the file will automatically write the defaults to the file. I guess it is more of a problem in TOML though.

I would personally prefer all keys being set in the config, but whichever way the majority wants is fine with me.

Xaeroxe:
Personally I would prefer the all keys approach.  This will allow us to change defaults in the future without breaking existing configurations.  Furthermore having all keys in the config file can act as a form of discoverability, by allowing users to review their config files and hopefully learn about new settings.

ebkalderon:
I also agree that we should have all keys set in the config. However, would anyone object to these warnings still being present in debug builds and turned off when in release mode?

Xaeroxe:
For what reason?

ebkalderon:
@Xaeroxe Since you mentioned the possibility of us changing the default configuration going forward, I was thinking it might be useful to retain the messages so that if someone happens to have a setting missing in the config file, the engine can alert them about it so they can add it back to the file.

Xaeroxe:
Makes sense, but why would we also not want them in release mode?

ebkalderon:
@ahmedcharles Heads-up, it looks like #259 conflicts with your PR. When you have time, could you please rebase against the latest `develop` code? Thanks!

Xaeroxe:
@ahmedcharles as determined by @ebkalderon please either close this PR for modify it to emit the warnings when compiled in debug mode only.

ahmedcharles:
I rebased and came up with a solution that I think has a better experience, which is to only warn when fields aren't explicitly optional. This is similar to serde's behavior when dealing with optional types in it's data model, which is that if they are missing, they become defaulted to whatever that type declares it's none value to be.

ahmedcharles:
Is there anything else I should do for this comment to make it ready?

ebkalderon:
@ahmedcharles Since no one has raised objections these past few days, I think the pull request is safe to merge. Thanks for the contribution! 🎉

ahmedcharles:
Thanks. Glad to have helped.

PR #259, opened 2017-07-20
Moves amethyst_config to root

Aceeri:


Xaeroxe:
@Aceeri Can we adjust the path in /Cargo.toml too?  It contains a workspace member for this crate at the moment.

PR #260, opened 2017-07-20
[Discussion] Support multiple file-type extensions in amethyst_assets

lschmierer:
- Multiple media-types with multiple extensions each
- Throws exception if multiple files with the same names, but different extensions exists

### DIFF
--- a/amethyst_assets/examples/asset.rs
+++ b/amethyst_assets/examples/asset.rs
@@ -29,8 +29,6 @@ impl Asset for DummyAsset {
     }
 }
 
-const DUMMY_EXTS: &[&str] = &["dum", "dummy"];
-
 struct DummyFormat;
 
 impl Format for DummyFormat {
@@ -38,7 +36,8 @@ impl Format for DummyFormat {
     type Error = Utf8Error;
 
     fn extensions() -> &'static [&'static str] {
-        DUMMY_EXTS
+        const DUMMY_EXTENSIONS: &[&str] = &["dum", "dummy"];
+        DUMMY_EXTENSIONS
     }
 
     fn parse(&self, bytes: Vec<u8>) -> Result<Self::Data, Self::Error> {
--- a/amethyst_assets/examples/asset.rs
+++ b/amethyst_assets/examples/asset.rs
@@ -29,14 +29,16 @@ impl Asset for DummyAsset {
     }
 }
 
+const DUMMY_EXTS: &[&str] = &["dum", "dummy"];
+
 struct DummyFormat;
 
 impl Format for DummyFormat {
     type Data = String;
     type Error = Utf8Error;
 
-    fn extension() -> &'static str {
-        "dum"
+    fn extensions() -> &'static [&'static str] {
+        DUMMY_EXTS
     }
 
     fn parse(&self, bytes: Vec<u8>) -> Result<Self::Data, Self::Error> {
@@ -58,6 +60,11 @@ fn main() {
     loader.register::<DummyAsset>(">> ");
 
     let dummy = loader.load("whatever", DummyFormat);
+    let dummy: DummyAsset = dummy.wait().expect("Failed to load dummy asset");
+
+    println!("dummy: {:?}", dummy);
+
+    let dummy = loader.load("whatelse", DummyFormat);
     let dummy: DummyAsset = dummy.wait().expect("Failed to load dummy asset");
 
     println!("dummy: {:?}", dummy);
--- a/amethyst_assets/examples/asset.rs
+++ b/amethyst_assets/examples/asset.rs
@@ -29,18 +29,23 @@ impl Asset for DummyAsset {
     }
 }
 
+const MEDIA_TYPE_DUMMY: &MediaType = "text/dummy";
+
 struct DummyFormat;
 
 impl Format for DummyFormat {
     type Data = String;
     type Error = Utf8Error;
 
-    fn extensions() -> &'static [&'static str] {
-        const DUMMY_EXTENSIONS: &[&str] = &["dum", "dummy"];
+    fn media_extensions() -> &'static [(&'static str, &'static [&'static str])] {
+        const DUMMY_EXTENSIONS: &[(&str, &[&str])] = &[(MEDIA_TYPE_DUMMY, &["dum", "dummy"])];
         DUMMY_EXTENSIONS
     }
 
-    fn parse(&self, bytes: Vec<u8>) -> Result<Self::Data, Self::Error> {
+    fn parse(&self, media_type: &MediaType, bytes: Vec<u8>) -> Result<Self::Data, Self::Error> {
+        if media_type != MEDIA_TYPE_DUMMY {
+            panic!("unsupported media type");
+        }
         from_utf8(bytes.as_slice()).map(|s| s.to_owned())
     }
 }
--- a/amethyst_assets/examples/cache.rs
+++ b/amethyst_assets/examples/cache.rs
@@ -50,18 +50,23 @@ impl Asset for DummyAsset {
     }
 }
 
+const MEDIA_TYPE_DUMMY: &MediaType = "text/dummy";
+
 struct DummyFormat;
 
 impl Format for DummyFormat {
     type Data = String;
     type Error = Utf8Error;
 
-    fn extensions() -> &'static [&'static str] {
-        const DUMMY_EXTENSIONS: &[&str] = &["dum", "dummy"];
+    fn media_extensions() -> &'static [(&'static str, &'static [&'static str])] {
+        const DUMMY_EXTENSIONS: &[(&str, &[&str])] = &[(MEDIA_TYPE_DUMMY, &["dum", "dummy"])];
         DUMMY_EXTENSIONS
     }
 
-    fn parse(&self, bytes: Vec<u8>) -> Result<Self::Data, Self::Error> {
+    fn parse(&self, media_type: &MediaType, bytes: Vec<u8>) -> Result<Self::Data, Self::Error> {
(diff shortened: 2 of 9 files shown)


lschmierer:
I think i will replace the extension list `&[&str]` by a mapping mime type -> extensions `Map<&str, &[&str]>`.

amethyst_assets/src/loader.rs
```diff
@@ -69,10 +71,9 @@ impl<A, E> Future for AssetFuture<A, E> {
                 // reference gets dropped and we enter this branch.
                 // Thus, we are the only one with access to the cell.
                 unsafe {
-                    x.value
-                        .into_inner()
-                        .expect("Thread panicked")
-                        .map(Async::Ready)
+                    x.value.into_inner().expect("Thread panicked").map(
+                        Async::Ready,
```
torkleyy: This formatting looks really weird to me because it does not clearly show that we're passing a function.

amethyst_assets/src/store/dir.rs
```diff
@@ -33,34 +34,66 @@ impl Directory {
...
+            match fs::metadata(path.with_extension(ext)) {
+                Ok(meta) => {
+                    return Ok(
+                        meta.modified()?
+                            .duration_since(UNIX_EPOCH)
+                            .unwrap()
+                            .as_secs(),
+                    );
+                }
+                Err(_) => (),
```
torkleyy: The new `modified` method could return metadata for `file.otherext` if the user lacks permission to access the matadata of `file.correctext`'.

amethyst_assets/src/store/dir.rs
```diff
@@ -33,34 +34,66 @@ impl Directory {
...
-        let mut v = Vec::new();
-        File::open(&path)?.read_to_end(&mut v)?;
+        for ext in exts {
+            match File::open(&path.with_extension(ext)) {
+                Ok(mut file) => {
+                    let mut v = Vec::new();
+                    file.read_to_end(&mut v)?;
+                    return Ok(v);
+                }
+                Err(_) => (),
```
torkleyy: We're discarding the real error here.

lschmierer: Yes that is intended. However, I will change to check for the ErrorKind and return the error if appropriated.

amethyst_assets/src/store/dir.rs
lschmierer: Okay, I will check the ErrorKind

amethyst_assets/src/loader.rs
lschmierer: That's what rustfmt does so I am fine with that.

lschmierer: Is there any way to tell rustfmt to keep specific lines as they are?

amethyst_assets/examples/asset.rs
```diff
@@ -29,14 +29,16 @@ impl Asset for DummyAsset {
     }
 }
 
+const DUMMY_EXTS: &[&str] = &["dum", "dummy"];
```
torkleyy: Could make that a local constant.

amethyst_assets/examples/cache.rs
```diff
@@ -50,14 +50,16 @@ impl Asset for DummyAsset {
     }
 }
 
+const DUMMY_EXTS: &[&str] = &["dum", "dummy"];
```
torkleyy: Same here.

amethyst_assets/src/loader.rs
```diff
@@ -219,85 +233,99 @@ impl Loader {
...
+    N: Into<String>,
+    S: Store,
 {
     let name = name.into();
 
     let store_id = storage.store_id();
-    let spec = AssetSpec::new(name.clone(), F::extension(), store_id);
+    let spec = AssetSpec::new(name.clone(), F::extensions(), store_id);
 
     A::retrieve(context, &spec)
```
torkleyy: Another reason why I didn't do this is that it is very weird with multiple files, same name, but different extensions (belonging to the same format). I don't know how exactly we should handle it, but I don't think having `foo.jpg` and `foo.jpeg` files in the same directory makes too much sense (nor handling it in our asset management).

amethyst_assets/src/store/dir.rs
```diff
@@ -33,34 +34,74 @@ impl Directory {
...
+            match fs::metadata(path.with_extension(ext)) {
+                Ok(meta) => {
+                    return Ok(
+                        meta.modified()?
+                            .duration_since(UNIX_EPOCH)
+                            .unwrap()
+                            .as_secs(),
+                    )
+                }
+                Err(err) => {
```
torkleyy: Suggesting 

```rust
    Err(err) if err.kind() == IoErrorKind::NotFound => {}
    Err(err) => return Err(err),
```

amethyst_assets/src/loader.rs
torkleyy: Yes, but I don't think it's very nice here.

lschmierer:
`Assets` now support extensions by media type. The media type is given back to the `parse` method of the `Format`. This allows us to handle different file types differently.

An `ImageFormat` which uses a `Texture` as `Data` for example could parse "image/png" and "image/jpeg".

The `Store` determines the media type by file extension.

amethyst_assets/src/loader.rs
lschmierer: It was to decide if the asset management needs to have knowledge about file-types at all (https://gitter.im/amethyst/engine?at=597070482723db8d5e2c01b9).

In the current approach, we identify an asset only by name, not by extension. Therefore only one asset of any asset type should be existent.

lschmierer: We can think about throwing an exception if multiple files with some name and different extensions exists.

amethyst_assets/src/asset.rs
```diff
@@ -155,15 +166,15 @@ where
...
     /// ## Examples
     ///
-    /// * `"png"`
-    /// * `"obj"`
-    /// * `"wav"`
-    fn extensions() -> &'static [&'static str];
+    /// * `[("image/png", ["png"), ("image/jpeg", ["jpg", "jpeg"])]`
+    /// * `[("text/plain", ["obj"])]`
+    /// * `[("audio/wav", ["wav"])]`
+    fn media_extensions() -> &'static [(&'static MediaType, &'static [&'static str])];
```
torkleyy: Err..why would you have a single format with different `MediaType`s?

torkleyy: Why not just have multiple format types?

lschmierer: Adapted that from the old asset management, where `AssetLoader` for image for example could load png, jpg, tga and bmp. 
https://github.com/amethyst/amethyst/blob/develop/src/asset_manager/asset_manager.rs#L352

We can change that behaviour. But since png, jpg, ... are also the same kind of "image" assets, that would make sense.

lschmierer: When using https://github.com/lgvz/imagefmt for image laoding, we would have 4 times the same `Format` which are implemented the same.

torkleyy: I'm in favor of the current approach. I just think it's not necessary to do this automatically.

lschmierer: What do you mean by "automatically"?

torkleyy:
I'm really unsure how exactly this is beneficial.

1) We now have two loops for trying out several different formats. Why do more slow IO given that the user clearly knows which one they want?

2) We make assumptions about the files in the directory; it would be easy to have weird bugs here. Additionally, we check if there are multiple files with the same name but different extensions in the folder.

3) The distinction which actual formats should be supported by one `Format` is unclear. When we want to map format identifiers (like `"png"` and `"obj"`) to the format objects, how would we do that with this design? It seems that it makes everything more complex.

After all, it's not a big problem to define multiple format types. And for the image formats: With a `Format` type only supporting for example `.png`, the `image` crate doesn't have to find out about that first.

I'm sorry, but I just don't think this PR is useful. Please tell me if there's something I don't see.

lschmierer:
I added the media types and the error if multiple files exist to encourage discussion and for trying out.
You can give your opinions. That is what a PR is for.

But why is the whole PR not useful? The multi extension part is.

torkleyy:
> The multi extension part is.

Could you explain why that is? It's an additional performance cost and just defining a second format calling the same function is not a problem at all. The mapping of format identifier -> format is just very easy (because you can use the extension). Additionally, there is no weirdness with `foo.png` and `foo.tga` being in the same folder.

nchashch:
I think it would be good to defer this until 0.6.0, because this change is non-trivial and controversial, and making the right decision might take a long time. The decision taking a long time is undesirable because we want to publish 0.5.0 sooner rather than later. Currently the only major thing that is left before 0.5 release is #233.

lschmierer:
> Could you explain why that is? It's an additional performance cost and just defining a second format calling the same function is not a problem at all. The mapping of format identifier -> format is just very easy (because you can use the extension). Additionally, there is no weirdness with foo.png and foo.tga being in the same folder.

@torkleyy I was referring to multiple extensions for a single type (first commit).
Like ".jpg" or ".jpge".

torkleyy:
@lschmierer Me too. Could we just do what @nchashch proposed?

Xaeroxe:
So basically everything this depends on has changed since it was opened.  This PR would have to be rewritten at this point.  I'm going to go ahead and close it.  Feel free to re open this if you feel I was wrong in closing it.

Xaeroxe:
@lschmierer Similar changes were made in https://github.com/amethyst/amethyst/pull/320

PR #261, opened 2017-07-21
Input rebinding phase 2

Xaeroxe:
Things done in this PR

- Use smallvec for InputHandler internals, reducing allocations required.
- Add functions
    - down_buttons
    - released_buttons
    - down_keys
    - released_keys
    - down_mouse_buttons
    - released_mouse_buttons

These functions will be helpful with assigning bindings in the first place when we just need to generically query for "whatever the user pressed."

- Add functions
    - axes
    - actions

These functions will be helpful for figuring out what to display on input rebinding screens.
- Enhance actions API to support multiple bindings per action
- Move private functions into closures
- Remove the opaque wrappers over iterators, return slice iterators where possible and in other places return a single `ButtonIterator` type.
- Change actions and axes to use string keys instead of integer keys.

Eventually there is going to be a phase 3 to make this interact with the asset system, which I plan on doing as soon as [Ron](https://github.com/kvark/ron) is ready for use.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,11 +36,11 @@ gfx_device_gl = "0.13"
 gfx_window_glutin = "0.14"
 glutin = "0.7"
 imagefmt = "4.0"
+smallvec = "0.4"
 specs = "0.9.1"
 rayon = "0.7"
 ticketed_lock = "0.1"
 wavefront_obj = "5.0"
-smallvec = "0.4"
 
 thread_profiler = { version = "0.1", optional = true }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ specs = "0.9.1"
 rayon = "0.7"
 ticketed_lock = "0.1"
 wavefront_obj = "5.0"
+smallvec = "0.4"
 
 thread_profiler = { version = "0.1", optional = true }
 
--- a/src/ecs/resources/input.rs
+++ b/src/ecs/resources/input.rs
@@ -168,18 +168,18 @@ impl InputHandler {
     }
 
     /// Returns a vector containing pressed down keys.
-    pub fn pressed_keys(&self) -> &[VirtualKeyCode] {
-        &*self.pressed_keys
+    pub fn pressed_keys(&self) -> Iter<VirtualKeyCode> {
+        self.pressed_keys.iter()
     }
 
     /// Returns a vector containing keys pressed on this frame.
-    pub fn down_keys(&self) -> &[VirtualKeyCode] {
-        &*self.down_keys
+    pub fn down_keys(&self) -> Iter<VirtualKeyCode> {
+        self.down_keys.iter()
     }
 
     /// Returns a vector containing keys released on this frame.
-    pub fn released_keys(&self) -> &[VirtualKeyCode] {
-        &*self.released_keys
+    pub fn released_keys(&self) -> Iter<VirtualKeyCode> {
+        self.released_keys.iter()
     }
 
     /// Checks if the given key is being pressed.
@@ -208,18 +208,18 @@ impl InputHandler {
     }
 
     /// Returns a vector containing pressed down mouse buttons.
-    pub fn pressed_mouse_buttons(&self) -> &[MouseButton] {
-        &*self.pressed_mouse_buttons
+    pub fn pressed_mouse_buttons(&self) -> Iter<MouseButton> {
+        self.pressed_mouse_buttons.iter()
     }
 
     /// Returns a vector containing mouse buttons pressed on this frame.
-    pub fn down_mouse_buttons(&self) -> &[MouseButton]  {
-        &*self.down_mouse_buttons
+    pub fn down_mouse_buttons(&self) -> Iter<MouseButton>  {
+        self.down_mouse_buttons.iter()
     }
 
     /// Returns a vector containing mouse buttons released on this frame.
-    pub fn released_mouse_buttons(&self) -> &[MouseButton] {
-        &*self.released_mouse_buttons
+    pub fn released_mouse_buttons(&self) -> Iter<MouseButton> {
+        self.released_mouse_buttons.iter()
     }
 
     /// Checks if the given mouse button is being pressed.
--- a/src/ecs/resources/input.rs
+++ b/src/ecs/resources/input.rs
@@ -32,9 +32,9 @@ impl From<MouseButton> for Button {
 /// Represents an axis made up of digital inputs, like W and S or A and D.
 /// Two of these could be analogous to a DPAD.
 pub struct Axis {
-    /// Positive button, when pressed down the axis value will return 1 if `neg` is not pressed down.
+    /// Positive button, when pressed down axis value will return 1 if `neg` is not pressed down.
     pub pos: Button,
-    /// Negative button, when pressed down the axis value will return -1 if `neg` is not pressed down.
+    /// Negative button, when pressed down axis value will return -1 if `neg` is not pressed down.
     pub neg: Button,
 }
 
@@ -309,7 +309,7 @@ impl InputHandler {
         }
     }
 
-    /// Checks if the all the given buttons are being pressed and at least one was pressed this frame.
+    /// Checks if the all given buttons are being pressed and at least one was pressed this frame.
     pub fn buttons_down(&self, buttons: &[Button]) -> bool {
         buttons.iter().any(|&b| self.button_down(b)) &&
         buttons.iter().all(|&b| self.button_is_pressed(b))
@@ -380,7 +380,7 @@ impl InputHandler {
             .map(|ref buttons| buttons.iter().any(|&b| self.button_released(b)))
     }
 
-    /// Checks if the all the given actions are being pressed and at least one was pressed this frame.
+    /// Checks if the all given actions are being pressed and at least one was pressed this frame.
     ///
     /// If any action in this list is invalid this will return the id of it in Err.
     pub fn actions_down(&self, actions: &[i32]) -> Result<bool, Vec<i32>> {
--- a/src/ecs/resources/input.rs
+++ b/src/ecs/resources/input.rs
@@ -333,9 +333,9 @@ impl InputHandler {
     }
 
     /// Returns the value of an axis by the i32 id, if the id doesn't exist this returns None.
-    pub fn axis_value(&self, id: i32) -> Option<f32> {
+    pub fn axis_value(&self, id: &str) -> Option<f32> {
         self.axes
-            .get(&id)
+            .get(id)
             .map(|a| {
                 let pos = self.button_is_pressed(a.pos);
                 let neg = self.button_is_pressed(a.neg);
@@ -447,7 +447,7 @@ impl InputHandler {
 
     /// Get's a list of all axes
     pub fn get_axes(&self) -> Vec<String> {
-        self.axes.keys().map(|&k| k).collect::<Vec<String>>()
+        self.axes.keys().map(|k| k.clone()).collect::<Vec<String>>()
     }
 
     /// Add a button to an action.
@@ -496,6 +496,6 @@ impl InputHandler {
 
     /// Get's a list of all action bindings
     pub fn get_actions(&self) -> Vec<String> {
-        self.actions.keys().map(|&k| k).collect::<Vec<String>>()
+        self.actions.keys().map(|k| k.clone()).collect::<Vec<String>>()
     }
 }
--- a/src/ecs/resources/input.rs
+++ b/src/ecs/resources/input.rs
@@ -83,8 +83,8 @@ pub struct InputHandler {
     released_mouse_buttons: SmallVec<[MouseButton; 8]>,
     mouse_position: Option<(i32, i32)>,
     previous_mouse_position: Option<(i32, i32)>,
-    axes: HashMap<i32, Axis>,
-    actions: HashMap<i32, SmallVec<[Button; 8]>>,
+    axes: HashMap<String, Axis>,
+    actions: HashMap<String, SmallVec<[Button; 8]>>,
     text_this_frame: String,
 }
 
@@ -353,27 +353,27 @@ impl InputHandler {
     // can't think of a use case for it.
 
     /// Checks if the given button is currently pressed.
-    pub fn action_is_pressed(&self, action: i32) -> Option<bool> {
+    pub fn action_is_pressed(&self, action: &str) -> Option<bool> {
         self.actions
-            .get(&action)
+            .get(action)
             .map(|ref buttons| buttons.iter().any(|&b| self.button_is_pressed(b)))
     }
 
     /// Checks if all the given actions are pressed.
     ///
     /// If any action in this list is invalid this will return the id of it in Err.
-    pub fn actions_are_pressed(&self, actions: &[i32]) -> Result<bool, Vec<i32>> {
+    pub fn actions_are_pressed(&self, actions: &[&str]) -> Result<bool, Vec<String>> {
         let mut all_buttons_are_pressed = true;
         let mut bad_values = Vec::new();
         for action in actions {
-            if let Some(buttons) = self.actions.get(action) {
+            if let Some(buttons) = self.actions.get(*action) {
                 if all_buttons_are_pressed {
                     if !buttons.iter().any(|&b| self.button_is_pressed(b)) {
                         all_buttons_are_pressed = false;
                     }
                 }
             } else {
-                bad_values.push(*action);
+                bad_values.push(action.to_string());
             }
         }
         if !bad_values.is_empty() {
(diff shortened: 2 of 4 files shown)


src/ecs/resources/input.rs
```diff
@@ -267,61 +154,74 @@ impl InputHandler {
         self.text_this_frame.as_str()
     }
 
-    /// Returns an iterator for all the pressed down keys.
-    pub fn pressed_keys(&self) -> PressedKeys {
-        self.current_frame.pressed_keys()
+    /// Returns a vector containing pressed down keys.
+    pub fn pressed_keys(&self) -> &SmallVec<[VirtualKeyCode; 16]> {
```
torkleyy: Could we just coerce this to a slice?

Xaeroxe: We can and that idea is like 1000x better.  Thank you!

src/ecs/resources/input.rs
```diff
@@ -195,18 +208,18 @@ impl InputHandler {
...
-    pub fn down_mouse_buttons(&self) -> &SmallVec<[MouseButton; 8]> {
-        &self.down_mouse_buttons
+    pub fn down_mouse_buttons(&self) -> &[MouseButton]  {
+        &*self.down_mouse_buttons
     }
 
     /// Returns a vector containing mouse buttons released on this frame.
-    pub fn released_mouse_buttons(&self) -> &SmallVec<[MouseButton; 8]> {
-        &self.released_mouse_buttons
+    pub fn released_mouse_buttons(&self) -> &[MouseButton] {
```
torkleyy: I would find it more consistent to return an iterator everywhere or a slice everywhere. Since the latter seems to be impossible, let's go for the former. You can just use `slice::Iter` here.

src/ecs/resources/input.rs
```diff
@@ -486,29 +445,57 @@ impl InputHandler {
...
+            self.actions.remove(&id);
+        }
     }
-}
 
-fn mouse_button_to_button(mb: &MouseButton) -> Button {
-    Button::Mouse(*mb)
-}
+    /// Returns an action's bindings.
+    pub fn get_action_bindings(&self, id: i32) -> Option<SmallVec<[Button; 8]>> {
```
torkleyy: By convention, getters should not start with `get` if possible.

src/ecs/resources/mod.rs
```diff
@@ -11,6 +11,6 @@ mod broadcaster;
 
 pub use self::broadcaster::Broadcaster;
 pub use self::camera::{Camera, Projection};
-pub use self::input::{Axis, Button, InputHandler, PressedButtons, PressedKeys, PressedMouseButtons};
+pub use self::input::{Axis, Button, InputHandler, ButtonIterator};
```
torkleyy: Please don't forget to order these imports.

src/lib.rs
```diff
@@ -71,6 +71,7 @@ pub extern crate amethyst_renderer as renderer;
 pub extern crate amethyst_config as config;
 
 extern crate cgmath;
+extern crate smallvec;
```
torkleyy: Please move it down so the alphabetical order is maintained.

src/ecs/resources/input.rs
```diff
@@ -333,9 +333,9 @@ impl InputHandler {
     }
 
     /// Returns the value of an axis by the i32 id, if the id doesn't exist this returns None.
-    pub fn axis_value(&self, id: i32) -> Option<f32> {
+    pub fn axis_value(&self, id: &str) -> Option<f32> {
```
ebkalderon: Since this is a user-facing API, could we use `AsRef<str>` instead of `&str` in all the `axis_` methods so users can easily pass in different kinds of string types?

Cargo.toml
```diff
@@ -40,6 +40,7 @@ specs = "0.9.1"
 rayon = "0.7"
 ticketed_lock = "0.1"
 wavefront_obj = "5.0"
+smallvec = "0.4"
```
ebkalderon: Please move this entry upward to retain alphabetical order.

src/ecs/resources/input.rs
```diff
@@ -433,82 +384,118 @@ impl InputHandler {
...
+    pub fn insert_action_binding(&mut self, id: &str, binding: Button) {
+        let mut make_new = false;
+        match self.actions.get_mut(id) {
+            Some(action_bindings) => {
+                if action_bindings.iter().all(|&b| b != binding) {
+                    action_bindings.push(binding);
+                }
+            }
+            None => {
+                make_new = true;
```
ebkalderon: Why have a `make_new` variable? Can't lines 469-471 be copied into this branch verbatim and be easier to read?

src/ecs/resources/input.rs
```diff
@@ -433,82 +384,118 @@ impl InputHandler {
...
+    /// Removes an action binding that was assigned previously.
+    pub fn remove_action_binding(&mut self, id: &str, binding: Button) {
+        let mut kill_it = false;
+        if let Some(action_bindings) = self.actions.get_mut(id) {
+            let index = action_bindings.iter().position(|&b| b == binding);
+            if let Some(index) = index {
+                action_bindings.swap_remove(index);
+            }
+            if action_bindings.len() == 0 {
+                kill_it = true;
```
ebkalderon: Same comment as with the `make_new` variable in `insert_action_binding`.

src/ecs/resources/input.rs
```diff
@@ -267,61 +167,74 @@ impl InputHandler {
         self.text_this_frame.as_str()
     }
 
-    /// Returns an iterator for all the pressed down keys.
-    pub fn pressed_keys(&self) -> PressedKeys {
-        self.current_frame.pressed_keys()
+    /// Returns a vector containing pressed down keys.
+    pub fn pressed_keys(&self) -> Iter<VirtualKeyCode> {
```
ebkalderon: Could this be made into a type alias called `KeyCodes`?

src/ecs/resources/input.rs
```diff
@@ -267,61 +167,74 @@ impl InputHandler {
...
+    /// Checks if all the given keys are being pressed.
+    pub fn keys_are_pressed(&self, keys: &[VirtualKeyCode]) -> bool {
+        keys.iter().all(|key| self.key_is_pressed(*key))
     }
 
-    /// Checks if the given mouse button is being pressed.
-    pub fn mouse_button_is_pressed(&self, button: MouseButton) -> bool {
-        self.current_frame.mouse_button_is_pressed(button)
+    /// Returns a vector containing pressed down mouse buttons.
+    pub fn pressed_mouse_buttons(&self) -> Iter<MouseButton> {
```
ebkalderon: Could this be made into a type alias called `MouseButtons`?

src/ecs/resources/input.rs
```diff
@@ -335,30 +248,66 @@ impl InputHandler {
...
+            .iter()
+            .map((|&mb| Button::Mouse(mb)) as fn(&MouseButton) -> Button);
+        let keys = self.pressed_keys.iter().map((|&k| Button::Key(k)) as fn(&VirtualKeyCode) -> Button);
+        ButtonIterator {
+            iterator: mouse_buttons.chain(keys)
+        }
+    }
+
+    /// Returns a vector containing the buttons that were pressed this frame
+    pub fn down_buttons(&self) -> ButtonIterator {
```
ebkalderon: Could `ButtonIterator` be renamed to `Buttons` to remain consistent with other iterator names?

src/ecs/resources/input.rs
Xaeroxe: No, because it makes the borrow checker mad.  self.actions is already borrowed thanks to the match statement.

src/ecs/resources/input.rs
torkleyy: I disagree. It's more difficult to read and you'll almost always pass a literal. If we argue with "it can save the user a method call sometimes", then Rust's choice being explicit about conversions doesn't make sense.

src/ecs/resources/input.rs
ebkalderon: Ahh, I see. I wonder if the upcoming non-lexical lifetimes in Rust will help with these kinds of issues.

src/ecs/resources/input.rs
ebkalderon: The `AsRef` trait exists precisely for this reason, though. It's meant to eliminate boilerplate in case users decide to pass in a `String`, `CString`, `OsString`, or perhaps even a simple array of bytes, and more importantly, it can prevent extra heap allocations on the user's end.

torkleyy: The last part is only valid for `Into<String>`, not for `AsRef`.

ebkalderon: Good point, I looked into it and I must have indeed been thinking about `Into<String>`. But my first point still stands. Even the Rust standard library makes frequent use of `AsRef` in exactly this way to avoid such boilerplate, and I think that users might appreciate an API that assists with zero runtime cost.

Xaeroxe: If we use `Into<String>` we'll likely be introducing an unnecessary allocation, as the hashmap prefers to take keys by reference rather than by value.  Most usage of this is going to be string literals.

ebkalderon: @Xaeroxe Sorry, perhaps I wasn't clear enough. I am suggesting we use `AsRef<str>`, not `Into<String>`. I had mistakenly pointed out that `AsRef<str>` would reduce unnecessary allocations, confusing it for `Into<String>`. However, I still stand by my point that `AsRef<str>` will behave identically to `&str` except it will allow users to pass in other string slice types like `OsStr` and `CStr` with less boilerplate.

Xaeroxe:
@ebkalderon Changes you requested have been made.

src/ecs/resources/input.rs
```diff
@@ -267,61 +173,74 @@ impl InputHandler {
         self.text_this_frame.as_str()
     }
 
-    /// Returns an iterator for all the pressed down keys.
-    pub fn pressed_keys(&self) -> PressedKeys {
-        self.current_frame.pressed_keys()
+    /// Returns a vector containing pressed down keys.
```
torkleyy: Doc comment seems outdated.

src/ecs/resources/input.rs
```diff
@@ -267,61 +173,74 @@ impl InputHandler {
...
+    pub fn pressed_keys(&self) -> KeyCodes {
+        self.pressed_keys.iter()
     }
 
-    /// Checks if the given key is being pressed.
-    pub fn key_is_pressed(&self, key: VirtualKeyCode) -> bool {
-        self.current_frame.key_is_pressed(key)
+    /// Returns a vector containing keys pressed on this frame.
```
torkleyy: Same here

src/ecs/resources/input.rs
```diff
@@ -267,61 +173,74 @@ impl InputHandler {
...
+    pub fn down_keys(&self) -> KeyCodes {
+        self.down_keys.iter()
     }
 
-    /// Checks if all the given keys are being pressed.
-    pub fn keys_are_pressed(&self, keys: &[VirtualKeyCode]) -> bool {
-        self.current_frame.keys_are_pressed(keys)
+    /// Returns a vector containing keys released on this frame.
```
torkleyy: Same here

src/ecs/resources/input.rs
```diff
@@ -267,61 +173,74 @@ impl InputHandler {
...
 
-    /// Checks if all the given mouse buttons are being pressed.
-    pub fn mouse_buttons_are_pressed(&self, buttons: &[MouseButton]) -> bool {
-        self.current_frame.mouse_buttons_are_pressed(buttons)
+    /// Returns a vector containing mouse buttons pressed on this frame.
+    pub fn down_mouse_buttons(&self) -> MouseButtons  {
+        self.down_mouse_buttons.iter()
+    }
+
+    /// Returns a vector containing mouse buttons released on this frame.
```
torkleyy: Same here.

src/ecs/resources/input.rs
```diff
@@ -404,25 +359,27 @@ impl InputHandler {
...
-            .map(|&b| self.button_is_pressed(b))
+            .get(action.as_ref())
+            .map(|ref buttons| buttons.iter().any(|&b| self.button_is_pressed(b)))
     }
 
     /// Checks if all the given actions are pressed.
     ///
     /// If any action in this list is invalid this will return the id of it in Err.
-    pub fn actions_are_pressed(&self, actions: &[i32]) -> Result<bool, Vec<i32>> {
+    pub fn actions_are_pressed<T: AsRef<str>>(&self, actions: &[T]) -> Result<bool, Vec<String>> {
```
torkleyy: Can we remove the verb here like it is the case above with the mouse buttons?

src/ecs/resources/input.rs
```diff
@@ -404,25 +359,27 @@ impl InputHandler {
...
         let mut all_buttons_are_pressed = true;
         let mut bad_values = Vec::new();
         for action in actions {
```
torkleyy: Could you use iterators to make this code nicer? I don't think there is a need to report multiple ill-named actions at once, given that is pretty rare, so that could be another simplification.

src/ecs/resources/input.rs
```diff
@@ -433,82 +390,118 @@ impl InputHandler {
     }
 
     /// Checks if the given action was pressed on this frame.
-    pub fn action_down(&self, action: i32) -> Option<bool> {
-        self.actions.get(&action).map(|&b| self.button_down(b))
+    pub fn action_down<T: AsRef<str>>(&self, action: T) -> Option<bool> {
+        self.actions
+            .get(action.as_ref())
+            .map(|ref buttons| buttons.iter().any(|&b| self.button_down(b)))
```
torkleyy: Why do we need the `ref` here?

src/ecs/resources/input.rs
```diff
@@ -433,82 +390,118 @@ impl InputHandler {
...
+            Ok(all_actions_are_pressed && any_action_is_pressed_this_frame)
         }
     }
 
     /// Assign an axis to an ID value
     ///
     /// This will insert a new axis if no entry for this id exists.
     /// If one does exist this will replace the axis at that id and return it.
-    pub fn insert_axis(&mut self, id: i32, axis: Axis) -> Option<Axis> {
+    pub fn insert_axis(&mut self, id: String, axis: Axis) -> Option<Axis> {
```
torkleyy: Sorry if it has been mentioned before, but why not be consistent and accept `Into<String>` here? I mean, this is the place where it would actually make the code shorter in most cases (because you most likely have a `&str`).

src/ecs/resources/input.rs
```diff
@@ -433,82 +390,118 @@ impl InputHandler {
...
+        self.axes.get(id.as_ref())
     }
 
-    /// Assign an action to an ID value
-    ///
-    /// This will insert a new action if no entry for this id exists.
-    /// If one does exist this will replace the action at that id and return it.
-    pub fn insert_action(&mut self, id: i32, binding: Button) -> Option<Button> {
-        self.actions.insert(id, binding)
+    /// Get's a list of all axes
```
torkleyy: I don't think we need the `'`.

src/ecs/resources/input.rs
```diff
@@ -433,82 +390,118 @@ impl InputHandler {
...
+    pub fn axes(&self) -> Vec<String> {
+        self.axes.keys().map(|k| k.clone()).collect::<Vec<String>>()
```
torkleyy: Why collect it?

src/ecs/resources/input.rs
```diff
@@ -433,82 +390,118 @@ impl InputHandler {
...
 
-    /// Removes an action, this will return the removed action if successful.
-    pub fn remove_action(&mut self, id: i32) -> Option<Button> {
-        self.actions.remove(&id)
+    /// Add a button to an action.
+    ///
+    /// This will insert a new binding between this action and the button.
+    pub fn insert_action_binding<T: AsRef<str>>(&mut self, id: T, binding: Button) {
+        let mut make_new = false;
+        match self.actions.get_mut(id.as_ref()) {
```
torkleyy: I think we can use the Entry API here.

src/ecs/resources/input.rs
```diff
@@ -433,82 +390,118 @@ impl InputHandler {
...
-        self.actions.get(&id)
+    /// Removes an action binding that was assigned previously.
+    pub fn remove_action_binding<T: AsRef<str>>(&mut self, id: T, binding: Button) {
+        let mut kill_it = false;
+        if let Some(action_bindings) = self.actions.get_mut(id.as_ref()) {
+            let index = action_bindings.iter().position(|&b| b == binding);
+            if let Some(index) = index {
+                action_bindings.swap_remove(index);
+            }
+            if action_bindings.len() == 0 {
```
torkleyy: Let's use late binding here: Change the line at the beginning of the method to `let kill_it;` and change this line to `kill_it = action_bindings.is_empty()`.

src/ecs/resources/mod.rs
```diff
@@ -11,6 +11,6 @@ mod broadcaster;
 
 pub use self::broadcaster::Broadcaster;
 pub use self::camera::{Camera, Projection};
-pub use self::input::{Axis, Button, InputHandler, PressedButtons, PressedKeys, PressedMouseButtons};
+pub use self::input::{Axis, Button, Buttons, KeyCodes, MouseButtons, InputHandler};
```
torkleyy: Sorry, but this is not ordered alphabetically. Hope I don't annoy you with that.

kvark: Alphabetical ordering together with the maximum line width is super annoying. Gotta redo quite a few lines just to make a simple change :(

torkleyy: So what do you suggest? I just think that alphabetical order helps when you read them (and it looks really clean and tidy).

src/ecs/resources/input.rs
Xaeroxe: We can, I used it, it mandated an extra allocation in situations where it wasn't necessary, so I took it out.

src/ecs/resources/input.rs
Xaeroxe: Verb is used to visually distinguish this function name, otherwise the function names get too similar and people get confused.

src/ecs/resources/input.rs
Xaeroxe: It is pretty rare, so I don't want to create a new opaque Iterator type to handle this situation.  This should only happen in development environments.

src/ecs/resources/input.rs
Xaeroxe: Cuz I don't want another opaque Iterator type for this function, and this function is called very infrequently.

src/ecs/resources/input.rs
Xaeroxe: Because I didn't think of that :P

src/ecs/resources/input.rs
Xaeroxe: Not possible, if the given action isn't valid then we can't run `kill_it = action_bindings.is_empty()` which leaves kill_it uninitialized.

src/ecs/resources/input.rs
Xaeroxe: I guess we don't.  Artifact of previous version of this code.

src/ecs/resources/input.rs
torkleyy: I was talking about iterating over `actions`.

Xaeroxe: Right, I knew that.  What I'm getting at is it's not necessary to aggressively optimize this because the error case (an action is invalid) should never happen in production code.

Xaeroxe: Oh!  You mean actions the function argument!  Yeah I thought about that, but I can optimize two passes into one pass by doing it this way.

Xaeroxe:
@torkleyy Requesting either your approval or additional input

PR #265, opened 2017-07-23
Add Audio support

Xaeroxe:
# Things done in this PR:
- Add rodio dependency
- Add amethyst_assets dependency to amethyst
- Add modules:
   - audio
      - dj
          - pub struct Dj
      - endpoint
          - pub struct Endpoint
          - pub fn default_endpoint()
          - pub fn endpoints()
      - play
          - pub fn once(souce: &Source, endpoint: &Endpoint)
      - source
          - pub struct Source, implements amethyst_assets::Asset
          - pub(crate) struct AudioDataReference
          - pub struct WavFormat, implements amethyst_assets::Format
          - pub struct OggFormat, implements amethyst_assets::Format
          - pub struct FlacFormat, implements amethyst_assets::Format
- Add AudioSystem which manages position updating for 3D sounds.
- Add AudioEmitter component for playing sounds
- Add AudioListener component for listening to those sounds.
- Make pong example have sfx.


Xaeroxe:
Linux developers: the libasound2-dev package will be required to compile amethyst once this package is merged, rodio interfaces with alsa which makes this necessary.

vitvakatu:
@Xaeroxe 
> Linux developers: the libasound2-dev package will be required to compile amethyst once this package is merged, rodio interfaces with alsa which makes this necessary.

Add information about it in readme, please

Cargo.toml
```diff
@@ -39,11 +40,15 @@ imagefmt = "4.0"
 smallvec = "0.4"
 specs = "0.9.1"
 rayon = "0.7"
+rodio = "0.5.1"
 ticketed_lock = "0.1"
 wavefront_obj = "5.0"
 
 thread_profiler = { version = "0.1", optional = true }
 
+[dev_dependencies]
```
torkleyy: I thought it was `dev-dependencies`?

src/audio/endpoint.rs
```diff
@@ -0,0 +1,25 @@
+use rodio::Endpoint as REndpoint;
+use rodio::*;
+
+/// A speaker(s) through which audio can be played.
+pub struct Endpoint {
```
torkleyy: Is there a reason for this wrapper?

src/audio/mod.rs
```diff
@@ -0,0 +1,10 @@
+//! Loading and playing of audio files.
+
+/// Provides structures used to load audio files.
+pub mod source;
+
+/// Provides functions used to play audio.
```
torkleyy: Would prefer to have module docs in the actual module with `//!`.

src/audio/endpoint.rs
Xaeroxe: Mostly just making it so users of the API don't also have to depend on rodio.

torkleyy: Could we give it another name? I think something like `Output` or `OutputDevice` is probably clearer to people who didn't deal with `cpal` stuff before.

Cargo.toml
Xaeroxe: I guess it is.  Woops.  For some reason it still worked when I added this but I'll fix it.

src/audio/endpoint.rs
ebkalderon: I agree, either `Output` or `Sink` are good name choices IMO.

ebkalderon:
Will this pull request eventually include an `AudioSystem` and components for audio listeners and emitters?

Xaeroxe:
@ebkalderon Yes it will, I'm not going to be able to work on this for a few days as my life is a little busy recently but I'll get back to this PR as soon as I can

src/audio/source.rs
```diff
@@ -0,0 +1,84 @@
...
+        "audio"
+    }
+
+    fn from_data(data: Vec<u8>, _: &()) -> Result<Self, Self::Error> {
+        Ok(Source { reference: AudioDataReference { raw: Arc::new(data) } })
+    }
+}
+
+/// Loads audio from wav files.
+pub struct WavFormat;
```
torkleyy: Could we have a module `assets` which reexports `*` from `amethyst_assets` and add the amethyst-specific functionality in there similar to how we do it with `amethyst::ecs`?

So like

```rust
// src/assets/mod.rs

use amethyst_assets::*;

pub mod formats;
```

and then keep the formats in the `formats` module.

torkleyy: But I'm also not sure whether or not formats should belong to `amethyst_assets`.

Xaeroxe: @torkleyy I'd prefer to keep the formats in Amethyst.  I also like your first proposal, @ebkalderon Does this sound good to you?  Bunching up all asset formats into `assets::formats` would help with asset type discoverability.

ebkalderon: Yes, keeping the assets in a `formats` module sounds good to me. 👍

alteous: It may be worth dropping the `Format` suffix if all the format types are put into one module.

torkleyy:
I think this PR is worth a line in the `CHANGELOG.md`.

examples/06_audio/assets/config.yml
```diff
@@ -0,0 +1,10 @@
+# ---
```
torkleyy: Should be a RON file.

src/assets/mod.rs
```diff
@@ -0,0 +1,5 @@
+//! Re-exports amethyst_assets and provides formats specific to the engine.
+
+pub use amethyst_assets::*;
+
+pub mod formats;
```
torkleyy: Nice, exactly the structure I would have used.

src/audio/audio_context.rs
```diff
@@ -0,0 +1,37 @@
+//! Provides a context for loading a file.
+
+use super::source::Source;
+use assets::*;
+
+use std::sync::Arc;
+
+/// A context for loading audio files
+pub struct AudioContext {
+    inner: SimpleContext<Source, Vec<u8>, NoError, fn(Vec<u8>) -> Result<Source, NoError>>,
```
torkleyy: Not sure about that dynamic dispatch here..I think we should replace that. Pretty sure a change in `amethyst_assets` is required.

src/audio/mod.rs
```diff
@@ -0,0 +1,9 @@
+//! Loading and playing of audio files.
+pub mod play;
+pub mod output;
+
+mod audio_context;
+mod source;
+
+pub use self::audio_context::AudioContext;
```
torkleyy: I guess we don't need such a deep hierarchy. I would move source and context into this file, actually.

src/audio/output.rs
```diff
@@ -0,0 +1,27 @@
...
+}
+
+/// Get the default output, returns none if no outputs are available.
+pub fn default_output() -> Option<Output> {
+    get_default_endpoint().map(|re| Output { endpoint: re })
+}
+
+/// Get a list of outputs available to the system.
+pub fn outputs() -> Vec<Output> {
+    get_endpoints_list().map(|re| Output {endpoint: re}).collect::<Vec<Output>>()
```
torkleyy: Is that allocation necessary?

src/audio/play.rs
```diff
@@ -0,0 +1,27 @@
...
+        Ok(sink) => {
+            sink.detach();
+            Ok(())
+        }
+
+        /// There is one and only one error that can be returned, which is unrecognized format
+        /// See documentation for DecoderError here:
+        /// https://docs.rs/rodio/0.5.1/rodio/decoder/enum.DecoderError.html
+        Err(_) => {
+            Err(())
```
torkleyy: Could we still return a proper error here?

src/audio/audio_context.rs
Xaeroxe: @torkleyy Just to clarify: Dynamic dispatch usually refers to a vtable lookup occurring, this doesn't involve a vtable lookup though, it just uses a hard coded function pointer so this isn't really that expensive.

src/audio/output.rs
Xaeroxe: I guess it isn't.  (I'm really looking forward to `impl Trait` so that I can have efficient interfaces that aren't a pain to write)

src/audio/mod.rs
Xaeroxe: I like to keep my mod files as clean as possible, make them only describe the hierarchy and exports.  Makes the code a little easier to read in my opinion.  What do you think of that?

examples/06_audio/assets/config.yml
Xaeroxe: Woops, yah sorry copy paste error.

CHANGELOG.md
```diff
@@ -10,12 +10,14 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 ## [Unreleased]
 ### Changed
 * Asset management rewrite (pull request [#244]).
+* Add audio support ([#265])
```
Xaeroxe: Thanks!

src/audio/audio_context.rs
torkleyy: Oh yeah sorry, I'm pretty inaccurate about that. The problem with a runtime function pointer is that inlining isn't possible. And it is a bit ugly in a type signature IMO.

src/audio/play.rs
torkleyy: Not `()` :smile: Just think about the case where a user unwraps it or returns `Box<Error>` or similar cases. A unit type `DecoderError` is much more descriptive here.

src/ecs/components/audio/audio_listener.rs
```diff
@@ -0,0 +1,13 @@
+use ecs::{Component, VecStorage};
+
+/// An audio listener, add this component to the local player character.
+pub struct AudioListener {
+    /// Position of the left_ear in the world.
+    pub left_ear: [f32; 3],
```
torkleyy: Not sure if this should be a global position.

src/ecs/components/audio/audio_listener.rs
```diff
@@ -0,0 +1,13 @@
...
+    /// Position of the right ear in the world.
+    pub right_ear: [f32; 3],
+}
+
+impl Component for AudioListener {
+    type Storage = VecStorage<Self>;
```
torkleyy: As far as I see it, there is only one `AudioListener`, so `VecStorage` is not very efficient here.

src/ecs/components/audio/audio_emitter.rs
```diff
@@ -0,0 +1,31 @@
...
+    /// Creates a new AudioEmitter component initialized to the given positions.  These positions will stay synced with Transform
+    /// if the Transform component is available on this entity.
+    pub fn new(output: &Output, emitter_position: [f32; 3], left_ear: [f32; 3], right_ear: [f32; 3]) -> AudioEmitter {
+        AudioEmitter {
+            sink: SpatialSink::new(&output.endpoint, emitter_position, left_ear, right_ear),
+        }
+    }
+
+    /// Adds an audio source to the queue of sounds to be played from this emitter.
+    pub fn append(&self, source: &Source) {
```
torkleyy: Shouldn't a source also allow playing multiple sounds at once? If that's already the case, I would ask you to improve naming or docs.

src/ecs/components/audio/audio_listener.rs
Xaeroxe: Yeah I had the same thought after I pushed the commit.  I'll fix it.

src/ecs/components/audio/audio_listener.rs
Xaeroxe: Alright, I'll check out the other storage types and see if I can find a better one.

src/ecs/components/audio/audio_emitter.rs
Xaeroxe: Current architecture doesn't permit multiple sounds at once, but it probably should.  I'll think about ways to accomplish this.

src/ecs/systems/audio.rs
```diff
@@ -0,0 +1,117 @@
...
+                let right_ear_position = listener_transform
+                    .transform_point(Point3::from(listener.right_ear))
+                    .into();
+                for (transform, mut audio_emitter) in (&transform, &mut audio_emitter).join() {
+                    let x = transform.0[3][0];
+                    let y = transform.0[3][1];
+                    let z = transform.0[3][2];
+                    let emitter_position = [x, y, z];
+                    // Remove all sinks whose sounds have ended.
+                    for i in 0..audio_emitter.sinks.len() {
```
Xaeroxe: This is broken and will cause panics.  I was trying to emulate the `retain` functionality here since that doesn't exist on SmallVec.  I've opened a PR to add `retain` to SmallVec [here](https://github.com/servo/rust-smallvec/pull/59)

src/audio/output.rs
```diff
@@ -1,27 +1,35 @@
 //! Provides structures and functions used to get audio outputs.
 
+// We have to use types from this to provide an output iterator type.
+extern crate cpal;
```
torkleyy: Could we have crate declarations in `src/lib.rs`?

src/audio/output.rs
```diff
@@ -0,0 +1,35 @@
+//! Provides structures and functions used to get audio outputs.
+
+// We have to use types from this to provide an output iterator type.
+extern crate cpal;
```
torkleyy: Somehow my comment was marked as outdated... Could you please move this declaration to `src/lib.rs`?

Xaeroxe: I was reluctant to add a dependency on this crate at all, but it was necessary to write the OutputIterator type, as soon as `impl Trait` is stabilized I'd like to remove this dependency, so I chose to limit the scope of it.  Do you feel my decision making process here was sound?

src/audio/output.rs
```diff
@@ -0,0 +1,35 @@
...
+
+impl Output {
+    /// Gets the name of the output
+    pub fn get_name(&self) -> String {
+        self.endpoint.get_name()
+    }
+}
+
+/// An iterator over outputs
+pub type OutputIterator = Map<EndpointsIterator, fn(Endpoint) -> Output>;
```
torkleyy: Could you just create a struct which owns `EndpointIterator` and map it in the `next` method? Would probably be more efficient.

src/audio/output.rs
torkleyy: Yes, definitely.

src/audio/output.rs
Xaeroxe: Sure.

src/audio/play.rs
```diff
@@ -0,0 +1,26 @@
...
+use std::io::Cursor;
+
+use super::source::Source;
+use super::output::Output;
+use super::DecoderError;
+
+/// Play a sound once.
+///
+/// This will return an Error if the loaded audio file in source could not be decoded.
+pub fn once(source: &Source, endpoint: &Output) -> Result<(), DecoderError> {
```
torkleyy: Calling this function from ECS is a bit ugly, since you got two unwraps above and the default ouptut has to be retrieved every time. Not sure, but I don't think a game should really crash / exit even if such a thing failed. Especially during development, it's enough to see an error in the console IMO.

src/audio/output.rs
```diff
@@ -0,0 +1,43 @@
...
+use rodio::*;
+
+/// A speaker(s) through which audio can be played.
+pub struct Output {
+    pub(crate) endpoint: Endpoint,
+}
+
+impl Output {
+    /// Gets the name of the output
+    pub fn get_name(&self) -> String {
```
torkleyy: Sorry about me nitpicking, but getters shouldn't have a `get_` unless necessary.

examples/04_pong/main.rs
```diff
@@ -202,18 +210,21 @@ impl<'a> System<'a> for PongSystem {
...
                 ball.position[1] = top_bound - ball.size / 2.;
                 ball.velocity[1] = -ball.velocity[1];
+                play::once(&self.bounce_sfx, &output::default_output().unwrap()).unwrap();
             }
 
             // Check if the ball is above the bottom boundary, if it is not deflect it
             if ball.position[1] - ball.size / 2. < bottom_bound {
                 ball.position[1] = bottom_bound + ball.size / 2.;
                 ball.velocity[1] = -ball.velocity[1];
+                play::once(&self.bounce_sfx, &output::default_output().unwrap()).unwrap();
```
torkleyy: Do we really need this 6 times? Probably we can improve that once we got proper collision detection added.

src/ecs/components/audio/audio_emitter.rs
```diff
@@ -0,0 +1,39 @@
...
+        self.sound_queue.push(
+            Decoder::new(Cursor::new(source.clone()))
+                .map_err(|_| DecoderError)?,
+        );
+        Ok(())
+    }
+}
+
+impl Component for AudioEmitter {
+    type Storage = VecStorage<Self>;
```
torkleyy: I would use something like `BTreeMap` for this component.

src/ecs/components/audio/audio_listener.rs
```diff
@@ -0,0 +1,16 @@
+use audio::output::Output;
+use ecs::{Component, HashMapStorage};
+
+/// An audio listener, add this component to the local player character.
+pub struct AudioListener {
```
torkleyy: `Debug` implementation would probably be nice here.

src/ecs/systems/audio.rs
```diff
@@ -0,0 +1,122 @@
...
+        self.input.total_duration()
+    }
+}
+
+impl<'a> System<'a> for AudioSystem {
+    type SystemData = (ReadStorage<'a, TransformComponent>,
+     ReadStorage<'a, AudioListener>,
+     WriteStorage<'a, AudioEmitter>);
+
+    fn run(&mut self, (transform, listener, mut audio_emitter): Self::SystemData) {
```
torkleyy: Looks amazing, especially because it uses `Transform` now! Have to look at this code later.

src/audio/mod.rs
```diff
@@ -0,0 +1,13 @@
...
+pub use self::source::Source;
+
+/// An error occurred while decoding the source.
+#[derive(Debug)]
+pub struct DecoderError;
```
torkleyy: This should probably implement `Error`.

examples/04_pong/main.rs
Xaeroxe: Yeah I was just working in the existing framework of the pong example :P I'm happy to refactor this to use a physics system as soon as we have it.

src/audio/mod.rs
Xaeroxe: It does now!

src/audio/play.rs
Xaeroxe: I fixed how the pong example was handling this, can you take a look at it and tell me what you think?

torkleyy:
I think it is better than before, but the function should probably part of the library and I'm not sure if we should get the default output every frame. Do you know how expensive that is?

Xaeroxe:
@torkleyy I'll address your other comments later today, but I did want to say I made the Debug implementation for Output behave a little more like a Debug implementation.  `.debug_struct` requires the fields to be `Debug` which the underlying `cpal::Endpoint` is not.  So this is the closest I can do.

Xaeroxe:
Documentation generated on this branch: https://xaeroxe.github.io/amethyst/doc/amethyst/index.html

torkleyy:
It's really hard to review this given you merged all the commits of `renderer` into this branch.

Xaeroxe:
I know :/ I wasn't expecting anyone to review this until renderer was merged.  Which by the way while I have your attention, is there any reason for us to not merge renderer yet?  Everything seems to be up to speed on that branch after @omni-viral 's PR is merged.

torkleyy:
Well, let's do this :)

src/ecs/systems/mod.rs
```diff
@@ -1,19 +1,23 @@
...
+pub use self::audio::AudioSystem;
+pub use self::dj::DjSystem;
 pub use self::rendering::RenderSystem;
 pub use self::transform::TransformSystem;
 
 // use config::Config;
 use error::Result;
 use ecs::{System, World};
 
-mod rendering;
```
torkleyy: Actually, the standard seems to be to put `use` before `mod`, although I did it the other way around before I read about it.

torkleyy:
Oh and could you please try to not create any merge commits in PRs? It is pretty confusing when you're looking at the commit log.

Xaeroxe:
The two commit histories are so intertwined at this point it's basically impossible to rebase, hence the merge commits.

torkleyy:
Guess we could also just sqash this into three commits: low level, ecs stuff and audio.

Xaeroxe:
I am trying and failing.  These two could probably be made to work together without merges but it would involve a lot of manual labor during which I'd probably make a few mistakes and break something.

torkleyy:
I can do it if you want. Shouldn't be hard thanks to `git`.

Xaeroxe:
By all means go ahead

Xaeroxe:
Ready to merge as far as I'm concerned.  Thanks @torkleyy for your help straightening out the commit history.

PR #267, opened 2017-07-31
Move amethyst_renderer crate to amethyst_renderer directory.

ahmedcharles:


ebkalderon:
@ahmedcharles Thank you so much for the PR! However, this functionality will already be provided in #233, so merging this might be a little redundant.

Issue #266, opened 2017-07-28
Asset hot reloading


torkleyy:
I've figured most parts out already, however I'm stuck at the last one. After a file has changed, it's been read, parsed and turned into an asset, how should we replace the old assets?

Most likely assets live as components in the world, but still, how should we track them? Or is limiting it to the `World` a good idea at all?

I'm not sure whether or not we need this feature before 0.5 release, probably that needs discussion, too.

Xaeroxe:
In order to prevent memory corruption issues we'll have to pause execution while the replacement occurs.  My first reaction is to put assets in an Arc<RwLock<T>> so we can track them and still hand out references to them.  We could keep some kind of Boolean "dirty" bit representing that assets are ready to be replaced, and check that every frame.  If they are ready to be replaced then pop the new assets from a queue and replace each one until the queue is empty.

omni-viral:
Global pauses are unnecessary. There data structures which permit update of value while it is read.
We should use one that works for one writer and many readers.

Xaeroxe:
@omni-viral I don't know of any such data structures.  Can you provide one for me to read about?

omni-viral:
@Xaeroxe I'll find link  to the paper to post it here.
Our case has quiescent points. So we can search in that direction.

omni-viral:
Maybe there is no need of fancy data structures either if we can update assets between systems run

omni-viral:
Maybe there is no need in fancy data structures if we can update assets between systems run

torkleyy:
Still, how would we trace the assets?

Xaeroxe:
We would need some kind of central `AssetManager` structure that keeps a `HashMap<String, Arc<RwLock<T>>>` so it can be referred to when the time to replace the asset comes.

Xaeroxe:
(Said structure would also require a way to remove references so that the memory the assets consume can be freed)

omni-viral:
Put all updates in a queue. Replace the queue atomically in maintenance phase. Flush updates into the world.

omni-viral:
@Xaeroxe no locks please)

Xaeroxe:
Then the update will have to occur in unsafe code, which is fine but we'll have to be much more careful about memory corruption.

omni-viral:
Well. We should wrap unsafe code into nice and safe type and test it heavily

torkleyy:
Not sure if the result is something we want, but what you're asking for
sounds like shred's `TrustCell`.

[...]

PR #272, opened 2017-08-07
Asset Hot Reloading [1/2]

torkleyy:
This is the first of two PRs for Asset Hot Reloading. It implements the low level logic for it; high-level concepts and examples are going to be in PR number two.

I'm going to assign @Xaeroxe for this since you took part in the discussion ([#266](https://github.com/amethyst/amethyst/issues/266)). Feel free to assign somebody else in case you don't have time.

### DIFF
--- a/amethyst_assets/src/asset.rs
+++ b/amethyst_assets/src/asset.rs
@@ -32,6 +32,17 @@ where
     /// The error that may be returned from `Self::Context::from_data`.
     type Error: Error;
 
+    /// Pushes an update to a queue.
+    /// The updated version will be applied by calling
+    /// `update`.
+    ///
+    /// **Note:** The updated version only gets pushed on a single
+    /// asset.
+    fn push_update(&self, updated: Self);
+
+    /// Applies a previously pushed update.
+    fn update(&mut self);
+
     /// Returns `true` if another asset points to the same
     /// internal data.
     ///
@@ -94,6 +105,19 @@ where
         self.map.read().get(spec).map(Clone::clone)
     }
 
+    /// Accesses a cached asset, locking the internal `RwLock` to get read access to the hash map.
+    /// In case the asset exists, `f` gets called with a reference to the cached asset and this
+    /// method returns `true`.
+    pub fn access<F: FnOnce(&T)>(&self, spec: &AssetSpec, f: F) -> bool {
+        if let Some(a) = self.map.read().get(spec) {
+            f(a);
+
+            true
+        } else {
+            false
+        }
+    }
+
     /// Deletes all cached values, except the ones `f` returned `true` for.
     /// May be used when you're about to clear unused assets (see `Asset::clear`).
     ///
@@ -124,10 +148,10 @@ pub trait Context {
     type Asset: Asset;
     /// The `Data` type the asset can be created from.
     type Data;
-    /// The error that may be returned from `from_data`.
+    /// The error that may be returned from `create_asset`.
     type Error: Error;
 
-    /// A small keyword for which category this asset belongs to.
+    /// A small keyword for which category these assets belongs to.
     ///
     /// ## Examples
     ///
@@ -151,6 +175,13 @@ pub trait Context {
     fn retrieve(&self, _spec: &AssetSpec) -> Option<Self::Asset> {
         None
     }
+
+    /// Updates an asset after it's been reloaded.
+    ///
+    /// This usually just puts the new asset into a queue;
+    /// the actual update happens by calling `update` on the
+    /// asset.
+    fn update(&self, spec: &AssetSpec, asset: Self::Asset);
 
     /// Gives a hint that several assets may have been released recently.
     ///
@@ -187,3 +218,4 @@ where
     /// Reads the given bytes and produces asset data.
     fn parse(&self, bytes: Vec<u8>) -> Result<Self::Data, Self::Error>;
 }
+
--- a/amethyst_assets/src/simple.rs
+++ b/amethyst_assets/src/simple.rs
@@ -20,6 +20,12 @@ use {Asset, AssetSpec, Cache, Context};
 /// `Asset` by calling these methods. Methods for the asset
 /// can then be implemented on that wrapper struct by getting
 /// the inner asset with `inner` and `inner_mut`.
+///
+/// The type parameter `A` is the type of the asset handle
+/// (examples: texture handle, shader id, ..). To avoid unnecessarily
+/// duplicated buffer allocations, make sure your handle is reference-counted,
+/// so wrap it with an `Arc` in case the handle doesn't have this functionality
+/// itself.
 #[derive(Clone)]
 pub struct AssetPtr<A> {
     inner: A,
--- a/amethyst_assets/src/simple.rs
+++ b/amethyst_assets/src/simple.rs
@@ -4,8 +4,79 @@
 use std::borrow::Cow;
 use std::error::Error;
 use std::marker::PhantomData;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+use parking_lot::RwLock;
 
 use {Asset, AssetSpec, Cache, Context};
+
+/// An `AssetPtr` which provides `push_update`, `update`
+/// and `is_shared` methods. These can simply be called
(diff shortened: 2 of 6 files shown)


amethyst_assets/src/asset.rs
```diff
@@ -85,6 +96,15 @@ where
         self.map.read().get(spec).map(Clone::clone)
     }
 
+    /// Retrieves an asset, locking the internal `RwLock` to get read access to the hash map.
+    /// In case this asset has been inserted previously, it will be cloned and returned.
+    /// Otherwise, you'll receive `None`.
```
Xaeroxe: Doc comment seems inaccurate as this function returns `()`

amethyst_assets/src/asset.rs
```diff
@@ -85,6 +96,14 @@ where
         self.map.read().get(spec).map(Clone::clone)
     }
 
+    /// Accesses a cached asset, locking the internal `RwLock` to get read access to the hash map.
+    /// In case the asset exists, `f` gets called with a reference to the cached asset.
+    pub fn access<F: FnOnce(&T)>(&self, spec: &AssetSpec, f: F) {
```
Xaeroxe: We probably ought to signal to the user of this function if a cached asset wasn't found.  Can we return a bool from here?  True if operation succeeded, false otherwise.  Sorry I didn't mention that on my first review.

omni-viral:
I have concerns about calling `AssetPtr::update`. 
Who shall call it?

torkleyy:
@omni-viral Most likely the `Context`, on `Context::update`.

amethyst_assets/src/simple.rs
```diff
@@ -4,9 +4,80 @@
...
+///
+/// The recommended strategy is to create a struct for an
+/// asset which simply wraps `AssetPtr` and implements
+/// `Asset` by calling these methods. Methods for the asset
+/// can then be implemented on that wrapper struct by getting
+/// the inner asset with `inner` and `inner_mut`.
+#[derive(Clone)]
+pub struct AssetPtr<A> {
+    inner: A,
+    update: Arc<(AtomicUsize, RwLock<Option<A>>)>,
```
Xaeroxe: Yeah, because otherwise this isn't a pointer.  Right?

omni-viral: I was under impression that asset (`A` here) should be shared itself via `Arc`. Thus I don't understand why update slot shared separately. Where am I wrong?

torkleyy: Yeah, `A` is most likely `Arc<SoundBuffer>`. However, some handles are automatically wrapped in an `Arc` or somebody may want to use another pointer (like `froggy::Pointer`), so I left that choice to the user.

amethyst_assets/src/simple.rs
```diff
@@ -4,9 +4,80 @@
...
+    where A: Clone
+{
+    /// Pushes an update to the shared update container;
+    /// this update can then be applied to all asset pointers by calling
+    /// `update` on them.
+    pub fn push_update(&self, updated: Self) {
+        let &(ref count, ref lock) = &*self.update;
+
+        *lock.write() = Some(updated.inner);
+        count.fetch_add(1, Ordering::SeqCst);
```
omni-viral: `Ordering::SeqCst` is too heavy. Asset data `A` is wrapped in `RwLock` so it is already properly synchronized and there is no need in memory fences. Thus I suggest to use `Ordering::Relaxed` here and in `update` below.
`RwLock` looks as overkill too. But I'm not ready to provide alternative yet.

amethyst_assets/src/simple.rs
```diff
@@ -4,9 +4,80 @@
...
+    pub fn new(data: A) -> Self {
+        AssetPtr {
+            inner: data,
+            update: Arc::new((AtomicUsize::new(0), RwLock::new(None))),
+            version: 0,
+        }
+    }
+
+    /// Borrows the inner asset.
+    pub fn inner(&self) -> &A {
```
omni-viral: I would like it to implement `Deref` and `DerefMut` insted of special functions.

amethyst_assets/src/simple.rs
omni-viral: Now I get it. Each `Asset` implementer puts `AssetPtr` with real data within itself and then implement `push_update` and `update` functions of `Asset` by calling functions from inner `AssetPtr`. Right?

amethyst_assets/src/simple.rs
torkleyy: Can do that, given it's a smart pointer in some way. But in that case we cannot have any methods on it...

amethyst_assets/src/simple.rs
torkleyy: `RwLock` is not overkill here, it's necessary. Will check how strict ordering has to be, thanks.

amethyst_assets/src/simple.rs
omni-viral: Why can't? `Rc` implements `Deref` and has methods by itself.

torkleyy: Please look it up, it doesn't.

amethyst_assets/src/simple.rs
```diff
@@ -4,9 +4,80 @@
...
 
+/// An `AssetPtr` which provides `push_update`, `update`
+/// and `is_shared` methods. These can simply be called
+/// in order to implement the `Asset` trait.
+///
+/// The recommended strategy is to create a struct for an
+/// asset which simply wraps `AssetPtr` and implements
+/// `Asset` by calling these methods. Methods for the asset
+/// can then be implemented on that wrapper struct by getting
+/// the inner asset with `inner` and `inner_mut`.
```
Xaeroxe: Since A is recommended to be some kind of smart pointer can we include that in the docs?  Otherwise I could see people who don't know any better accidentally storing a copy of their asset multiple times in memory.  Something to the effect of "A will be cloned a lot, so keep it small.  Arc<T> would be a good fit here."

Xaeroxe: 👍 Thanks @torkleyy for making that change!

amethyst_assets/src/simple.rs
torkleyy: If we use `Relaxed` here, the asset could perform unnecessary updates. Why do you think it should be `Relaxed`? I see that it doesn't have to be sequentially consistent, but relaxed isn't strict enough, is it?

Xaeroxe: I think ordering is somewhat of a moot point here as there are two operations being performed in a single thread.  I'm with @torkleyy on this one, data should always be available before we signal the data is available.

Xaeroxe: (Not to mention that since we're working with an IO driven piece of code RAM manipulation is the least of our performance concerns.)

omni-viral: In any case `Release` in `push_update` and `Acquire` in `update` would be enough.

> RAM manipulation is the least of our performance concerns.)
At the side of IO code - absolutely agree.
At the side of logic when no updates occurs ever paying for `SeqCst` for each asset each frame - I'm concerned.

torkleyy: Exactly, fully agreed.

amethyst_assets/src/simple.rs
omni-viral: Ok. I see. It doesn't have methods. Not anymore at least. But I don't know why. What is the problem with methods and `Deref`?

omni-viral: Also. Look at `Vec<T>`. It implements `Deref` and has methods. I've just checked 😄

torkleyy: They don't deref to `T`. The reason smart poimters shouldn't have methods is ambigiuity. I think a getter is just fine, a bit explicitness makes the user code easier to understand.

amethyst_assets/src/simple.rs
Xaeroxe: @torkleyy If you make the changes to `Release` and `Acquire` then I consider this ready to merge.

Xaeroxe:
CI passes on stable and beta, merging.

PR #273, opened 2017-08-07
A little tweak in amethyst_assets public interface

omni-viral:
Make `Asset` type appear in public methods parameters instead of its `Context`

### DIFF
--- a/amethyst_assets/src/asset.rs
+++ b/amethyst_assets/src/asset.rs
@@ -23,6 +23,15 @@ pub trait Asset
 where
     Self: Clone + Sized,
 {
+    /// The `Context` type that can produce this asset
+    type Context: Context<Asset=Self, Data=Self::Data, Error=Self::Error>;
+
+    /// The `Data` type the asset can be created from.
+    type Data;
+
+    /// The error that may be returned from `Self::Context::from_data`.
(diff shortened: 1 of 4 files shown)


amethyst_assets/src/asset.rs
```diff
@@ -23,6 +23,15 @@ pub trait Asset
 where
     Self: Clone + Sized,
 {
+    /// The `Context` type that can produce this asset
+    type Context: Context<Asset=Self, Data=Self::Data, Error=Self::Error>;
+
+    /// The `Data` type the asset can be created from.
+    type Data;
```
torkleyy: If we want these types here, could we remove them from `Context`?

omni-viral: They are just like aliases to write `A::Error` instead of `<A::Context as Context::Error`.

omni-viral: Same as `Item` in `std::IntoIterator`

torkleyy: Okay, fair.

torkleyy:
This is definitely an improvement and makes type inference better. Seems there are no other concerns, so I'm going to merge this PR soon.

PR #277, opened 2017-08-12
Add default DisplayConfig

Xaeroxe:
Adds an improved implementation of Default for DisplayConfig.

### DIFF
--- a/src/gfx_device/display_config.rs
+++ b/src/gfx_device/display_config.rs
@@ -3,9 +3,10 @@
 /// Graphical display configuration.
 ///
 /// These are fed in when calling `video_init()`.
-#[derive(Debug, Default, Deserialize, Serialize)]
+#[derive(Debug, Deserialize, Serialize)]
 pub struct DisplayConfig {
     /// Name of the application window.
+    #[serde(default = "default_title")]
     pub title: String,
     /// Enables or disables fullscreen mode.
     #[serde(default)]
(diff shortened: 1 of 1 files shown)


src/gfx_device/display_config.rs
```diff
@@ -20,14 +21,31 @@ pub struct DisplayConfig {
...
+            vsync: default_vsync(),
+            multisampling: default_multisampling(),
+            visibility: default_visibility(),
+        }
+    }
+}
+fn default_title() -> String { "Amethyst game".to_string() }
+fn default_vsync() -> bool { true }
+fn default_multisampling() -> u16 { 1 }
+fn default_visibility() -> bool { true }
```
torkleyy: Would prefer a newline at the end

Xaeroxe: Not sure that was worth a review comment but sure there ya go.

Xaeroxe:
@Aceeri if you approve of this please merge.

Aceeri:
I still don't really like this loading scheme, I prefer the old macro because it requires a lot less boilerplate to express this. Is there really that much benefit to just simplifying this to a `#[derive(Serialize, Deserialize)]`?

Xaeroxe:
@Aceeri Having a Default implementation seems to be an improvement over not having one right?  I do agree that there is a lot of boilerplate here. I was considering writing my own macro to help with this, probably similar to the one you wrote but for ron instead.  I haven't reviewed your old macro code personally, how easy do you think it would be to adapt to this situation?  My other concern with introducing a macro is those tend to be a bit more difficult to identify why they aren't compiling (if they won't compile) so in a way the simpler approach might be worth it as it could make problems easier to identify, but then there's also tools like `cargo-expand` that make debugging macros much easier.  Basically I have no strong feelings one way or another, I just want default to behave in a reasonable way.

torkleyy:
@Aceeri Using serde attributes directly allows for more control. Another point is that not everything has a reasonable default; some configs may not be able to provide a fallback. I figured that it would just be easier to read this code instead of a macro. We can readd it if you consider it a requirement, but I tend to avoid macros on the user side.

Aceeri:
@torkleyy I can't really think of anything without a reasonable default for configurations. If there is a situation where there is no simple one, there is always the option of using `Option<T>` and panicking when it is `None`. I'd rather have a usage that fits the most common case (which is I'm guessing not panicking and not really caring if the config can't load in a field or not).

PR #279, opened 2017-08-15
Improve axis doc comment

Xaeroxe:
cc @Binero


torkleyy:
Could we also make it future proof by changing it to an enum and adding `__NonExhaustive`?

Xaeroxe:
@torkleyy sure but that change will have to wait until tonight (Approx. 11 hours from now.)

Xaeroxe:
Audio is taking a lot of my time, so I won't be able to get to making these input enhancements until later.  However this PR makes improvements to the docs that are needed.  So I'm going to go ahead and merge it.

PR #281, opened 2017-08-15
Some fixes for renderer

torkleyy:


src/app.rs
```diff
@@ -143,12 +137,23 @@ impl<'a, 'b, T: State + 'a> ApplicationBuilder<'a, 'b, T> {
     /// Creates a new ApplicationBuilder with the given initial game state and
     /// display configuration.
     pub fn new(initial_state: T) -> Self {
+        use ecs::components::*;
+        use prelude::*;
+
+        let mut world = World::new();
+
+        world.register::<Transform>();
```
torkleyy: Hold on..this doesn't seem right, I'll have to remove it.

omni-viral: Yeap. Shouldn't be here. Good eye.

PR #283, opened 2017-08-15
Fix bug regarding chunks

torkleyy:


amethyst_renderer/src/scene.rs
```diff
@@ -66,7 +68,7 @@ impl Scene {
     /// Iterates through all stored lights in parallel in chunks.
     pub fn par_chunks_lights(&self, count: usize) -> LightsChunks {
         let size = self.lights.len();
-        self.lights.par_chunks(((size - 1) / count) + 1)
+        self.lights.par_chunks(max(size / count, 1))
```
omni-viral: `((size - 1) / count) + 1` was intended in order to ensure that count of chunks would not exceed `count`
Not it will if `size % count != 0`

Issue #190, opened 2017-02-15
Modularization of Amethyst


lschmierer:
Tackling Issue #34 is long overdue in my opinion.
This issue is more or less the first step of **Part 1: Features** of the RFC #34.

Before adding more features (aka "shards") to the engine, current functionality should be modularized as much as possible.
Features can be enabled/disabled using *Cargos features*.

## Spotted Features (aka Shards)
Feature                                                    | depends on            | mandatory
------------------------------------------ | ------------------- | ------------
engine (core)                                            |                               |
config (project #152)                               | engine                    | yes
renderer (amethyst_renderer, gfx_device)  | engine, config  | no, enabled by default
transform system                                     | engine,             | no, enabled by default
asset manager                                          | engine                   | no, enabled by default

A little cleanup of the code base is probably required. See #191.

Feedback welcomed!

ebkalderon:
@lschmierer I believe you might have misunderstood the point of shards. `ecs` should not be considered a shard, it is a required feature of the engine. I have modified your table so it adheres more closely to the original shards concept:

Shard Name     | Depends on      | Crates needed                                     | Required? | Enabled by default?
---------------|-----------------|---------------------------------------------------|-----------|--------------------
engine         |                 | amethyst                                          | Yes       | Yes
assets         | engine          | amethyst                                          | No        | Yes
audio          | engine          | amethyst                                          | No        | Yes
physics        | engine          | amethyst                                          | No        | Yes
project (#152) | engine, assets  | amethyst, amethyst_project (?)                    | No        | Yes
rendering      | engine, project | amethyst, amethyst_project (?), amethyst_renderer | No        | Yes

Please note that shards are *not* equivalent to the engine's crate dependency graph! Please do not get them confused. Below is an incomplete crate dependency graph in table form for comparison:

Crate name        | Required dependencies | Optional dependencies
------------------|-----------------------|----------------------------------------
amethyst          | specs                 | amethyst_project (?), amethyst_renderer
amethyst_project  | yaml_rust             |
amethyst_renderer | gfx                   | gfx_device_dx11, gfx_device_gl, etc.

lschmierer:
@ebkalderon Thanks for the clarification, that shards are independent form the crate depndency graph was clear to me.

Your table looks pretty good. We should probably maintain it at some prominent place for reference.

ebkalderon:
Agreed, I'm on it! 👍 I'll put it up on the wiki right now.

agmcleod:
[...]

PR #285, opened 2017-08-15
Renderer rewrite

torkleyy:
This is the PR merging `renderer` into `develop`. It is WIP, but shows the overall progress.

## TODO

* Fix examples
* Check docs, crate structure
* Properly handle configs


derekdreery:
I'm finding this new version of the renderer much easier to understand! :)

Can I suggest taking `Error` and `Result` out of the prelude. If I want to create my own lib on top of amethyst, I probably want to define my own error types, using something like [error-chain]. The types in the prelude conflict with this.

They are still accessible easily at `amthyst::Error` and `amethyst::Result`.

[error-chain]: https://github.com/rust-lang-nursery/error-chain

torkleyy:
Thanks for the suggestion! Makes sense to me, so feel free to create a PR ;)

src/ecs/resources/input/input_handler.rs
```diff
@@ -36,8 +36,8 @@ pub struct InputHandler {
     pressed_mouse_buttons: SmallVec<[MouseButton; 16]>,
     down_mouse_buttons: SmallVec<[MouseButton; 16]>,
     released_mouse_buttons: SmallVec<[MouseButton; 16]>,
-    mouse_position: Option<(i32, i32)>,
-    previous_mouse_position: Option<(i32, i32)>,
+    mouse_position: Option<(f64, f64)>,
```
Xaeroxe: Why do these need to change?  mouse positions on screen are integral values, so I don't see what value a float adds here, much less an f64 as those are more expensive in 32 bit architectures.  Many games are only released in 32 bit mode.  I see in winit these are f64 values now.  I still think we shouldn't use f64 values here, and we should instead cast the values from winit to i32.

src/ecs/resources/input/input_handler.rs
```diff
@@ -260,7 +285,7 @@ impl InputHandler {
     }
 
     /// Returns the value of an axis by the string id, if the id doesn't exist this returns None.
-    pub fn axis_value<T: AsRef<str>>(&self, id: T) -> Option<f32> {
+    pub fn axis_value<T: AsRef<str>>(&self, id: T) -> Option<f64> {
```
Xaeroxe: we don't gain any value from using an f64 here, it's just more empty bits as this function can only return -1, 0, and 1.

omni-viral: No empty bits as additional 32 are used for precision. But yeah, maybe unnecessary precise values

omni-viral: I think the same function will be used for joysticks axes. But yeah, f64 seems unnecessary precise.

src/ecs/resources/input/input_handler.rs
torkleyy: I would be in favor of reverting these changes.

derekdreery:
I'd quite like to help writing documentation to go with this, but I might need some help understanding it myself! Is it ready to start having docs written (what's the situation with promise-driven asset loading for instance), and if so is there some example of a working program using the code. It doesn't have to be well documented or anything, I just need a starting point :).

src/ecs/resources/input/input_handler.rs
Xaeroxe: Alright so I took a look at the winit git history and did some of my own research to make sure I understood why they were using f64 values.  Some gaming mice are much more precise than your average monitor, so being able to represent "Sub-pixel" movement can actually be helpful if the game is for example a first person shooter.  So with that in mind I'm actually okay with keeping these changes, because a videogame oftentimes has bigger problems to worry about than two f64 values if those values are actually serving a purpose.

Xaeroxe: An additional note: Windows doesn't actually support this kind of precision.  Linux and Mac OS do though.

src/ecs/resources/input/input_handler.rs
Xaeroxe: [Xinput documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/microsoft.directx_sdk.reference.xinput_gamepad(v=vs.85).aspx) shows they use a short for axis values, which an f32 is more than capable of adequately representing.

torkleyy:
@derekdreery I really appreciate that you want to help out with the docs! We really need them ;) I would recommend to start writing them as soon as this PR has been merged. Examples should be a basic starting point then.

src/ecs/resources/input/input_handler.rs
OvermindDL1: I definitely vote for a floating point of some format, it is what I used in my engine.

Interestingly I did not know abut mice doing subprecision, apparently my mouse does this, I just noticed it.  ^.^

amethyst_renderer/src/lib.rs
```diff
@@ -282,6 +282,32 @@ impl<'a> RendererBuilder<'a> {
...
+    pub fn with_config(&mut self, config: Config) -> &mut Self {
+        self.config = config;
+        let mut wb = self.winit_builder.clone();
+        wb = wb.with_title(self.config.title.clone())
+            .with_visibility(self.config.visibility);
+
+        if self.config.fullscreen {
+            wb = wb.with_fullscreen(winit::get_primary_monitor());
+        }
+        match self.config.dimensions {
```
stenverbois: These could be made a bit cleaner with `if let` bindings.

Binero:
There is still a few issues with this PR. 

- It exposes a lot of `cgmath`, for instance inside `renderer::Camera`.
- It appears to me that the renderer and the asset manager are tightly linked through `AssetPtr`. It was decided in #190 that both systems are optional, so the asset manager must work without the renderer, and the renderer must work without the asset manager.

torkleyy:
@Binero I don't know when that decision was made, but having a game without rendering doesn't make much sense to me.

Xaeroxe:
[Zork!](https://en.m.wikipedia.org/wiki/Zork) although I must say it seems unlikely a text adventure would have any need for Amethyst. :P

Binero:
@torkleyy Well, the issue is linked in the post. One of the rationales I quite like is that the components need to be reusable and replaceable. If I want to use a custom renderer, that should be possible. If I want to use a custom asset manager, that should be possible as well.

Xaeroxe:
The fundamental problem with that idea is that it's in direct conflict with our desire for a hot reloading feature.  The two have to be coupled to make that work.  The only way I can see around this would be to use cargo features to expose a different API without hot reloading capabilities when there isn't an "assets" feature present.  So if modularity is truly that high of a priority for us we can do that but I feel that should be a bigger discussion with more people involved.  My views personally are that we eventually have to make some architectural decisions for our end users in order to provide useful and usable software.  Too many choices can hamper ease of use.

torkleyy:
The way I see it, renderer and assets crate are separate, thus reusable. The main crate is just the glue, so if you want to e.g. use another renderer, you can still use the other crates, but you need your own glue.

Binero:
One must consider that because they are so tightly coupled at the moment, it is impossible to generate meshes at runtime, as is needed in for example voxel engines or other volumetric engines. 

This is because `MeshComponent` is an `Asset`, which is loaded from `Loader`. I think the following approach would be much more flexible:

1. When you load a `Mesh` from the `Loader`, it gives you an `AssetFuture<T> where T: Mesh` which implements `Component` whenever `T` does.
2. This `AssetFuture<Mesh>` can be attached to an entity.
3. There is an `AssetSystem` which tracks whenever these futures have been completed, and replaced the `AssetFuture<T>` components on these entities with `T`. 

As an added bonus, this system could leave behind a `LoadedAsset<T>` component, as to keep track of which entities use which assets. Hot reloading can then easily be implemented by simply iterating over these entities and adding new `AssetFuture`s. 

This approach has the advantage that when the asset system is not required, as it isn't for instance when you are generating meshes, it doesn't get in the way as it does now, preventing you from implementing your features.

omni-viral:
Wow. Checks are passed :tada:

PR #287, opened 2017-08-16
Add Factory abstraction

torkleyy:
cc @omni-viral


src/ecs/resources/rendering.rs
```diff
@@ -0,0 +1,110 @@
...
+        self.execute(move |f| mb.build(f))
+    }
+
+    /// Execute a closure which takes in the real factory.
+    pub fn execute<F, T, E>(&self, fun: F) -> FactoryFuture<T, E>
+        where F: FnOnce(&mut ::renderer::Factory) -> Result<T, E> + 'static,
+              T: 'static,
+              E: 'static,
+    {
+        let (f, cell) = FactoryFuture::new();
```
torkleyy: Now the optimization here would be to just execute it immediately in case the factory is thread-safe.

src/ecs/resources/rendering.rs
```diff
@@ -0,0 +1,110 @@
...
+    type Item = A;
+    type Error = E;
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        match Arc::try_unwrap(self.inner.take().unwrap()) {
+            Ok(x) => {
+                // As soon as the worker thread finished executing the closure, the second
+                // reference gets dropped and we enter this branch.
+                // Thus, we are the only one with access to the cell.
+                unsafe {
```
omni-viral: Move this unsafe block into `FactoryFutureCell` and make safe consuming method to extract the result.

src/ecs/resources/rendering.rs
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+}
+
+struct FactoryFutureCell<A, E> {
+    value: UnsafeCell<Option<Result<A, E>>>,
+}
+
+impl<A, E> FactoryFutureCell<A, E> {
+    fn set(&self, res: Result<A, E>) {
```
omni-viral: Such method itself is unsafe.
Common technique to have safe `set` method is to make special type that holds `Arc` and does not implement `Sync`, only `Send`

torkleyy: This is just private API and definitely not worth building any abstractions.

src/ecs/resources/rendering.rs
```diff
@@ -0,0 +1,110 @@
...
+            let r = fun(factory);
+            cell.set(r);
+        }));
+
+        f
+    }
+}
+
+/// A mesh which may not have been created yet.
+pub type MeshFuture = FactoryFuture<Mesh, Error>;
```
torkleyy: I'm not sure what you're talking about, will look into it later.

src/ecs/resources/rendering.rs
torkleyy: Tried to, but that doesn't work since `poll` only gives me `&mut`, so I have to use `Option` and `take`. I don't think we should waste our time on a stylistic issue right now.

torkleyy:
I won't be able to finish this PR today. Feel free to work on this in the meantime. I will be able to merge PRs for this branch, in case @omni-viral wants to create one.

src/ecs/resources/rendering.rs
omni-viral: I don't think it stylistic. One should be very careful with unsafe blocks. Putting related unsafe blocks together helps

src/ecs/resources/rendering.rs
```diff
@@ -0,0 +1,81 @@
...
+            Async::Ready(x) => x.map(Async::Ready),
+            Async::NotReady => Ok(Async::NotReady),
+        }
+    }
+}
+
+/// The factory abstraction, which allows to access the real
+/// factory and returns futures.
+pub struct Factory {
+    jobs: MsQueue<Box<FnOnce(&mut ::renderer::Factory)>>,
```
omni-viral: Do `Box<FnOnce(_)>` works already?
If no - use `Box<FnBox()>`

torkleyy: I guess not since it takes `self`, not `self: Box<Self>`; given that we cannot move unsized types, I should change that. Thanks!

torkleyy: It's nightly only, so I'm going to add just a dummy trait for that..

src/ecs/resources/rendering.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub fn execute<F, T, E>(&self, fun: F) -> FactoryFuture<T, E>
+        where F: FnOnce(&mut ::renderer::Factory) -> Result<T, E> + 'static,
+              T: 'static,
+              E: 'static,
+    {
+        let (send, recv) = channel();
+
+        self.jobs.push(Box::new(move |factory| {
+            let r = fun(factory);
+            let _ = send.send(r);
```
omni-viral: `let _ = ` here to ignore returned `Result<(), Result<T, E>>`?

omni-viral:
LGTM
`crossbeam::MsQueue` is a little bit heavy for task but not too much.
Add support in `RenderSystem` and place `Send` markers where needed.

torkleyy:
We definitely want to have this abstraction and given that this is blocking other PRs (like fixing examples), I just merged this for now. Feel free to fix nits if you spot any ;)

PR #288, opened 2017-08-17
Remove one level of indirection

omni-viral:


amethyst_assets/src/loader.rs
```diff
@@ -108,9 +108,9 @@ unsafe impl<A, E> Sync for AssetFutureCell<A, E> {}
 /// `ThreadPool`.
 pub struct Loader {
     contexts: FnvHashMap<TypeId, Box<Any>>,
-    directory: Arc<Box<AnyStore>>,
+    directory: Arc<AnyStore>,
```
torkleyy: Oh... I didn't have in mind that `Arc` allocates.

amethyst_assets/src/store/mod.rs
```diff
@@ -50,24 +50,26 @@ impl<T> AnyStore for T
     }
 }
 
-impl<T> Store for Box<T>
-    where T: AnyStore + ?Sized
+impl<T, S> Store for T
+    where T: ::std::ops::Deref<Target=S>,
```
torkleyy: Could you please add an import?

amethyst_assets/src/store/mod.rs
```diff
@@ -50,24 +50,26 @@ impl<T> AnyStore for T
...
+          S: AnyStore + ?Sized
 {
     type Error = BoxedErr;
 
     fn modified(&self, category: &str, id: &str, ext: &str) -> Result<u64, Self::Error> {
-        T::modified(self, category, id, ext)
+        S::modified(self.deref(), category, id, ext)
```
brendanzab: Why not use `*self` here instead of calling `deref` directly?

omni-viral: I have no preferences here. I change if people want.
`self` would be enough thanks to deref coercion.

amethyst_assets/src/loader.rs
omni-viral: It even doesn't matter if it allocates. What does is implementation of `CoerceUsized`.

PR #289, opened 2017-08-17
Improve ApplicationBuilder

torkleyy:


src/app.rs
```diff
@@ -186,6 +199,35 @@ impl<'a, 'b, T: State + 'a> ApplicationBuilder<'a, 'b, T> {
...
+            forward: [0.0, 0.0, 1.0].into(),
+            right: [1.0, 0.0, 0.0].into(),
+            up: [0.0, 1.0, 0.0].into(),
+        };
+
+        let render_sys = RenderSystem::new(&self.events, pipe)?;
+
+        let this = self.add_resource(cam)
+            .add_resource(Factory::new())
+            .register::<LightComponent>()
```
omni-viral: Those `register` calls are fine while `SystemExt` isn't there.
Good to have them in one place at least.

PR #291, opened 2017-08-18
Make `Context` return some `IntoFuture` instead of `Result`

omni-viral:


amethyst_assets/src/asset.rs
```diff
@@ -1,7 +1,9 @@
 use std::error::Error;
 
 use fnv::FnvHashMap;
+use futures::IntoFuture;
 use parking_lot::RwLock;
+use rayon::ThreadPool;
```
omni-viral: Unused now

amethyst_assets/src/asset.rs
```diff
@@ -19,19 +21,19 @@ use StoreId;
...
+    type Context: Context<Asset=Self, Data=Self::Data, Error=Self::Error, Result=Self::Result>;
 
     /// The `Data` type the asset can be created from.
     type Data;
 
     /// The error that may be returned from `Self::Context::from_data`.
     type Error: Error;
 
+    /// The result type of `Self::Context::create_asset`
+    type Result: IntoFuture<Item=Self, Error=Self::Error>;
```
torkleyy: I'm not sure if we should really keep aliases for the associated types here.

amethyst_assets/src/asset.rs
```diff
@@ -163,16 +169,16 @@ pub trait Context {
     fn category(&self) -> &str;
 
     /// Provides the conversion from the data format to the actual asset.
-    fn create_asset(&self, data: Self::Data) -> Result<Self::Asset, Self::Error>;
+    fn create_asset(&self, data: Self::Data) -> Self::Result;
```
torkleyy: Looks similar to my solution ;)

amethyst_assets/src/asset.rs
```diff
@@ -163,16 +169,16 @@ pub trait Context {
...
     /// Notifies about an asset load. This is can be used to cache the asset.
     /// To return a cached asset, see the `retrieve` function.
-    fn cache(&self, _spec: AssetSpec, _asset: &Self::Asset) {}
+    fn cache(&self, _spec: AssetSpec, _asset: Self::Result) {}
 
     /// Returns `Some` cached value if possible, otherwise `None`.
     ///
     /// For a basic implementation of a cache, please take a look at the `Cache` type.
-    fn retrieve(&self, _spec: &AssetSpec) -> Option<Self::Asset> {
+    fn retrieve(&self, _spec: &AssetSpec) -> Option<Self::Result> {
```
torkleyy: I don't think we want the `cache` and the `retrieve` methods to deal with futures..

amethyst_assets/src/asset.rs
omni-viral: They may be removed. Sometimes they may confuse the compiler...

amethyst_assets/src/loader.rs
```diff
@@ -15,93 +16,114 @@ use asset::AssetSpec;
 use store::AnyStore;
 use {Allocator, Asset, BoxedErr, Context, Directory, Format, AssetError, LoadError, Store};
 
+enum AssetFutureInner<A, F, S>
+    where A: Asset
+{
+    Prepared(Receiver<Result<A::Result, LoadError<A::Error, F, S>>>),
```
torkleyy: Could we just do the same we have in `amethyst::ecs::resources::rendering`?

amethyst_assets/src/loader.rs
```diff
@@ -15,93 +16,114 @@ use asset::AssetSpec;
...
-            }
-        }
+        *inner = AssetFutureInner::Received(res);
+        Ok(Async::NotReady)
     }
 }
 
-struct AssetFutureCell<A, E> {
-    value: UnsafeCell<Option<Result<A, E>>>,
+impl<A, F, S> Component for AssetFuture<A, F, S>
```
torkleyy: This is obsolete, so my PR removes that impl.

amethyst_assets/src/loader.rs
```diff
@@ -314,42 +325,67 @@ pub fn load_asset_future<A, F, N, S>(context: Arc<A::Context>,
...
 /// Like `reload_asset`, but loads the asset on a worker thread and returns
 /// an `AssetFuture` immediately.
-pub fn reload_asset_future<A, F, N, S>
-    (context: Arc<A::Context>,
-     format: F,
-     name: N,
-     storage: S,
-     thread_pool: &ThreadPool)
-     -> AssetFuture<A, AssetError<A::Error, F::Error, S::Error>>
+pub fn reload_asset_future<A, F, N, S>(context: Arc<A::Context>,
```
torkleyy: I would like to get rid of `_future` functions. Asset management should consistently use futures, no need for two flavors everywhere.

amethyst_assets/src/simple.rs
```diff
@@ -86,14 +87,15 @@ impl<A> AssetPtr<A>
 
 /// A simple implementation of the `Context` trait.
 pub struct SimpleContext<A, D, E, T> {
-    cache: Cache<A>,
+    cache: Cache<Result<A, E>>,
```
torkleyy: I don't think we should do that.

amethyst_assets/src/asset.rs
torkleyy: I created a typedef for the `AssetError` then since it will be `<A as Blah>::` like 4 times then..

amethyst_assets/src/simple.rs
omni-viral: How does your solution populates cache?

torkleyy: I'd cheat and used `map` ;) Didn't do the caching yet.

amethyst_assets/src/loader.rs
omni-viral: If you upgrade `Format` and `Store` to use `IntoFuture` than it will be good idea

torkleyy:
So..you still working on this? Or do you want me to finish mine?

torkleyy:
Don't think I'll be able to do it in the next hours :(

PR #295, opened 2017-08-22
Renderer asset loading

Xaeroxe:
cc @omni-viral 

This PR aims to provide formats and context for textures, meshes, and materials.  Work thus far:

- Add ImageData wrapper which an eventual TextureContext will interpret.
- Add Jpeg/Jpg format which emits this ImageData structure
- Add BmpFormat
- Add PngFormat


Cargo.toml
```diff
@@ -33,6 +33,7 @@ futures = "0.1.14"
 gfx = "0.16"
 num_cpus = "1.2"
 genmesh = "0.4"
+image = "0.15"
```
omni-viral: we already have _imagefmt_. Is this one better?
If that's the case then remove _imagefmt_ from dependencies.

src/assets/formats/textures.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+/// Allows loading of Jpeg files.
+pub struct JpegFormat;
+
+impl Format for JpegFormat {
+    type Data = ImageData;
+    type Error = ImageError;
+    type Result = ImageFuture;
```
omni-viral: Simple `Result<ImageData, ImageError>` will fit here perfectly IMHO

src/assets/formats/textures.rs
```diff
@@ -0,0 +1,89 @@
...
+        } else {
+            Ok(Async::NotReady)
+        }
+    }
+}
+
+fn parse_jpeg(bytes: Vec<u8>) -> ImageFuture {
+    let result = Arc::new(Mutex::new(None));
+    let result_clone = result.clone();
+    spawn(move || {
```
omni-viral: Don't use `thread::spawn`. If you need to start long-run operation - use `ThreadPool` instead.

omni-viral:
And thank you for fixing warnings I left. I was forced to finish my work ASAP

Cargo.toml
alteous: N.B. The decoding functions in `image` perform very slowly in debug mode.

src/assets/formats/textures.rs
torkleyy: No, we want to load this asynchronously. That's exactly what `SpawnedFuture` is designed for.

omni-viral: @torkleyy makes sense for formats other than _bmp_ etc. But with `ThreadPool` and not by `thread::spawn`

Cargo.toml
Xaeroxe: I don't really have an opinion towards either crate.  I was mostly just looking for a way to decode images in Rust, image was the first crate I found.  I totally didn't realize that we also had imagefmt which does the same thing but with fewer formats.  @alteous you said that image runs slowly in debug mode, can you elaborate on how you came across that information?  Have you used the image crate before?  Are there any pull requests I could send to the image crate in order to improve this situation?

src/assets/formats/textures.rs
Xaeroxe: Side note: even bmp would benefit from a future, as there's actually a few ways the underlying data can be represented and a conversion might still be necessary.

omni-viral: Most of bmp color formats natively supported by modern gpus. But of cource `ThreadPool ` may be used if conversion is required

src/assets/formats/textures.rs
```diff
@@ -0,0 +1,90 @@
...
+pub use image::ImageError;
+
+/// ImageData provided by formats, can be interpreted as a texture.
+#[derive(Clone, Debug)]
+pub struct ImageData {
+    raw: ImageBuffer<Rgba<u8>, Vec<u8>>,
+}
+
+/// A future which will eventually have an image available.
+pub struct ImageFuture {
```
omni-viral: Use `SpawnedFuture` insted of bicycle :smirk:

Cargo.toml
alteous: Recording CPU cycles with `perf` on the [PBR example of three](https://github.com/kvark/three-rs/blob/master/examples/pbr.rs) suggests that the major bottleneck is image decoding which is achieved using the `image` crate. Most notably, PNG decoding and converting between RGB / RGBA (etc.) is markedly slow in debug mode - about 10 seconds on my machine for the PBR example. In release mode however the load time is negligible.

The runtime performance and compile time impacts on iteration time, so I try to avoid the `image` crate where feasible. Not all parts are slow, for example `save_buffer` is quick and effective in debug mode. I haven't looked into ways of improving the crate - it's much less effort for me to use `stb_image` + `image_buffer` instead.

Xaeroxe:
@torkleyy @omni-viral please review https://github.com/amethyst/amethyst/pull/295/commits/8b637775ecdaa70a488404db2930a778521c3c64 and make sure I'm using this the way it was intended.

amethyst_assets/src/loader.rs
```diff
@@ -248,18 +248,19 @@ fn load_asset_inner<C, F, S>(context: Arc<C>,
     let spec_asset_err = spec.clone();
     let context_clone = context.clone();
     let pool = pool.clone();
+    let pool_clone = pool.clone();
```
torkleyy: Do we really need the clone when we're just passing references?

omni-viral: Yeap. Or closure wouldn't be `'static`

torkleyy:
Merging as requested by @omni-viral

PR #297, opened 2017-08-23
TextureContext and stuff

omni-viral:
What done in this PR:

- [x] Add `TextureComponent` and `TextureContext`
- [x] Implement `Asset` for `TextureContext` and `Context` for `TextureContext`
- [x] Implement `Asset` for `MeshComponent` and `Context` for `MeshContext`
- [x] Make `AssetPtr` easily wrappable
- [x] Implement `SystemExt` for `RenderSystem`
    - [x] Move `Component`s registering and `Resource`s creation from `Application::with_renderer` to `RenderSystem::build`
- [x] Add `Application::with_loader` method to add `Loader` to the `World`
- [x] Add `Application::register_asset<A: Asset + Component>` method to register `Asset` implementation as `Component` in `World` and add `Asset`s associated `Context` to the `Loader`
- [x] Register `AssetFuture<A>` in `World` and add system `specs::common::Merge<AssetFuture<A>>` in `Application::register_asset<A: Asset + Component>` method

cc @Xaeroxe , @torkleyy


amethyst_assets/src/loader.rs
```diff
@@ -86,7 +86,7 @@ impl Loader {
     /// Registers an asset and inserts a context into the map.
     pub fn register<A, C>(&mut self, context: C)
         where A: Asset + 'static,
-              C: Context<Asset=A> + 'static,
+              C: Context<Asset=A> + Send + Sync + 'static,
```
torkleyy: Could we have these bounds directly on the trait?

amethyst_assets/src/loader.rs
```diff
@@ -163,10 +163,11 @@ impl Loader {
     fn context<C>(&self) -> &Arc<C>
         where C: Context + 'static,
     {
-        let context = self.contexts
+        let context: &Box<Any + Send + Sync> = self.contexts
             .get(&TypeId::of::<C::Asset>())
             .expect("Assets need to be registered with `Loader::register`.");
 
+        let context: &Any = context; // Won't compile without this one
```
torkleyy: Can't we just do `context.as_ref().do..`?

amethyst_renderer/src/tex.rs
```diff
@@ -67,7 +68,7 @@ impl<D, T> TextureBuilder<D, T>
     /// Creates a new `TextureBuilder` with the given raw texture data.
     pub fn new(data: D) -> Self {
         use gfx::SHADER_RESOURCE;
-        use gfx::format::SurfaceType;
+        ;
```
Xaeroxe: Looks like ya dropped a spare semicolon here.

src/ecs/resources/rendering.rs
```diff
@@ -98,6 +106,9 @@ impl Factory {
     }
 }
 
+/// A mesh which may not have been created yet.
```
Xaeroxe: Woops, this should probably say texture instead of mesh

amethyst_assets/src/loader.rs
omni-viral: I don't want to overflow trait with bounds.
But if we not going to use it without `Loader` then it's OK I think

amethyst_assets/src/loader.rs
omni-viral: No.
`Any + Send + Sync` doesn't has `downcast_ref` method
`Any` has it, `Any + Send` has it, even `Any + Send + 'static` has it. But not `Any + Send + Sync`
So I have to manually cast it to `&Any`
It can be `(context as &Any).downcast_ref()` or as I done here.

amethyst_renderer/src/tex.rs
omni-viral: In case someone would need a semicolon 😄

amethyst_assets/src/loader.rs
Xaeroxe: For an engine with a high emphasis on multi threading I think it's OK to litter Send + Sync restrictions.  This way we can do more with out context types without worrying about breaking code in the future.

amethyst_assets/src/loader.rs
torkleyy: How does this make any sense?

amethyst_assets/src/loader.rs
```diff
@@ -163,10 +163,11 @@ impl Loader {
     fn context<C>(&self) -> &Arc<C>
         where C: Context + 'static,
     {
-        let context = self.contexts
+        let context: &Box<Any + Send + Sync> = self.contexts
             .get(&TypeId::of::<C::Asset>())
             .expect("Assets need to be registered with `Loader::register`.");
 
+        let context: &Any = context; // `Any + Send + Sync` doesn't has `downcast_ref`
```
torkleyy: I wonder if there is a good explanation for that.. In case we want to keep that comment, it should be "have" instead of "has"

src/ecs/components/rendering/texture.rs
```diff
@@ -0,0 +1,235 @@
...
+impl TextureContext {
+    pub(crate) fn new(factory: Arc<Factory>) -> Self {
+        TextureContext {
+            cache: Cache::new(),
+            factory: factory,
+        }
+    }
+}
+
+impl Context for TextureContext {
```
torkleyy: Why do we need a custom implementation instead of `SimpleContext`?

src/ecs/systems/rendering.rs
```diff
@@ -15,25 +21,26 @@ pub struct RenderSystem {
     #[derivative(Debug = "ignore")]
     renderer: Renderer,
     scene: Scene,
+    #[derivative(Debug = "ignore")]
+    factory: Arc<Factory>,
```
torkleyy: `Factory` should be a resource.

src/ecs/systems/rendering.rs
```diff
@@ -54,19 +61,47 @@ impl<'a> System<'a> for RenderSystem {
     }
 }
 
-impl RenderSystem {
+impl<'a, 'b> SystemExt<'a, (&'b EventsLoop, PipelineBuilder)> for RenderSystem {
```
torkleyy: I don't see how the `SystemExt` trait is useful; one can just make an associated method which has the advantage that we don't need to include a trait, don't have to use a tuple for the real arguments and don't have any restrictions wrt return type, number of flavors,...

amethyst_assets/src/loader.rs
Xaeroxe: In order to deepen the mystery further, I've found this also compiles.

```Rust
    fn context<C>(&self) -> &Arc<C>
        where C: Context + 'static,
    {
        let context: &Box<Any + Send + Sync> = self.contexts
            .get(&TypeId::of::<C::Asset>())
            .expect("Assets need to be registered with `Loader::register`.");

        Any::downcast_ref(context).unwrap()
    }
```

Xaeroxe: I think we might be hitting a bug in rustc here honestly.  It just doesn't like that one incredibly specific syntactic construct.  Perhaps we should file an issue with [rust-lang/rust](https://github.com/rust-lang/rust)

src/ecs/components/rendering/texture.rs
omni-viral: Indeed...

amethyst_assets/src/loader.rs
omni-viral: Sorry. My engilsh is so bad. 😭

src/ecs/components/rendering/texture.rs
omni-viral: It's hard to specify `SimpleContext` type which needs non-empty closure

src/ecs/systems/rendering.rs
omni-viral: `Context` requires to be owner of `Factory`

src/ecs/systems/rendering.rs
omni-viral: `SystemExt` to be used with `Application`.
`Application::with_system` will require `S: for<'a> SystemExt<'a>`, it will accept building arguments for it and build it in place providing `World` to register any relevant `Component`s and put `Resource`s

amethyst_assets/src/loader.rs
omni-viral: @Xaeroxe no mistery. Your variant has explicit cast just as mine

Xaeroxe: I think the explicit cast shouldn't be necessary, rustc has enough information to figure out the intent.  I filed an issue [here](https://github.com/rust-lang/rust/issues/44064), let's see if the rust devs agree with me on this one or not.

Xaeroxe: Either way, we have a way to move forward with this code so we ought to keep it.

src/assets/formats/textures.rs
```diff
@@ -12,7 +12,7 @@ pub use imagefmt::Error as ImageError;
 /// ImageData provided by formats, can be interpreted as a texture.
 #[derive(Clone, Debug)]
 pub struct ImageData {
-    raw: Image<u8>,
+    pub(crate) raw: Image<u8>,
```
torkleyy: Stupid question, why isn't this field public?

amethyst_assets/src/loader.rs
omni-viral: Those are methods of type `Any` and not of the trait.
If they be methods of the trait - rustc would understad what your intent.

src/ecs/systems/rendering.rs
torkleyy: We should just clone the `Factory`. If I didn't do that already, the queue contained in it should be reference counted.

src/assets/formats/textures.rs
Xaeroxe: Because then we'd be exposing a type from imagefmt, which I'm not sure if we want to lock that crate in as part of our public API.

amethyst_assets/src/loader.rs
Xaeroxe: [Any](https://doc.rust-lang.org/beta/std/any/trait.Any.html) is a trait per the documentation.

src/assets/formats/textures.rs
torkleyy: Fair, then let's have getters for bytes and dimensions.

amethyst_assets/src/loader.rs
Xaeroxe: However, I guess the question is "should adding a provided method on a trait be a breaking change for multi-trait trait objects in the event that ambiguity between methods for multiple traits is introduced by the provided method's presence?"  Which I think the answer is no, so yeah it makes sense to prevent method calls in this way, but perhaps the compiler could give a better syntax hint as to what should be done instead.

omni-viral: There is a type for each trait. 
And `downcast_ref` is in `impl Any`, not `trait`

src/ecs/systems/rendering.rs
omni-viral: But who will fetch it?

src/assets/formats/textures.rs
omni-viral: And format. And that's all what `raw` is made of

src/ecs/components/rendering/texture.rs
Xaeroxe: @omni-viral Use a function pointer, closures that don't capture environment variables automatically coerce to function pointers.

I.e. `SimpleContext<TextureComponent, ImageData, TextureFuture, fn(ImageData) -> TextureFuture> `

src/ecs/components/rendering/mesh.rs
```diff
@@ -1,11 +1,133 @@
...
+    fn poll(&mut self) -> Poll<MeshComponent, RendererError> {
+        match self.0.poll() {
+            Ok(Async::Ready(mesh)) => Ok(Async::Ready(MeshComponent::new(mesh))),
+            Ok(Async::NotReady) => Ok(Async::NotReady),
+            Err(err) => Err(err),
+        }
+    }
+}
+
+/// Context to create textures from images
```
Rhuagh: This comment is bad.

Binero:
Given this PR already does so much regarding asset reloading, it might be worth addressing the problem where new asset stores cannot be added to the loader.

The `assets::Directory` store requires an `assets::Allocator` to be initialised. The `assets::Allocator` is created inside `Engine::new`, and used to create the first `assets::Directory`, but is discarded after. As such, the only way to later add a new store to the asset loader, is to also create a new allocator, which means the ids can conflict, which on its end means that `assets::AssetSpec` may no longer be unique.

amethyst_assets/src/simple.rs
```diff
@@ -135,22 +148,30 @@ impl<A> AssetPtr<A>
...
 
 /// A simple implementation of the `Context` trait.
-pub struct SimpleContext<A, D, R, T> {
-    cache: Cache<AssetFuture<AssetPtr<A>>>,
+pub struct SimpleContext<A, D, T> {
+    cache: Cache<AssetFuture<SimpleAsset<A>>>,
     category: Cow<'static, str>,
     load: T,
-    phantom: PhantomData<(D, R)>,
+    phantom: PhantomData<*const D>,
```
torkleyy: Could you explain why you made that change?

omni-viral: `R` can be inferred
And `D` isn't owned by `SimpleContext`. If you have `PhantomData<D>` drop checker may think that while `SimpleContext` gets dropped it may access inner `D`
Changing it to `*const D` says that's not the case.
If my explanation confuse you (сhukchi is not a writer), you may read about it [here](https://doc.rust-lang.org/std/marker/struct.PhantomData.html#ownership-and-the-drop-check)

torkleyy: I know, but `D` is `'static`, so it shouldn't matter?

amethyst_assets/src/merge.rs
```diff
@@ -1,68 +0,0 @@
...
-    }
-}
-
-impl<'a, T, E, F> System<'a> for Merge<F>
-    where T: Component + Send + Sync + 'static,
-          F: Future<Item = T, Error = E> + Component + Send + Sync,
-          E: Error,
-{
-    type SystemData = (Entities<'a>,
-                       // FIXME: Here should be some Fetch<'a, Log> to log occured errors
```
torkleyy: That's exactly why we cannot have generic errors; there has to be one single place in the ECS for all errors, and that's [`specs::common::Errors`](https://github.com/slide-rs/specs/blob/master/src/common.rs#L71). Please note though that raising errors in the ECS should, in general, be avoided.

amethyst_assets/src/loader.rs
```diff
@@ -50,27 +51,78 @@ impl<A, E> Future for SpawnedFuture<A, E> {
...
 /// `ThreadPool`.
 pub struct Loader {
-    contexts: FnvHashMap<TypeId, Box<Any>>,
-    directory: Arc<AnyStore>,
+    contexts: FnvHashMap<TypeId, Box<Any + Send + Sync>>,
+    directory: StoreWithId<Directory>,
     pool: Arc<ThreadPool>,
-    stores: FnvHashMap<String, Arc<AnyStore>>,
+    stores: FnvHashMap<String, StoreWithId>,
+    allocator: Allocator,
```
torkleyy: :+1: Good idea, but let's make it `pub`lic.

amethyst_assets/src/merge.rs
omni-viral: Yeah. I think you're right.

amethyst_assets/src/asset.rs
```diff
@@ -25,7 +27,47 @@ pub trait Asset: Sized {
 }
 
 /// A future for an asset
-pub type AssetFuture<A, E = AssetError> = Shared<Box<Future<Item=A, Error=E>>>;
+pub struct AssetFuture<A>(Shared<Box<Future<Item=A, Error=BoxedErr>>>);
+
+impl<A> Component for AssetFuture<A>
```
torkleyy: Do we need this component implementation?

amethyst_assets/src/asset.rs
```diff
@@ -25,7 +27,47 @@ pub trait Asset: Sized {
...
+    where A: Component, Self: 'static
```
torkleyy: This is true for `A: 'static` which is true for `A: Component`, isn't it?

src/assets/formats/mod.rs
```diff
@@ -1,3 +1,8 @@
 //! Contains formats that amethyst can load as assets.
 
+
+/// Formats to load textures
```
torkleyy: Please put module docs into the modules.

src/ecs/components/rendering/texture.rs
```diff
@@ -0,0 +1,237 @@
...
+        }
+
+        let image = image.raw;
+        let fmt = match convert_color_format(image.fmt) {
+            Some(fmt) => fmt,
+            None => return TextureFuture::unsupported_color_format(image.fmt),
+        };
+
+        if image.w > u16::max_value() as usize || image.h > u16::max_value() as usize {
+            return TextureFuture::unsupported_size(image.w, image.h);
```
torkleyy: No need for `return`

src/ecs/components/rendering/texture.rs
torkleyy: I see using `SimpleContext` isn't perfect right now, but maybe we could just move its methods into `impl SimpleContext`, remove the implementation of `Context` and the `create_asset` method so we can just delegate to `SimpleContext` for everything except `create_asset`.

src/ecs/resources/rendering.rs
```diff
@@ -25,15 +27,16 @@ impl<A, E> Future for FactoryFuture<A, E> {
 
 /// The factory abstraction, which allows to access the real
 /// factory and returns futures.
+#[derive(Clone)]
```
torkleyy: :+1:

src/ecs/systems/rendering.rs
```diff
@@ -55,18 +58,34 @@ impl<'a> System<'a> for RenderSystem {
...
     /// Create new `RenderSystem`
     /// It creates window and do render into it
-    pub fn new(events: &EventsLoop, pipe: PipelineBuilder, config: DisplayConfig) -> Result<Self>
-        where Self: Sized
-    {
-        let mut renderer = Renderer::build(events)
-            .with_config(config)
-            .build()
-            .map_err(|_| Error::System)?;
+    fn build((events, pipe, config): (&'b EventsLoop, PipelineBuilder, Option<DisplayConfig>), world: &mut World) -> Result<Self> {
```
torkleyy: Why take `Option<DisplayConfig>`?

src/ecs/systems/rendering.rs
```diff
@@ -55,18 +58,34 @@ impl<'a> System<'a> for RenderSystem {
...
+            proj: Projection::perspective(1.3, Deg(60.0)).into(),
+            forward: [0.0, 0.0, 1.0].into(),
+            right: [1.0, 0.0, 0.0].into(),
+            up: [0.0, 1.0, 0.0].into(),
+        };
+
+        world.add_resource(Factory::new());
+        world.add_resource(cam);
+        world.register::<LightComponent>();
+        world.register::<MaterialComponent>();
```
torkleyy: Should we register assets that come with the renderer here?

src/app.rs
```diff
@@ -142,14 +142,24 @@ pub struct ApplicationBuilder<'a, 'b, T: State + 'a> {
...
-    pub fn new(initial_state: T) -> Self {
-        ApplicationBuilder {
-            base_path: format!("{}/resources", env!("CARGO_MANIFEST_DIR")).into(),
+    pub fn new(initial_state: T) -> Result<Self> {
+        use num_cpus;
+        use rayon::Configuration;
+
+        let num_cores = num_cpus::get();
+        let cfg = Configuration::new().num_threads(num_cores);
+        let pool = ThreadPool::new(cfg).map(|p| Arc::new(p)).map_err(|_| Error::Application)?;
```
Binero: We shouldn't discard errors like this.

torkleyy:
Could we address these reviews?

amethyst_assets/src/asset.rs
omni-viral: Yes. Or you wont be able to attach it to entity.

PR #298, opened 2017-08-23
Propagating config to renderer/window, and example fixes

Rhuagh:


Xaeroxe:
Definitely an improvement!  I'm going to merge this.

PR #299, opened 2017-08-24
Made the assets example compile.

Binero:
This moves the assets example to the new renderer as well as to the new input handler (although it does not use the new `actions` system yet).

This removes nearly all if not all asset related functionality. Asset loading has not been implemented yet. Most notably, these changes would be required:

 - `assets::Directory` cannot be instantiated as there is no way to get an `assets::Allocator` from the `assets::Loader`.
 - Object/mesh files and texture files are waiting on PR #297

All removed code has been replaced with appropriate 'TODO's.

Additionally, this commit reformatted the example and split it up into various functions, to make it easier to read. The initialisation of the lights and the camera are now fitted in their own function. The `run()` function prevents uses of `.unwrap()` and `.expect()`, and instead returns any errors to `main()` which can then print them out.

### DIFF
--- a/examples/05_assets/main.rs
+++ b/examples/05_assets/main.rs
@@ -1,227 +1,133 @@
-//! Demonstrates several assets-related techniques, including writing a custom
-//! asset loader, and loading assets from various paths.
-//!
-//! TODO: Rewrite for new renderer.
+//! Demostrates loading custom assets using the Amethyst engine.
+// TODO: Add asset loader directory store for the meshes.
 
 extern crate amethyst;
 extern crate cgmath;
 
-use amethyst::{Application, State, Trans};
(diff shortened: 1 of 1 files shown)


examples/05_assets/main.rs
```diff
@@ -1,227 +1,133 @@
...
+
+    camera.forward = [0., 0., -1.0].into();
+    camera.right = [1.0, 0.0, 0.0].into();
+    camera.up = [0., 1.0, 0.].into();
+}
+
+/// Adds lights to the scene.
+fn initialise_lights(world: &mut World) {
+    use amethyst::ecs::components::LightComponent;
+    use amethyst::renderer::light::PointLight;
```
torkleyy: I'm a bit worried about discoverability here.

PR #300, opened 2017-08-24
Adding ambient color propagation from ECS down to renderer effect.

Rhuagh:
Also fixing a bug with using specs Merge system. Need to create Errors resource for Merge to work correctly. And fixing the renderable example as much as possible.

### DIFF
--- a/src/ecs/systems/rendering.rs
+++ b/src/ecs/systems/rendering.rs
@@ -10,6 +10,7 @@ use ecs::resources::AmbientColor;
 use error::{Error, Result};
 use renderer::prelude::*;
 use renderer::Config as DisplayConfig;
+use renderer::Rgba;
 
 use super::SystemExt;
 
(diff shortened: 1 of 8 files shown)


src/ecs/systems/rendering.rs
```diff
@@ -83,10 +90,11 @@ impl<'a, 'b> SystemExt<'a, (&'b EventsLoop, PipelineBuilder, Option<DisplayConfi
 
         world.add_resource(Factory::new());
         world.add_resource(cam);
+        world.add_resource::<Option<AmbientColor>>(None);
```
torkleyy: Couldn't we also just add a default ambient light? If somebody doesn't want any, they can always just set it to `[0.0, 0.0, 0.0]`

Rhuagh: True, I'll fix that later tonight.

Xaeroxe:
@torkleyy and @omni-viral r?  If you both approve we should merge this.

PR #305, opened 2017-08-27
Reimplement examples

omni-viral:
- [x] hello_world
- [x] window
- [x] sphere
- [x] renderable
- [x] pong
- [x] assets


torkleyy:
The renderable example should probably be renamed since there are no
Renderables anymore.

omni-viral:
Assets and Renderable are very similar. I think they should be merged.

PR #306, opened 2017-08-27
Moving ECS modules around for a more functionality driven structure.

Rhuagh:
Also moved input rebinding to it's own crate as it can be used outside amethyst.


amethyst_input/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "amethyst_input"
+version = "0.1.0"
+authors = ["Rhuagh <seamonr@gmail.com>"]
```
Xaeroxe: I hate to be that guy, but given that I authored the majority of the functionality in this crate can you stick me as an additional author here?

Rhuagh: Wow, did I forget that? Soooorrry :~(

torkleyy:
Is there any need to merge this so fast? I planned to review this PR, but can't do that today anymore. I think we really have enough time to allow more reviews.

nchashch:
@torkleyy Sure! So we will merge this after your review.

Xaeroxe:
Examples still work with this commit so we're golden as far as I'm concerned.

PR #312, opened 2017-08-28
rustfmt on renderer crate

Rhuagh:


Xaeroxe:
cc @omni-viral @torkleyy @ebkalderon

amethyst_renderer/src/config.rs
```diff
@@ -45,10 +45,18 @@ impl Default for Config {
         }
     }
 }
-fn default_title() -> String { "Amethyst game".to_string() }
-fn default_vsync() -> bool { true }
-fn default_multisampling() -> u16 { 1 }
-fn default_visibility() -> bool { true }
+fn default_title() -> String {
```
Xaeroxe: Can we use `#[cfg_attr(rustfmt, rustfmt_skip)]` for these default functions?  I like the old formatting better.

Rhuagh: I'm fine with either.

omni-viral: I'd prefer to avoid skipping for consistency

Xaeroxe: I guess I can live with that.

Rhuagh: We can make it default to have single expression functions on a single line if you want.

Xaeroxe: That works!

Rhuagh: Can I fix that as a separate PR after all the current ones are merged? Getting messed up in the head from all the PRs..

Xaeroxe: Sure

PR #315, opened 2017-08-29
Configuring rustfmt to only reorder imports in group

Rhuagh:


amethyst_config/rustfmt.toml
```diff
@@ -1 +1,4 @@
 max_width = 200
+reorder_imports = true
+reorder_import_names = true
+reorder_imports_in_group = true
```
Xaeroxe: Can you also stick these in /rustfmt.toml?  There's no need for amethyst_config to have formatting different from the rest of the repository right?

Rhuagh: They are already, in atleast one of the other PRs.

Rhuagh: I can remove the config specific config file.

Xaeroxe: That would be preferable

Rhuagh:
Ok, config file is moved.

rustfmt.toml
```diff
@@ -1 +1,3 @@
 reorder_imports = true
+reorder_import_names = true
```
Xaeroxe: This was already done by a previous PR, can you remove this commit from the branch?

Rhuagh:
Removed :)

PR #320, opened 2017-08-31
Use multiple extensions for formats

Xaeroxe:
Some formats (such as jpeg) have multiple file extensions that could be considered valid.  This slight ergonomics improvement allows end users to not have to know exactly what extension their file is using if there's any room for ambiguity.  (Only works with Rust 1.20, which was made stable today.)

### DIFF
--- a/amethyst_assets/examples/asset.rs
+++ b/amethyst_assets/examples/asset.rs
@@ -43,13 +43,10 @@ impl Context for DummyContext {
 struct DummyFormat;
 
 impl Format for DummyFormat {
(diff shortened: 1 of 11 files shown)


Xaeroxe:
Yeah it does.  My thoughts are that those costs are likely to be insignificant as most filetypes only have one valid extension, and these changes allow you to put the most common extensions first, followed by the least common ones.

Xaeroxe:
@torkleyy Since you shepherd the assets system I'll let you push the merge button here if you so choose.

PR #322, opened 2017-09-01
[WIP] Rewrite book for 0.5.0

Xaeroxe:
I'm beginning a rewrite of the book for 0.5.0.  This is still work in progress but I'm opening this PR early for initial feedback of the content thus far.


book/src/getting_started.md
```diff
@@ -4,17 +4,32 @@ These instructions will help you set up a working Amethyst game development
...
+i686-pc-windows-msvc
+
+x86_64-pc-windows-gnu
+
+i686-pc-windows-gnu
+
+### Mac OS
+x86_64-apple-darwin
+
+i686-apple-darwin
```
torkleyy: Speaking of which, we should test OS X on Travis.

book/src/getting_started.md
```diff
@@ -25,17 +40,79 @@ Rust compiler. Here are the system requirements (they're pretty modest):
...
 
-There are two ways to get started working with Amethyst:
+## Rust project basics
+
+**If you consider yourself a rust veteran you can probably skip this section,
+just create a binary project and add a dependency for amethyst from crates.io.**
+
+Once you have that setup you'll need to make a new cargo binary project for your
+game to be written in.  You can do this with the following command
+```
```
torkleyy: Some stricter markdown specs ask for an empty line before and after each code block, so I'd recommend doing that.

book/src/getting_started.md
```diff
@@ -25,17 +40,79 @@ Rust compiler. Here are the system requirements (they're pretty modest):
...
+$ cargo init --bin my_cool_game
+```
+Where my_cool_game is the name of your project.
+
+Next you'll want to add a dependency to amethyst.  To do this visit
```
torkleyy: I think we can just use `amethyst = "*"`; that's what we have `Cargo.lock` after all.

book/src/getting_started.md
```diff
@@ -25,17 +40,79 @@ Rust compiler. Here are the system requirements (they're pretty modest):
...
+page.  Now go to the Cargo.toml file inside your project directory and add
+```toml
+[dependencies]
+amethyst = "x.x"
+```
+where x.x are the first two sections of the version number you grabbed from
+amethyst.  This way you can receive non-breaking patches from us with
+`cargo update` but you won't receive breaking changes unless you change your
+version number in Cargo.toml.
+This is because amethyst follows [semver](http://semver.org/spec/v2.0.0.html).
```
torkleyy: Could we have Amethyst written with an uppercase "A"?

book/src/getting_started.md
```diff
@@ -25,17 +40,79 @@ Rust compiler. Here are the system requirements (they're pretty modest):
...
+## Amethyst specific setup
+
+Next you'll need a resources folder that will contain the data for your game.
+You can call this anything, but our convention is just `resources`.  Stick it in
+your project directory next to Cargo.toml.
+
+### config.ron
+
+  The first file you'll want to create in here is `config.ron` this file is
+written in the [RON](https://github.com/ron-rs/ron) format.  It's contents will
```
torkleyy: Should be "Its"

book/src/getting_started.md
```diff
@@ -25,17 +40,79 @@ Rust compiler. Here are the system requirements (they're pretty modest):
...
+
+## Amethyst specific setup
+
+Next you'll need a resources folder that will contain the data for your game.
+You can call this anything, but our convention is just `resources`.  Stick it in
+your project directory next to Cargo.toml.
+
+### config.ron
+
+  The first file you'll want to create in here is `config.ron` this file is
```
torkleyy: We could add here that it's similar to Rust syntax, so it shouldn't be too alien.

book/src/getting_started.md
```diff
@@ -25,17 +40,79 @@ Rust compiler. Here are the system requirements (they're pretty modest):
...
+on startup.
 
-Since we're just getting started, it's fastest and highly recommended to use
-[Amethyst CLI][ac], which is included in the [amethyst_tools][at] crate. If
-you're of the intrepid type, you may go the vanilla Cargo route if you wish.
+`vsync: true,` If this is true then the game will use [vertical synchronization](https://en.wikipedia.org/wiki/Screen_tearing#V-sync).
 
-[ac]: https://github.com/amethyst/tools/tree/master/src/cli
-[at]: https://github.com/amethyst/tools
+This file is used to generate a `DisplayConfig` which we can then use to make a
```
torkleyy: The grammar here seems odd. Maybe "be used to make"?

book/src/getting_started.md
Xaeroxe: Oh, thanks!

book/src/getting_started.md
Xaeroxe: If you use that then I believe `cargo update` will retrieve breaking changes as well as patches.

book/src/getting_started.md
Xaeroxe: Sure

book/src/getting_started.md
Xaeroxe: Oh sure yeah :)

book/src/getting_started.md
Xaeroxe: Oh, thanks I always mess that one up.

book/src/getting_started.md
Xaeroxe: Yeah that's much better.  Thanks!

book/src/getting_started.md
torkleyy: I guess that's the intended behavior for binaries.

Xaeroxe: > I guess that's the intended behavior for binaries.
When you say "that" are you referring to "break when you run cargo update" or "don't break when you run cargo update"?

torkleyy: The former; for binaries you'd just update specific packages after you checked if the new version is any use to you.

book/book.toml
```diff
@@ -0,0 +1,3 @@
+title = "Amethyst Documentation"
+author = "Amethyst Developers"
+dest = "html"
```
ebkalderon: If the destination is set to `html` here, why is `.gitignore` also being changed to ignore `book/book` instead of `book/html`?

book/src/getting_started.md
ebkalderon: Agreed. Let's prepare a separate PR for that soon.

book/src/getting_started.md
ebkalderon: Should probably be changed to:
```
Next, we need to pull in the `amethyst ` crate as a dependency.
```

book/src/getting_started.md
```diff
@@ -25,17 +40,79 @@ Rust compiler. Here are the system requirements (they're pretty modest):
...
+```
+Where my_cool_game is the name of your project.
+
+Next you'll want to add a dependency to amethyst.  To do this visit
+[https://crates.io/crates/amethyst] and grab the latest version number from that
+page.  Now go to the Cargo.toml file inside your project directory and add
+```toml
+[dependencies]
+amethyst = "x.x"
+```
```
ebkalderon: Missing newline after code block.

book/src/getting_started.md
```diff
@@ -25,17 +40,79 @@ Rust compiler. Here are the system requirements (they're pretty modest):
...
+game to be written in.  You can do this with the following command
+```
+$ cargo init --bin my_cool_game
+```
+Where my_cool_game is the name of your project.
+
+Next you'll want to add a dependency to amethyst.  To do this visit
+[https://crates.io/crates/amethyst] and grab the latest version number from that
+page.  Now go to the Cargo.toml file inside your project directory and add
+```toml
```
ebkalderon: Missing newline before code block.

Xaeroxe:
So here's my priority list:

- UI
- 2D rendering
- The book

There's a really good chance 0.6 is going to be released before I get to the book for 0.5.  I also don't think this is necessarily a terrible loss, since as @torkleyy pointed out we probably won't be able to use this for any games until the 0.6 release.  Without the UI Amethyst isn't even fit for a Ludum Dare game.  So I'm going to close this.

I promise the book will be my absolute first priority when 0.6 is released.

PR #324, opened 2017-09-01
Fix several warnings and skip doc test

torkleyy:


Cargo.toml
```diff
@@ -29,16 +29,13 @@ amethyst_input = { path = "amethyst_input", version = "0.1" }
...
-dds-rs = "0.4"
 derivative = "1.0"
 fnv = "1.0"
 futures = "0.1.14"
 gfx = "0.16"
 num_cpus = "1.2"
 genmesh = "0.4"
 imagefmt = "4.0"
-serde = "1"
-serde_derive = "1"
```
Xaeroxe: Pretty sure this breaks things. This crate uses serde to create ron files.

amethyst_input/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-extern crate serde;
+//extern crate serde;
```
Xaeroxe: Same as above.

amethyst_renderer/src/lib.rs
```diff
@@ -75,7 +75,7 @@ extern crate gfx_core;
 extern crate gfx_macros;
 extern crate num_cpus;
 extern crate rayon;
-extern crate serde;
+//extern crate serde;
```
Xaeroxe: Why just comment this out? If we're going to remove it we should remove this line.

Cargo.toml
Xaeroxe: ... Apparently Travis thinks it's Ok?  I guess I need to look into this further but how can you implement Serialize without serde?

Xaeroxe: Oh woops my bad, I understand now.

PR #325, opened 2017-09-01
Improve Appveyor behaviour and add osx to Travis build

torkleyy:


torkleyy:
^ That would be nice if bors could do that for use ;)

PR #326, opened 2017-09-01
Release new versions **NON-BREAKING**

Xaeroxe:
@torkleyy Fixed many warnings for our crates in https://github.com/amethyst/amethyst/pull/324 so I propose we send new versions to cargo.


amethyst_assets/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "amethyst_assets"
-version = "0.1.0"
+version = "0.1.1"
```
torkleyy: Let's also use this version when specifying `amethyst_assets` as dependency in our core crate.

torkleyy:
I don't think this is worth a release, but I'm sure we can release a new version soon, so let's leave it open.

Xaeroxe:
Since we're delaying merging this I just want to make it clear to any future readers this only increments the patch versions, not the minor versions.

Xaeroxe:
There haven't really been any interesting non breaking changes made since 0.5.1 was released, and all work from this point forward will be breaking it seems, as I doubt anyone is interested in rebasing onto 0.5.  so we can either release 0.5.2 now or just never put one out at all.

torkleyy:
I don't think we need a 0.5.2 unless something is horribly broken there.

PR #328, opened 2017-09-01
Fix texture coordinates in pong example

torkleyy:
cc @jojolepro


Xaeroxe:
Examples are excluded from releases so it's probably not necessary to merge this into release-0.5.  So I changed the target to develop.  If you agree with this decision please merge.

torkleyy:
@Xaeroxe I just usually merge it into the release branch if there are no breaking changes, that's why.

PR #332, opened 2017-09-04
[wip] Implement error handling using error-chain

derekdreery:
This PR demonstrates the kind of way you would use `error-chain` to handle errors, and replaces all calls to expect/unwrap in the code and examples. Where I've created new error names to catch errors that were previously panicking, I'm not sure I've picked the right name, or that I'm capturing the right data, but at least this PR gives you a sense of how error handling could work.

It collects any errors in the parallel draw call into a vector.

The examples have been updated to use `quick_main` macro, which allows the "main" function to return a `Result`. If you run with `RUST_BACKTRACE=1 cargo run ...`, then whenever an error-chain error is created, a backtrace is attached, and reported if using quick_main.

### DIFF
--- a/amethyst_renderer/Cargo.toml
+++ b/amethyst_renderer/Cargo.toml
@@ -37,6 +37,7 @@ rayon = "0.8"
 serde = "1.0"
 serde_derive = "1.0"
 winit = "0.7"
+error-chain = "0.10"
 
 gfx_device_gl = { version = "0.14", optional = true }
 gfx_window_glutin = { version = "0.17", optional = true }
--- a/amethyst_renderer/examples/material.rs
+++ b/amethyst_renderer/examples/material.rs
@@ -1,5 +1,6 @@
 //! Launches a new renderer window.
 
+#[macro_use] extern crate error_chain;
 extern crate amethyst_renderer as renderer;
 extern crate cgmath;
 extern crate genmesh;
@@ -11,13 +12,14 @@ use genmesh::{MapToVertices, Triangulate, Vertices};
 use genmesh::generators::SphereUV;
 use renderer::prelude::*;
 use renderer::vertex::PosNormTangTex;
+use renderer::Result;
 
-fn main() {
+fn run() -> Result<()> {
     use std::time::{Duration, Instant};
     use winit::{Event, EventsLoop, WindowEvent};
 
     let mut events = EventsLoop::new();
-    let mut renderer = Renderer::new(&events).expect("Renderer create");
+    let mut renderer = Renderer::new(&events)?;
     let pipe = renderer
         .create_pipe(
             Pipeline::build().with_stage(
@@ -25,13 +27,10 @@ fn main() {
                     .clear_target([0.0, 0.0, 0.0, 1.0], 2.0)
                     .with_model_pass(pass::DrawShaded::<PosNormTangTex>::new()),
             ),
-        )
-        .expect("Pipeline create");
+        )?;
 
     let verts = gen_sphere(64, 64);
-    let mesh = renderer.create_mesh(Mesh::build(&verts)).expect(
-        "Mesh create",
-    );
+    let mesh = renderer.create_mesh(Mesh::build(&verts))?;
 
     let mut scene = Scene::default();
     let alb = Texture::from_color_val([1.0; 4]);
@@ -52,8 +51,7 @@ fn main() {
                         .with_albedo(alb.clone())
                         .with_roughness(rog)
                         .with_metallic(met),
-                )
-                .expect("Material create");
+                )?;
             let model = Model {
                 mesh: mesh.clone(),
                 material: mtl,
@@ -103,10 +101,13 @@ fn main() {
             _ => (),
         });
 
-        renderer.draw(&scene, &pipe, delta);
+        renderer.draw(&scene, &pipe, delta)?;
         delta = Instant::now() - start;
     }
+    Ok(())
 }
+
+quick_main!(run);
 
 fn gen_sphere(u: usize, v: usize) -> Vec<PosNormTangTex> {
     SphereUV::new(u, v)
--- a/amethyst_renderer/examples/sphere.rs
+++ b/amethyst_renderer/examples/sphere.rs
@@ -1,5 +1,6 @@
 //! Launches a new renderer window.
 
+#[macro_use] extern crate error_chain;
 extern crate amethyst_renderer as renderer;
 extern crate cgmath;
 extern crate genmesh;
@@ -11,13 +12,14 @@ use genmesh::{MapToVertices, Triangulate, Vertices};
 use genmesh::generators::SphereUV;
 use renderer::prelude::*;
 use renderer::vertex::PosNormTex;
+use renderer::Result;
 
-fn main() {
+fn run() -> Result<()> {
     use std::time::{Duration, Instant};
     use winit::{Event, EventsLoop, WindowEvent};
 
     let mut events = EventsLoop::new();
-    let mut renderer = Renderer::new(&events).expect("Renderer create");
+    let mut renderer = Renderer::new(&events)?;
     let pipe = renderer
         .create_pipe(
             Pipeline::build().with_stage(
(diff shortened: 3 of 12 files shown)


derekdreery:
I'm not sure of the impact on performance - you may want to silently ignore errors in production to run faster.

Rhuagh:
Seems it's not working very well with ECS. Could you look into that?

derekdreery:
Yes, at this point I'm just gathering opinions on the general strategy.

Rhuagh:
FWIW, I like it.

amethyst_renderer/examples/material.rs
```diff
@@ -1,5 +1,6 @@
 //! Launches a new renderer window.
 
+#[macro_use] extern crate error_chain;
```
Xaeroxe: Odd, I've been writing mine as

```Rust
#[macro_use]
extern crate error_chain;
```

Is it documented anywhere what style these should be?  I wasn't able to find documentation for this.

Xaeroxe: Note: I'm not requesting this be changed I just want to know if I've been doing mine wrong.

Aceeri: I typically have been doing it your way, but there are some merits to the one liner.

amethyst_renderer/examples/sphere.rs
```diff
@@ -70,11 +68,14 @@ fn main() {
...
         });
 
-        renderer.draw(&scene, &pipe, delta);
+        renderer.draw(&scene, &pipe, delta)?;
         delta = Instant::now() - start;
     }
+    Ok(())
 }
 
+quick_main!(run);
```
Aceeri: In the future we should replace this with main again: https://github.com/rust-lang/rust/issues/43301

omni-viral:
`error_chain` looks similar to `quick_error`.
But later does implement `From` im macro.
Is `error_chain` better in some way?

amethyst_renderer/src/error.rs
```diff
@@ -1,126 +1,131 @@
...
-            Error::PoolCreation(_) => "Failed to create thread pool!",
-            Error::ProgramCreation(_) => "Failed to create shader program!",
-            Error::ResViewCreation(_) => "Failed to create resource view!",
-            Error::TargetCreation(_) => "Failed to create render target!",
-            Error::TextureCreation(_) => "Failed to create texture!",
-            Error::WindowDestroyed => "Window has been destroyed!",
+error_chain! {
+    errors {
+        /// Failed to create a buffer.
+        BufferCreation(e: gfx::buffer::CreationError) {
```
omni-viral: According to docs [here](https://docs.rs/error-chain/0.11.0-rc.2/error_chain/index.html)
when you wrap another error (not a chain) you put it into `foreign_links` block.
This will add `From` implementations automatically.

omni-viral: If you will wrap this chain in main crate you need to add it in `links` block

amethyst_renderer/examples/material.rs
derekdreery: I'm completely indifferent - I'll change to the amethyst standard :)

derekdreery:
Thanks for the replys, I'll implement this for all the crates

amethyst_renderer/src/error.rs
derekdreery: If you use foreign_links, you gain ergonomics, but lose the ability to control the description (description is forwarded from contained error). I'll implement it using foreign_links, but want to note the trade-off.

amethyst_renderer/Cargo.toml
```diff
@@ -37,6 +37,7 @@ rayon = "0.8"
 serde = "1.0"
 serde_derive = "1.0"
 winit = "0.7"
+error-chain = "0.10"
```
torkleyy: There already is `0.11` btw

amethyst_renderer/examples/material.rs
torkleyy: Did you do `cargo fmt`?

amethyst_renderer/examples/material.rs
```diff
@@ -103,11 +101,14 @@ fn main() {
...
         });
 
-        renderer.draw(&scene, &pipe, delta);
+        renderer.draw(&scene, &pipe, delta)?;
         delta = Instant::now() - start;
     }
+    Ok(())
 }
 
+quick_main!(run);
```
torkleyy: Examples should, if possible, avoid macros IMO.

Emilgardis: [quick_main!](https://github.com/rust-lang-nursery/error-chain/blob/master/src/quick_main.rs) could easily be replaced with how it is recommended to be done: [link](https://github.com/rust-lang-nursery/error-chain/blob/master/examples/quickstart.rs)


'''
fn main() {
    if let Err(ref e) = run() {
        use std::io::Write;
        let stderr = &mut ::std::io::stderr();
        let errmsg = "Error writing to stderr";

        writeln!(stderr, "error: {}", e).expect(errmsg);

        for e in e.iter().skip(1) {
            writeln!(stderr, "caused by: {}", e).expect(errmsg);
        }

        // The backtrace is not always generated. Try to run this example
        // with `RUST_BACKTRACE=1`.
        if let Some(backtrace) = e.backtrace() {
            writeln!(stderr, "backtrace: {:?}", backtrace).expect(errmsg);
        }

        ::std::process::exit(1);
    }
}
'''

amethyst_renderer/src/error.rs
```diff
@@ -1,126 +1,70 @@
 //! Renderer error types.
 
-use std::error::Error as StdError;
-use std::fmt::{Display, Formatter};
-use std::fmt::Result as FmtResult;
-use std::result::Result as StdResult;
-
 use gfx;
 use gfx_core;
```
Emilgardis: This is not recommended error_chain usage. It is advised to use paths from root, i.e `::gfx_core::*`

amethyst_renderer/src/lib.rs
```diff
@@ -65,6 +65,11 @@
 #![deny(missing_docs)]
 #![doc(html_logo_url = "https://tinyurl.com/jtmm43a")]
 
+// for error-chain
+#![recursion_limit="4096"]
```
Emilgardis: Why have 4096 as the limit?
error_chain recommends 1024, and if I remember correctly, newer rust versions have this limit already set to 1024.

amethyst_renderer/src/error.rs
Emilgardis: However, if it works I don't see any problems.

derekdreery:
I was having some issues with maxing out recursion. It might have been
before they increased the limit, so I will check without the config line
before I say everything's ready for review.

omni-viral:
This feature becomes rather important as hierarchy of errors in amethyst grows.

amethyst_renderer/Cargo.toml
derekdreery: ty

derekdreery:
I'm currently trying to understand the asset loader, so I can do an error chain. I need to have a think, because at the moment errors are not concrete types.

omni-viral:
> errors are not concrete types.

@derekdreery what do you mean?

derekdreery:
An example would be

```rust
#[derive(Clone, Debug)]
pub enum LoadError<A, F, S> {
    /// The conversion from data -> asset failed.
    AssetError(A),
    /// The conversion from bytes -> data failed.
    FormatError(F),
    /// The storage was unable to retrieve the requested data.
    StorageError(S),
}
```

I'm still exploring where these are specialised into concrete types. It might be easier if I ask some questions, maybe on gitter? I'll have to see if I have time to work on this today.

omni-viral:
@derekdreery `LoadError` is instantiated for `Loader` functions

Xaeroxe:
Out of 12 files changed 10 now have conflicts.  It'd probably be easier to rewrite this PR rather than try and re-merge it at this point.  If you disagree with me @derekdreery I'll re-open this but for now I'm closing it.

For future PR authors: Large architecture overhauls that hit multiple function signatures and existing processes have to be completed quickly or they rapidly become changes to outdated code.  This PR has been open for almost a month now.

derekdreery:
I agree.

PR #340, opened 2017-09-09
Add Loading state

torkleyy:
That's about the concept I imagined. I did not try it out e.g. by creating examples yet.

One thing I noticed (which should be done anyways, even without this PR) is that we should have the music (IIRC in the pong example) as a resource (instead of inside a system). It would be better to just have a `Vec` of `Source`s IMO, which would also allow adding the music later.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -1,7 +1,54 @@
 //! Utilities for game state management.
 
+use std::error::Error;
+
+use futures::Future;
+
 use engine::Engine;
 use event::Event;
+
+/// A state which allows to display a loading screen until a future
+/// is `Ready`.
+#[derive(Debug)]
+pub struct Loading<C, F> {
+    create_state: Option<C>,
+    future: F,
+}
+
+impl<C, F> Loading<C, F> {
+    /// Creates a new `Loading` state.
+    /// `create_state` is a closure which accepts the polled
+    /// item from the future; it's responsible for creating the new
+    /// `State`, which will be switched to.
+    pub fn new(future: F, create_state: C) -> Self {
+        Loading {
+            create_state: Some(create_state),
+            future,
+        }
+    }
+}
+
+impl<C, F, S, T> State for Loading<C, F>
+where C: FnOnce(T) -> S,
+      F: Future<Item = T>,
+      F::Error: Error,
+      S: State + 'static,
+{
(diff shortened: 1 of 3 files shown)


Xaeroxe:
I didn't see your status label there.  What's left to do?

torkleyy:
Oh uh it's unclear if this is really the way to go (there may be other ways to go, this PR sort of influences the engine design, that's why I added `type: RFC`) and there's no example showing whether this is practical yet.

Xaeroxe:
Hm well it seemed quite practical to me and an improvement on our existing design but sure I'll wait for others to weigh in.

Rhuagh:
I have an update of the pong example for the music, which moves the sources into a resource. But I think it's way outdated now (a week later :P). I can take a look at updating only the Pong example in a day or two if you want me to.

Rhuagh:
I think this is a great start btw, there's just one thing that I think needs addressing. 

With the proposed solution the state it's transitioning from would have to create the future, i.e. It would have to know what data the state it's transiting to needs. I would propose that we add a way for a state to create the future and defer its own start. This could possibly be done with an on_init that returns an optional future+fnonce, if that is Some, the Loading state is pushed.

Rhuagh:
On_init could just return Option<Loading>

src/state.rs
```diff
@@ -1,8 +1,55 @@
...
+
 use engine::Engine;
 use event::Event;
 
+/// A state which allows to display a loading screen until a future
+/// is `Ready`.
+#[derive(Debug)]
+pub struct Loading<C, F> {
+    create_state: Option<C>,
+    future: F,
```
omni-viral: Why not `Future::map` those two?

Xaeroxe:
@torkleyy So we got some extra comments from @Rhuagh and @omni-viral.  I too like the `on_init` idea as it lets a state define its own asset needs.  This PR is also going to need to be rebased a bit as it has a conflict now.

Xaeroxe:
(Although I might call `on_init` `load` instead.)

Rhuagh:
I don't mind what it's called tbh. `load` or `on_load` is better than `on_init` though, singals more clearly what it's about.

torkleyy:
This needs both a rebase and some more thoughts, so I'm closing it.

PR #341, opened 2017-09-09
Add Loader::load_data

torkleyy:
This allows loading assets from their asset data.

In order to rewrite the examples, we need to implement `Asset` for `MaterialComponent`.

### DIFF
--- a/amethyst_assets/src/asset.rs
+++ b/amethyst_assets/src/asset.rs
@@ -1,7 +1,7 @@
 use std::error::Error;
 
 use futures::{Async, Poll};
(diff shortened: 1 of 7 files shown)


Xaeroxe:
Why would an end user prefer this over Asset::load?  I thought the Asset::Data type was mostly an implementation detail.

torkleyy:
Not really. Let's say you got the vertices and the indices of the mesh and you wanna build a mesh out of that. That's what `Loader::load_data` is for.

torkleyy:
Travis is still working private repo's osx builds, so this might take while. But there's no rush.

Xaeroxe:
I'm not worried about these changes so I'm gonna override bors.

Issue #249, opened 2017-06-19
Plan for 0.6 release


nchashch:
# Why should we plan?
- Avoid duplication of effort
- Do some architecting in advance
- Implement the most valuable features first
- Set clear goals after accomplishing which we can release 0.6
- Get an idea of how much work is ahead
# How should we plan?
To make it simple I propose:
1. List out all the features we want (it would also be nice to rank them by value/effort ratio)
2. Let developers explicitly claim them (and if they chose to drop the project, the feature would again become "claimable" - with whatever work has already been done before the original developer dropped it already done)

This would serve two purposes:
1. Make duplication of effort very difficult (because areas of work are explicitly claimed)
2. Plan things out in English before writing any code (so we can apply scarce open source developer hours for maximum effect)
# Conclusion
So I think it would be nice to discuss what features we want to implement before 0.6 release in this issue's comments. Then we could compile a final list of features (ranked by value/effort) in the issue itself, and then let people claim features they would like to work on and then during development negotiate dropping and reclaiming the dropped features.

Xaeroxe:
If possible I'd like to get sound effect and music APIs into 0.6.  My first goal is to get 3D audio working in rodio, then integrate rodio into Amethyst.  I know we said in https://github.com/amethyst/amethyst/issues/243 that we didn't really want to make 3D audio in Rust however I wasn't really terribly impressed by the alternatives, as OpenAL isn't exactly Open anymore.  I've got a branch of rodio with super rudimentary audio attenuation that I really just need to upgrade into a more robust and better quality implementation.

Value: High, videogames without sound are kind of a hard sell.
Effort: High, I've found one source with good documentation on how to do 3D audio and it'll be quite a bit of reading for me to sift through before I really know how to do it.

vitvakatu:
Well, I don't know whether it would be possible to bring it into 0.6 release or not, but I'd like to see basic collision detection and basic physics subsystem. I want to use `collision-rs` and maybe `froggy` for that task.

_Value:_ High, we  don't have any collision detection/physics yet and there is no evidence that we can use external libs for this purpose.
_Effort:_ Above average/High. We already have some algorithms in `collision-rs`, we can use experience of `nphysics`. I'm sure I can do this, but I don't know how long will it take.

minecrawler:
[...]

Issue #176, opened 2017-02-08
Question: more functionality on specs Camera


derekdreery:
I want to quickly implement a free-flow camera, and to do this I need to be able to rotate the view. The specs camera makes this difficult (you'd need to move the lookat value in a circle around the eye).

I'm happy to implement some methods on the Camera (something like below) if the project owners think this is valuable?

```rust
impl Camera {
    /// Rotates the camera relative to the current view direction
    pub fn view_rotate(&mut self, pitch: f32, yaw: f32, roll: f32) { ... }

    /// Rotates the camera relative to the world coordinate system
    pub fn world_rotate(&mut self, x: f32, y: f32, z: f32) { ... }
}
```

OvermindDL1:
If you want to rotate relative then taking a quaternion instead of pitch/yaw/roll or x/y/z would remove ambiguities in the transformation (especially if you want to lerp it over time).

derekdreery:
Ok I've read some literature and see why quaternions are useful. Maybe the API could be:

```rust
impl Camera {
    /// Rotate camera using given quaternion
    pub fn rotate_quaternion(&mut self, q: cgmath::Quaternion<f32>) { ... }

    /// Rotate camera about given axis by given amount (clockwise)
    pub fn rotate(&mut self, angle: f32, axis: cgmath::Vector3<f32>) { ... }
}

derekdreery:
Or could take a `[f32; 4]` instead of a quaternion for compatibility.

OvermindDL1:
> Or could take a `[f32; 4]` instead of a quaternion for compatibility.

That would not have guaranteed alignment though, a nice thing about quaternions as well is the SSE'abilities of them.  Plus it is documentative.  :-)

I like the two you shown though, a `rotate_quaternion` method is the one that shoul dbe used, but your `rotate` method taking an angle and an axis (really need vectors for documentative purposes as well I'd say...) is also not bad, but it does calculate slower than a quaternion (a quaternion is just a 4d unit vector, so you can use traditional vector optimizations on them, including SSE commands).

However, in my opinion a `rotate` function should not take alternatives, it should take 'one source of truth', and instead the quaternion should have a helper on it that takes an angle and axis to build a quaternion from that.

ebkalderon:
@derekdreery I think this might be a better question for the `specs` developers to handle, since we are essentially using their `Camera` resource verbatim (discounting the completely separate `Camera` in the renderer crate which serves an entirely different purpose).

[...]

Issue #124, opened 2016-10-25
Add logging system


nchashch:
There is a need for a logging system, to uniformly handle errors/provide meaningful debug messages for the user. Currently errors are reported using `println!`s and `panic!`s.

Probably an instance of `Logger` should be inside of the `Context` struct (or it should be a resource in the `World` as suggested in #125), so it is available globally.

Emilgardis:
Yes, I believe the wisest move is to use [slog](https://github.com/slog-rs/slog).

Aceeri:
Would we even need to put the instance of the Logger inside of the Context/resources? `slog` seems to work as a global variable similar to `println!` and the `log` crate.

Emilgardis:
It can work as that but it can also provide an interface "per-module", see https://github.com/slog-rs/slog/blob/master/crates/example-lib/lib.rs

Aceeri:
@Emilgardis Yeah, but I feel having a logger as a global is a big plus rather than having to pass it to everything.

Emilgardis:
That sounds good to me. :+1:

Emilgardis:
I'm still working on this, it has just been slow, but now that the context rewrite is done I feel like it can be done correctly.

Emilgardis:
I'm going to do this by not having the logger as a resource, but also make clones of the root logger that can then be fetched with ´slog_scope::logger()`.

Binero:
I think having the logger be accessible globally encourages bad design. One of the features of slog is to create 'forks' of loggers, and most importantly, to filter on them. It's best if somehow every System can have access to their own, and only their own logger.

torkleyy:
@Binero could you elaborate on how "having the logger be accessible globally" encourages bad design?

Binero:
@torkleyy I briefly touched the topic in the original post, but I will elaborate.

`slog` has many interesting features like filtering what is logged. If different systems share a common `Logger`, we cannot filter log messages by system anymore. This one one of the main features of `slog`. Usually when debugging, we debug one system and not other systems that might be unrelated. 

Every system should only have access to a specific fork of the global `Logger`, as that considerably increases the power of the logging framework by unlocking this powerful filtering capability.

[...]

PR #342, opened 2017-09-10
Implement Advanced Rotation and Movement

minecrawler:
Well, since I already told you that I need rotation for my project, I figured I can just contribute this feature so it arrives faster (hopefully). I will start by doing the core camera modification and then try to come up with a way to add all the different ideas and wishes the best I can. Feel free to make suggestions anytime.

I am still quite the Rust beginner and this is my first quaternion camera implementation, so I hope that I don't make too many stupid mistakes. Well, this PR should help get early feedback and keep up the quality :)

I test the implementation against my project (so I get hands-on feedback on rotation-behavior), however I will also write unit tests later on. If you have a 2D project with orthographic camera, feel free to do a test for your project and give feedback for your use-cases, for example do you need any specific helper-functions?

Related discussions are happening in #176 and #249 .


amethyst_renderer/src/cam.rs
```diff
@@ -53,25 +53,128 @@ impl From<Projection> for Matrix4<f32> {
...
+        //self.rotation = Quaternion::from::<f32>(Euler {
+        //    x: x,
+        //    y: y,
+        //    z: z,
+        //});
+        
+        self
+    }
+    
     /// Calculates the view matrix from the given data.
```
Rhuagh: I would do this as
```
impl Into<Matrix<f32>> for Camera {
    fn into(self) -> Matrix<f32> {
        [...]
    }
}
```

src/ecs/rendering/systems.rs
```diff
@@ -58,10 +58,12 @@ impl<'a> System<'a> for RenderSystem {
             self.scene.add_light(light.0.clone());
         }
 
-        self.scene.add_camera(camera.clone());
+        let mut cam = camera.clone();
```
Rhuagh: Why the clone? You should be able to dereference with `*cam`.

Rhuagh: Ah, I see you copied the previous code. Your code makes it so only the version going into the renderer have the updated values, which is not what you want I think.

```
camera.update();
self.scene.add_camera(camera.clone());
```

amethyst_renderer/src/cam.rs
```diff
@@ -53,25 +53,128 @@ impl From<Projection> for Matrix4<f32> {
...
 impl Camera {
+    /// Create a new camera from the given values.
+    pub fn new(
+        projection: Matrix4<f32>,
+        position: Point3<f32>,
+        look_at: Point3<f32>,
+        up: Vector3<f32>
+    ) -> Self {
+        if position == look_at {
+            panic!("Position and LookAt cannot have the same coordinates!");
```
torkleyy: I think Amethyst should only panic in very special situations, which most likely don't even occur. Just think about the fact that this data might come from a config file or a script, which shouldn't cause a panic.

amethyst_renderer/src/cam.rs
```diff
@@ -53,25 +53,128 @@ impl From<Projection> for Matrix4<f32> {
...
+        }
+        
+        let base_forward = (look_at - position);
```
torkleyy: No need for `(` parentheses `)`.

amethyst_renderer/src/cam.rs
```diff
@@ -53,25 +53,128 @@ impl From<Projection> for Matrix4<f32> {
 }
 
 /// Camera struct.
-///
-/// TODO: Add more convenience methods, refine API.
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 pub struct Camera {
+    /// Initial forward vector of the camera.
+    pub base_forward: Vector3<f32>,
```
torkleyy: Can't we just make this `(0.0, 0.0, -1.0)`?

amethyst_renderer/src/cam.rs
```diff
@@ -53,25 +53,128 @@ impl From<Projection> for Matrix4<f32> {
...
+    /// Calculated center point of the camera.
+    center: Vector3<f32>,
+    /// Trigger re-calculations if values were changed.
+    pub dirty: bool,
     /// Location of the camera in three-dimensional space.
     pub eye: Point3<f32>,
```
torkleyy: I think we should convert the camera to a component and use the `Transform` position and rotation.

amethyst_renderer/src/cam.rs
torkleyy: This method doesn't actually consume `self`, so we should not use `to` or `into`.

amethyst_renderer/src/cam.rs
minecrawler: The old implementation let the user set their own up- and right-vectors, so I just left it there. If we set the base-forward, we might as well define the up vector, too. Though we have to communicate that decision clearly in the book and docs.

I don't care either way, since I don't know of any specific use case, which would require the user to deviate (other than preference).

torkleyy: I guess we shouldn't use these directional vectors any longer and use transform stuff instead.

minecrawler:
@torkleyy very good idea. I will move the helper functions there, if that is all-right with you

amethyst_renderer/src/cam.rs
Rhuagh: True.

amethyst_renderer/src/cam.rs
Xaeroxe: @minecrawler In instances where the structure creation can fail it's typically considered appropriate to have new return a `Result` or `Option`

amethyst_renderer/src/cam.rs
```diff
@@ -53,25 +53,128 @@ impl From<Projection> for Matrix4<f32> {
...
+            dirty: false,
+            eye: position,
+            proj: projection,
+            rotation: Quaternion::look_at(base_forward, up).normalize(),
+            up: up,
+        }
+    }
+    
+    /// Move the camera relatively to its current position, but independently from its orientation.
+    pub fn move_absolute(&mut self, direction: Vector3<f32>) -> &mut Self {
```
omni-viral: Camera should be just an entity with a pair: `Transform` and `Projection` components.
I like those functions but they all belong to `Transform` component.

omni-viral: To know what camera to use there should be special `Resource` with `EntityId`.
In future there will be many viewports and cameras in use. Viewports will have `EntityId` of the camera.

minecrawler: I am currently wondering, if those methods shouldn't go into `LocalTransform`? As for multi-cameras, I guess I can prepare that by using an entity with components and then getting the first camera/transform component-pair in the rendering system to shove into the `Scene`. I don't see any viewports, yet, and I think they are out of scope for this PR...

Rhuagh: For now, I'd just add a CameraEntity resource:

```
pub struct CameraEntity {
    pub entity: Entity,
}
```

and in RenderSystem: `Fetch<'a, CameraEntity>`.

Rhuagh: And yes, those methods should probably go in LocalTransform.

omni-viral: @Rhuagh, `CamearEntity` resource is what I would recomend. And yes, `LocalTransform` is the right place

Rhuagh: I agree.

torkleyy: Why a camera resource? You can just filter for it, and check if it's the only entity with a `Camera` component:

```rust
let mut cams = (&transforms, &cameras).join();
let cam = cams.next();
if cams.next().is_some() {
    // Log error: "Engine only allows one camera at the moment"
}

if let Some(cam) = cam {
    ...
}
```

torkleyy: This would at least be more obvious and extensible.

minecrawler: I also prefer the `join()`-way. It feels more ECS-y imho. After all, we already have Entities and components. Also, having more than one way to retrieve the correct camera might lead to inconsistencies and bugs later on (which one to use - the resource or the camera?) until the viewports are implemented and define the only way again...

As for the error: I would rather only log a warning. It's not an error we have to handle when there are multiple cameras, the others are just not used. By the way, do we even have a logging system, yet? I think we only have an issue (#124) at the moment... I will leave a `// todo` in place.

omni-viral: I'd like to have multiple cameras before multiple viewports. Switching active camera is better than storing data elsewhere and overwrite it.

omni-viral: That's why I suggest to have active camera id as a resource.

torkleyy: Why not just introduce both at once? It isn't really more work and for now, the solution I suggested is simpler and more obvious.

minecrawler: ok, will do. Maybe I can finally push all those changes tonight or tomorrow, so you can kill me :D

Xaeroxe: We like you alive :P

minecrawler: @torkleyy so you want to postpone this PR until someone starts working on multiple viewports?

Well, I'd like to keep it simple and I implemented the solution you present here in the scene in order to find the first camera. All others are silently dropped atm (though it might print a warning as soon as the logging system is in place). If @omni-viral needs the resource on top of that, it shouldn't be a problem to implement it. The more I think about it, the more sense it makes to actually use a resource if you want to have multiple cameras now. Later on, we just have to remove the global resource and store the active camera entity on the viewport instead...

torkleyy: No. I'd like to do what @omni-viral mentioned in a future PR. I'm not sure if we need multiple cameras now. What would be the use of that? And btw, if we really wanted that the "proper" solution would be to have marker components e.g. `ActiveCamera` (using a `NullStorage`).

minecrawler: Yeah, I also wanted to write about an `ActiveCamera` marker, however how would you mark that for multiple VPs later on? Well, I guess, no matter what we do now, it's temporary.

Ah, and btw, maybe he needs them for his own project? idk and idc why he needs it, but if he says it is a requirement for him, why not implement it for now?

nchashch: Maybe rename `move_absolute` to `move_global` and `move_relative` to `move_local`? Because `move_absolute` would suggest setting the position of the camera and `move_relative` would suggest shifting the position of the camera relative to it's initial position by adding a `direction` vector to it. This unwarranted suggestion would add unnecessary cognitive load on API users. Same change would apply to all of the other `*_absolute`/`*_relative` methods below for consistency if you will decide to make this change.

Also, I think it would be better to change `direction` to `shift` or `translation` because `direction` would suggest a unit vector, which only defines direction, when in fact it is an arbitrary not-necessarily-unit vector representing a shift either in the global coordinate system or in the local coordinate system of the camera.

And it would be nice to add more detailed doc comments for these methods because they wouldn't be at all trivial to a person learning to use this API for the first time. Maybe something like:
```rust
/// Move the camera by `shift` vector in the global coordinate system,
/// the resulting move doesn't depend on camera's orientation.
pub move_global(...) {...}
/// Move the camera by `shift`vector in the local coordinate system, i.e. if `shift == Vec3(dx, dy, dz)`
/// the camera will move by `dx` in the `forward` direction, by `dy` in the `right` direction and,
/// `dz` in the `up` direction. Of course this means that the resulting move depends on the orientation of
/// this `forward`, `right`, `up` triplet of vectors.
pub fn move_local(...) { ... }
```

amethyst_renderer/src/cam.rs
```diff
@@ -53,25 +53,128 @@ impl From<Projection> for Matrix4<f32> {
...
+    
+    /// Move the camera relatively to its current position and orientation.
+    pub fn move_relative(&mut self, direction: Vector3<f32>) -> &mut Self {
+        self.dirty = true;
+        self.eye += self.rotation * direction;
+        self
+    }
+    
+    /// Pitch the camera relatively to the world.
+    pub fn pitch_absolute(&mut self, amount: Deg<f32>) -> &mut Self {
```
nchashch: Maybe `angle` instead of `amount` will be more informative? Same would apply to all of the below methods dealing with angles.

amethyst_renderer/src/cam.rs
omni-viral: @torkleyy. I'm not sure when we will implement multiple viewports. But multiple cameras can be implemented right now. So why should we delay. Idc what resource or marker component will point to active camera. Both will be ok. But I think resource is more clear and unambiguous that there is only one resource hence only one active camera.

omni-viral: Both hands for those names. "Global" and "local" are widely used to describe basis. In 3d modeling software for example.

minecrawler: I agree, I will change the names of the methods and parameters.

torkleyy: @omni-viral I don't really see any obvious use case for the complication, but if you insist on it, I vote for the component marker.

minecrawler:
Sorry I am falling behind. RL struck. Well, for now, I try to do stuff bit-by-bit. It's really not much and I want to get this done :)

I added the stuff we talked about. I did not try to compile it yet (probably needs a fix or two) and the tests are still missing, too. I haven't checked the examples, yet, they will probably need to be updated to use the methods instead of changing the rotation directly.

torkleyy:
Nice to see progress here!

minecrawler:
I am not very happy with the current `Camera::to_view_matrix()`. It requires the `LocalTransform` passed to it, however does not require the camera at all. From this perspective it would make sense to also move it over to `LocalTransform`. However imho, the view matrix calculation does not make a whole lot of sense there from an abstraction point of view. It's only needed for the camera.

What do you think?

Xaeroxe:
Pinging @omni-viral.  I think you're the only one who might have an opinion on @minecrawler's comment.

src/ecs/transform/components/local_transform.rs
```diff
@@ -3,19 +3,25 @@
...
+    /// Right vector [x, y, z]
+    pub right: [f32; 3],
     /// Quaternion [w (scalar), x, y, z]
     pub rotation: [f32; 4],
     /// Scale vector [x, y, z]
     pub scale: [f32; 3],
+    /// Translation/position vector [x, y, z]
+    pub translation: [f32; 3],
+    /// Up vector [x, y, z]
+    pub up: [f32; 3],
```
Aceeri: Are all these really necessary? This nearly doubles the size of the transform.

Aceeri: I think you can just get these from the quaternion:
```rust
let w = transform.rotation[0];
let x = transform.rotation[1];
let y = transform.rotation[2];
let z = transform.rotation[3];
let up = [
    2 * (x * y - w * z), 
    -2 * (x * x + z * z), 
    2 * (y * z + w * x),
];
let right = [
    1 - 2 * (y * y + z * z),
    2 * (x * y + w * z),
    2 * (x * z - w * y),
];
let forward = [
    2 * (x * z + w * y),
    2 * (y * x - w * x),
    1 - 2 * (x * x + y * y),
];
```

Xaeroxe: Which approach we go with kind of depends on how often these are accessed.  If we can it'd be nice to just use results from previous computations, so we can avoid needlessly repeating calculations, but if they're accessed infrequently it might be better to just do the calculation on demand.

minecrawler: yes, for movement and rotation, I have to know how the player set up the orientation of things. Alternatively, we could define a static orientation.... Previously, these fields were only on the camera, but we want to have movement freedom on all objects, right :)

tbh, if you have a better idea on how to implement it, or if you think, that we do not need that kind of functionality, I am open for discussions. For now, I put everything in there which we talked about in the issues and here.

Aceeri: You could always just add a secondary component if you are using these a lot, if we add them to the `LocalTransform` though then we are doubling the size of the component (which leads to less cache efficiency) and if the user doesn't use these then it is wasted.

minecrawler: btw, your above calculations only give you the local system, not the global (as in the base orientation).

I already thought about making another component. Doing the separation might be interesting for static meshes, which don't change their rotation and position during the simulation. The only downside I can see is that the user would have to pass the matching LocalTransform to every movement or use a movement-builder (which passes the LocalTransform). I think, the downside is not too bad, so I will probably implement that.

omni-viral:
Passes fetches `Camera` as a resource. You need to make them fetch it from from entity instead.

src/ecs/transform/components/local_transform.rs
minecrawler: or... we could set the global orientation for _all_ objects as a resource. That would mean another parameter - in case you don't have a better idea on how to get the resource from within LocalTransform...

Well, thinking about this stuff, I am wondering. LocalTransform is a component. Is it even the right place for those calculations? That movement-builder seems better and better (though how good is it performance-wise?)

minecrawler:
@omni-viral yes, I haven't touched the passes, yet. There are still a ton of things I have to fix before it even compiles. Then I have to get the examples running and test the implementation in my application (which is a FP 3D game, it requires those methods). Maybe I'll also add another example featuring a little 3D parcour (from blocks), so we can use it for physics, too, later on...

minecrawler:
@omni-viral What is your take on my question concerning `Camera::to_view_matrix()`?

omni-viral:
@minecrawler you can't calculate view matrix from `LocalTransform` alone.
You need rotation + translation + scale + projection. First three is in `LocalTransform` but you only have projection matrix in `Camera`

minecrawler:
@omni-viral in both, the current implementation and my fork, the projection is not used to calculate the View Matrix. It is calculated using `Matrix4::look_at()`, which has the following [signature](https://docs.rs/cgmath/0.15.0/cgmath/struct.Matrix4.html#method.look_at): 

```rust
fn look_at(eye: Point3<S>, center: Point3<S>, up: Vector3<S>) -> Matrix4<S>
```

what you probably mean is the camera-view-matrix, which is not calculated in `Camera::to_view_matrix()`.

omni-viral:
@minecrawler Indeed. Projection matrix and view matrix both get send into vertex shader.
I think it is good idea to put `to_view_matrix` method into `LocalTransform`. But maybe it should be renamed.

Rhuagh:
View matrix and projection matrix should be kept separate. The view matrix is used in other scenarios, not just for rendering (frustum culling is one).

omni-viral:
@Rhuagh How do you gonna do frustum culling without projection?

Rhuagh:
I shouldn't comment on things while having a fever... Regardless, having the view matrix separate is useful

omni-viral:
Whoops

amethyst_renderer/src/pass/flat.rs
```diff
@@ -67,7 +68,10 @@ where
     N: Component + AsRef<Material> + Send + Sync,
 {
     type Data = (
-        Option<Fetch<'a, Camera>>,
+        Fetch<'a, Orientation>,
+        ReadStorage<'a, ActiveCamera>,
+        ReadStorage<'a, Camera>,
+        ReadStorage<'a, LocalTransform>,
```
Binero: Shouldn't this use `Transform` to build the view matrix, as opposed to `LocalTransform`? What if the camera gets attached to another object?

Binero:
I would prefer it if rendering without a hierarchy also works without `LocalTransform`. I am currently writing a project where I do not use `LocalTransform` at all. My entities are located not with a 3D position, but rather with an angle and a radius. As such I have my own `LocalTransform`-type structure, and a system that converts it into a `Transform`.

amethyst_renderer/src/pass/flat.rs
minecrawler: The camera stores its translation and rotation in a `LocalTransform`. That's per definition. If the camera does not have a `LocalTransform` components, the join further down will return `None`, which means you do not have a camera.

minecrawler:
If I understand you correctly, you don't want to use the Amethyst standard way of handling things because you have an edge-case. I don't know how I can cover each and every edge-case, so imho we should focus on the Amethyst standard. Ultimately, I would like to hear the opinion of the Amethyst Members or hear a proposal on how to architecture it, if you need it.

Binero:
@minecrawler Edge-case or not, this is just one possible edge case. A more flexible design is always a plus, and what you call 'the amethyst standard way' can easily be implemented on top of a more flexible interface. (Being MVP)

amethyst_renderer/src/pass/flat.rs
Binero: That shouldn't be the case at all, and I don't see how it is the case here either. 

`LocalTransform` holds the transform of the entity relative to the transform of the parent. `Transform` holds the absolute transformation of an entity. Entities are able to function without `LocalTransform` just fine.

Xaeroxe:
It is my understanding that all entities with a LocalTransform should also have a Transform.  LocalTransform is merely a convenience structure for describing entities positions relative to their parents.  It contains the same data as Transform if the entity has no parent.  So, use of Transform here should be fine.

@Binero Sounds like you needed radial coordinates, which isn't really that unusual of a use case.  Although it would be nice to have to radial coordinates be part of Amethyst's math offerings itself so I'll see if I can get that integrated into our ecosystem somewhere.

Binero:
@Xaeroxe `Transform` works for all use cases. If we make `LocalTransform` also just a wrapper around a model matrix, we can have one system which generates the `Transform` of every entity regardless of its coordinate system.

We could further specify a new component `CartesianCoordinates` and a new component `PolarCoordinates`, and a system to convert both of those into `LocalTransform` respectively. If someone uses a more exotic coordinate system, they could likewise create a new component to represent it, and a system to go with it.

Xaeroxe:
We should probably make an issue for Polar Coordinates so we don't pollute this pull request.

Xaeroxe:
Just to eliminate any doubt that Transform is the right thing to use in this context, I point to our pong example:

https://github.com/amethyst/amethyst/blob/develop/examples/04_pong/main.rs#L465

minecrawler:
Interesting. I remember talking about the opposite further up. Well, I will have to take a look at `Transform` then. The bad thing is that there really is not much documentation about how to use it. It's a Matrix4, however how is the information (rotation, translation, scale) stored in there?

Do I have to take anything into consideration for the Polar Coordinates?

Xaeroxe:
Polar Coordinates are out of scope for this pull request, I'm probably not the best person to ask about matrices as I'm still learning them myself, but @omni-viral and @Rhuagh might be good resources to you here.

OvermindDL1:
> The bad thing is that there really is not much documentation about how to use it. It's a Matrix4, however how is the information (rotation, translation, scale) stored in there?

I'd assume it's in the standard Translation Matrix form (3 2-vectors stating rotation, 1 stating location, along with some wasted space that makes the math much more efficient).

You'd perform the rotations as a translation to that matrix from your polar coordinates, it is fairly simple (just some sin/cos) and it has to be done anyway as the GPU uses transformation matrices.  'Transform' is definitely the right thing to use in almost all cases.

Rhuagh:
I was under the impression that `Transform` was for local to world transformation, while `LocalTransform` was for local to parent transform?

Binero:
@minecrawler It contains a Transformation matrix. If you multiply a point by that matrix, it will be transformed from its own coordinate system to world coordinates. 

A `LocalTransform` can be turned into a transformation matrix, using the[ `LocalTransform::matrix(&self)`](https://github.com/amethyst/amethyst/blob/develop/src/ecs/transform/components/local_transform.rs#L67) function. 

If we multiply that `LocalTransform` transformation matrix with the transformation matrix of the parent entity, we get the `Transform` of the child entity. This is also exactly how the `TransformSystem` [implements this](https://github.com/amethyst/amethyst/blob/develop/src/ecs/transform/systems.rs#L131).

Rhuagh:
Camera transformation is most likely a local to world transform and the view matrix is the inverse of that usually. I'm not sure if polluting the `Transform` interface with that is a good idea atm,  I'd have to think that through for a while. I do appreciate it will simplify fetching in the renderer.

Binero:
@Rhuagh Not at all. `Transform` contains the absolute transformation of an entity. That is, a point in a coordinate system relative to the current entity, multiplied with `Transform`, gives you that same point but in world coordinates. 

If we take another point in a coordinate system relative to the current entity, and multiply it with `LocalTransform`, we get that point's position relative to the entity of the current parent. If the current entity has no parent, that relative position is also the absolute position.

Rhuagh:
Yeah, that's what I meant

Rhuagh:
I'm just not sure why anyone would use `LocalTransform` if they weren't using it for node hierarchical data, instead of using `Transform` directly.

Rhuagh:
Ah, I've read the rest of the thread here now. Yes, we should support rendering without `LocalTransform` if you don't need hierarchy.

Binero:
@Rhuagh Always using `LocalTransform` is beneficial when entities are attached and detached to other entities (e.g. a player entering a vehicle). The `Camera` should use `Transform` however, as only `Transform` contains the actual transformation of the camera, in absolute world coordinates.

Rhuagh:
I agree. The question remains however, on where we should put the view matrix calculation. Personally I'm in favor of having that in a specialized location (i.e. `Camera` etc).

Binero:
There is only one view matrix, as it translates world space to camera-space (where the camera is the origin). I am fine with having the render system calculate the view matrix based on the camera's `Transform`.

Rhuagh:
I'm not sure that's a good idea, that requires the `RenderSystem` to calculate the inverse of an arbitrary transform, which could cause issues for some rotations.

Binero:
@Rhuagh When would this be an issue? Could you elaborate? As far as I am aware, inverting the translation matrix of the camera is the correct way to retrieve the view matrix.

Rhuagh:
The issue comes if the determinant of the transformation matrix is close to zero, which can happen for specific orientations. In these cases the inverse is undefined, which would cause components of the inverse to be NaN.

Rhuagh:
It's usually beneficial to calculate the view matrix using the inverse of the rotation, and the inverse of translation of the camera separately.

Rhuagh:
After thinking it through a bit more, you'd have the same issue with creating the view matrix outside if your rotation matrix is degenerate, it would just be more obvious where the issue was. 

We could add the view matrix calculation inside the renderer passes that use a camera, but we'd have to be very vocal if the inverse calculation failed. 

So, the `Camera` abstraction should contain utilities for controlling the camera position/orientation then only I guess?

Binero:
@Rhuagh The `Camera` position and orientation would depend on its `Transform`. The `Camera` structure would for now be an empty structure. Once we have multiple viewports, it would contain information on what viewport to use.

Rhuagh:
Well, I thought the point of this PR was to provide easier to use abstractions over the `Transform`? 

I.e. we might want to supply a ThirdPersonCamera, that will float "X units behind, Y units above" the `Entity`, except when colliding with something, then it's skewed etc.

Binero:
@Rhuagh This PR does not provide that functionality as-is. Furthermore, a 3rd person camera would be as trivial as creating a camera entity with a parent, and a LocalTransform transforming the camera right behind the parent.

Rhuagh:
@Binero so what is your proposal for this PR? Should the utilities be on `Transform`, on `LocalTransform`, on both or none? Or should we remove the utilities and let the users handle it on their own?

Binero:
@Rhuagh The utilities should be on higher level abstractions, see #392.

minecrawler:
@Binero which would mean that I would have to wait for the PR implementing those abstractions, then implementing the transform helpers on that. Am I right?

Xaeroxe:
@Binero I'd like to avoid blocking this PR on implementation of #392, unless we can finalize and build those modifications relatively quickly.  It seems we have several unanswered questions on that so let's skip that for now.  Let's leave the transform helpers as is for now and we'll relocate them when we implement #392.

Xaeroxe:
So a lot has happened since this PR was opened.  It'd probably be easier to rewrite it starting from head at this point.  If you disagree with me @minecrawler please tell me and I'll re-open this.

jojolepro:
;(

minecrawler:
Well, I try to keep this branch up2date. I already merged most stuff (for example the bundles), and merging more upstream isn't that big of a problem. Is there a particular PR you are concerned about?

If you think that I am taking too long, then you are very right D: I am still struggling with some of the mathematical and architectural decisions, which only became apparent during discussion. If that's a reason for you to close this PR, then that's ok with me. I received the initial feedback I need, so when I come back with a complete implementation, I hope that there will only be minor issues.

Xaeroxe:
Here we go, go ahead and merge upstream

torkleyy:
Phew, I just clicked my notification and saw tons of changes (several thousands lines), but then saw it's just a merge commit :smile:

minecrawler:
wow. All the camera stuff was driving me crazy about how to get the math right. Now that it's gone (thanks to @Rhuagh), the PR is very small indeed and just adds the helpers to `LocalTransform`. I guess, that's really all I initially wanted, however, the least I got from going through such a lot of code iterations and lines of source is a deeper understanding of Amethyst's code and a bit of practical insight into how you guys work with Rust.

Well, what's still missing is testing the implementation on my application and writing a few test cases. Also, the examples are broken, but by now, that's outside of the scope of this PR.

Other than that: review time?

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -1,22 +1,32 @@
...
     /// Scale vector [x, y, z]
     pub scale: [f32; 3],
+    /// Translation/position vector [x, y, z]
+    pub translation: [f32; 3],
 }
 
 impl LocalTransform {
+    /// Rotate to look at a point in space (without rolling)
+    pub fn look_at(&mut self, orientation: &Orientation, position: Vector3<f32>) -> &mut Self {
+        let cam_quat = Quaternion::from(self.rotation);
```
Xaeroxe: Unused variable

Xaeroxe:
Also: This desperately needs squashing.

amethyst_core/src/transform/components/local_transform.rs
minecrawler: wow, not even the compiler displays a warning about that one, but true, it's unused!

Xaeroxe: It did here: https://travis-ci.org/amethyst/amethyst/jobs/291734426#L806

minecrawler: strange, my local compiler didn't show that one (only a warning in `amethyst_renderer\src\formats\texture.rs:219:5`)~

Rhuagh: Run with nightly compiler.

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -56,63 +55,65 @@ impl LocalTransform {
...
     /// Rotate relatively to the world
     #[inline]
     pub fn rotate_global(&mut self, axis: Vector3<f32>, angle: Deg<f32>) -> &mut Self {
-        let axis_normalized = Vectro3::from(axis).normalize();
-        let q = Quaternion::from::<f32>(Euler {
-            x: axis_normalized.x * angle,
-            y: axis_normalized.y * angle,
-            z: axis_normalized.z * angle,
+        let axis_normalized = Vector3::from(axis).normalize();
+        let q = Quaternion::from(Euler {
```
Rhuagh: `Quaternion::from_axis_angle`

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -124,10 +125,10 @@ impl LocalTransform {
         let rel_axis_normalized = Quaternion::from(self.rotation)
             .rotate_vector(Vector3::from(axis))
             .normalize();
-        let q = Quaternion::from::<f32>(Euler {
-            x: rel_axis_normalized.x * angle,
-            y: rel_axis_normalized.y * angle,
-            z: rel_axis_normalized.z * angle,
+        let q = Quaternion::from(Euler {
```
Rhuagh: See above.

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -136,20 +137,18 @@ impl LocalTransform {
...
-    pub fn set_rotation<D: Into<Deg<f32>>>(&mut self, x: D, y: D, z: D) -> &mut Self {
-        self.rotation = Quaternion::from::<f32>(Euler {
-            x: x,
-            y: y,
-            z: z,
-        }).into();
-
-        self.flag(true);
+    pub fn set_rotation<D: Into<Deg<f32>> + Angle>(&mut self, x: D, y: D, z: D) -> &mut Self {
+        let rotation =
```
Rhuagh: Here, you can use `Quaternion::from(Euler {})`, should be safer.

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -56,63 +55,65 @@ impl LocalTransform {
...
-        self.flag(true);
         self
     }
 
     /// Move relatively to its current position and orientation.
     #[inline]
     pub fn move_local(&mut self, axis: Vector3<f32>, amount: f32) -> &mut Self {
-        self.translation += Quaternion::from(self.rotation).conjugate() * axis.normalize() * amount;
+        let delta = Quaternion::from(self.rotation).conjugate() * axis.normalize() * amount;
+
```
Rhuagh: `self.translation = (Vector3::from(self.translation) + delta).into();`

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -158,22 +157,22 @@ impl LocalTransform {
         let forward = orientation.forward;
         let trans = self.translation;
         let center = Point3::new(
```
Rhuagh: `let center = Point3::from(trans) + Vector3::from(forward);`

amethyst_core/src/transform/components/local_transform.rs
minecrawler: The problem is that the new version of cgmath changed traits on the Euler struct (to ` Angle`), leading to unimplemented trait errors for conversion between cgmath types

Rhuagh: Ugh. Ok.

Rhuagh:
Could you please squash the commits for this PR ?

minecrawler:
tbh, I am not sure how to squash the commits once I have pushed them to the feature branch for this PR. afaik, I can only squash them on merge (which is hidden in the GitHub button, however not supported by bors). Should I create a new PR from a clean branch?

Rhuagh:
Do you have a commandline you can use ?

git rebase -i HEAD~somenumber
change all pick except the first one to squash, and write a new message.

Xaeroxe:
@minecrawler 

@Rhuagh is going to sleep now so be sure to ping me with any questions you have about squashing.

minecrawler:
uh, I am pretty sure this is not how it is supposed to look like. Will apply the changes to a fresh PR tomorrow. I copied the branch before pushing the changes, so nothing is lost

Rhuagh:
Ok, just make a new PR from a fresh branch then

PR #345, opened 2017-09-11
Add volume parameter to play module functions

Xaeroxe:
Allows users to dynamically control sfx volume levels.

### DIFF
--- a/src/audio/play.rs
+++ b/src/audio/play.rs
@@ -2,19 +2,20 @@
 
 use std::io::Cursor;
 
-use rodio::Decoder;
-use rodio::play_once as rplay_once;
+use rodio::{Decoder, play_once as rplay_once, Sink, Source as RSource};
 
 use super::DecoderError;
 use super::output::Output;
 use super::source::Source;
 
-/// Play a sound once.
+/// Play a sound once.  A volume of 1.0 is unchanged, while 0.0 is silent.
 ///
 /// This will return an Error if the loaded audio file in source could not be decoded.
-pub fn try_play_once(source: &Source, endpoint: &Output) -> Result<(), DecoderError> {
-    match rplay_once(&endpoint.endpoint, Cursor::new(source.clone())) {
-        Ok(sink) => {
+pub fn try_play_once(source: &Source, volume: f32, endpoint: &Output) -> Result<(), DecoderError> {
+    let mut sink = Sink::new(&endpoint.endpoint);
+    match Decoder::new(Cursor::new(source.clone())) {
+        Ok(source) => {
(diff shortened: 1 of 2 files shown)


Xaeroxe:
Just realized: I need to update the doc comments for this.  Don't merge yet.

Xaeroxe:
Ready to merge.

src/audio/play.rs
```diff
@@ -2,19 +2,20 @@
...
 ///
 /// This will return an Error if the loaded audio file in source could not be decoded.
-pub fn try_play_once(source: &Source, endpoint: &Output) -> Result<(), DecoderError> {
-    match rplay_once(&endpoint.endpoint, Cursor::new(source.clone())) {
-        Ok(sink) => {
+pub fn try_play_once(source: &Source, volume: f32, endpoint: &Output) -> Result<(), DecoderError> {
+    let mut sink = Sink::new(&endpoint.endpoint);
+    match Decoder::new(Cursor::new(source.clone())) {
+        Ok(source) => {
+            sink.append(source.amplify(volume));
```
torkleyy: Is there any difference between `sink.set_volume` and `source.amplify`?

src/audio/play.rs
```diff
@@ -29,46 +30,39 @@ pub fn try_play_once(source: &Source, endpoint: &Output) -> Result<(), DecoderEr
...
 
-/// Play a sound n times.
+/// Play a sound n times.  A volume of 1.0 is unchanged, while 0.0 is silent.
 ///
 /// This may silently fail, in order to get error information use `try_play_n_times`.
-pub fn play_n_times(source: &Source, endpoint: &Output, n: u16) {
+pub fn play_n_times(source: &Source, volume: f32, endpoint: &Output, n: u16) {
     // This absurd construct is here to keep warnings about an unused result from appearing.
-    if let Ok(()) = try_play_n_times(source, endpoint, n) {}
+    let _ = try_play_n_times(source, volume, endpoint, n);
```
torkleyy: Could we still have a printed error message here?

src/audio/play.rs
Xaeroxe: Not really no, they both have the same impact on emitted samples.  So I'm just going to leave this as is because there's no reason to go either direction.

src/audio/play.rs
Xaeroxe: Sure!

src/audio/play.rs
Xaeroxe: So I guess there is one subtle difference.  Sink volumes can take up to 5 ms to be propagated to the sink, so source amplify doesn't have this 5 ms delay.

Issue #329, opened 2017-09-02
Automatically create formatting PR


torkleyy:

Binero:
Can't this be closed now?

Xaeroxe:
Nope, because we still don't enforce formatting via CI.  I tried a few times but nothing got merged because I was indecisive.

torkleyy:
My idea would still be to have a bot which just creates a PR every three days in case the codebase is not fully formatted. We could even auto-merge it in case bors doesn't have concerns.

zoonage:
Maybe it's also worth having travis check that running `cargo fmt` doesn't result in any changes?

Rhuagh:
It's harder than it sounds, we've tried a few times to do this.

zoonage:
What about something like `cargo fmt -- --write-mode diff` inside an include-only matrix for rust stable and then checking the return code?

torkleyy:
The problem with checking formatting on CI is that it makes contributions
harder and makes our builds break (and feature PRs incredibly large due to
formatting changes).
I prefer pure formatting PRs because of that.

jojolepro:
Then we don't want to enforce formatting and we can close this?

zoonage:
I think I agree with @torkleyy with the periodic formatting PRs, should we re-name the ticket to implement this? As @jojolepro said the ticket title is wrong

jojolepro:
From @torkleyy 
>My idea would still be to have a bot which just creates a PR every three days in case the codebase is not fully formatted. We could even auto-merge it in case bors doesn't have concerns.

torkleyy:
Thanks to @magnonellie's work on CI and efforts to stabilize rustfmt, this can be closed now.

PR #348, opened 2017-09-12
Enforce rustfmt via travis.

Xaeroxe:
This PR will cause travis to fail if rustfmt detects any unformatted files.

Fixes https://github.com/amethyst/amethyst/issues/329

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,24 @@
-# Set up the Rust toolchain.
+---
+env: 
+  - FORMATTING=true
+  - FORMATTING=false
 language: rust
-rust:
+matrix: 
+  allow_failures: 
+      - env: "FORMATTING=false"
+        rust: nightly
+  exclude: 
+      - env: "FORMATTING=true"
+        rust: stable
+      - env: "FORMATTING=true"
+        rust: beta
+      - env: "FORMATTING=true"
+        os: osx
+        rust: nightly
+rust: 
   - stable
   - beta
   - nightly
-
-# Allow for occasional compiler breakage on nightly Rust.
-matrix:
-  allow_failures:
-    - rust: nightly
 
 # Build the Bors staging and trying branches, plus PRs to develop, master, and
 # the latest release branch.
@@ -30,6 +40,7 @@ addons:
   apt:
     packages:
       - libasound2-dev
+      - cmake
 
 # Since this build doesn't require sudo, we can turn it off and improve build
 # performance.
@@ -43,13 +54,7 @@ cache: cargo
 before_script:
   - export PATH=$PATH:/home/travis/.cargo/bin
 
-# Generate documentation, compile the engine, run tests.
-script: |
-  cargo doc --all --no-deps -v &&
-  # Build and test without profiler
-  cargo test --all -v &&
-  # Build and test with profiler
-  cargo test --all --features profiler -v
+script: ./travis_script.sh 
 
 # Push notifications to `amethyst/general` and `amethyst/engine` Gitter chats.
 notifications:
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,3 +1,4 @@
 reorder_imports = true
 reorder_import_names = true
 reorder_imports_in_group = true
+error_on_line_overflow = false
--- a/src/app.rs
+++ b/src/app.rs
@@ -20,7 +20,6 @@ use timing::{Stopwatch, Time};
 
 /// User-friendly facade for building games. Manages main loop.
 ///
-
 #[derive(Derivative)]
 #[derivative(Debug)]
 pub struct Application<'a, 'b> {
@@ -57,7 +56,7 @@ impl<'a, 'b> Application<'a, 'b> {
     /// # Type Parameters
     ///
     /// - `S`: A type that implements the `State` trait. e.g. Your initial
-    ///        game logic. 
+    ///        game logic.
     ///
     /// # Lifetimes
     ///
@@ -76,9 +75,9 @@ impl<'a, 'b> Application<'a, 'b> {
     /// use amethyst::prelude::*;
     ///
     /// struct NullState;
-    /// 
+    ///
     /// impl State for NullState {}
-    /// 
+    ///
     /// fn main() {
     ///     let mut game = Application::new(NullState).expect("Failed to initialize");
     ///     game.run();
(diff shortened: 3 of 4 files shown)


.travis.yml
```diff
@@ -46,9 +46,15 @@ before_script:
...
 script: |
-  cargo install rustfmt &&
-  cargo install cargo-update &&
-  cargo install-update rustfmt &&
+  # Check if cargo-update is present
+  if ![cargo install-update --version]; then
+    cargo install cargo-update;
+  fi &&
+  if ![rustfmt --version]; then
+    cargo install rustfmt;
```
torkleyy: Should be rustfmt-nightly

.travis.yml
```diff
@@ -45,6 +46,16 @@ before_script:
 
 # Generate documentation, compile the engine, run tests.
 script: |
+  # Check if cargo-update is present
+  if ![cargo install-update --version]; then
```
torkleyy: Could simplify to `which appname || install`.

.travis.yml
Xaeroxe: From the rustfmt readme (emphasis mine)

(These instructions use the Syntex version of Rustfmt. If you want to use the nightly version replace install rustfmt with install rustfmt-nightly, **however you must then only run this with the nightly toolchain**)

Meaning this would break our stable and beta builds.

.travis.yml
Xaeroxe: This has complicated semantics with cargo subcommands so I prefer to just check if the command works.

Xaeroxe:
It's already working :D base branch isn't formatted quite right, so this is failing on Travis.  I'll fix the formatting soon.

Xaeroxe:
https://travis-ci.org/amethyst/amethyst/jobs/274710091#L510 this needs to be fixed though, so I'll address that.

Xaeroxe:
@torkleyy r?

.travis.yml
FaultyRAM: We could run it as a separate [build stage](https://docs.travis-ci.com/user/build-stages/). It'd have to run after tests though.

Xaeroxe:
That makes sense to me.  I'll look into what our options are with travis, as I'd still like to keep the nightly rustc as an "allowed failure"

Xaeroxe:
I promise I'll squash all this once I have a working version.

Xaeroxe:
Phew, got it.  I hate yaml.

Xaeroxe:
image

torkleyy:
@Xaeroxe Doesn't seem to work. I don't know such a syntax, but I also don't know why CI failed before.

Xaeroxe:
Yeah I noticed.  I'm trying something new.

torkleyy:
Maybe Travis has some special behavior here. Try extracting it into a script file (which you can then also test locally)?

Xaeroxe:
I'll do that if my latest commit doesn't work.

.travis.yml
```diff
@@ -1,15 +1,25 @@
...
+      - env: "FORMATTING=false"
+        rust: nightly
+  exclude: 
+      - env: "FORMATTING=true"
+        rust: stable
+      - env: "FORMATTING=true"
+        rust: beta
+      - env: "FORMATTING=true"
+        os: osx
+        rust: nightly
```
kvark: this looks rather unclear to me. Is the intention only test the formatting on Linux-nightly? If so, wouldn't it be simpler to just have an "if Linux && nightly { test formatting }" as opposed to encoding that into the build matrix?

Xaeroxe: That is the intention, but the reason it was done this way is because standard Linux nightly is an allowed failure, we don't want formatting failures to be an allowed failure, so a dedicated machine is added for it that is not an allowed failure.

torkleyy: The intention is to create another build machine. Providing two versions of one env var multiplies them, so we have to exclude all the duplicates except one.

kvark: :/ not sure if the confusing config is worth an extra machine here. rustfmt check is supposedly light enough to be done together with some building job

Xaeroxe: That was the original intent, but here's the constraints I've been given:

- We must use rustfmt-nightly, because rustfmt isn't maintained
- Formatting failure must fail the build, otherwise this does us no good
- Failing to build on nightly must not fail the build, as nightly can fail for reasons unrelated to the PR in question

With those constraints this was the solution I came up with.

kvark: Ok, I understand. Is it safe to assume that nightly failures are infrequent? In this case, I'd prefer the rustfmt checks to be failable as well. I don't see a point in strict enforcement of formatting, since failure to format doesn't block anyone from working/testing the product. Hopefully, people just keep an eye on nightly failures and react accordingly. Common sense can save quite a bit of time/nerves from PRs rejected purely on the formatting basis.

Xaeroxe: Then you disagree with the premise of the pull request, which is that formatting of the code base should be enforced.  A constantly formatted code base won't see changes when people run formatting tools on their code, so people can both easily submit well formatted code and not incur additional reviewer work in sorting out the "just formatting" changes.  That's my motivation here.

Xaeroxe:
@torkleyy I'm pretty sure it has to do with yaml interpreting the indentation levels of the script.  I'll move this into its own file.

Xaeroxe:
@kvark I'd hate to act against your wishes, so can I get a yes or no from you on this?

kvark:
@Xaeroxe I abstain

jojolepro:
I approve the change if travis ends with a success. Please note that I looked at the comments, but not at the code. Enforcing rustfmt will make the code cleaner, and will reduce the chances of merges that happens when someone forgets to rustfmt, then in the next commit rustfmt was used on file that have been modified since then.

Xaeroxe:
bors r-

Looks like rustfmt-nightly disagrees with rustfmt

Xaeroxe:
@torkleyy I'm less certain about this change now.  rustfmt-nightly and rustfmt disagree on some things.  I personally don't usually have a nightly compiler installed.  Perhaps we could defer this until rustfmt-nightly works with stable?

Xaeroxe:
And to clarify: I'm fine getting over myself and installing a nightly compiler, but do we want to ask all contributors to do that?

Xaeroxe:
I don't think travis has permission to do that, heck sometimes we don't if they didn't hit the right checkbox while  making the PR.

torkleyy:
I mean push to the Amethyst repo after the PR has been merged. This would require a GH token. But I didn't think about disadvantages too much. Let's continue discussion tomorrow.

Xaeroxe:
This PR isn't terribly useful as is, so I'm going to close it.

PR #349, opened 2017-09-12
Create PULL_REQUEST_TEMPLATE.md

Xaeroxe:
# Purpose

Add a pull request template to Amethyst for new contributors.  Github will automatically populate a PR body with this file when opening a new pull request to Amethyst.

# API changes

No API changes were made, as this is all meta work.

# PR checklist
Write an x in the below boxes if they are true.
- [x] I agree to dual-license my contribution under the MIT license and the Apache License version 2.0
- [ ] I've tested my code on at least one supported platform
- [ ] I ran rustfmt on my code
- [ ] There were breaking changes in my PR so I:
    - [ ] Updated the book to reflect these changes.
    - [ ] Updated the examples to use these changes correctly.
    - [ ] Updated the corresponding doc comments if necessary.
- [ ] I am providing an optimization to amethyst, and I have benchmarks proving it is an improvement.

- [ ] This PR contains platform-specific code (possibly in a dependency) and I have tested it on:
    - [ ] Windows
    - [ ] OSX
    - [ ] Linux

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,22 @@
+# Purpose
+
+<Write your PR's purpose here>
+
+# API changes
+
+<Write the functions and structures you added or removed, if any>
+
+# PR checklist
+Write an x in the below boxes if they are true.
+- [ ] I agree to dual-license my contribution under the MIT license and the Apache License version 2.0
+- [ ] I ran rustfmt on my code
+- [ ] There were breaking changes in my PR so I:
+    - [ ] Updated the book to reflect these changes.
+    - [ ] Updated the examples to use these changes correctly.
+    - [ ] Updated the corresponding doc comments if necessary.
+- [ ] I am providing an optimization to amethyst, and I have benchmarks proving it is an improvement.
+
+- [ ] This PR contains platform-specific code and I have tested it on:
+    - [ ] Windows
+    - [ ] OSX
+    - [ ] Linux
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -9,6 +9,7 @@
 # PR checklist
 Write an x in the below boxes if they are true.
 - [ ] I agree to dual-license my contribution under the MIT license and the Apache License version 2.0
+- [ ] I've tested my code on at least one supported platform
 - [ ] I ran rustfmt on my code
 - [ ] There were breaking changes in my PR so I:
     - [ ] Updated the book to reflect these changes.
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -17,7 +17,7 @@ Write an x in the below boxes if they are true.
     - [ ] Updated the corresponding doc comments if necessary.
 - [ ] I am providing an optimization to amethyst, and I have benchmarks proving it is an improvement.
 
-- [ ] This PR contains platform-specific code and I have tested it on:
+- [ ] This PR contains platform-specific code (possibly in a dependency) and I have tested it on:
     - [ ] Windows
     - [ ] OSX
     - [ ] Linux


jojolepro:
>  This PR contains platform-specific code and I have tested it on:

Shouldn't there also be a case for just testing your changes, even if it doesn't contain platform-specific code?

Xaeroxe:
In order to not bog down the process the checklist is supposed to feel "light" and only include things people are likely to forget.  

Side note: If you haven't even tested your changes on one platform you don't really have any business making a PR to Amethyst.  I expect reviewers to not just trust the person making the PR but they should also test it themselves.

Xaeroxe:
Eh I kind of proved your point with my little rant.  I'll add it.

omni-viral:
If it possible I'd like it to be in comment following PR body where description of the PR should reside.

Xaeroxe:
@omni-viral Why?  That's why I have the "Purpose" header, so the description can go there.

stenverbois:
I feel like `CONTRIBUTING.md` is a more appropriate place for such a checklist. A contributor is already encouraged to read it when submitting PRs and most of the things you mentioned are in there.

A relatively minor but possibly annoying downside of a literal checklist is that Github displays it as "progress" while it is not used that way in this case. This may clutter the overview and distract from PRs that actually have tasks to complete.

image

Xaeroxe:
I like the checkboxes as they can give a quick overview of the contents of the PR, but I don't like the progress bar.  One way to bypass that would be to move the checklist into a comment rather than in the PR body.  I dislike this as it adds more maintenance burden on contributors, but maybe it's the way to go.  What do you guys think?

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
...
+# PR checklist
+Write an x in the below boxes if they are true.
+- [ ] I agree to dual-license my contribution under the MIT license and the Apache License version 2.0
+- [ ] I've tested my code on at least one supported platform
+- [ ] I ran rustfmt on my code
+- [ ] There were breaking changes in my PR so I:
+    - [ ] Updated the book to reflect these changes.
+    - [ ] Updated the examples to use these changes correctly.
+    - [ ] Updated the corresponding doc comments if necessary.
+- [ ] I am providing an optimization to amethyst, and I have benchmarks proving it is an improvement.
```
torkleyy: I really think this is the job of the reviewers. I mean why would you merge an optimization which either isn't proven to be one or isn't obvious. E.g. there's no need to write a benchmark if a call to `.clone()` has been eliminated, but if somebody rewrites an algorithm, it's just common sense to ask for a benchmark, isn't it?

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
...
+<Write your PR's purpose here>
+
+# API changes
+
+<Write the functions and structures you added or removed, if any>
+
+# PR checklist
+Write an x in the below boxes if they are true.
+- [ ] I agree to dual-license my contribution under the MIT license and the Apache License version 2.0
+- [ ] I've tested my code on at least one supported platform
```
torkleyy: I feel like this should be the job of CI / reviewers, too.

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
...
+- [ ] I ran rustfmt on my code
+- [ ] There were breaking changes in my PR so I:
+    - [ ] Updated the book to reflect these changes.
+    - [ ] Updated the examples to use these changes correctly.
+    - [ ] Updated the corresponding doc comments if necessary.
```
torkleyy: Yep, this is a useful remainder (esp. lines 15 and 17).

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
+# Purpose
+
+<Write your PR's purpose here>
+
+# API changes
+
+<Write the functions and structures you added or removed, if any>
```
torkleyy: This should especially be written into the CHANGELOG, but a summary as PR description is indeed nice.

PULL_REQUEST_TEMPLATE.md
Xaeroxe: Fair point.

Xaeroxe:
My only problem with the bot comment is that if the author isn't a member they won't be able to check the boxes.  However this would probably work better as a comment underneath the PR body, so I'll refactor this into that.

PR #352, opened 2017-09-14
Remove documention build, set warnings as errors

Xaeroxe:
By removing the documentation build our travis becomes faster, and by preventing Travis from passing with warnings we keep our code cleaner.


Xaeroxe:
Hm yeah keeping some warnings is good for WIP so I'm closing this too.

PR #353, opened 2017-09-14
Standardize builder naming

Xaeroxe:


Xaeroxe:
bors r+

@torkleyy I couldn't find any methods that borrow the Builder mutably, otherwise I would have done that.

torkleyy:
Well the idea was to add those, because we don't have them yet.

omni-viral:
@torkleyy for each moving `with_` add mutating `add_` method?

jojolepro:
What about ::default and ::new?
Time::default()
InputHandler::new()
LocalTransform::default()
Transform::default()
TransformSystem::new()
pass::DrawFlat::<PosNormTex>::new()
Dj::new(&output)
MaterialBuilder::new()
MeshComponent::new()
BoxedErr::new

Xaeroxe:
@jojolepro Those are interesting and probably deserve to be addressed, but they're out of the scope of this pull request

Xaeroxe:
Overriding.

Issue #229, opened 2017-04-17
Provide some sort of event system


torkleyy:
A more general purpose event system would be nice, probably by changing `handle_events` to receive not just window events.

I'd need this for asset management, where the `State` will be informed when an `AssetSet` has finished.

vitvakatu:
I can work on it. Do you have some ideas?
Or maybe you want to do it yourself?

torkleyy:
@vitvakatu Sure, you can do it, if you want. I hoped to have some discussion about what it should offer. For example, do we want a fixed set of events (enum) or unlimited?

vitvakatu:
I suppose, unlimited (user defined) set of events is better. Another question, could we implement it in pretty way.

torkleyy:
OTOH why would the user emit an event if the only receiver is the active state?

Lionex:
[Looking back at the inspiration,](http://gameprogrammingpatterns.com/event-queue.html) it seems prudent to provide a way to create new event queues and listeners for things like the audio system, or potentially a tutorial system, combat system, etc.

vitvakatu:
Hmm... @torkleyy you're write. Messages (or events, it's the same) are belong to OOP, it's not the Rust way. We should have specs and our state system in mind while planning the event system.

OvermindDL1:
As an example, since I keep popping up here and want amethyst to eventually replace my custom-made C++ ECS engine, I guess I should state how my events worked.  :-)

First the gist, the event system had a couple of parts, but the same basic interface for it all, and it was designed to be both very efficient (some paths could even be inlined by the compiler for static knowledge) as well as easy to use via scripting interfaces.

Internally it had a few parts, the first of which was the static dispatching, these were for performance sensitive parts used by things like collision, this was fast, low level, but could still be interacted through via the generic dynamic interfaces.  The rest was built on top of the static dispatching with a generic path for generic events that were entirely arbitrary and user defined, the interface through this part was the same no matter the event type, which was to send:
```cpp
context.sendEvent(Atom eventName, VariantMap &eventData);
// -- or --
context.sendEvent(NodePtr node, Atom eventName, VariantMap &eventData);
// -- or --
context.sendEvent(EntityID entity, Atom eventName, VariantMap &eventData);
```
Callback functions were always of the form:
```cpp
void callbackFunction(Atom eventName, VariantMap &eventData);
```
And registration would be something like:
```cpp
[...]

PR #355, opened 2017-09-15
Add events resource and event enum

Xaeroxe:
Fixes https://github.com/amethyst/amethyst/issues/229

Note: I opened https://github.com/amethyst/amethyst/pull/354 before but that was when my branch was malformed.  This one is correct.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -2,6 +2,7 @@
 
 use std::sync::Arc;
 
+use event::{Event, Events};
 use input::InputHandler;
 use rayon::ThreadPool;
 use renderer::Config as DisplayConfig;
@@ -9,7 +10,7 @@ use renderer::PipelineBuilder;
 use shred::{Resource, ResourceId};
 #[cfg(feature = "profiler")]
 use thread_profiler::{register_thread_with_profiler, write_profile};
-use winit::{Event, EventsLoop};
+use winit::{Event as WinitEvent, EventsLoop};
 
 use assets::{Asset, Loader, Store};
 use ecs::{Component, Dispatcher, DispatcherBuilder, System, World};
@@ -20,15 +21,13 @@ use timing::{Stopwatch, Time};
 
 /// User-friendly facade for building games. Manages main loop.
 ///
-
 #[derive(Derivative)]
 #[derivative(Debug)]
 pub struct Application<'a, 'b> {
     /// The `engine` struct, holding world and thread pool.
     #[derivative(Debug = "ignore")]
     pub engine: Engine,
-
-    #[derivative(Debug = "ignore")]
+   #[derivative(Debug = "ignore")]
     dispatcher: Dispatcher<'a, 'b>,
     #[derivative(Debug = "ignore")]
     events: EventsLoop,
@@ -57,7 +56,7 @@ impl<'a, 'b> Application<'a, 'b> {
     /// # Type Parameters
     ///
     /// - `S`: A type that implements the `State` trait. e.g. Your initial
-    ///        game logic. 
+    ///        game logic.
     ///
     /// # Lifetimes
     ///
@@ -76,9 +75,9 @@ impl<'a, 'b> Application<'a, 'b> {
     /// use amethyst::prelude::*;
     ///
     /// struct NullState;
-    /// 
+    ///
     /// impl State for NullState {}
-    /// 
+    ///
     /// fn main() {
     ///     let mut game = Application::new(NullState).expect("Failed to initialize");
     ///     game.run();
@@ -86,7 +85,7 @@ impl<'a, 'b> Application<'a, 'b> {
     /// ~~~
     pub fn new<S>(initial_state: S) -> Result<Application<'a, 'b>>
     where
-        S: State + 'a
+        S: State + 'a,
     {
         ApplicationBuilder::new(initial_state)?.build()
     }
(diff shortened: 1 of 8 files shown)


src/event.rs
```diff
@@ -1,4 +1,18 @@
...
-pub use winit::{ElementState, Event, KeyboardInput, MouseButton, VirtualKeyCode, WindowEvent};
+pub use winit::{ElementState, Event as WinitEvent, KeyboardInput, MouseButton, VirtualKeyCode,
+                WindowEvent};
+
+/// Events that the engine might send to the user.
+pub enum Event {
+    /// An event from the Winit crate
+    WinitEvent(WinitEvent),
+    /// Unused event, enables _ in match statements
+    UnusedEvent,
```
torkleyy: I don't get why we need this variant.

src/event.rs
```diff
@@ -1,4 +1,18 @@
...
 //! types.
 
-pub use winit::{ElementState, Event, KeyboardInput, MouseButton, VirtualKeyCode, WindowEvent};
+pub use winit::{ElementState, Event as WinitEvent, KeyboardInput, MouseButton, VirtualKeyCode,
+                WindowEvent};
+
+/// Events that the engine might send to the user.
+pub enum Event {
+    /// An event from the Winit crate
+    WinitEvent(WinitEvent),
```
torkleyy: Should we make this implementation detail part of the public API?

src/event.rs
omni-viral: As I understand it there is only one variant for now. And @Xaeroxe use `_` arm in matches so those matches will stay valid when new variants will be added here.

omni-viral: Although I was thinking `_` arm could be added even if match have arm for all variants.

src/event.rs
Xaeroxe: We already have actually.  Our events thus far have come from Winit.  Now the next logical question is should we leave it that way?  I think so personally, as reimplementing and converting the events could be a pain and cost us cycles unnecessarily.  Winit is also a cross platform library that should work for every use case we have thus far.

src/event.rs
omni-viral: Well. I've checked. It is possible but causes warning.

src/event.rs
torkleyy: I still see some issues. E.g. the hierarchy is deeper.

src/event.rs
torkleyy: Well we should make this enum future proof and add a `__NonExhaustive` variant instead.

src/event.rs
Xaeroxe: Yeah the hierarchy is getting kind of excessive at this point.  Perhaps I can see what I can do about that.

src/event.rs
Xaeroxe: Huh, I actually didn't know about `__NonExhaustive` thanks!

stenverbois: It does not matter what it is called as it is not a standard feature. `__Nonexhaustive` is a std-only hack. The best we can do at the moment is a `#[doc(hidden)]` variant.

[#[non_exhaustive] is on its way](https://github.com/rust-lang/rust/issues/44109)

torkleyy: @stenvorbis Yeah I don't really care how it's called, but prefixing with `__` makes sense and this name is somehow a known pattern. Ye, we should have `#[doc(hidden)]`

src/app.rs
```diff
@@ -163,10 +163,18 @@ impl<'a, 'b> Application<'a, 'b> {
                 )
                 {
                     let mut input = engine.world.write_resource::<InputHandler>();
-                    if let Event::WindowEvent { ref event, .. } = event {
-                        input.send_event(&event);
+                    if let WinitEvent::WindowEvent { ref event, .. } = event {
+                        input.send_event(event);
                     }
                 }
+                let event = match event {
```
torkleyy: I'd actually make the hierarchy even flatter, but it's definitely an improvement.

src/event.rs
```diff
@@ -1,4 +1,23 @@
...
+#[derive(Clone)]
+pub enum Event {
+    /// A WindowEvent from the winit crate
+    WindowEvent(WindowEvent),
+    /// A DeviceEvent from the winit crate
+    DeviceEvent(DeviceId, DeviceEvent),
+    /// Awakened event from the winit crate
+    Awakened,
+    // Add more events here as needed.
+    #[doc(hidden)]
```
torkleyy: Could we please have docs before attributes? I think that's most used order in this crate and also the one I prefer.

src/event.rs
```diff
@@ -1,4 +1,23 @@
-//! This module contains the `WindowEvent` type and re-exports glutin event
+//! This module contains the `Event` enum and re-exports glutin event
 //! types.
 
-pub use winit::{ElementState, Event, KeyboardInput, MouseButton, VirtualKeyCode, WindowEvent};
+pub use winit::{DeviceEvent, DeviceId, ElementState, KeyboardInput, MouseButton, VirtualKeyCode, WindowEvent};
+
+/// Events that the engine might send to the user.
```
LucioFranco: Do we want to either give a bit more detail on what is within `WindowEvent` or link to the winit crate? Might make it easier for newer people to figure out how to use the input system.

Xaeroxe: Input system is managed automatically by the amethyst crate, so users aren't required to pump events themselves.  However as @torkleyy said I think I'm going to flatten this hierarchy anyways.

Xaeroxe:
This is quite inferior to an approach using https://github.com/Rhuagh/shrev-rs so I'm going to close this and rewrite it with that later.

PR #359, opened 2017-09-15
Small fixes to 03_renderable and 04_pong examples [non breaking]

nchashch:
* Get rid of magic numbers in `ExampleSystem::run` method in 03_renderable
* Rename `Plank` to `Paddle` in 04_pong

### DIFF
--- a/examples/04_pong/main.rs
+++ b/examples/04_pong/main.rs
@@ -50,16 +50,16 @@ enum Side {
     Right,
 }
 
-struct Plank {
+struct Paddle {
     pub position: f32,
     pub velocity: f32,
     pub dimensions: [f32; 2],
     pub side: Side,
 }
 
-impl Plank {
-    pub fn new(side: Side) -> Plank {
-        Plank {
+impl Paddle {
+    pub fn new(side: Side) -> Paddle {
+        Paddle {
             position: 0.,
             velocity: 1.,
             dimensions: [1., 1.],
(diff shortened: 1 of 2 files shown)


examples/04_pong/main.rs
```diff
@@ -68,8 +68,8 @@ impl Plank {
     }
 }
 
-impl Component for Plank {
-    type Storage = VecStorage<Plank>;
+impl Component for Paddle {
+    type Storage = VecStorage<Paddle>;
```
torkleyy: While we're at it, maybe we could use another storage?

nchashch: Like `DenseVecStorage`?

Rhuagh: Yes

nchashch: I don't think it will make any difference. Because there are just 2 paddles with `VecStorage` there will be just a `Vec` of length 2, which is ok.

Maybe merge this, and if I missed something @torkleyy will create another PR changing the storage?

nchashch:
Merging this as there are 4 approvals.

examples/04_pong/main.rs
torkleyy: It doesn't make a difference in this particukar case. But we should not use `VecStorage` in our example, because I already see people using it as the default even though it can be very heavy if used incorrectly.

torkleyy:
Please don't merge manually, but use bors for it in the future.

examples/04_pong/main.rs
jojolepro: I might have only used VecStorage until now? :thinking:

PR #364, opened 2017-09-17
Add ECS bundle concept

Rhuagh:
Opening this early to get feedback. Is this a viable concept or should I trash it immediately?


jojolepro:
@Rhuagh Basically they are code wrappers to make the code less heavy and reduce the boiler plate?
I'd like to know how that handles conflicts between dependencies and component registration too.

Rhuagh:
Well, they're a generification of with_renderer in the ApplicationBuilder, so instead of adding functions there for every single concept (with_audio, with_transform), we can add bundles instead.

Conflicts are handled by specs for the most part. Resources are overridden, component registration is essentially a NOP for any duplicate registration. System registration is a panic if it already exists. When it comes to dependencies, each bundle handles that on it's own. Look at the TransformBundle for example, it has support for dependencies.

Rhuagh:
An alternative would be to have the build() function take the ApplicationBuilder as argument and return it wrapped in a Result.

src/app.rs
```diff
@@ -569,6 +570,38 @@ impl<'a, 'b, T: State + 'a> ApplicationBuilder<'a, 'b, T> {
...
+    ///
+    /// # Errors
+    ///
+    /// This function creates systems and resources, which use any number of dependent
+    /// crates or APIs, which could result in any number of errors.
+    /// See each individual bundle for a description of the errors it could produce.
+    ///
+    pub fn with_bundle<B>(mut self, bundle: B) -> Result<Self>
+    where
+        B: for<'c> ECSBundle<'a, 'b, (&'c EventsLoop)>,
```
omni-viral: RenderSystem is special as it the only one that need access to `EventsLoop` during building.
What should we do if there will be new bundle that will require some other data from `Application` to build?

omni-viral: Parentheses are unnecessary. If you want to make a tuple with single element you need to make it like this `(A,)`

Rhuagh: Yeah, it's not perfect, but I haven't figured out a good solution to that yet.

Rhuagh: Updated, the bundle trait now works on the ApplicationBuilder directly.

Rhuagh:
Refactored to use ApplicationBuilder directly.

src/ecs/mod.rs
```diff
@@ -7,14 +7,15 @@ pub mod transform;
...
-/// Extension trait for all systems.
-pub trait SystemExt<'a, A>: System<'a> {
-    /// Constructs a new system with the given arguments.
-    /// Registers all supported components with the World.
-    /// Puts resources into World.
-    fn build(args: A, world: &mut World) -> Result<Self>
-    where
-        Self: Sized;
+/// Describes a bundle of ECS components, resources and systems
+pub trait ECSBundle<'a, 'b, T: ::state::State + 'a> {
```
omni-viral: This bound shouldn't be here. Please, remove it from `ApplicationBuilder`'s functions where `T: State` is not required. I believe the only function that requires `T: State` is `ApplicationBuilder::build`

Rhuagh: Well, ApplicationBuilder requires it at the moment, and that's a separate change from this PR imho.

Rhuagh: I fixed it.

Xaeroxe:
Looks good so far!  What would you consider the remaining tasks on this PR?

Rhuagh:
A bit of cleanup, and making a bundle for the pong  stuff. And more documentation

Rhuagh:
Ready to go after rereview.

Xaeroxe:
Override.

src/app.rs
```diff
@@ -585,65 +585,15 @@ impl<'a, 'b, T: State + 'a> ApplicationBuilder<'a, 'b, T> {
...
-        use ecs::SystemExt;
-        use ecs::rendering::RenderSystem;
-        let render_sys = RenderSystem::build((&self.events, pipeline, config), &mut self.world)?;
-        Ok(
-            self.with_thread_local(render_sys)
-                .register_mesh_asset()
-                .register_texture_asset()
-                .register_material_not_yet_asset(),
-        )
+    pub fn with_bundle<B>(self, bundle: B) -> Result<Self>
```
torkleyy: If a bundle offers multiple systems, how would one manage dependencies? Should maybe the documentation list the system names it adds?

src/ecs/audio/bundle.rs
```diff
@@ -0,0 +1,71 @@
...
+        &self,
+        mut builder: ApplicationBuilder<'a, 'b, T>,
+    ) -> Result<ApplicationBuilder<'a, 'b, T>> {
+        // Remove option here when specs get support for optional fetch in
+        // released version
+        if !builder.world.res.has_value(
+            ResourceId::new::<Option<Output>>(),
+        )
+        {
+            builder = builder.with_resource(default_output());
```
torkleyy: We should really have an entry API here.

src/ecs/audio/bundle.rs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        let dj = match *builder.world.read_resource::<Option<Output>>() {
```
torkleyy: Should use `map`.

src/ecs/audio/bundle.rs
```diff
@@ -0,0 +1,71 @@
...
+            Some(ref audio_output) => Some(Dj::new(&audio_output)),
+
+            None => None,
+        };
+
+        if let Some(dj) = dj {
+            builder = builder.with_resource(dj).with(
```
torkleyy: That's where we need `add_*`.

src/ecs/input/bundle.rs
```diff
@@ -0,0 +1,48 @@
...
+
+use app::ApplicationBuilder;
+use config::Config;
+use ecs::ECSBundle;
+use ecs::input::{Bindings, InputHandler};
+use error::Result;
+
+/// Bundle for adding the InputHandler and input bindings
+///
+/// # Errors
```
torkleyy: Should be `## Errors`

src/ecs/input/bundle.rs
```diff
@@ -0,0 +1,48 @@
+//! ECS input bundle
+
+use app::ApplicationBuilder;
+use config::Config;
+use ecs::ECSBundle;
+use ecs::input::{Bindings, InputHandler};
+use error::Result;
+
+/// Bundle for adding the InputHandler and input bindings
```
torkleyy: Should be `InputHandler`.

src/ecs/input/bundle.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+impl InputBundle {
+    /// Create a new input bundle with no bindings
+    pub fn new() -> Self {
+        Self { bindings: None }
+    }
+
+    /// Use the provided bindings with the InputHandler
+    pub fn with_bindings(mut self, bindings: Bindings) -> Self {
```
torkleyy: For the sake of a flatter syntax, could we have `new(Bindings)` and `from_file` instead? I wonder if we'd want no bindings at all.

src/ecs/input/bundle.rs
```diff
@@ -0,0 +1,48 @@
...
+        self.bindings = Some(bindings);
+        self
+    }
+
+    /// Load bindings from file
+    pub fn with_bindings_from_file(self, file: &str) -> Self {
```
torkleyy: Should be using path API.

src/ecs/rendering/bundle.rs
```diff
@@ -0,0 +1,127 @@
...
+use ecs::rendering::resources::{Factory, AmbientColor};
+use ecs::rendering::systems::RenderSystem;
+use ecs::transform::components::*;
+use error::{Error, Result};
+
+/// Rendering bundle
+///
+/// Will register all necessary components needed for rendering, along with any resources.
+/// Will also register asset contexts with the asset Loader, and add systems for merging
+/// AssetFutures into their respective component.
```
torkleyy: Should be `AssetFuture`.

src/ecs/rendering/bundle.rs
```diff
@@ -0,0 +1,127 @@
...
+///
+/// Will add RenderSystem as a thread local system.
```
torkleyy: Should be `RenderSystem`

src/ecs/rendering/bundle.rs
```diff
@@ -0,0 +1,127 @@
...
+///
+/// # Errors
```
torkleyy: Should be `## Errors`

src/ecs/rendering/bundle.rs
```diff
@@ -0,0 +1,127 @@
...
+///
+/// Returns errors related to:
+///
+/// * Renderer creation
```
torkleyy: :+1:

src/ecs/rendering/bundle.rs
```diff
@@ -0,0 +1,127 @@
...
+            .with_resource(Factory::new())
+            .with_resource(cam)
+            .with_resource(AmbientColor(Rgba::from([0.01; 3])))
+            .register::<Transform>()
+            .register::<LightComponent>()
+            .register::<MaterialComponent>()
+            .register::<MeshComponent>()
+            .register::<TextureComponent>();
+
+        // asset stuff, enable/disable flag for this?
```
torkleyy: Please add a `FIXME` or `TODO` for these comments.

src/ecs/rendering/mod.rs
```diff
@@ -1,9 +1,11 @@
 //! `amethyst` rendering ecs module
 
+pub use self::bundle::RenderBundle;
```
torkleyy: I think we should decide on whether we want to have a public module or reexported items. Given these bundles all have the same postfix, a conflict is impossible. I just see this inconsistency existed before the PR :worried:

src/ecs/transform/bundle.rs
```diff
@@ -0,0 +1,52 @@
...
+use ecs::ECSBundle;
+use ecs::transform::*;
+use error::Result;
+
+/// Transform bundle
+///
+/// Will register transform components, and the TransformSystem.
+/// TransformSystem will be registered with name "transform_system".
+///
+/// # Errors
```
torkleyy: Same as above

src/ecs/transform/bundle.rs
```diff
@@ -0,0 +1,52 @@
...
+///
+/// No errors will be returned by this bundle.
+///
+/// # Panics
```
torkleyy: Should be `## Panics`

src/ecs/transform/bundle.rs
```diff
@@ -0,0 +1,52 @@
...
+///
+/// Panics in TransformSystem registration if the bundle is applied twice in the same dispatcher.
```
torkleyy: `TransformSystem`

src/ecs/transform/bundle.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+impl<'a> TransformBundle<'a> {
+    /// Create a new transform bundle
+    pub fn new() -> Self {
+        Self { dep: &[] }
+    }
+
+    /// Set dependencies for the TransformSystem
+    pub fn with_dep(mut self, dep: &'a [&'a str]) -> Self {
```
torkleyy: This is something that makes the new bundle system more complex and gives the user less control. Also see my comment above. But well, let's see how it works out.

src/ecs/transform/bundle.rs
Xaeroxe: That's what he wrote?

src/ecs/rendering/bundle.rs
Xaeroxe: That's what he wrote?

src/ecs/input/bundle.rs
Xaeroxe: That's what he wrote?

src/ecs/transform/bundle.rs
torkleyy: Yep, just without backticks.

src/ecs/rendering/bundle.rs
omni-viral: Maybe he means absence of grave acent "`"?

src/app.rs
Rhuagh: I'd put all the information that the user could need in the documentation for each bundle.

src/ecs/audio/bundle.rs
Rhuagh: Yes.

src/ecs/audio/bundle.rs
Rhuagh: Yep.

src/ecs/audio/bundle.rs
Rhuagh: Not sure I follow.

Rhuagh: Got it now.

src/ecs/input/bundle.rs
Rhuagh: I just followed what @csherratt has done with the ApplicationBuilder etc.

src/ecs/input/bundle.rs
Rhuagh: From what I've understood, the InputHandler only use bindings for actions/axes, but the user can also access the internal storages directly for a lot of things, so you aren't required to use bindings, which is why I made that a separate function.

src/ecs/transform/bundle.rs
Rhuagh: You are not forced to use the bundle system, you can do all it does on it's own, and the code is open. 
I feel this is a way of lowering the threshold to start using the engine, without giving anything up really.

PR #368, opened 2017-09-19
Remove OSX from travis builds.

Xaeroxe:
Travis can take in excess of 5 hours to start OSX builds and most PRs aren't likely to break OSX builds specifically, so I propose we remove this from our travis and just make sure we test on mac ourselves if we're worried about a PR breaking OSX.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,10 +20,9 @@ branches:
     - master
     - /release-.*/
 
-# Test on both Linux and OS X.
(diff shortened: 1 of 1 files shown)


Xaeroxe:
Ironically this won't start building because our build queue is backlogged because we're waiting for OSX machines.

Xaeroxe:
bors r+

It hasn't been 24 hours but we have a pretty good consensus from active members and this is trivial to revert if anyone has any problems, so I'm going to merge it.

Xaeroxe:
Overriding.

PR #371, opened 2017-09-19
Made it possible to define external texture formats

jojolepro:
Just setting the raw image data to pub instead of pub(crate) allows external users to define custom formats and makes my life easier to test the new TTF loader and format.

This change would be temporary as @Xaeroxe wants to make a more complex wrapper. But, because it is clearly no a priority, this allows the feature to be exposed with minimal effort.

You if don't want this included, refuse the pr and close it.

Thanks!


Xaeroxe:
So to be clear: I changed my mind about this idea, I think it's an improvement I just never got around to making it.  Thanks!

jojolepro:
yay

jojolepro:
Thanks!

PR #372, opened 2017-09-19
Fix horizontally flipped textures

Xaeroxe:
Fixes an issue reported by @jojolepro

### DIFF
--- a/amethyst_renderer/src/pass/shaders/vertex/basic.glsl
+++ b/amethyst_renderer/src/pass/shaders/vertex/basic.glsl
@@ -24,6 +24,7 @@ void main() {
     vertex.position = model * vec4(position, 1.0);
     vertex.normal = mat3(model) * normal;
     vertex.tangent = mat3(model) * tangent;
+    // Horizontally flip the texture to compensate for OpenGL's inverted V axis.
     vertex.tex_coord = vec2(tex_coord.x, 1.0 - tex_coord.y);
     gl_Position = proj * view * vertex.position;
 }
--- a/amethyst_renderer/src/pass/shaders/vertex/basic.glsl
+++ b/amethyst_renderer/src/pass/shaders/vertex/basic.glsl
@@ -24,6 +24,6 @@ void main() {
     vertex.position = model * vec4(position, 1.0);
     vertex.normal = mat3(model) * normal;
     vertex.tangent = mat3(model) * tangent;
-    vertex.tex_coord = tex_coord;
+    vertex.tex_coord = vec2(tex_coord.x, 1.0 - tex_coord.y);
     gl_Position = proj * view * vertex.position;
 }


jojolepro:
yay !

OvermindDL1:
I'm not so sure about adding that work there.  The better question might be 'why' is it flipped somewhere?

Xaeroxe:
@OvermindDL1 This is an OpenGL specific problem, so it's getting an OpenGL specific fix.  This solution has been pretty extensively discussed on gitter and was even the recommended workaround from a member of gfx.

Xaeroxe:
The primary problem is that OpenGL uses a different UV coordinate system than every other backend there is.

amethyst_renderer/src/pass/shaders/vertex/basic.glsl
```diff
@@ -24,6 +24,6 @@ void main() {
     vertex.position = model * vec4(position, 1.0);
     vertex.normal = mat3(model) * normal;
     vertex.tangent = mat3(model) * tangent;
-    vertex.tex_coord = tex_coord;
+    vertex.tex_coord = vec2(tex_coord.x, 1.0 - tex_coord.y);
```
omni-viral: Please add a comment about why this line looks like this.

jojolepro:
Wait, I thought it was vertically flipped?

Xaeroxe:
@jojolepro, I think we had our terminology wrong on this from the start, you reported upside down textures which would mean they were horizontally flipped.

Xaeroxe:
@torkleyy @OvermindDL1 So to summarize here is the scenario:  OpenGL uses an inverted V axis when compared to other graphics backends.  We want to flip the displayed image to make it consistent with other backends.  There's two ways to do this that I know of:

Option 1:
Load images upside down with the asset loader
Pros:
- Don't have to add another floating operation to the GPU
Cons:
- Downstream custom image loaders such as the one @jojolepro made have to load images horizontally flipped but if and only if they are using the opengl feature for amethyst.

Option 2:
Tell OpenGL to flip images horizontally using shader code (This PR in its current state)
Pros:
- Same underlying image data works for all backends
- Custom loaders don't have to implement weird backend specific logic.
Cons:
- Adds another floating point operation to the GPU
- Won't work for custom shaders

Personally I don't consider the custom shaders problem to be a dilemma, because shaders written in GLSL should be expecting the V axis to be inverted, so if we feed the data to those shaders horizontally flipped it then becomes we who are in error.

I also don't really consider introducing an extra floating point operation to OpenGL to be a major concern, as high performance applications should be using next-gen graphics APIs such as Vulkan, DX12, or Metal.

jojolepro:
I'm for option 2

Xaeroxe:
Ok I think I just thought of a way to make everyone happy.

Option 3:

Update the code that sends the image to the GPU to invert the image just prior to sending to the GPU, only if the opengl feature is enabled.

This occurred to me because Option 2 isn't correct for shaders either.  We're already passing GLSL bad data, but we don't want to complicate loader logic, so the solution is to horizontally flip the image as it's being transferred to the GPU.

Xaeroxe:
I'll close this PR now since the commits made in it aren't going to be merged.

OvermindDL1:
> This occurred to me because Option 2 isn't correct for shaders either. We're already passing GLSL bad data, but we don't want to complicate loader logic, so the solution is to horizontally flip the image as it's being transferred to the GPU.

This is what was really bugging me about it pretty hard.  Shaders should not be used for such fixups.

Also, isn't OpenGL correct in how it does it?  Doesn't Vulkan do the same (at least from my C++ testing of vulkan)?  I have no clue about DirectX but I do know DirectX is not as mathematically correct as OpenGL so it would not surprise me if it was quite stupid on things...

Xaeroxe:
I'm not going to dispute what's the "correct" orientation of the V axis because to be frank I don't care, all I know is OpenGL does it different from all the other backends so it's easiest to patch OpenGL.  Vulkan uses the same V orientation as the other backends: https://github.com/gfx-rs/gfx/tree/master/src/backend/vulkan

OvermindDL1:
https://matthewwellings.com/blog/the-new-vulkan-coordinate-system/

Blah, that bugs the crap out of me, it is *wrong*.  Who was the idiot of a moron that thought going *down* with positive numbers was correct?!?

omni-viral:
@OvermindDL1 what the mathematical law says that up is positive? 😏

Rhuagh:
> Blah, that bugs the crap out of me, it is wrong. Who was the idiot of a moron in the standard body that thought going down with positive numbers was correct?!?

They did it that way, because _all_ other backends do it that way, except for GL.

msiglreith:
> They did it that way, because all other backends do it that way, except for GL.

Only Vulkan uses a left-handed coordinate system for NDCs. All others use a right-handed (including GL). GL differs in UV coordinate system and depth-range ([-1,1] instead of the common and better [0, 1]).

msiglreith:
Regarding the issue, I would still advocate to flip only texture coordinates used for accessing render target access.

Rhuagh:
> Only Vulkan uses a left-handed coordinate system for NDCs

Oh, _that_ I don't understand.

torkleyy:
@msiglreith Maybe texture coordinates could be flipped on upload?

OvermindDL1:
> @OvermindDL1 what the mathematical law says that up is positive? :smirk:

Not a law, but a standard, look at any 2d+ axis (untransformed).

jojolepro:
If I jump, my height would decrease :thinking: Nice logic.

OvermindDL1:
> If I jump, my height would decrease :thinking: Nice logic.

How would that happen?  Wouldn't the height of your character be based off of its bottom point within object local-space?

Xaeroxe:
I think height in this context meant "my character's y position"

msiglreith:
> Maybe texture coordinates could be flipped on upload?

Too much preprocessing overhead involved imo.

Xaeroxe:
@msiglreith Do we not introduce the same overhead to the GPU by doing it per draw call?

msiglreith:
@Xaeroxe I would argue that the overhead of a single VALU instruction is negligible atm. Flipping textures and texture coordinates on the CPU is probably more noticable and complex.

Rhuagh:
I world argue that loading times are just as important as fps.

Issue #357, opened 2017-09-15
Transparency not implemented


jojolepro:
The transparency is simply not implemented.

Please fix. :)

Xaeroxe:
Partial-transparency should be supported too.  It's quite common in videogames.

jojolepro:
Of course :)

PR #373, opened 2017-09-20
Add blending for transparency

Xaeroxe:
This PR updates the amethyst renderer blending algorithm to permit transparency.  It also removes a couple warning from amethyst_renderer in regards to unnecessary mutability of self parameters.

Fixes https://github.com/amethyst/amethyst/issues/357

### DIFF
--- a/amethyst_renderer/src/pipe/effect/mod.rs
+++ b/amethyst_renderer/src/pipe/effect/mod.rs
@@ -13,7 +13,7 @@ use gfx::preset::depth::{LESS_EQUAL_TEST, LESS_EQUAL_WRITE};
 use gfx::pso::buffer::{ElemStride, InstanceRate};
 use gfx::shade::{ProgramError, ToUniform};
 use gfx::shade::core::UniformValue;
-use gfx::state::{Rasterizer, Stencil};
+use gfx::state::{Blend, BlendValue, ColorMask, Equation, Factor, Rasterizer, Stencil};
 use gfx::traits::Pod;
 
 use self::pso::{Data, Init, Meta};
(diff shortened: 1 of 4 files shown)


Xaeroxe:
cc @msiglreith if you wouldn't mind taking the time to review this.

amethyst_renderer/src/pipe/effect/mod.rs
```diff
@@ -226,7 +226,7 @@ impl<'a> EffectBuilder<'a> {
                 Stencil::default(),
             ));
         }
-        self.init.out_colors.push(name);
+        self.init.out_colors.push((name, ColorMask::all(), Blend::new(Equation::Add, Factor::One, Factor::OneMinus(BlendValue::SourceAlpha))));
```
omni-viral: I'm not sure about this blend equation

omni-viral: Aren't sum of `Factors` should be 1?

msiglreith: You can use some of our existing [presets](https://docs.rs/draw_state/0.7.1/draw_state/preset/blend/constant.ALPHA.html).
Ideally, let the user (implementer of the graphics passes) manually set the blending settings. Similar to the depth settings above.

omni-viral:
As it was discussed on gitter we need separete passes for opaque and transparent objects. Pass for transparent objects have to draw them in sorted order.

amethyst_renderer/src/pipe/effect/mod.rs
omni-viral: It should be possible for custom passes. Passes for simple transparency should use specific blend function

Xaeroxe:
Well whatever we're supposed to do to accomplish this, this isn't it.  So I'll close this PR and open it again when I have a working solution.

PR #374, opened 2017-09-20
Clear textures and samplers from data for all passes

omni-viral:
`textures` and `samplers` vectors from `pso::Data` should be cleared for each pass

### DIFF
--- a/amethyst_renderer/src/pipe/effect/mod.rs
+++ b/amethyst_renderer/src/pipe/effect/mod.rs
@@ -100,7 +100,7 @@ impl Effect {
     }
 
     /// FIXME: Update raw buffer without transmute, use `Result` somehow.
-    pub fn update_buffer<N, T>(&self, name: N, data: &[T], enc: &mut Encoder)
+    pub fn update_buffer<N, T>(&mut self, name: N, data: &[T], enc: &mut Encoder)
     where
         N: AsRef<str>,
         T: Pod,
@@ -115,7 +115,7 @@ impl Effect {
     }
 
     /// FIXME: Update raw buffer without transmute.
-    pub fn update_constant_buffer<N, T>(&self, name: N, data: &T, enc: &mut Encoder)
+    pub fn update_constant_buffer<N, T>(&mut self, name: N, data: &T, enc: &mut Encoder)
     where
         N: AsRef<str>,
         T: Copy,
@@ -126,6 +126,11 @@ impl Effect {
         }
         // FIXME: Don't silently ignore unknown update.
         // maybe `.expect(...)` would fit here
+    }
+
+    pub fn clear(&mut self) {
+        self.data.textures.clear();
+        self.data.samplers.clear();
     }
 
     pub fn draw(&self, mesh: &Mesh, enc: &mut Encoder) {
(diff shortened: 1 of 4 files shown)


amethyst_renderer/src/pipe/effect/mod.rs
```diff
@@ -100,7 +100,7 @@ impl Effect {
     }
 
     /// FIXME: Update raw buffer without transmute, use `Result` somehow.
-    pub fn update_buffer<N, T>(&self, name: N, data: &[T], enc: &mut Encoder)
+    pub fn update_buffer<N, T>(&mut self, name: N, data: &[T], enc: &mut Encoder)
```
Xaeroxe: Why is mut necessary here?

amethyst_renderer/src/pipe/effect/mod.rs
```diff
@@ -115,7 +115,7 @@ impl Effect {
     }
 
     /// FIXME: Update raw buffer without transmute.
-    pub fn update_constant_buffer<N, T>(&self, name: N, data: &T, enc: &mut Encoder)
+    pub fn update_constant_buffer<N, T>(&mut self, name: N, data: &T, enc: &mut Encoder)
```
Xaeroxe: Same as above.

omni-viral: I know it isn't required by compiler but you don't want to update the same constant buffer simultaneously.

omni-viral: I know it isn't required by compiler but you don't want to update the same constant buffer simultaneously.

amethyst_renderer/src/pipe/effect/mod.rs
omni-viral: Same as below

amethyst_renderer/src/pipe/effect/mod.rs
Xaeroxe: 👍 Gotcha thanks

amethyst_renderer/src/pipe/effect/mod.rs
```diff
@@ -100,7 +100,7 @@ impl Effect {
     }
 
     /// FIXME: Update raw buffer without transmute, use `Result` somehow.
```
jojolepro: I don't know if its important, but it seems like a potential optimisation

amethyst_renderer/src/pipe/effect/mod.rs
```diff
@@ -115,7 +115,7 @@ impl Effect {
     }
 
     /// FIXME: Update raw buffer without transmute.
```
jojolepro: same

amethyst_renderer/src/pipe/effect/mod.rs
omni-viral: How exactly this is optimisation?

jojolepro: I didn't knew what transmute was, I just assume from the way the comment was written that it was a bad thing. :)

Xaeroxe: @jojolepro It's extremely dangerous to use, but if used correctly it can be quite helpful.

Xaeroxe:
bors r+

It's a few hours early but I doubt anyone is actually opposed to this.

PR #375, opened 2017-09-20
Use serde to (de)serialize entities

omni-viral:
Add (de)serialization support for entities to Amethyst.
Main parts are:
* `Marker` trair.
`Marker` is used for several purposes:
  * Mark which entities should be serialized.
  * It contains ID to replace `Entity` stored in `Component`s during serialization process.
  * It may contain additional data to be used for tracking owner of the component
(future `NetMarker` 🌍)
* `MarkerAllocator` trait for allocating `Marker`s, storing `Marker -> Entity` mapping and other stuff.
* `SerializableComponent` trait that is auto implemented by components if they implements `Serialize`, `DeserializeOwned` and `Copy` (later is rather strict, maybe it should be replaced with `Clone` or removed somehow).
Mainly `Component` implements `SerializableComponent` manually when it contains `Entity`.
`Component` should replace `Entity` with provided `Marker` in `SerializableComponent::save` function. And replace `Marker` with `Entity` in `SerializableComponent::load` function.
* `Components` trait is implemented for tuples of `SerializableComponent`s
It allows to save all existing and mentioned in tuple components of one entity from `ReadStorage`s into serializable structure. And load it into `WriteStorage`s again.

Helper types `WorldSerialize` and `WorldDeserialize` are good example how to use this facility.

JFF. Pong 🏓 example:
- [x] Fixed corner camping due to linear ball velocity never changing angle
- [x] Added cheat functions. Saving by `F5` key and loading by `F7`

### DIFF
--- a/examples/04_pong/main.rs
+++ b/examples/04_pong/main.rs
@@ -2,6 +2,11 @@
 
 extern crate amethyst;
 extern crate futures;
+extern crate rayon;
+extern crate ron;
+extern crate serde;
+#[macro_use]
+extern crate serde_derive;
 
 use amethyst::{ApplicationBuilder, Result};
 use amethyst::assets::{AssetFuture, BoxedErr};
@@ -15,15 +20,19 @@ use amethyst::ecs::{Component, DenseVecStorage, ECSBundle, Fetch, FetchMut, Join
 use amethyst::ecs::audio::DjBundle;
 use amethyst::ecs::input::{InputBundle, InputHandler};
 use amethyst::ecs::rendering::{Factory, MaterialComponent, MeshComponent, RenderBundle};
+use amethyst::ecs::saveload::{self, WorldDeserialize, WorldSerialize};
 use amethyst::ecs::transform::{LocalTransform, Transform, TransformBundle};
 use amethyst::prelude::*;
 use amethyst::renderer::Config as DisplayConfig;
 use amethyst::renderer::prelude::*;
 use amethyst::timing::Time;
 use futures::{Future, IntoFuture};
+use serde::de::DeserializeSeed;
+use serde::ser::Serialize;
 
 struct Pong;
 
+#[derive(Clone, Copy, Serialize, Deserialize)]
 struct Ball {
     pub position: [f32; 2],
     pub velocity: [f32; 2],
@@ -44,11 +53,13 @@ impl Component for Ball {
     type Storage = DenseVecStorage<Self>;
 }
 
+#[derive(Clone, Copy, Serialize, Deserialize)]
 enum Side {
     Left,
     Right,
 }
 
+#[derive(Clone, Copy, Serialize, Deserialize)]
 struct Paddle {
     pub position: f32,
     pub velocity: f32,
@@ -60,7 +71,7 @@ impl Paddle {
     pub fn new(side: Side) -> Paddle {
         Paddle {
             position: 0.,
-            velocity: 1.,
+            velocity: 0.,
             dimensions: [1., 1.],
             side: side,
         }
@@ -99,10 +110,14 @@ impl<'a, 'b, T> ECSBundle<'a, 'b, T> for PongBundle {
         &self,
         builder: ApplicationBuilder<'a, 'b, T>,
     ) -> Result<ApplicationBuilder<'a, 'b, T>> {
+        use saveload::marker::{U64Marker, U64MarkerAllocator};
         Ok(
             builder
+                .with_resource(U64MarkerAllocator::new())
+                .with_resource(Saved::default())
                 .with_resource(Score::new())
                 .with_resource(Time::default())
+                .register::<U64Marker>()
                 .register::<Ball>()
                 .register::<Paddle>()
                 .with(PongSystem, "pong_system", &[]),
@@ -133,10 +148,12 @@ impl<'a> System<'a> for PongSystem {
         // Properties of left paddle.
         let mut left_dimensions = [0.0, 0.0];
         let mut left_position = 0.0;
+        let mut left_velocity = 0.0;
 
         // Properties of right paddle.
         let mut right_dimensions = [0.0, 0.0];
         let mut right_position = 0.0;
+        let mut right_velocity = 0.0;
 
         let delta_time = time.delta_time.subsec_nanos() as f32 / 1.0e9;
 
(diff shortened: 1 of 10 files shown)


OvermindDL1:
Is it serializing the entity ID's?  In my C++ ECS I serialized entities out not including their integral ID's (instead they were assigned new ones inside the serialized format starting at 0, then when loaded then were loaded into whatever ID's were appropriate at the time).  Anytime I had a component reference another entity it would check if that entity had a SerializeKey component (it had an integer on it was all, uniquely assigned from the world) and would serialize it based on that key (if it was included in the set then it was included, else if not then it was just 'pointed' to by that key, when reloaded if it was a keyed load then it looked up in the existing world for an entity with that key and used its entity id, else it nulled it).  If it has a Transient component (empty tag component) then it was not serialized and the entity id reference was nulled (set to 0, the invalid entity). If it did not have either of those components then it would serialize it as a full entity in the serialization set with a new internal unique ID that was then referenced.  Individual fields could be marked as `[Transient]` to skip them on serialization entirely (or have other attributes to only serialize on file save/load, or to serialize or not on client->server or server->client and so forth).  It allowed me to serialize 1 thing if I wanted, or up to entire sets of entities (to JSON, a binary packing, or whatever else I wished, json and binary were the two built-in default formats) of an arbitrary tree or parent/child relationship.

omni-viral:
@OvermindDL1 I'm not serializing `Entity`. Right now I'm working on it.
If some components want to be serializable and it want to store Entity it has to put uid of the entity into serialized data. This means that entity has to have such uid. The component with uid has to implement special trait and be serializable as well.
Component with uid can have additional features like rejecting deserialization.

src/ecs/mod.rs
```diff
@@ -1,5 +1,6 @@
 //! `amethyst` engine built-in types for `specs`.
 
+pub use shred::*;
```
torkleyy: We should not use two glob imports here. This could easily cause breakage.

omni-viral: Even one global import from another crate could break compilation.
Maybe we shouldn't import globally at all?
Replacing all `ecs::{foo, bar, baz}` with `specs::{foo, bar, baz}` doesn't look bad.

torkleyy:
This could be implemented super-easily if we had specialization...

torkleyy:
But I'm not sure if Amethyst should have its own strategy. Why not let Specs do that?

src/saveload/mod.rs
```diff
@@ -0,0 +1,446 @@
...
+    /// Convert this component into deserializable form (`Data`) using entity to marker mapping function
+    fn load<'a, F>(data: Self::Data, ids: F) -> Result<Self, Self::Error>
+        where F: FnMut(M) -> Option<Entity>;
+}
+
+/// An error type which cannot be instantiated.
+/// Used as a placeholder for associated error types if
+/// something cannot fail.
+#[derive(Debug)]
+pub enum NoError {}
```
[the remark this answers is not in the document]
omni-viral: It will be eventually replaced with `!`.
Current solution can be shared if we would have some crate to import in all other amethyst crates

omni-viral:
@torkleyy I try to make solution for serialization as generalized as possible.
I agree that this can be implemented at specs side.

omni-viral:
:tada: Saving and loading works in pong example
Press F5 to save. Press F7 to load.
If you attempt to load before saving first it will yell at you in terminal.

examples/04_pong/main.rs
```diff
@@ -100,10 +110,14 @@ impl<'a, 'b, T> ECSBundle<'a, 'b, T> for PongBundle {
...
         builder: ApplicationBuilder<'a, 'b, T>,
     ) -> Result<ApplicationBuilder<'a, 'b, T>> {
+        use saveload::marker::{U64Marker, U64MarkerAllocator};
         Ok(
             builder
+                .with_resource(U64MarkerAllocator::new())
+                .with_resource(Saved::default())
                 .with_resource(Score::new())
                 .with_resource(Time::default())
+                .register::<U64Marker>()
```
Xaeroxe: These could probably do with a bundle.

Xaeroxe:
@omni-viral I notice you still have this marked as WIP, other than fixing the merge conflicts, what remains to be done in this PR?

examples/04_pong/main.rs
omni-viral: Idk. They are tied to save/load systems. And those are tied to pong bundle

omni-viral:
@Xaeroxe. Rebase and some cleanup, that's all

examples/04_pong/main.rs
Xaeroxe: Oh!  Perhaps we ought to add that marker implementation to Amethyst itself.  U64 seems pretty generic.

omni-viral: It is there

Xaeroxe: Hm, so why can't a bundle for this be added to Amethyst if the marker structures are part of Amethyst?

omni-viral: Because anyway you will include it in bundle with systems using it.

omni-viral: Can one bundle add another?

Xaeroxe: Yep, just use `ApplicationBuilder::with_bundle` from inside the bundle.

Aceeri:
Any reason for not using the existing storage implementation of Serialize/Deserialize?

src/ecs/saveload/marker.rs
```diff
@@ -0,0 +1,122 @@
...
+
+/// Basic marker allocator
+#[derive(Clone, Debug)]
+pub struct U64MarkerAllocator {
+    index: u64,
+    mapping: HashMap<u64, Entity>,
+}
+
+impl U64MarkerAllocator {
+    /// Crate new `U64MarkerAllocator` which will yield `U64Marker`s starting with `0`
```
LucioFranco: should be Create 😄

src/ecs/saveload/marker.rs
```diff
@@ -0,0 +1,122 @@
...
+    type Allocator: MarkerAllocator<Self>;
+
+    /// Get this marker id
+    fn id(&self) -> Self::Identifier;
+
+    /// Update marker with new value
+    fn update(&mut self, _other: Self) {}
+}
+
+/// This allocator is used with `Marker` trait
```
LucioFranco: Might be nice to provide a small example of implementing this trait if we are gonna let people implement it themselves.

src/ecs/saveload/ser.rs
```diff
@@ -0,0 +1,135 @@
...
+use serde::ser::{self, Serialize, Serializer, SerializeSeq};
+
+use shred::{Resources, ResourceId, SystemData};
+use specs::{Entities, FetchMut, Join, ReadStorage, WriteStorage};
+
+use super::{Components, EntityData, Storages};
+use super::marker::{Marker, MarkerAllocator};
+
+
+/// Serialize entities
```
LucioFranco: Since this a public function, could we get a more in-depth description on what it does?

src/ecs/saveload/ser.rs
```diff
@@ -0,0 +1,135 @@
...
+            }
+        }
+        to_serialize = add;
+    }
+    serseq.end()
+}
+
+
+
+/// Serialize entities
```
LucioFranco: Same as above.

src/ecs/saveload/marker.rs
omni-viral: `NetMarker` will implement this function. Maybe I can add it as an example into doc

Aceeri:
Any reason why this doesn't use the serialization implementation for `Storage` in `specs`? There is the small difference of it just serializing storages in a sequence rather than as entities.

omni-viral:
@Aceeri because this serializes with help of `SerializableComponent` trait.
`Serialize` implementation of `specs::Storage` can't help here.

Xaeroxe:
We're at 16 commits in this PR now, can you squash them?

torkleyy:
I would have really liked to review this PR..

torkleyy:
Also, I still don't understand why we don't implement this in Specs.

Xaeroxe:
@torkleyy 

@LucioFranco is pretty eager to get serialization going here and this seems to be a viable solution so let's make this fairly quick.

Xaeroxe:
Since we've switched gears on this and are making this PR to specs instead I'm going to close this.  Please re-open it if I'm in the wrong.

omni-viral:
Example changes are left here

PR #377, opened 2017-09-20
Add OpenGL patch for vertical flipping

Xaeroxe:
Implements the idea discussed in https://github.com/amethyst/amethyst/pull/372

Fixes horizontal flipping of textures in OpenGL.


OvermindDL1:
Which reminds me, is it possible to load vulkan and opengl in the same system and pick them depending on hardware?

Xaeroxe:
(I know this currently has a few problems, I'm fixing them now.

Xaeroxe:
@OvermindDL1 Currently no, but we'll address that in the future.

Xaeroxe:
Note: this code doesn't currently have an OpenGL switch, that's because the amethyst crate always uses OpenGL as of this moment.

src/ecs/rendering/components/texture.rs
```diff
@@ -187,7 +187,16 @@ impl Context for TextureContext {
         if image.w > u16::max_value() as usize || image.h > u16::max_value() as usize {
             return TextureFuture::unsupported_size(image.w, image.h);
         }
-
+        
+        // Flip the image horizontally for OpenGL, when other backends are added
+        // DO NOT execute this for statement on them.
```
torkleyy: We have the cfg already, so why not use it right now?

Xaeroxe: It's present in amethyst_renderer but not in amethyst.

torkleyy: Hm.. I would still be in favor of doing it right now. Maybe we could just add the flag to the main crate?

Xaeroxe: I agree, although can we do that in a follow up pull request? I'd like to also add the necessary shaders so we can turn on all of our backends.

omni-viral: I'm not sure you need to do such low-level rendering related stuff in main crate.
'amethyst_renderer/src/tex.rs' is better place for do such hacks 😄

Xaeroxe: @omni-viral The patch requires the ability to mutate the underlying buffer, and I'd prefer to not mutate a buffer unless we're actively consuming it.  `TextureBuilder::new` doesn't currently consume the data.  Should it be made to?

Xaeroxe: Alternatively: Make it just take a mutable reference, flip it, send it to the GPU, flip it back, release the borrow.

Side note: This is getting more hacky by the second.

omni-viral: @Xaeroxe you can flip it in `TextureBuilder::build`

omni-viral: Each row should be flipped or rows order?

Xaeroxe: @omni-viral just the row order.  Thanks for the idea, I'll do it in `build`

msiglreith: nit: flipping along the y-axis is called 'vertical flip' (I always mess this up always)

Xaeroxe: Oh, thanks @msiglreith

Xaeroxe:
(Formatted with latest nightly)

src/app.rs
```diff
@@ -144,9 +142,10 @@ impl<'a, 'b> Application<'a, 'b> {
         {
             let engine = &mut self.engine;
             let states = &mut self.states;
-            if engine.world.res.has_value(
-                ResourceId::new::<InputHandler>(),
-            )
+            if engine
+                .world
+                .res
+                .has_value(ResourceId::new::<InputHandler>())
```
Xaeroxe: rustfmt may have had a bit too much to drink here.

Xaeroxe:
@omni-viral @torkleyy Ready for review.

Rhuagh:
No idea if this is a good  way, so I'll abstain.

amethyst_renderer/src/tex.rs
```diff
@@ -136,10 +136,30 @@ where
         use gfx::memory::cast_slice;
         use gfx::texture::ResourceDesc;
 
+
+        // This variable has to live here to make sure the flipped
+        // buffer lives long enough. (If one exists)
+        let mut v_flip_buffer;
+        let mut data = self.data.as_ref();
+
+        if cfg!(feature = "opengl") {
```
torkleyy: Thanks for addressing!

amethyst_renderer/src/renderer.rs
```diff
@@ -51,27 +51,27 @@ impl Renderer {
...
+              T: Pod + Copy,
     {
         tb.build(&mut self.factory)
     }
 
     /// Builds a new material resource.
     pub fn create_material<DA, TA, DE, TE, DN, TN, DM, TM, DR, TR, DO, TO, DC, TC>(&mut self, mb: MaterialBuilder<DA, TA, DE, TE, DN, TN, DM, TM, DR, TR, DO, TO, DC, TC>) -> Result<Material>
         where DA: AsRef<[TA]>,
-              TA: Pod,
+              TA: Pod + Copy,
```
torkleyy: Could this be `Clone`?

Xaeroxe: So it could, but I figured Copy would be more appropriate.  I checked all the types that implement Plain Old Data and all of them also implement Copy, so this doesn't really introduce restrictions, it just means I don't have to write out ".clone()"

torkleyy: I don't have a strong opinion here, I just tend to not restrict types unnecessarily.

Xaeroxe: @torkleyy If you don't have a strong opinion can you please provide an approval so we can merge this?

omni-viral: `Copy` tells us that cloning of array can be done with simple memcpy. So the overhead at least is not bigger than copying memory.

Xaeroxe:
bors r+

This has been open for 48 hours, reviewed by 4 people, two of which approved, and two of which expressed no major qualms.  So despite the fact that I'm missing half an approval I'm going to merge this.

Xaeroxe:
And there's our last review.  Thanks @torkleyy

Issue #379, opened 2017-09-21
RFC: State specific Systems and Scenes


Xaeroxe:
As a player traverses through menus and game levels each will likely have its own `State` associated with it.

As the game transitions between States it may be desirable for the game programmer to introduce and remove Systems, and to remove Entities when the context in which they are necessary is no longer present.  Simultaneously however, some Systems and Entities will need to be global.  So in order to achieve these goals I propose two additions to the main crate:

1. State specific dispatchers.  Each State has a Dispatcher associated with it in the StateMachine, the State can populate the Dispatcher with Systems in `on_start` and this dispatcher will be executed after the primary global Dispatcher.  This Dispatcher will only be disposed of if the State structure associated with it is also disposed of, either by `Trans::Pop` or `Trans::Switch`

2. State specific `Scenes`.  Each state will have access to a `Scene` structure which it can modify in any of functions on the `State` trait.  A `Scene` is simply a vector of Entities correlated with this `State`.  When the `State` is disposed of, either by `Trans::Pop` or `Trans::Switch` all entities in the Scene will be deleted from the specs::World.

EDIT: In order to keep the `Scene` vector from growing indefinitely it will be iterated through every frame and any entities found that are no longer [alive](https://docs.rs/specs/0.9.5/specs/struct.EntitiesRes.html#method.is_alive) will be `swap_remove`d from the vector.

Unresolved question: Should entities that are part of a Scene be made "inactive" when the State is paused, and if so how would this be implemented?

Xaeroxe:
cc @amethyst/engine-devs This will impact all of us so please review when you have time.

jojolepro:
Unresolved question: Should entities that are part of a Scene be made "inactive" when the State is paused, and if so how would this be implemented?
I'd say yes. How to do it? Remove them from specs on pause, and put them back in on resume. You already talked about a scene. From what I understand, its a buffer that tracks which entities were created by this State. We could make the Systems owned by a State register their new Entities as either owned by the Global game, or by the local/current State.

Also as a reference, android components look really closely to what we are trying to achieve with State. The only big difference that I see is that we also have to sync our data with a multithreaded ecs, which is both very efficient, and very complex. We'll have to innovate quite a bit here ;)
![image](https://user-images.githubusercontent.com/4755185/30675021-08545854-9e4c-11e7-9ebc-8ab282e23708.png)

Xaeroxe:
[...]

PR #380, opened 2017-09-22
[WIP] State specific dispatchers and scenes

Xaeroxe:
Implementation demonstration of https://github.com/amethyst/amethyst/issues/379

### DIFF
--- a/examples/03_renderable/main.rs
+++ b/examples/03_renderable/main.rs
@@ -14,6 +14,7 @@ use amethyst::assets::{AssetFuture, BoxedErr, Context, Format, Loader};
 use amethyst::assets::formats::meshes::ObjFormat;
 use amethyst::assets::formats::textures::PngFormat;
 use amethyst::config::Config;
+use amethyst::ecs::DispatcherBuilder;
 use amethyst::ecs::{Fetch, FetchMut, Join, System, WriteStorage};
 use amethyst::ecs::rendering::{LightComponent, MaterialComponent, AmbientColor, TextureContext,
                                Factory, TextureComponent, MeshComponent, MeshContext, RenderBundle};
@@ -82,7 +83,7 @@ impl<'a> System<'a> for ExampleSystem {
 struct Example;
 
 impl State for Example {
-    fn on_start(&mut self, engine: &mut Engine) {
+    fn on_start<'a, 'b>(&mut self, engine: &mut Engine, _: &mut Scene) -> Option<DispatcherBuilder<'a, 'b>> {
 
         initialise_camera(&mut engine.world.write_resource::<Camera>());
 
@@ -199,9 +200,10 @@ impl State for Example {
             camera_angle: 0.0,
             pipeline_forward: true,
         });
+        None
     }
 
-    fn handle_event(&mut self, engine: &mut Engine, event: Event) -> Trans {
+    fn handle_event(&mut self, engine: &mut Engine, _: &mut Scene, event: Event) -> Trans {
         let w = &mut engine.world;
         // Exit if user hits Escape or closes the window
         let mut state = w.write_resource::<DemoState>();
--- a/src/state.rs
+++ b/src/state.rs
@@ -3,6 +3,8 @@
 use engine::Engine;
 use event::Event;
 
+use specs::{ Dispatcher, DispatcherBuilder, EntitiesRes, Entity };
+
 /// Types of state transitions.
 pub enum Trans {
     /// Continue as normal.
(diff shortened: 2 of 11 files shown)


Xaeroxe:
Couple things missing from this that I'll add later:

- This PR kind of mucked up the signature for `on_start` and I don't really like that, so I'm going to add another function to the `State` trait called `dispatcher` with a default implementation that returns `None`.  That way new users can learn about custom dispatchers a little later instead of at "hello_world"
- There's also no way to inactivate entities from the `Scene`s of inactive states at the moment.  I'll work on how to do this later.

Rhuagh:
Looking good so far. The only concern I have is how to handle synchronization of the entity list in a generic way. This is why I proposed to just interface with the world directly from the users side.

Xaeroxe:
> Synchronize the entity list

Can you elaborate on this?  What specifically would need to be synchronized?

src/state.rs
```diff
@@ -77,8 +104,8 @@ impl<'a> StateMachine<'a> {
     /// Panics if no states are present in the stack.
     pub fn start(&mut self, engine: &mut Engine) {
         if !self.running {
-            let state = self.state_stack.last_mut().unwrap();
-            state.on_start(engine);
+            let &mut (ref mut state, ref mut dispatcher, ref mut scene) = self.state_stack.last_mut().unwrap();
```
omni-viral: I think it would look nicer like this
```
let (ref mut state, ref mut dispatcher, ref mut scene) = *self.state_stack.last_mut().unwrap();
```

src/state.rs
```diff
@@ -99,7 +126,17 @@ impl<'a> StateMachine<'a> {
     pub fn fixed_update(&mut self, engine: &mut Engine) {
         if self.running {
             let trans = match self.state_stack.last_mut() {
-                Some(state) => state.fixed_update(engine),
+                Some(&mut (ref mut state, _, ref mut scene)) => {
+                    let mut i = 0;
+                    while i < scene.vec.len() {
```
omni-viral: Just `scene.vec.retain(|entity| engine.world.read_resource::<EntitiesRes>().is_alive(entity))`

src/state.rs
```diff
@@ -134,41 +176,43 @@ impl<'a> StateMachine<'a> {
...
-                state.on_pause(engine);
+            if let Some(&mut (ref mut state, _, ref mut scene)) = self.state_stack.last_mut() {
+                state.on_pause(engine, scene);
             }
 
-            self.state_stack.push(state);
-            let state = self.state_stack.last_mut().unwrap();
-            state.on_start(engine);
+            let mut scene = Scene::new();
+            let dispatcher = state.on_start(engine, &mut scene).map(|db| db.with_pool(engine.pool.clone()).build());
```
omni-viral: The same `State` will be pushed and popped again and again during the game.
Building dispatchers each time may be slow. Can't we cache 'em in some way?

torkleyy:
Marking this as `type: RFC` since this is engine design. I'm sorry I cannot give it a proper review right now.

src/state.rs
Xaeroxe: We could, but that introduces its own problems because any state the systems may contain will be preserved between instances of the state.  Why would building a dispatcher be slow?  It doesn't depend on external data or operate on large data sets.

examples/03_renderable/main.rs
```diff
@@ -82,7 +83,7 @@ impl<'a> System<'a> for ExampleSystem {
 struct Example;
 
 impl State for Example {
-    fn on_start(&mut self, engine: &mut Engine) {
+    fn on_start<'a, 'b>(&mut self, engine: &mut Engine, _: &mut Scene) -> Option<DispatcherBuilder<'a, 'b>> {
```
torkleyy: That's the exact opposite of what I asked for earlier. Why not have one very flexible dispatcher? After all, one could even wrap a dispatcher in a system and add it.

src/state.rs
```diff
@@ -3,6 +3,8 @@
 use engine::Engine;
 use event::Event;
 
+use specs::{ Dispatcher, DispatcherBuilder, EntitiesRes, Entity };
```
torkleyy: Did you run `cargo fmt`?

Xaeroxe:
@torkleyy The decision is still left to the user as none of this introduces mandatory architecture.

\<rant\>
The Rust gamedev ecosystem is infected with this idea that no architecture is preferable to any architecture.  I think this is fool-hardy, user friendliness means introducing some "guided tour" through the capabilities of the software.  It's a way forward for people who can't be bothered to engineer their own architecture.  Ideally said architecture should not restrict the user, but it merely provides them with blatant and obvious options, which they can choose to ignore.  When you start building a game in Unreal or Unity you don't start with designing state and scene management, you start with coding your entities, you start with what makes your game unique.  That's because the engine does that scene management for you.  What game wouldn't need something like this?  What game would be inhibited by something like this being present?  Open source software is not known for being particularly user friendly and this is why.  The instant you try and provide any kind of "recommended workflow" not even a "mandatory workflow" but a "recommended" one everyone just loses their mind.  Our growth as an engine depends on user's first experience with the software and if their first experience is 

Q: "How do I do this very basic and common thing?"
A: "Figure it out yourself"

They're not going to want to continue using the software.  They'll go run off and use the commercial alternatives, because they had a satisfactory answer to that question.  This attitude is the number one cause of FOSS failing to make a significant impact on the market.  Every time open source saw widespread usage it was because it entered the hands of a company who was willing to provide satisfactory answers to extremely common user questions.  I appreciate architectural decisions can be kind of scary because if we mess them up we cause serious problems downstream.  This specific proposal however, does not introduce inflexibility.  No internal engine systems are dependent on this, and you are not barred from hardware access if you don't use this.  There is room for improvement to this proposal as it stands but simply "not making a decision" will lead to decreased usage of the engine.  Are we Piston?  Or are we a competitor to commercial game engines?
\</rant\>

If we absolutely won't introduce these kinds of benefits to the main Amethyst crate itself then I plan on providing an "example project" demonstrating how to do these kinds of things.  I won't even put it under the amethyst organization if it really bothers you that much.

jojolepro:
> The decision is still left to the user as none of this introduces mandatory architecture.
:+1:

torkleyy:
@Xaeroxe I fully agree on guided user experience. I just favor the approach of expressibility over fixed concepts.

> If we absolutely won't introduce these kinds of benefits to the main Amethyst crate itself then I plan on providing an "example project" demonstrating how to do these kinds of things. I won't even put it under the amethyst organization if it really bothers you that much.

Presenting common patterns is a good idea and I don't have any objections there.

> The decision is still left to the user as none of this introduces mandatory architecture.

The way I see it, we got 1 dispatcher + 0..1 dispatchers per state now and the concept of scenes. This does seem mandatory to me.

---

Please give me some time to come up with a good answer; I fully agree that having every user rewriting the world is not a good idea ;)

torkleyy:
So could we please just pause the discussion for a moment?

Xaeroxe:
@torkleyy Alright, just tell me when your proposal is ready.  I've got plenty of other work to do anyways.

Xaeroxe:
@torkleyy has much better ideas on how to do this, so I'm closing this PR.

Issue #383, opened 2017-09-23
Amethyst executing at 1400 FPS and eating all the CPU


Xaeroxe:
@jojolepro 's profiler indicated the pong example was running at 1400 FPS causing amethyst to consume all of the CPU resources.  We should provide some kind of FPS cap.

jojolepro:
I already knew that, I can use the steam fps counter, or the fps counter I added to amethyst.
To cap the fps, a while(true) checking the time in a single thread environment (outside of the dispatcher) could do the job easily.

Rhuagh:
Don't do while(true), that will use up one thread all the time. Use yield or something like it instead.

jojolepro:
I read somewhere a while ago that using any kind of pause will result in imprecision, because the os will lower the thread priority while its in pause, and will check less often if it should unpause it.

Rhuagh:
For processes yes, this is for threads. Yield will "give the cpu" to other threads in the same process.

Rhuagh:
Yield is basically a way to say "someone else can go now, but if noone else with the same or higher priority wanna go I'll keep going".

omni-viral:
Usually you wait in few steps steps. First you yield to another thread then you yield to another process. If you need to wait for too long - sleep

PR #384, opened 2017-09-24
Enforce maximum framerate

Xaeroxe:
This pull request will repeatedly yield the thread until sufficient time has passed to advance to the next frame.

cc @Rhuagh @jojolepro 

Fixes https://github.com/amethyst/amethyst/issues/383

### DIFF
--- a/amethyst_renderer/src/config.rs
+++ b/amethyst_renderer/src/config.rs
@@ -29,9 +29,6 @@ pub struct Config {
     /// Sets the visibility of the window.
     #[serde(default = "default_visibility")]
     pub visibility: bool,
-    /// FPS will be restricted to this value. Defaults to 144.
-    #[serde(default = "default_max_fps")]
-    pub max_fps: u32,
 }
 
 impl Default for Config {
@@ -45,7 +42,6 @@ impl Default for Config {
             vsync: default_vsync(),
             multisampling: default_multisampling(),
             visibility: default_visibility(),
-            max_fps: default_max_fps(),
         }
     }
 }
@@ -60,9 +56,6 @@ fn default_multisampling() -> u16 {
 }
 fn default_visibility() -> bool {
     true
-}
-fn default_max_fps() -> u32 {
-    144
 }
 
 impl From<Config> for WindowBuilder {
--- a/amethyst_renderer/src/config.rs
+++ b/amethyst_renderer/src/config.rs
@@ -29,6 +29,9 @@ pub struct Config {
     /// Sets the visibility of the window.
     #[serde(default = "default_visibility")]
     pub visibility: bool,
+    /// FPS will be restricted to this value. Defaults to 144.
+    #[serde(default = "default_max_fps")]
+    pub max_fps: u32,
 }
 
 impl Default for Config {
@@ -42,6 +45,7 @@ impl Default for Config {
             vsync: default_vsync(),
             multisampling: default_multisampling(),
             visibility: default_visibility(),
+            max_fps: default_max_fps(),
         }
     }
 }
@@ -56,6 +60,9 @@ fn default_multisampling() -> u16 {
 }
 fn default_visibility() -> bool {
     true
+}
+fn default_max_fps() -> u32 {
+    144
 }
 
 impl From<Config> for WindowBuilder {
--- a/amethyst_renderer/src/renderer.rs
+++ b/amethyst_renderer/src/renderer.rs
@@ -11,9 +11,9 @@ use config::Config;
 use error::{Error, Result};
 use mesh::{Mesh, MeshBuilder};
 use mtl::{Material, MaterialBuilder};
-use pipe::{ColorBuffer, DepthBuffer, PolyPipeline, PipelineBuild, PipelineData, Target};
+use pipe::{ColorBuffer, DepthBuffer, PipelineBuild, PipelineData, PolyPipeline, Target};
 use tex::{Texture, TextureBuilder};
-use types::{ColorFormat, Device, DepthFormat, Window, Encoder, Factory};
+use types::{ColorFormat, DepthFormat, Device, Encoder, Factory, Window};
 use vertex::VertexFormat;
 use winit::{self, EventsLoop, WindowBuilder};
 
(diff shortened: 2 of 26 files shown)


Rhuagh:
Why don't we just say minimum duration between frames, that way we can store a `Duration`, and just do a simple comparison.

torkleyy:
Shouldn't this be configured via config file?

Aceeri:
There isn't a clear "fps" that a game should run at. For example, most people cap counterstrike at 300 fps because it could still allow for better input responsiveness, while other games might cap at 60/120/144 depending on monitor refresh rate. So it definitely should be a variable amount that can be set.

src/app.rs
```diff
@@ -114,6 +117,10 @@ impl<'a, 'b> Application<'a, 'b> {
         while self.states.is_running() {
             self.timer.restart();
             self.advance_frame();
+            // Enforce maximum FPS
+            while self.timer.elapsed() * self.max_fps < Duration::from_secs(1) {
+                yield_now();
```
torkleyy: A) Why don't we use `sleep` here?
B) To ensure we're not using 100% CPU, we should yield / sleep at least once per frame.

Xaeroxe:
r? @torkleyy @Rhuagh @omni-viral 

This is now set via the config, the frame rate is enforced in the renderer rather than in the main application, and we are yielding at least once every frame.

This has also had the latest rustfmt-nightly executed on it.

Xaeroxe:
@Rhuagh The reason for not using the minimum duration at this point is that it's more convenient to specify in terms of FPS rather than Seconds Per Frame.

Xaeroxe:
@torkleyy I just installed it :(

```
$ cargo +nightly fmt -- --version
0.2.7-nightly ( )
```

As for the sleep question: [sleep](https://doc.rust-lang.org/std/thread/fn.sleep.html) documentation states that sleep is imprecise.  @Rhuagh mentioned on gitter Linux is notorious for sleeping longer than expected.

torkleyy:
@Xaeroxe That's why my usual approach is to only sleep for 90% of the time.

Xaeroxe:
So here's some code to play around with:

https://play.rust-lang.org/?gist=88ec2d8d24de0d283acacd52cfc64f6d&version=stable

So if I graph out my results from that I get this (time is in micro seconds):

image

So at best within our time range this has a variance of +- 1%.  So if we only sleep when > 1 ms of time is needed, and only sleep for 95% of that duration, then yield the rest I think that will give us the best results.

amethyst_renderer/src/config.rs
```diff
@@ -57,6 +61,9 @@ fn default_multisampling() -> u16 {
 fn default_visibility() -> bool {
     true
 }
+fn default_max_fps() -> u32 {
+    144
```
omni-viral: I think this limit should be a bit higher than display framerate. If we wait longer than needed (ocasionally) then fps will be lower than specified here.

amethyst_renderer/src/renderer.rs
```diff
@@ -115,6 +119,16 @@ impl Renderer {
 
         self.device.cleanup();
 
+        // # Enforce maximum FPS
+        // Sleep if the time needed is > 10 ms for 9.5 ms
+        let time_remaining = Duration::from_secs(1) / self.max_fps - start.elapsed();
+        if time_remaining > Duration::new(0, 10000000) {
```
omni-viral: I fear sleeping precision is worse than that.
In fact it may be so bad we shouldn't use it here at all.
Or do it for at least 100ms period.

amethyst_renderer/src/renderer.rs
```diff
@@ -115,6 +119,16 @@ impl Renderer {
...
+            sleep(time_remaining * 95 / 100);
```
omni-viral: I'd recommend to wake up for `N` earlier where `2*N` is your treshold for sleeping.
E.g. subtract constant instead of multiplying.

amethyst_renderer/src/renderer.rs
```diff
@@ -115,6 +119,16 @@ impl Renderer {
...
+        }
+        // Yield for the remainder of the time.
+        while start.elapsed() * self.max_fps < Duration::from_secs(1) {
```
omni-viral: It is faster to store `1 sec / max_fps` in field as you always do this operation.

amethyst_renderer/src/renderer.rs
```diff
@@ -91,6 +94,7 @@ impl Renderer {
         use gfx::Device;
         #[cfg(feature = "opengl")]
         use glutin::GlContext;
+        let start = Instant::now();
```
omni-viral: We should check against `Instant` from last time this funtion was called.
Store it in struct and update after waiting.

amethyst_renderer/src/renderer.rs
omni-viral: You can have max fps in config yet store inverse in structure.

amethyst_renderer/src/renderer.rs
omni-viral: Well. Since you've tested it you may have more actual info about it than my guesding

Rhuagh:
Can we just please remove sleep?  I read up yesterday, and worst case for Windows will sleep 15.6ms I you do sleep(1). For 144 fps that's 2 frames..

Aceeri:
Just tested this and it panics because you subtract a duration with one after it (overflows due to signed integer).

Xaeroxe:
@Aceeri Also confirmed sleep is sleeping for too long, causing the frame rate to be much slower than intended.  So I'm going to go back to full yielding.  @Aceeri has already prepared a PR against my branch to do this which I'll be merging later today.

Xaeroxe:
So: This isn't configurable at the moment.  That's because we don't have any kind of generalized configuration.  This has to live outside the renderer and we don't have an Application level config file.  So I'm going to make another followup PR to introduce that and possibly refactor our growing list of needed configuration files.  I'd like to keep that PR separate from this bug fix PR though.

r? @omni-viral

examples/03_renderable/main.rs
```diff
@@ -31,17 +32,18 @@ struct DemoState {
     point_light: bool,
     directional_light: bool,
     camera_angle: f32,
-    #[allow(dead_code)]
-    pipeline_forward: bool, // TODO
+    #[allow(dead_code)] pipeline_forward: bool, // TODO
```
torkleyy: Could we change that in the rustfmt config?

src/app.rs
```diff
@@ -217,8 +224,8 @@ pub struct ApplicationBuilder<'a, 'b, T> {
     pub world: World,
     pool: Arc<ThreadPool>,
     /// Allows to create `RenderSystem`
-    // TODO: Come up with something clever
```
torkleyy: Why remove this line?

src/app.rs
```diff
@@ -110,12 +112,17 @@ impl<'a, 'b> Application<'a, 'b> {
...
     pub fn run(&mut self) {
         self.initialize();
-
+        self.timer.start();
         while self.states.is_running() {
-            self.timer.restart();
             self.advance_frame();
+            // We are done with this frame, we can just rest for now.
+            while self.timer.elapsed() < self.dur_per_frame {
+                ::std::thread::yield_now();
```
torkleyy: style: import `std::thread` and call `thread::yield_now`

Xaeroxe:
Oh also side note: I tested this with @jojolepro 's FPS counter and it seems to work perfectly.

src/app.rs
```diff
@@ -655,6 +663,20 @@ impl<'a, 'b, T> ApplicationBuilder<'a, 'b, T> {
...
+    /// Sets the maximum frames per second of this game.
+    ///
+    /// # Parameters
+    ///
+    /// `max_fps`: the maximum frames per second this game will run at.
+    ///
+    /// # Returns
+    ///
+    /// This function returns the ApplicationBuilder after modifying it.
+    pub fn with_max_fps(mut self, max_fps: u32) -> Self {
```
torkleyy: One might expect that setting `max_fps = 0` removes the cap. (How) should we address that?

src/app.rs
Xaeroxe: It communicates absolutely 0 useful information.  It's the worst todo I've ever seen.

src/app.rs
Xaeroxe: Well removing the cap is quite dangerous as we've proven.  I'd rather not build around enabling that.  If someone really hates the cap they could set it to something absurd like 9999.

examples/03_renderable/main.rs
Xaeroxe: Why?  I kind of like it this way.  It communicates the impact of the #[] more clearly.

src/app.rs
torkleyy: Yes, but atm using `0` will cause the result of the division to be infinite. So my concern is that the convention to use `0` for limits as a way to disable it could cause confusion here. So if we don't implement that, we need documentation for that behavior.

Xaeroxe: Yeah that's a good point.  Especially if this is going in the config file we ought to handle for that.

src/app.rs
torkleyy: Well, somebody added it there as a remainder to fix something. If we remove that comment, we should 1) fix it or 2) open an issue in case it's a bigger problem.

examples/03_renderable/main.rs
torkleyy: But shouldn't the focus be on the field name and its type rather than attributes? And I also had a case where the line got super long because of that formatting and thus wasn't very nice to read.

src/app.rs
Xaeroxe: Git blame says @omni-viral wrote this, @omni-viral mind telling us what the intended task for this TODO was?  Maybe I can make it say something more useful.

src/app.rs
Xaeroxe:
```rust
if max_fps == 0 {
    self.max_fps = std::u32::MAX;
} else {
    self.max_fps = max_fps
}
```

That seems like a real simple way to enable using 0 while not building around having no FPS cap.

examples/03_renderable/main.rs
Xaeroxe: I don't feel this really removes any visibility of that information, but I also have no strong feelings on formatting other than "it shouldn't be absolutely awful to read and write".  So sure I'll add it to the config.

src/app.rs
Xaeroxe: Fixed.

Xaeroxe:
So uh, we need to make sure we're running cargo fmt with the `--all` parameter.  Our sub crates needed a lot of formatting changes.

Xaeroxe:
Because I already know this question is coming:

```
$ cargo +nightly fmt -- --version
0.2.7-nightly ( )
```

src/app.rs
stenverbois: Such special cases could use documentation anyway.

omni-viral:
@Xaeroxe just a moment, I'm scrolling the diff.

src/app.rs
omni-viral: It was about public access to `EventsLoop`. It isn't need to be pub anymore

Xaeroxe:
Alright sounds good, I'm gonna merge this as is.

bors r+

PR #386, opened 2017-09-30
Edit window properties through events

jojolepro:
Added event handler to RenderSystem. 

Consume events that contains a closure of type fn(&mut Window)->()

See an example in pong, where I set the cursor to be hidden.

### DIFF
--- a/amethyst_renderer/src/lib.rs
+++ b/amethyst_renderer/src/lib.rs
@@ -77,4 +77,4 @@ mod mesh;
 mod mtl;
 mod renderer;
 mod tex;
-mod types;
+pub mod types;
--- a/amethyst_renderer/src/lib.rs
+++ b/amethyst_renderer/src/lib.rs
@@ -67,7 +67,7 @@ pub mod pass;
 pub mod prelude;
 pub mod pipe;
 pub mod vertex;
-
+pub mod types;
 
 mod cam;
 mod color;
@@ -77,4 +77,3 @@ mod mesh;
 mod mtl;
 mod renderer;
 mod tex;
-pub mod types;
--- a/amethyst_renderer/src/renderer.rs
+++ b/amethyst_renderer/src/renderer.rs
@@ -41,7 +41,7 @@ impl Renderer {
     }
 
     /// Gets a mutable window to modify
-    pub fn window_mut(&mut self)->&mut Window{
+    pub fn window_mut(&mut self) -> &mut Window {
         &mut self.window
     }
 
--- a/amethyst_renderer/src/renderer.rs
+++ b/amethyst_renderer/src/renderer.rs
@@ -40,6 +40,11 @@ impl Renderer {
         RendererBuilder::new(el)
     }
 
+    /// Gets a mutable window to modify
+    pub fn window_mut(&mut self)->&mut Window{
+        &mut self.window
+    }
+
     /// Builds a new mesh from the given vertices.
     pub fn create_mesh<D, T>(&mut self, mb: MeshBuilder<D, T>) -> Result<Mesh>
     where
--- a/examples/04_pong/main.rs
+++ b/examples/04_pong/main.rs
@@ -2,6 +2,7 @@
 
 extern crate amethyst;
 extern crate futures;
+extern crate winit;
 
 use amethyst::{ApplicationBuilder, Result};
 use amethyst::assets::{AssetFuture, BoxedErr};
@@ -15,10 +16,12 @@ use amethyst::ecs::{Component, DenseVecStorage, ECSBundle, Fetch, FetchMut, Join
 use amethyst::ecs::audio::DjBundle;
 use amethyst::ecs::input::{InputBundle, InputHandler};
 use amethyst::ecs::rendering::{Factory, MaterialComponent, MeshComponent, RenderBundle};
+use amethyst::ecs::rendering::events::WindowModifierEvent;
 use amethyst::ecs::transform::{LocalTransform, Transform, TransformBundle};
 use amethyst::prelude::*;
 use amethyst::renderer::Config as DisplayConfig;
 use amethyst::renderer::prelude::*;
+use amethyst::shrev::EventHandler;
 use amethyst::timing::Time;
 use futures::{Future, IntoFuture};
 
@@ -335,6 +338,20 @@ impl State for Pong {
                 bounce_sfx,
                 score_sfx,
             });
+            // Sets the cursor hidden when the mouse is hovering the pong window.
+            if let Some(mut window_events) = engine
+                .world
+                .res
+                .try_fetch_mut::<EventHandler<WindowModifierEvent>>(0)
+            {
+                window_events.write_single(WindowModifierEvent {
+                    modify: |win| {
+                        win.set_cursor_state(winit::CursorState::Hide)
+                            .ok()
+                            .expect("could not grab mouse cursor")
+                    },
+                });
+            }
 
             let have_output = engine.world.read_resource::<Option<Output>>().is_some();
 
--- a/examples/04_pong/main.rs
+++ b/examples/04_pong/main.rs
@@ -345,11 +345,11 @@ impl State for Pong {
                 .try_fetch_mut::<EventHandler<WindowModifierEvent>>(0)
             {
                 window_events.write_single(WindowModifierEvent {
-                    modify: |win| {
+                    modify: Box::new(|win| {
                         win.set_cursor_state(winit::CursorState::Hide)
                             .ok()
-                            .expect("could not grab mouse cursor")
-                    },
+                            .expect("Could not hide mouse cursor")
+                    }),
                 });
             }
 
--- a/src/ecs/rendering/bundle.rs
+++ b/src/ecs/rendering/bundle.rs
@@ -95,7 +95,7 @@ where
             up: [0.0, 1.0, 0.0].into(),
         };
 
-        let window_events = EventHandler::<WindowModifierEvent>::new();
+        let window_events = EventHandler::<WindowModifierEvent>::with_capacity(20);
         let reader_id = window_events.register_reader();
 
         builder = builder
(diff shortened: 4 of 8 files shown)


src/ecs/rendering/systems.rs
```diff
@@ -12,25 +14,46 @@ pub struct RenderSystem<P> {
...
+        //Read all window-modifying events
+        match events.read(&mut self.reader_id) {
+            Ok(EventReadData::Data(data)) => {
+                let mut win = self.renderer.window_mut();
+                for ev in data.collect::<Vec<_>>() {
+                    let closure = ev.modify;
+                    closure(win);
+                }
+            }
+            _ => {}
```
Rhuagh: Overflow variant might be good, I don't foresee it as a problem, but for completeness sake.

src/ecs/rendering/systems.rs
```diff
@@ -12,25 +14,46 @@ pub struct RenderSystem<P> {
     pipe: P,
     #[derivative(Debug = "ignore")]
     renderer: Renderer,
+    reader_id: ReaderId,
```
Rhuagh: More descriptive name of the reader maybe.

amethyst_renderer/src/renderer.rs
```diff
@@ -40,6 +40,11 @@ impl Renderer {
         RendererBuilder::new(el)
     }
 
+    /// Gets a mutable window to modify
+    pub fn window_mut(&mut self)->&mut Window{
```
Rhuagh: Formatting.

examples/04_pong/main.rs
```diff
@@ -335,6 +338,20 @@ impl State for Pong {
                 bounce_sfx,
                 score_sfx,
             });
+            // Sets the cursor hidden when the mouse is hovering the pong window.
+            if let Some(mut window_events) = engine
+                .world
+                .res
+                .try_fetch_mut::<EventHandler<WindowModifierEvent>>(0)
+            {
+                window_events.write_single(WindowModifierEvent {
```
Xaeroxe: This section ends up looking a bit verbose with use of shrev.  This might look a bit cleaner if we used a resource like

```rust
pub struct WindowCommands {
    inner: SmallVec<[Box<FnOnce(&mut winit::Window)>; 5]>
}

impl WindowCommands {
    pub fn send_command(&mut self, command: Box<FnOnce(&mut winit::Window)>) {
        self.inner.push(command);
    }
```

Then just drain that vec in the renderer.

Xaeroxe: To put it another way: shrev is really good if you might want more than one ReaderId, if you only have one ReaderId though it offers little benefit and mostly just complicates the implementation.

src/ecs/rendering/systems.rs
```diff
@@ -12,25 +14,46 @@ pub struct RenderSystem<P> {
...
+        <P as PipelineData<'a>>::Data,
+        FetchMut<'a, EventHandler<WindowModifierEvent>>,
+    );
+
+    fn run(&mut self, (factory, data, events): Self::SystemData) {
+        //Read all window-modifying events
+        match events.read(&mut self.reader_id) {
+            Ok(EventReadData::Data(data)) => {
+                let mut win = self.renderer.window_mut();
+                for ev in data.collect::<Vec<_>>() {
```
Xaeroxe: `.collect` here isn't really necessary, for loops work on iterators.

src/ecs/rendering/bundle.rs
```diff
@@ -92,10 +95,14 @@ where
             up: [0.0, 1.0, 0.0].into(),
         };
 
+        let window_events = EventHandler::<WindowModifierEvent>::new();
```
Xaeroxe: `new` defaults to a buffer size of 200, I never see us needing more than 10 for this so use `with_capacity` instead.

src/ecs/rendering/events.rs
```diff
@@ -0,0 +1,9 @@
+//! Events of the rendering system
+
+use renderer::types::Window;
+
+/// Even to modify a window.
+pub struct WindowModifierEvent {
+    /// The closure that modifies the window
+    pub modify: fn(&mut Window) -> (),
```
Xaeroxe: This should be of type Box<FnMut(&mut Window) -> ()>, just to grant better flexibility.

src/ecs/rendering/bundle.rs
Rhuagh: Make it 20 just top be on the safe side

amethyst_renderer/src/lib.rs
```diff
@@ -77,4 +77,4 @@ mod mesh;
 mod mtl;
 mod renderer;
 mod tex;
-mod types;
+pub mod types;
```
torkleyy: Please keep pub modules in groups.

examples/04_pong/main.rs
```diff
@@ -335,6 +338,20 @@ impl State for Pong {
...
+                    modify: |win| {
+                        win.set_cursor_state(winit::CursorState::Hide)
+                            .ok()
+                            .expect("could not grab mouse cursor")
```
torkleyy: How would this be handled in a real world application? Shouldn't we use best practices in our examples?

examples/04_pong/main.rs
```diff
@@ -335,6 +338,20 @@ impl State for Pong {
...
             });
+            // Sets the cursor hidden when the mouse is hovering the pong window.
+            if let Some(mut window_events) = engine
+                .world
+                .res
+                .try_fetch_mut::<EventHandler<WindowModifierEvent>>(0)
+            {
+                window_events.write_single(WindowModifierEvent {
+                    modify: |win| {
+                        win.set_cursor_state(winit::CursorState::Hide)
```
torkleyy: How does the cursor become visible again?

src/ecs/rendering/events.rs
torkleyy: So this isn't really an event but a more a message to the window which is necessary because `Window` is thread-local?

src/ecs/rendering/mod.rs
```diff
@@ -2,10 +2,12 @@
...
 pub use self::components::*;
+pub use self::events::*;
 pub use self::resources::*;
 pub use self::systems::*;
 
 pub mod components;
 pub mod resources;
 pub mod systems;
 pub mod bundle;
+pub mod events;
```
torkleyy: Should be reordered, not sure why rustfmt didn't do that.

torkleyy:
We should eventually make this similar to `FactoryRes`: create a thread-safe resource which just passes the jobs to the real window.

src/ecs/rendering/events.rs
Rhuagh: Yes

amethyst_renderer/src/lib.rs
Rhuagh: Might not have run cargo fmt --all

Rhuagh:
Do we want to keep the `FactoryRes`?

torkleyy:
@Rhuagh why not? What would you use instead?

Rhuagh:
I thought someone said they wanted to replace that abstraction with something better. I have no set opinion myself.

omni-viral:
@torkleyy the problem with `FactoryRes` is that you have to wait two frames before the mesh or texture will be attached to the component. It is better to abstract away usage of factory. The mesh component should be created without `Factory`. `RenderSystem` then can use `Factory` to create internal gpu resource if it isn't created yet. This will simplify creating procedural meshes. Current threadsafe `Factory` abstraction can be used still to work with assets if it fits better.

amethyst_renderer/src/lib.rs
jojolepro: forgot --all

examples/04_pong/main.rs
jojolepro: You need to send a new event to the window. The code is the same, you just replace winit::CursorState::Hide by winit::CursorState::Normal

https://docs.rs/winit/0.8.2/winit/enum.CursorState.html

examples/04_pong/main.rs
jojolepro: I'm not sure how that could actually crash. What I'm sure though, is that the error message is not the right one.

Xaeroxe:
Thanks for addressing my comments, LGTM.

examples/04_pong/main.rs
```diff
@@ -335,6 +338,20 @@ impl State for Pong {
...
+            // Sets the cursor hidden when the mouse is hovering the pong window.
+            if let Some(mut window_events) = engine
+                .world
+                .res
+                .try_fetch_mut::<EventHandler<WindowModifierEvent>>(0)
+            {
+                window_events.write_single(WindowModifierEvent {
+                    modify: Box::new(|win| {
+                        win.set_cursor_state(winit::CursorState::Hide)
+                            .ok()
```
torkleyy: 1) Why call `.ok` just to unwrap later?
2) Didn't we agree on not exposing `winit` stuff or is my memory wrong?
3) Those are 14 lines for something that basic, so shouldn't we change something here?
4) @Rhuagh You previously said that you're opposed to callbacks, but that's essentially what this PR is introducing, isn't it?
5) Should we eventually discuss a strategy for things like `Factory`, `Window`, ..?

Rhuagh: Yes, I'm opposed to callbacks, which this is not, this is essentially a Command. I do think we can do this in a number of different ways, in a more data-driven way, but I'm not sure if it's worth it.

Yes, we should discuss a strategy for framework local resources that the user needs to be able to control, but have no direct access to.

jojolepro: I delete the whole thing, we keep it, or we modify it? :P

jojolepro:
I delete the whole thing, we keep it, or we modify it? :P

Xaeroxe:
Another PR addressed this, so I'm closing this.

PR #387, opened 2017-10-02
Add WindowMessages resource

Xaeroxe:
This is my take on https://github.com/amethyst/amethyst/pull/386

So there's a few benefits to this approach:

- We can be optimistic and only allocate space for 2 boxes rather than being forced to be pessimistic and allocate space for 20 boxes.  Thanks to SmallVec, 99.99% of the time this won't even be an extra allocation.
- Resulting syntax is much less verbose
- We can use the same winit::Window type rather than having the type change every time the backend changes
- We can use FnMut instead of Fn.  Technically there's no reason we couldn't use FnOnce but rustc stable doesn't support it at the moment.  This is a small advantage as most of the time it won't be needed, but it's there in case my imagination has failed me.

cc @Rhuagh @jojolepro

### DIFF
--- a/amethyst_renderer/src/renderer.rs
+++ b/amethyst_renderer/src/renderer.rs
@@ -15,7 +15,7 @@ use pipe::{ColorBuffer, DepthBuffer, PipelineBuild, PipelineData, PolyPipeline, 
 use tex::{Texture, TextureBuilder};
 use types::{ColorFormat, DepthFormat, Device, Encoder, Factory, Window};
 use vertex::VertexFormat;
-use winit::{self, EventsLoop, WindowBuilder};
+use winit::{self, EventsLoop, Window as WinitWindow, WindowBuilder};
 
 /// Generic renderer.
 pub struct Renderer {
@@ -134,6 +134,25 @@ impl Renderer {
         self.window
             .swap_buffers()
             .expect("OpenGL context has been lost");
+    }
+
+    /// Retrieves an immutable borrow of the window.
+    ///
+    /// No operations require a mutable borrow as of 2017-10-02
+    #[cfg(feature = "opengl")]
+    pub fn window(&self) -> &WinitWindow {
+        self.window.window()
+    }
+
+    #[cfg(feature = "metal")]
+    #[cfg(feature = "vulkan")]
+    pub fn window(&self) -> &WinitWindow {
+        &self.window.0
+    }
+
+    #[cfg(feature = "d3d11")]
+    pub fn window(&self) -> &WinitWindow {
+        &*self.window.0
     }
 }
 
--- a/src/ecs/rendering/resources.rs
+++ b/src/ecs/rendering/resources.rs
@@ -8,6 +8,8 @@ use futures::sync::oneshot::{channel, Receiver, Sender};
 use gfx::traits::Pod;
 use renderer::{Error, Material, MaterialBuilder, Mesh, MeshBuilder, Texture, TextureBuilder};
 use renderer::Rgba;
+use smallvec::SmallVec;
+use winit::Window;
 
 pub(crate) trait Exec: Send + Sync {
     fn exec(self: Box<Self>, factory: &mut ::renderer::Factory);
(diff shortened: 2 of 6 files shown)


amethyst_renderer/src/renderer.rs
```diff
@@ -135,6 +135,23 @@ impl Renderer {
...
             .expect("OpenGL context has been lost");
     }
+
+    /// Retrieves an immutable borrow of the window.
+    ///
+    /// No operations require a mutable borrow as of 2017-10-02
+    #[cfg(feature = "opengl")]
+    pub fn window(&self) -> &WinitWindow {
+        self.window.window()
+    }
```
Rhuagh: Newline?

Xaeroxe: Added.

src/ecs/rendering/resources.rs
```diff
@@ -135,3 +137,33 @@ impl AsRef<Rgba> for AmbientColor {
         &self.0
     }
 }
+
+/// This specs resource with id 0 permits sending commands to the
+/// renderer internal window.
+pub struct WindowCommander {
+    // It's unlikely we'll get more than two commands per frame
+    // 2 Boxes also make this the same size as a Vec, so this costs
```
torkleyy: How is this true?

src/ecs/rendering/resources.rs
```diff
@@ -135,3 +137,33 @@ impl AsRef<Rgba> for AmbientColor {
...
+    // no more space in the structure than a Vec would.
+    //
+    // NOTE TO FUTURE AUTHORS: This could be an FnOnce but that's not possible
```
torkleyy: I solved this in https://github.com/slide-rs/specs/pull/214/files , maybe you can use that.

src/ecs/rendering/resources.rs
```diff
@@ -135,3 +137,33 @@ impl AsRef<Rgba> for AmbientColor {
...
+
+impl WindowCommander {
+    /// Create a new WindowCommander
+    pub fn new() -> Self {
+        Self {
+            queue: SmallVec::new(),
+        }
+    }
+
+    /// Execute this closure on the winit::Window next frame.
```
torkleyy: Missing backticks

src/ecs/rendering/resources.rs
```diff
@@ -135,3 +137,33 @@ impl AsRef<Rgba> for AmbientColor {
...
+    // no more space in the structure than a Vec would.
+    //
+    // NOTE TO FUTURE AUTHORS: This could be an FnOnce but that's not possible
+    // right now as of 2017-10-02 because FnOnce isn't object safe.  It might
+    // be possible as soon as FnBox stabilizes.  For now I'll use FnMut instead.
+    pub(crate) queue: SmallVec<[Box<FnMut(&Window) + Send + Sync + 'static>; 2]>,
+}
+
+impl WindowCommander {
+    /// Create a new WindowCommander
```
torkleyy: nit: missing backticks

src/ecs/rendering/resources.rs
```diff
@@ -135,3 +137,33 @@ impl AsRef<Rgba> for AmbientColor {
...
+    // 2 Boxes also make this the same size as a Vec, so this costs
+    // no more space in the structure than a Vec would.
+    //
+    // NOTE TO FUTURE AUTHORS: This could be an FnOnce but that's not possible
+    // right now as of 2017-10-02 because FnOnce isn't object safe.  It might
+    // be possible as soon as FnBox stabilizes.  For now I'll use FnMut instead.
+    pub(crate) queue: SmallVec<[Box<FnMut(&Window) + Send + Sync + 'static>; 2]>,
+}
+
+impl WindowCommander {
```
torkleyy: Meh, this name sounds more like a strategy RTS game's name. Could we rename it? Maybe `WindowMessages`?

Xaeroxe: Hahaha 😆 "Coming soon to stores near you!  Window Commander!!  Tell panes of glass to do things which they'll proceed to ignore because they're just glass!  Copyright 2017"

I'll rename it :P

src/ecs/rendering/resources.rs
Xaeroxe: Every vector is a combination of ptr, length, and capacity.  Each of these is `WORD` size long.  When a SmallVec is inline it does not store a ptr, or capacity, it just stores the given array type.  Box is `WORD` size long so if you make the inline array type (2 * `WORD` size) then even the inline variant doesn't take any more space than it would if it were on the heap.  Structure size isn't bloated, and you still get your smallvec optimization.

torkleyy: Oops, right. This comment was stupid.

torkleyy: The two-element array somehow made me think it's two times pointer-size.

src/ecs/rendering/resources.rs
Xaeroxe: Ha!  Not bad!  I like it.  I'll give it a shot.

src/ecs/rendering/resources.rs
omni-viral: Hmm. So the `SmallVec` copies data from stack to heap if you push more elements than array can hold... Is it a good choice? Why not storing only exceeding amount of elements on heap?
Two boxed traits have size of 4 pointers though.

Xaeroxe: Oh hm I guess that's true.  I'll remove my comment then.  As for storing what you can inline even when it spills, I had the same thought but then you can't operate on a SmallVec as if it were a slice because the memory is not contiguous.

Xaeroxe:
bors r=@torkleyy

Xaeroxe:
bors r=@torkleyy

Had to fix my commit message real quick.

PR #390, opened 2017-10-03
Refactor vertex format

omni-viral:
This little refactoring is aimed to increase use of assotiated consts, reduce overhead and simplify upcoming feature to use separate buffers for vertex attributes which will be preferable way to use meshes.

### DIFF
--- a/amethyst_renderer/src/vertex.rs
+++ b/amethyst_renderer/src/vertex.rs
@@ -1,50 +1,78 @@
 //! Built-in vertex formats.
 
-use std::any::Any;
-
-use gfx;
-use gfx::format::Format;
-use gfx::pso::buffer::Structure;
+use gfx::format::{ChannelType, Format, SurfaceType};
+use gfx::pso::buffer::Element;
 use gfx::traits::Pod;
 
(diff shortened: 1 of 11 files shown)


amethyst_renderer/src/vertex.rs
```diff
@@ -3,48 +3,66 @@
...
+use gfx::format::{ChannelType, Format, SurfaceType};
+use gfx::pso::buffer::Element;
 use gfx::traits::Pod;
 
-/// Handle to a vertex attribute.
-pub type Attribute = gfx::pso::buffer::Element<Format>;
+pub trait Attribute {
+    const NAME: &'static str;
+    const FORMAT: Format;
+    const SIZE: u32; // Has to be equal to `std::mem::size_of::<Self::Repr>() as u32`
```
torkleyy: Make comment a doc comment?

torkleyy: And in general, please add some docs.

omni-viral: I forgot to stage docs before commiting 😄

Issue #388, opened 2017-10-03
Support for vertex buffer per attribute


Rhuagh:
The most common scenario with gltf is that you have one buffer per vertex attribute, so we need support for that.

Rhuagh:
@omni-viral

omni-viral:
I'm on it

omni-viral:
https://github.com/amethyst/amethyst/pull/417

PR #391, opened 2017-10-04
Multibuffer

omni-viral:
Allow `Mesh` to have multiple buffers.
Passes fetches buffers by required formatted attributes.
Mostly it will be:
* One buffer with all attributes.
  Format differs for `PosNormTex` and `PosTex` even if only `Position` and `TexCoord` is queried by `Pass`. So if you have both you need `Pass` for each `VertexFormat`
* Each attribute in its own buffer.
  Single `Pass` can draw all meshes that have all attributes required. No matter if they have some more. No matter of ordering.
  @Rhuagh  reported that most of glTf assets contains buffer per attribute.

I haven't test it for more than one buffer yet 😛 

Bonus: I've removed some unnecessary `Sync` bounds.

### DIFF
--- a/amethyst_renderer/Cargo.toml
+++ b/amethyst_renderer/Cargo.toml
@@ -37,6 +37,7 @@ rayon = "0.7"
 rayon-core = "1.2"
 serde = "1.0"
 serde_derive = "1.0"
+smallvec = "0.4.2"
 winit = "0.7"
 
 gfx_device_gl = { version = "0.14", optional = true }
@@ -57,9 +58,6 @@ gfx_window_dxgi = { version = "0.7", optional = true }
 
 [dependencies.hetseq]
 version = "0.1.5"
-
-[dependencies.smallvec]
-version = "0.4.2"
 
 [dependencies.specs]
 version = "0.9.5"
--- a/amethyst_renderer/Cargo.toml
+++ b/amethyst_renderer/Cargo.toml
@@ -58,6 +58,9 @@ gfx_window_dxgi = { version = "0.7", optional = true }
 [dependencies.hetseq]
 version = "0.1.5"
 
+[dependencies.smallvec]
+version = "0.4.2"
+
 [dependencies.specs]
 version = "0.9.5"
 
--- a/amethyst_renderer/Cargo.toml
+++ b/amethyst_renderer/Cargo.toml
@@ -37,7 +37,6 @@ rayon = "0.7"
 rayon-core = "1.2"
 serde = "1.0"
 serde_derive = "1.0"
-smallvec = "0.4.2"
 winit = "0.7"
 
 gfx_device_gl = { version = "0.14", optional = true }
--- a/amethyst_renderer/src/mesh.rs
+++ b/amethyst_renderer/src/mesh.rs
@@ -1,6 +1,6 @@
 //! Mesh resource.
 
-use std::iter::{Chain, once, Once};
+use std::iter::{once, Chain, Once};
 use std::marker::PhantomData;
 
 use specs::{Component, DenseVecStorage};
@@ -35,15 +35,17 @@ pub trait VertexData {
 
 /// Construct new vertex data from raw data and vertex format
 pub fn vertex_data<D, V>(data: D) -> (D, PhantomData<V>)
-    where D: AsRef<[V]>,
-          V: VertexFormat,
+where
+    D: AsRef<[V]>,
+    V: VertexFormat,
 {
     (data, PhantomData)
 }
 
 impl<D, V> VertexData for (D, PhantomData<V>)
-    where D: AsRef<[V]>,
-          V: VertexFormat,
+where
+    D: AsRef<[V]>,
+    V: VertexFormat,
 {
     const ATTRIBUTES: Attributes<'static> = V::ATTRIBUTES;
 
@@ -53,7 +55,7 @@ impl<D, V> VertexData for (D, PhantomData<V>)
 
     fn build(&self, factory: &mut Factory) -> Result<VertexBuffer> {
         use gfx::Factory;
-        use gfx::memory::{Bind, cast_slice};
+        use gfx::memory::{cast_slice, Bind};
         use gfx::buffer::Role;
 
         let verts = self.0.as_ref();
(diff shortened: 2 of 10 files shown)


amethyst_renderer/Cargo.toml
```diff
@@ -58,6 +58,9 @@ gfx_window_dxgi = { version = "0.7", optional = true }
 [dependencies.hetseq]
 version = "0.1.5"
 
+[dependencies.smallvec]
```
torkleyy: Why aren't these in the `dependencies` table?

amethyst_renderer/src/mesh.rs
```diff
@@ -1,44 +1,173 @@
...
-    pub fn attributes(&self) -> &[(&'static str, AttributeFormat)] {
-        self.attrs
+        // None of the vertex buffers match requested attributes
+        None
     }
 
-    /// Returns the mesh's vertex buffer and associated buffer slice.
-    pub fn geometry(&self) -> (&RawBuffer, &Slice) {
-        (&self.vbuf, &self.slice)
+    /// Returns assotiated `Slice`
```
torkleyy: "assotiated"

amethyst_renderer/src/mesh.rs
```diff
@@ -51,29 +180,50 @@ impl Mesh {
     }
 }
 
+impl Component for Mesh {
```
torkleyy: :+1:

amethyst_renderer/src/vertex.rs
```diff
@@ -284,3 +296,36 @@ impl With<TexCoord> for PosNormTangTex {
...
+
+        impl_query!(@ $($a),*);
+    };
+    (@) => {};
+    (@ $head:ident $(,$tail:ident)*) => {
+        impl_query!($($tail),*);
+    };
+}
+
+impl_query!(A, B, C, D);
```
torkleyy: Sorry, I really don't know what `Query` is used for and what kind of maximum this macro invocation is responsible for. I lost track of what's going on in our renderer lately :(

amethyst_renderer/Cargo.toml
omni-viral: There was a feature. Now it's gone 😄

amethyst_renderer/src/vertex.rs
omni-viral: It just compose some static const slice of named attribute formats from`With` trait.
We have only 4 different attributes right now.
Maybe I can add some more here. There is no cost.

omni-viral: Look how passes use it

omni-viral: > I lost track of what's going on in our renderer lately :(

You can't keep up with everything 🏃

Xaeroxe:
Looks great!  I'll approve this as soon as we update the examples.

omni-viral:
> update the examples.

It is expected to be compatible

amethyst_renderer/src/mesh.rs
```diff
@@ -1,44 +1,181 @@
...
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug)]
 pub struct Mesh {
-    attrs: &'static [(&'static str, AttributeFormat)],
-    prim: Primitive,
     slice: Slice,
     transform: Matrix4<f32>,
-    vbuf: RawBuffer,
+    vbufs: SmallVec<[VertexBuffer; 4]>,
 }
```
Rhuagh: 4 is gonna be too low later on when we start looking at animation and morphing. For skeletal animation you need atleast bone ids and bone weights per vertex.

amethyst_renderer/src/mesh.rs
Rhuagh: Yay

amethyst_renderer/src/mesh.rs
omni-viral: 4 here isn't a limit. You can push more. They just will be on heap.

torkleyy: Still, we should be sure to handle the common cases. I think we're using too much `SmallVec` in our crate; I've seen a benchmark in the past which showed that you shouldn't use `SmallVec` in most cases. Cannot remember where though.

Xaeroxe: SmallVec is optimized for a very specific use case.  It should only be used when, for the overwhelming majority of the time, your collection will be under that many elements.  For example SmallVec works well for the input system because It is exceedingly rare that more than 8 keys will be pressed simultaneously.  It also works well for the WindowMessage system because it is incredibly unlikely you'll receive more than one WindowMessage per frame.  I don't think this however is an appropriate use case for SmallVec, as it sounds like the need for more VertexBuffers is going to grow quite rapidly.

omni-viral: Actually there is no difference what to use here `Vec` or `SmallVec`. Because each `push` is associated with creating of `gfx::RawBuffer` which is much slower operation.

omni-viral: Rapidly growing demand of vertex buffers isn't a good thing.
3 buffers with `[f32; 3]`, 1 with `[f32; 2]` will be the most case for static objects.
Additional 4 pairs of `u16` and `f32` for skeletal animation.
And that's it for 99.9% of meshes.

Rhuagh: If so, then 4 is fine

amethyst_renderer/src/mesh.rs
```diff
@@ -1,44 +1,181 @@
...
 
 impl Mesh {
     /// Builds a new mesh from the given vertices.
-    pub fn build<D, V>(verts: D) -> MeshBuilder<D, V>
-    where
-        D: AsRef<[V]>,
-        V: VertexFormat,
+    pub fn build<D, V>(verts: D) -> MeshBuilder<((D, PhantomData<V>), ())>
+        where D: AsRef<[V]>,
```
torkleyy: This is definitely not the new rustfmt style

Rhuagh: Add `--all` att the end

omni-viral: I like old format :frowning:

amethyst_renderer/src/vertex.rs
```diff
@@ -328,4 +328,31 @@ macro_rules! impl_query {
...
+    Q,
+    R,
+    S,
+    T,
+    U,
+    V,
+    W,
+    X,
+    Y,
+    Z
```
Xaeroxe: Going for a new high score in LOC? :P hahaha rustfmt is funny sometimes.

Xaeroxe:
@torkleyy r? UI Pass is going to have to rewrite some parts of the pipeline, since the pipeline doesn't currently support serial iteration.  So I'd like to get this merged as I'm planning on it being part of the UI PR.

omni-viral:
https://github.com/amethyst/amethyst/issues/388

PR #394, opened 2017-10-07
Filter axis-motion events

Xaeroxe:
@Binero reported their devices were sending a large volume of axis motion events, causing the winit event buffer to overflow.  This should help.  That degree of precision is completely useless as we can only meaningfully process one of those events per frame, so this PR ensures that only the most recent event per device, per axis, per frame, is actually sent to the buffer.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -204,38 +204,72 @@ impl<'a, 'b> Application<'a, 'b> {
     fn compress_events(vec: &mut Vec<Event>, new_event: Event) {
         match new_event {
             Event::WindowEvent { ref event, .. } => match event {
-                &WindowEvent::MouseMoved { .. } => for mut stored_event in vec.iter_mut() {
-                    if let &mut Event::WindowEvent {
-                        event: WindowEvent::MouseMoved { .. },
-                        ..
-                    } = stored_event
-                    {
-                        mem::replace(stored_event, new_event.clone());
-                        return;
+                &WindowEvent::MouseMoved { .. } => {
+                    let mut iter = vec.iter_mut();
+                    while let Some(stored_event) = iter.next_back() {
+                        match stored_event {
+                            &mut Event::WindowEvent {
+                                event: WindowEvent::MouseMoved { .. },
+                                ..
+                            } => {
+                                mem::replace(stored_event, new_event.clone());
+                                return;
+                            }
+
+                            &mut Event::WindowEvent {
+                                event: WindowEvent::AxisMotion { .. },
+                                ..
+                            } => {}
+
+                            &mut Event::DeviceEvent {
+                                event: DeviceEvent::Motion { .. },
+                                ..
+                            } => {}
+
+                            _ => {
+                                break;
+                            }
+                        }
                     }
-                },
+                }
 
                 &WindowEvent::AxisMotion {
                     device_id,
                     axis,
                     value,
-                } => for mut stored_event in vec.iter_mut() {
-                    if let &mut Event::WindowEvent {
-                        event:
-                            WindowEvent::AxisMotion {
-                                axis: stored_axis,
-                                device_id: stored_device,
-                                value: ref mut stored_value,
+                } => {
+                    let mut iter = vec.iter_mut();
+                    while let Some(stored_event) = iter.next_back() {
+                        match stored_event {
+                            &mut Event::WindowEvent {
+                                event:
+                                    WindowEvent::AxisMotion {
+                                        axis: stored_axis,
+                                        device_id: stored_device,
+                                        value: ref mut stored_value,
+                                    },
+                                ..
+                            } => if device_id == stored_device && axis == stored_axis {
+                                *stored_value += value;
+                                return;
                             },
-                        ..
-                    } = stored_event
-                    {
-                        if device_id == stored_device && axis == stored_axis {
-                            *stored_value += value;
-                            return;
+
+                            &mut Event::WindowEvent {
+                                event: WindowEvent::MouseMoved { .. },
+                                ..
+                            } => {}
+
+                            &mut Event::DeviceEvent {
+                                event: DeviceEvent::Motion { .. },
+                                ..
+                            } => {}
+
+                            _ => {
+                                break;
+                            }
                         }
                     }
-                },
+                }
 
                 _ => {}
             },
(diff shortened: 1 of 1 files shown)


src/app.rs
```diff
@@ -155,14 +156,94 @@ impl<'a, 'b> Application<'a, 'b> {
...
+            });
+
+            if let Some(mut event_handler) =
+                engine.world.res.try_fetch_mut::<EventHandler<Event>>(0)
+            {
+                for event in &events {
                     event_handler.write_single(event.clone());
                 }
+            }
+            for event in events.drain(..) {
```
torkleyy: nit: I think this could be just `events`

Xaeroxe:
I'm actually not sure of that, but if that's the case then this is still possible you just have to add the cumulative deltas together in one final event.

Rhuagh:
I'm 99% sure of it, and AxisMotion is in pixels, while DeviceEvent Motion is based on the controller. But yes, accumulate, just remember to split on device id for device, and stop and emit when seeing different event types, or you may get inaccuracies

src/app.rs
```diff
@@ -155,14 +156,94 @@ impl<'a, 'b> Application<'a, 'b> {
...
+            // against previous events and determine if we need to discard older events
+            //
+            // NOTE: Type is explicit here to make sure we're not pushing references in later code.
+            let mut events: Vec<Event> = Vec::new();
+            self.events.poll_events(|new_event| {
+                // This boolean helps us avoid writing the same piece of code over and over.
+                let mut write_new_event = true;
+                match new_event {
+                    Event::WindowEvent { ref event, .. } => match event {
+                        &WindowEvent::MouseMoved { .. } => for mut stored_event in &mut events {
```
torkleyy: You're saying that events are too many, but with this we have to iterate over all so far collected events *for every new event*. Is that a good thing?

src/app.rs
```diff
@@ -155,14 +156,94 @@ impl<'a, 'b> Application<'a, 'b> {
...
+                            if let &mut Event::WindowEvent {
+                                event: WindowEvent::MouseMoved { .. },
+                                ..
+                            } = stored_event
+                            {
+                                mem::replace(stored_event, new_event.clone());
+                                write_new_event = false;
```
omni-viral: Just return here.

src/app.rs
```diff
@@ -155,14 +156,94 @@ impl<'a, 'b> Application<'a, 'b> {
...
+                                        axis: stored_axis,
+                                        device_id: stored_device,
+                                        ..
+                                    },
+                                ..
+                            } = stored_event
+                            {
+                                if device_id == stored_device && axis == stored_axis {
+                                    mem::replace(stored_event, new_event.clone());
+                                    write_new_event = false;
```
omni-viral: Just return here.

src/app.rs
```diff
@@ -155,14 +156,94 @@ impl<'a, 'b> Application<'a, 'b> {
...
+                            device_id: stored_device,
+                            event:
+                                DeviceEvent::Motion {
+                                    axis: stored_axis, ..
+                                },
+                        } = stored_event
+                        {
+                            if device_id == stored_device && axis == stored_axis {
+                                mem::replace(stored_event, new_event.clone());
+                                write_new_event = false;
```
omni-viral: Just return here.

src/app.rs
```diff
@@ -155,14 +156,94 @@ impl<'a, 'b> Application<'a, 'b> {
...
+                if write_new_event {
+                    events.push(new_event);
+                }
+            });
+
+            if let Some(mut event_handler) =
+                engine.world.res.try_fetch_mut::<EventHandler<Event>>(0)
+            {
+                for event in &events {
                     event_handler.write_single(event.clone());
```
omni-viral: Isn't there a function to write multiple? It can be faster.

src/app.rs
omni-viral: No, you don't iterate over them. Because you don't push them here in the first place.

Xaeroxe: This vec should be kept small, the intent is that this vec contains all the events we write to the buffer, so instead of writing thousands we're now only writing one per input axis.

src/app.rs
```diff
@@ -184,20 +184,20 @@ impl<'a, 'b> Application<'a, 'b> {
                         },
 
                         &WindowEvent::AxisMotion {
-                            device_id, axis, ..
+                            device_id, axis, value
                         } => for mut stored_event in &mut events {
                             if let &mut Event::WindowEvent {
```
omni-viral: Maybe it's better to store different types of events separately, instead of `if let` each event?

Xaeroxe: Store them in more vecs?  That's probably more inefficient given this vec's purpose is to be small.

src/app.rs
```diff
@@ -268,6 +197,74 @@ impl<'a, 'b> Application<'a, 'b> {
         self.engine.world.maintain();
     }
 
+    /// Input devices can sometimes generate a lot of motion events per frame, these are
+    /// useless as the extra precision is wasted and these events tend to overflow our
+    /// otherwise very adequate event buffers.  So this function removes and compresses redundant
+    /// events.
+    fn compress_events(vec: &mut Vec<Event>, new_event: Event) {
```
torkleyy: Much better!

Xaeroxe:
@Rhuagh This PR now accumulates events with relative information.

Xaeroxe:
@Rhuagh Self-requested review on this, do not merge until he has approved.

src/app.rs
```diff
@@ -155,14 +156,24 @@ impl<'a, 'b> Application<'a, 'b> {
...
+            let mut events: Vec<Event> = Vec::new();
+            self.events.poll_events(|new_event| {
+                Self::compress_events(&mut events, new_event);
+            });
+
+            if let Some(mut event_handler) =
+                engine.world.res.try_fetch_mut::<EventHandler<Event>>(0)
+            {
+                if let Err(err) = event_handler.write_slice(&events) {
+                    eprintln!(
```
torkleyy: We should really add a logging system

Rhuagh:
Both `WindowEvent::AxisMotion` and  `DeviceEvent::Motion` give delta values. The first one does atleast, I've tested that quite thourougly in my own input handler, and based on @Binero's logs the same happens for the second one.

There is one issue with the given solution as I see it: It reorders the input events for the frame. A mouse click early in the list and a mouse click late in the list will both now be mouse clicks late in the list. Might not be that noticeable, I'm not sure.

Rhuagh:
MouseMoved gives an absolute position btw.

src/app.rs
```diff
@@ -186,6 +197,74 @@ impl<'a, 'b> Application<'a, 'b> {
...
+    /// Input devices can sometimes generate a lot of motion events per frame, these are
+    /// useless as the extra precision is wasted and these events tend to overflow our
+    /// otherwise very adequate event buffers.  So this function removes and compresses redundant
+    /// events.
+    fn compress_events(vec: &mut Vec<Event>, new_event: Event) {
+        match new_event {
+            Event::WindowEvent { ref event, .. } => match event {
+                &WindowEvent::MouseMoved { .. } => for mut stored_event in vec.iter_mut() {
+                    if let &mut Event::WindowEvent {
+                        event: WindowEvent::MouseMoved { .. },
```
omni-viral: There is `device_id` as well.

Xaeroxe: Gotta filter by mouse pointer so we can support this feature: https://youtu.be/UiLSiqyDf4Y :P 

I don't feel it's necessary to filter MouseMoved by device the event came from.

Xaeroxe:
@torkleyy I'd love to but it's a little difficult to produce these events in a headless environment, they require DeviceIds and WindowIds which are opaque types that can only be constructed within the winit crate.

PR #397, opened 2017-10-08
Use `FlaggedStorage` for `Transform` component.

Rhuagh:
`TransformSystem`will clear the flags. If `TransformSystem` is not used, the user is in charge of clearing the flags.


src/ecs/transform/systems.rs
```diff
@@ -76,10 +80,13 @@ impl<'a> System<'a> for TransformSystem {
 
         {
             // Compute transforms without parents.
-            for (ent, local, global, _) in (&*entities, &locals, &mut globals, !&children).join() {
+            for (ent, local, _, _) in (&*entities, &locals, &globals.check(), !&children).join() {
```
Aceeri: Probably use `RestrictedStorage` here to avoid checks.

amethyst_assets/Cargo.toml
```diff
@@ -25,5 +25,5 @@ fnv = "1"
 futures = "0.1.14"
 log = "0.3.8"
 parking_lot = "0.4.4"
-rayon = "0.7"
-specs = { version = "0.9.5", features = ["common"] }
+rayon = "0.8"
+specs = { version = "0.10", features = ["common"] }
```
torkleyy: We should also use the `serde` feature, which is what @omni-viral implemented

torkleyy: No reason to stop bors ^^

PR #398, opened 2017-10-10
Refactored pong example

Binero:
The example has been refactored to be split up into more modules. This will hopefully help new users to understand how the example works more easily. 

Most notably, the amount of (custom) systems went up from just one big, intimidating system to four small, easy to understand systems. A lot of functionality has been moved into helper functions as well.


Binero:
Maybe the most important change is that the ball is now a circle, not a square. 🎉

Xaeroxe:
This looks great!  Thanks!  However I have encountered a bug, sometimes the ball will just start colliding with whatever surface it runs into repeatedly, as if the collision is registering but the change in velocity doesn't cause it to escape the collision bounds.  This causes multiple problems, such as the music halting due to the massive sound overflow (side note I need to add limits for how many sounds can play at once.)  and the ball just generally being stuck where it is.  If you play for about 5 minutes or so I imagine you'll probably see it too.  Keep in mind that it's entirely possible for the ball to be inside an object further than the velocity per frame permits it to move which probably causes this problem.

examples/04_pong/systems/bounce.rs
```diff
@@ -0,0 +1,78 @@
...
+        // signficantly larger than the second one, as that will make the second time step
+        // unable to move the ball back out, causing its velocity to be negated again, and again...
+        for (ball, transform) in (&mut balls, &transforms).join() {
+            use ARENA_HEIGHT;
+
+            let ball_x = transform.translation[0];
+            let ball_y = transform.translation[1];
+
+            // Bounce at the top of the bottom of the arena.
+            if ball_y <= ball.radius {
```
Xaeroxe: I would actually take this a step further and also check to see if the ball is traveling towards the boundary in question by checking the current velocity.  For example if the ball is hitting the top wall make sure it has a positive Y velocity before executing the bounce code.  The only reason this matters is to avoid multiple bounce sound effects being played per collision.

examples/04_pong/systems/bounce.rs
```diff
@@ -0,0 +1,78 @@
...
+        // a ball being stuck inside the wall. This happens when the first time step is
+        // signficantly larger than the second one, as that will make the second time step
+        // unable to move the ball back out, causing its velocity to be negated again, and again...
+        for (ball, transform) in (&mut balls, &transforms).join() {
+            use ARENA_HEIGHT;
+
+            let ball_x = transform.translation[0];
+            let ball_y = transform.translation[1];
+
+            // Bounce at the top of the bottom of the arena.
```
Xaeroxe: I believe this should read "top or the bottom"

Xaeroxe:
@torkleyy I would have loved to fit pong into one file myself, but the sheer volume of code that was in that one file was making the one file not exactly a straightforward beginning to end read.  I think a lot of this will be alleviated by the addition of a proper physics system instead of having to code a very pong specific physics setup.  It could also be aided by the use of external files to describe meshes and materials.

examples/04_pong/pong.rs
```diff
@@ -0,0 +1,241 @@
...
+            radius: BALL_RADIUS,
+            velocity: [BALL_VELOCITY_X, BALL_VELOCITY_Y],
+        })
+        .with(local_transform)
+        .with(Transform::default())
+        .build();
+}
+
+
+fn load_asset<T, F>(world: &mut World, f: F) -> AssetFuture<T::Item>
```
jojolepro: I heard that there is a wrapper for this in amethyst now. Might be worth looking into it ;)

Binero: I heard [the same](https://github.com/amethyst/amethyst/pull/399#discussion_r143707137). I think it's best to wait till the two PRs are merged, and then fix all the examples at once.

PR #399, opened 2017-10-10
Refactored and updated sphere example to latest renderer.

Binero:
The shading works again on the sphere example. Took the opportunity to split different tasks into different functions so new users can have an easier time understanding the code.


omni-viral:
Great!
I guess it doesn't look like flag of Bangladesh anymore.

examples/02_sphere/main.rs
```diff
@@ -1,95 +1,51 @@
...
+const POINT_LIGHT_COLOUR: Rgba = Rgba(1.0, 1.0, 1.0, 1.0); // white
+const BACKGROUND_COLOUR: [f32; 4] = [0.0, 0.0, 0.0, 0.0]; // black
+const LIGHT_POSITION: [f32; 3] = [2.0, 2.0, -2.0];
+const LIGHT_RADIUS: f32 = 5.0;
+const LIGHT_INTENSITY: f32 = 3.0;
+
 struct Example;
 
-fn load_proc_asset<T, F>(engine: &mut Engine, f: F) -> AssetFuture<T::Item>
+fn load_proc_asset<T, F>(world: &mut World, f: F) -> AssetFuture<T::Item>
```
torkleyy: We should really replace this with the new asset management methods, which I think I've already added.

Binero: All examples use these functions. I think it might be better to solve them all in a separate PR after this and the pong PR merged.

torkleyy: As long as we can get rid of creating this function in every example, I'm fine with that.

PR #401, opened 2017-10-10
[WIP] Refactor asset system to use Streams instead of Futures

omni-viral:
This is a step toward simplifying assets API.


jojolepro:
Would it be possible to remove the overhead of declaring the Future by the user, while keeping it internally?
That would remove the overhead that we see here:
https://github.com/amethyst/amethyst/blob/develop/examples/06_material/main.rs#L25
and here
https://github.com/amethyst/amethyst/blob/develop/examples/05_assets/main.rs#L256
and here
https://github.com/amethyst/amethyst/blob/develop/examples/05_assets/main.rs#L273

Its annoying to declare manually so many futures, when in 99% of the use cases we want them to be used, and we don't provide a clean wrapper for them.

Rhuagh:
I'm all for simple API, but I'll let someone with more knowledge of futures and assets review this.

Rhuagh:
@jojolepro that is going away soon, we just need to fix the examples (the code to swap to already exists).

torkleyy:
Labeling as `type: RFC`. This PR should get more attention than usual PRs, which includes asking the non-members if they like it.

amethyst_assets/src/stream.rs
```diff
@@ -0,0 +1,319 @@
...
+        }
+    }
+}
+
+impl<S> Drop for UpdatesStream<S>
+where
+    S: Stream,
+{
+    fn drop(&mut self) {
+        let mut waiters = self.inner.notifier.waiters.lock().unwrap();
```
torkleyy: We need tests for this sort of stuff!

jojolepro:
I'm not 100% sure of everything that this change implies. I understand that it allows partial and async loading, but how does that allows hot swapping once the stream ended?

amethyst_assets/src/stream.rs
omni-viral: I've copied it from `futures` crate. Should be fine :smile:
`futures::future::Shared` is doing the same and we use them.
I have an idea for making all this much simplier and maybe faster but I need linear types to enforce required contract on type level.

omni-viral:
> I understand that it allows partial and async loading

Nope.
It's all about hot swapping. Each element in stream is full asset instance.

> hot reloading is only a development feature.

And streams would be simple futures with feature turned off. I haven't wrote it yet :smirk: Basically `AssetHandler` will be either `AssetFuture` or `AssetStream` (naming suugestions are welcome) depending on cargo feature.

omni-viral:
@jojolepro Streams aren't very handy for partial loading as they have no `seek` operation.

torkleyy:
We also need a `Merge` system for that, in the Specs `common` module.

torkleyy:
I'd really love an architecture where we don't need all the locking..

omni-viral:
> I'd really love an architecture where we don't need all the locking..

@torkleyy me too, but lock-free algorithms sometimes prove to be slower.
It always depends on use case.

torkleyy:
I wonder if there is a fundamental thing that forces us to actually use locks or any atomic thing. My feeling is that the assets crate as it currently is is rather heavy. I'd like a lightweight but still flexible one. What I'm saying here is not very related with this PR, I'm just wondering if we can do something like this. When I rewrote Amethyst's previous asset management, I did it because it was heavy for example. I failed at improving that point though.

Rhuagh:
I'm gonna close this as we have just completely reworked the asset system. Feel free to open again if you think I'm wrong.

PR #404, opened 2017-10-11
Reduce cpu usage

Aceeri:
Sleeps for a given amount of time and then yields until the frame is up.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -120,7 +120,16 @@ impl<'a, 'b> Application<'a, 'b> {
         self.timer.start();
         while self.states.is_running() {
             self.advance_frame();
+
             // We are done with this frame, we can just rest for now.
+            if self.timer.elapsed() < self.dur_per_frame {
+                let sleep_time = self.dur_per_frame - self.timer.elapsed();
+
+                if let Some(amount) = sleep_time.checked_sub(Duration::from_millis(1)) {
+                    thread::sleep(amount);
+                }
+            }
+
             while self.timer.elapsed() < self.dur_per_frame {
                 thread::yield_now();
             }


Rhuagh:
I've had 12ms sleep on Windows using a similar approach

torkleyy:
@Aceeri I had the very same idea before, it was rejected because of what @Rhuagh said.

OvermindDL1:
> I've had 12ms sleep on Windows using a similar approach

That is standard, the windows system event poller (which includes timers) only polls about once every 12ms.

Rhuagh:
So sleeping is out of the question on Windows, we can't have an engine that sleeps for 75% of the frame budget. We could do OS gating I guess?

Xaeroxe:
I'm wary of this as sleep accuracy historically hasn't been the best, but if we can collect data proving in what circumstances this does or does not work (preferably across multiple machines, operating systems, and processors) then prove this doesn't adversely impact frame scheduling then it sounds good to me.

Rhuagh:
Yeah, that's what we're doing currently, but it takes up 30% CPU in "idle mode" for other OS.

OvermindDL1:
> Yeah, that's what we're doing currently, but it takes up 30% CPU in "idle mode" for other OS.

Is that with yield?

I just did some more research, apparently on windows using `nanosleep(0)` just does a 'useless context switch that returns immediately' thus scheduling time to other programs but getting back to you asap (I.E. it does not put you on the end of the windows workqueue like `yield` does).

Rhuagh:
Yeah, we use yield in a loop until the min frame duration has passed.

OvermindDL1:
> Yeah, we use yield in a loop until the min frame duration has passed.

In that case it seems the cpu usage is just from the program being rescheduled instantly because nothing else needs to run, seems fine then.  You can always add a configuration option to sleep instead if you want (just be sure to 'catch up' if it sleeps longer than expected) for power-sensitive devices like laptops.

Rhuagh:
Yeah, that's what I'm thinking. We will have to do catchup logic though, and not just look on the current frame duration.

OvermindDL1:
> Yeah, that's what I'm thinking. We will have to do catchup logic though, and not just look on the current frame duration.

Just be sure to take into account what happens if the system is put to sleep or hibernated too!

(I had a lot of special checks and tests for if the count suddenly got 'too high' in my C++ engine, among other things, it is a bit of a rat's nest over the years, but it works...)

Rhuagh:
A cap on catchup logic should handle that OK.

Aceeri:
I'm not seeing these 12ms sleeps that you are getting, tested it on 4 computers too (Windows, mac, and 2 linux machines)

OvermindDL1:
> I'm not seeing these 12ms sleeps that you are getting, tested it on 4 computers too (Windows, mac, and 2 linux machines)

12ms is the 'max' it should be, it will be between 0ms to 12ms.  And that is only on windows.
Linux does not have such an issue, it has a better timing system.
No clue about Mac.  ^.^;

Rhuagh:
Mac is close to Linux, with the exception that yield  suck if there is nothing to yield to

Xaeroxe:
@Aceeri Since you approved the PR made by @Rhuagh I'm going to close this, please re-open if I'm in the wrong.

PR #406, opened 2017-10-11
Making RenderSystem responsible for pumping input events into the event channel

Rhuagh:
Also going up to shrev 0.6.


examples/04_pong/systems/paddle.rs
```diff
@@ -9,10 +9,12 @@ use amethyst::timing::Time;
 pub struct PaddleSystem;
 
 impl<'s> System<'s> for PaddleSystem {
-    type SystemData = (WriteStorage<'s, Paddle>,
-     WriteStorage<'s, LocalTransform>,
-     Fetch<'s, Time>,
-     Fetch<'s, InputHandler<String, String>>);
+    type SystemData = (
+        WriteStorage<'s, Paddle>,
```
torkleyy: Somebody changed this before, now it's changed back. The question is whose formatting tool is the one we agreed on.

examples/04_pong/systems/paddle.rs
```diff
@@ -35,10 +37,9 @@ impl<'s> System<'s> for PaddleSystem {
                 transform.translation[1] += paddle.velocity * delta_time * movement as f32;
 
                 // We make sure the paddle remains in the arena.
-                transform.translation[1] = transform.translation[1].max(0.0).min(
```
torkleyy: Same here. It seems people are using different versions.

examples/04_pong/systems/paddle.rs
Rhuagh: I think @Binero might have forgotten to do it after the last changes of the Pong PR.

src/ecs/rendering/systems.rs
```diff
@@ -49,5 +56,135 @@ where
...
+            );
+        }
+    }
+}
+
+/// Input devices can sometimes generate a lot of motion events per frame, these are
+/// useless as the extra precision is wasted and these events tend to overflow our
+/// otherwise very adequate event buffers.  So this function removes and compresses redundant
+/// events.
+fn compress_events(vec: &mut Vec<Event>, new_event: Event) {
```
jojolepro: We could have that in a generic input crate some day.

Rhuagh: No, I think compression should be as close to the source as possible to not flood the system with unnecessary events.

Xaeroxe:
So I know this is technically out of scope but there's one more thing I was wondering if you could squeeze in here, I would like it if the winit event buffer was 1000 events large instead of 200.  @jojolepro has found that for frames that take an exceptionally long time 200 isn't quite enough.  At the time when he filed the issue I said that game performance when the game is running at 5 FPS isn't really something I'm concerned with.  I've had a little bit of a change of heart though, as input events getting lost could cause the input to become "stuck" and one of my least favorite bugs from Minecraft is when my input spontaneously gets "stuck" after a single long frame despite being released.  I figure if at the engine level we can avoid that kind of behavior we probably should.  @jojolepro when you were running at 5 FPS how many events were you losing per frame? Do you think 1000 is enough for the buffer?

Rhuagh:
Do we want the same on `InputEvent`?

Rhuagh:
I managed to screw up so it got closed.

Xaeroxe:
Thanks @Rhuagh!  Looks great!

jojolepro:
I don't know how many events I was losing, I wasn't getting warnings in the console :/

Rhuagh:
If you didn't get messages on the console, it wasn't overflowing.

PR #407, opened 2017-10-11
Adding a configurable frame limiter.

Rhuagh:


src/util/frame_limiter.rs
```diff
@@ -0,0 +1,123 @@
+//! Frame limiter
+
+use std::thread::{sleep, yield_now};
+use std::time::{Duration, Instant};
+
+/// Frame rate limiting strategy
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub enum FrameRateLimitStrategy {
+    /// No limit, spin as fast as possible
```
Xaeroxe: Would like to add "WARNING: THIS MAY PRODUCE SYSTEM INSTABILITY DUE TO CPU STARVATION"

src/app.rs
```diff
@@ -120,10 +119,9 @@ impl<'a, 'b> Application<'a, 'b> {
         self.timer.start();
         while self.states.is_running() {
             self.advance_frame();
+
             // We are done with this frame, we can just rest for now.
```
Xaeroxe: Might want to update this comment since we're not necessarily resting here.

src/util/frame_limiter.rs
Xaeroxe: Alternatively: If the strategy for this variant was to call `yield_now` it might not be as much of a concern.

Rhuagh:
Fixed @Xaeroxe.

Rhuagh:
If needed, we can even do separate impl Default for different platforms.

PR #417, opened 2017-10-16
Adding loading of Mesh with any combination of separate vertex buffers

Rhuagh:
Fixes https://github.com/amethyst/amethyst/issues/388


Rhuagh:
It's @omni-viral's macro btw, don't want to take credit for that :P

torkleyy:
Which combinations are there we diddn't cover?

Rhuagh:
The old implementation only have support for interleaved in the same vertex buffer.

Rhuagh:
And the current solution also blocks the user from inventing their own vertex attributes

Rhuagh:
I realised this is missing pass support, so I'm gonna work some more on this, so wait with reviewing it.

torkleyy:
Wow, do we need this for every pass? And if so, can we use the new passes from the old ones?

Rhuagh:
No, because attributes in the "old" passes are interleaved in a single vertex buffer, the new ones use a vertex buffer per attribuets.

This is ready for review now.

omni-viral:
Related issue: https://github.com/amethyst/amethyst/issues/388

Xaeroxe:
40 minutes, from 24 hours, 4 approvals, good enough.

bors r+

PR #423, opened 2017-10-17
Allow passes to fetch conflicting `SystemData`

torkleyy:
This is needed in order to implement caching as said by @omni-viral


omni-viral:
Won't it panic in case there is real conflict?

torkleyy:
You can have conflicts between passes and that'll be just fine I think, since they're sequential, aren't they?

torkleyy:
Or at least that was what I had assumed, maybe I'm wrong.

omni-viral:
No. They all run in parallel

torkleyy:
Okay, sorry. I thought that they'd only be themselves parallel, but apperantly they're also executed in parallel.

PR #425, opened 2017-10-17
[rs] 06: Moving timing and the frame limiter to core, and moving circular buffer and fps counter to a new utils crate

Rhuagh:


Rhuagh:
I know the formatting is off a little on imports, I will remedy that in the next PR set. I didn't want to incur conflicts with the other active PR.

src/util/time.rs
```diff
@@ -1,13 +0,0 @@
-//! Util time conversion functions
-
-use std::time::Duration;
-
-///Converts a Duration to the time in seconds.
-pub fn duration_to_secs(duration: Duration) -> f32 {
```
Xaeroxe: I like this function, seems like it didn't get added back.  Can we put it back somewhere?

Rhuagh: Ok, it wasn't used anywhere was all.

Xaeroxe: Mostly a visibility problem :/ it should be used.

Rhuagh: A grep on the project showed no uses. The other method was used in one place only, so I moved it there. But I can return this module in the core crate no worries. Will be tomorrow though.

Xaeroxe: What I meant was it works great in places like this: https://github.com/amethyst/amethyst/blob/develop/examples/04_pong/systems/move_balls.rs#L18 it just wasn't used there because we all forgot it existed.

Xaeroxe: (Last night while working on my prototype I was wanting something like this, I totally forgot we already had it)

Rhuagh: Oki, will remedy tomorrow.

Rhuagh: Will move it to the timing module

Rhuagh:
Review comments addressed, please review again @Xaeroxe

Xaeroxe:
Handled in #430

PR #432, opened 2017-10-19
[feat] Adding utilities to Camera interface for quickly building a simple camera

Rhuagh:


Rhuagh:
I agree, but that'll be changed when we do the camera refactoring

PR #437, opened 2017-10-21
Standardize both default and new methods

### LINKED ISSUE #356: Incoherence between using ::new and ::default
From a piece of code I wrote, not all of them are here.

```
Components:
LocalTransform::default()
Transform::default()
MeshComponent::new()

Resources:
Time::default()
InputHandler::new()

Systems:
TransformSystem::new()

Other stuff:
pass::DrawFlat::::new()
Dj::new(&output)
MaterialBuilder::new()
BoxedErr::new
```

Aceeri:
Fixes #356 

As a side note, I noticed that a lot of the `amethyst_input` impl blocks were kind of over restricted (requiring `Clone` when only 2 methods really need it), should we just move those out to new blocks with the extra restrictions?


amethyst_input/src/bindings.rs
```diff
@@ -19,10 +19,20 @@ where
     pub(super) actions: HashMap<AC, SmallVec<[Button; 4]>>,
 }
 
+impl<AX, AC> Default for Bindings<AX, AC>
```
torkleyy: Use derivative?

amethyst_input/src/bindings.rs
```diff
@@ -19,10 +19,20 @@ where
...
+where
+    AX: Hash + Eq,
+    AC: Hash + Eq,
+{
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
torkleyy: Please place trait impls after struct impl

amethyst_input/src/bundle.rs
```diff
@@ -29,6 +29,7 @@ use {Bindings, InputEvent, InputHandler, InputSystem};
 ///
 /// No errors returned from this bundle.
 ///
+#[derive(Default)]
```
torkleyy: Should use derivative

amethyst_input/src/input_handler.rs
```diff
@@ -15,7 +15,6 @@ use super::event::InputEvent::*;
 ///
 /// For example, if a key is pressed on the keyboard, this struct will record
 /// that the key is pressed until it is released again.
-#[derive(Default)]
```
torkleyy: Should use derivative

amethyst_renderer/src/pass/flat/interleaved.rs
```diff
@@ -29,6 +29,16 @@ pub struct DrawFlat<V> {
     _pd: PhantomData<V>,
 }
 
+impl<V> Default for DrawFlat<V>
+where
+    V: Query<(Position, TexCoord)>,
+    Self: Pass,
+{
+    fn default() -> Self {
+        Self::new()
```
torkleyy: See above, swap new and default impl

amethyst_renderer/src/pass/flat/separate.rs
```diff
@@ -24,6 +24,15 @@ use vertex::{Position, Separate, TexCoord, VertexFormat};
 #[derive(Clone, Debug, PartialEq)]
 pub struct DrawFlatSeparate;
 
+impl Default for DrawFlatSeparate
+where
+    Self: Pass,
+{
+    fn default() -> Self {
+        Self::new()
```
torkleyy: Same here

amethyst_renderer/src/pass/pbm/interleaved.rs
```diff
@@ -31,6 +31,15 @@ pub struct DrawPbm<V> {
     _pd: PhantomData<V>,
 }
 
+impl<V> Default for DrawPbm<V>
+where
+    V: Query<(Position, Normal, Tangent, TexCoord)>,
+{
+    fn default() -> Self {
+        Self::new()
```
torkleyy: Same here

amethyst_renderer/src/pass/pbm/separate.rs
```diff
@@ -24,13 +24,13 @@ use types::Encoder;
 use vertex::{Normal, Position, Separate, Tangent, TexCoord, VertexFormat};
 
 /// Draw mesh with physically based lighting
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Default, Clone, Debug, PartialEq)]
 pub struct DrawPbmSeparate;
```
torkleyy: We should add a private dummy field here

amethyst_renderer/src/pipe/pipe.rs
```diff
@@ -227,6 +227,12 @@ pub struct PipelineBuilder<Q> {
     targets: Vec<TargetBuilder>,
 }
 
+impl Default for PipelineBuilder<Queue<()>> {
+    fn default() -> Self {
+        Self::new()
```
torkleyy: Same here

amethyst_assets/src/storage.rs
```diff
@@ -36,6 +36,12 @@ pub struct AssetStorage<A: Asset> {
     unused_handles: MsQueue<Handle<A>>,
 }
 
+impl<A: Asset> Default for AssetStorage<A> {
+    fn default() -> Self {
+        Self::new()
```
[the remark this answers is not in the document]
Aceeri: The standard seems to be reverse, except if deriving Default, but who knows.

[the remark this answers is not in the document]
torkleyy: There's not really a standard. Clippy recommends it the way I prefer, rust-lang does it the other way around. This pattern is used in Specs, has been used so far in Amethyst AFAIK and you also used it in some places, so I thought unifying it would be best.

Rhuagh:
I prefer new called from default. I think conceptually default is a special case of new.

torkleyy:
`new` is a convenience function for `default` :wink:

torkleyy:
Also, if you derive `Default`, you'll have to do it that way. So it's more consistent.

Rhuagh:
That may be, I still feel it's backwards.

Xaeroxe:
>  I noticed that a lot of the `amethyst_input` impl blocks were kind of over restricted 

The clone restrictions are necessary to generate the events, the whole event channel falls apart without it.

amethyst_input/src/bindings.rs
```diff
@@ -9,7 +9,8 @@ use smallvec::SmallVec;
 use super::{Axis, Button};
 
 /// Used for saving and loading input settings.
-#[derive(Serialize, Deserialize, Clone)]
+#[derive(Derivative, Serialize, Deserialize, Clone)]
+#[derivative(Default(bound = "AX: Hash + Eq, AC: Hash + Eq"))]
```
torkleyy: Why does derive not work here?

Aceeri: Because `fn new()` is implemented for those bounds.

Aceeri: Because otherwise we'd have different semantics between Default and `fn new()` depending on bounds. I don't know if we want to unrestrict `fn new()` or not?

Rhuagh:
r? @torkleyy 

Also needs rebase.

amethyst_input/src/bindings.rs
torkleyy: I usually don't put bounds on the new function except for when I don't want to create another impl block.

Issue #396, opened 2017-10-08
Make crates more self-contained and introduce `amethyst_core`


torkleyy:
A few days ago I came up with an idea, hey why don't we split up Amethyst and make the creates just utility crates for Specs. While some liked the idea, @Xaeroxe was quite skeptical. And I share that skepticism. That's why I came up with a compromise:

Right now we have our crates `amethyst_assets`, `amethyst_renderer`, both of which contain their own functionality but are then wrapped in the main crate. We should instead put everything related to assets in the assets crate and the same for the renderer. This would not only solve constant complications with the orphan rules, but also allow us to make those crates optional, while still allowing easy integration and a cohesive experience thanks to the concept of ECS. Now, there are however things many Amethyst crates will share, e.g. `Transform`. These things should go into `amethyst_core`.

## What should happen to the main crate?

I'm not sure if a main crate should remain, I've experimented with making [a more generic state machine lately](https://github.com/rustgd/machina) and the `ecs` module would essentially get dropped. Timer stuff probably goes into the core crate and fps counters and profiling tools could go into an `amethyst_tools` crate.

omni-viral:
The link is slightly broken

torkleyy:
Fixed now, sorry

Rhuagh:
Yay!

Xaeroxe:
@torkleyy I guess I'm not entirely certain how this proposal solves our orphan rules problem.  The purpose of the renderer crate is to render assets on screen, while the purpose of the assets crate is to allow new asset types to be defined.  So when @omni-viral wants to implement `amethyst_assets::Asset` for `amethyst_renderer::Mesh` I'm still not sure how he'd do that under this proposal.  Does the proposal just accept such an implementation won't be possible?  Because if so we have two choices:

- Wrap Mesh ourselves in `amethyst_core` and implement Asset for the wrapper (similar to what we've already one)
- Ask downstream users to do the same for their own projects, which is essentially asking them to write the same piece of code for every project just to satisfy our need to separate concerns.

Basically I'm questioning the wisdom of trying to separate concerns too heavily, because these concerns might not be that separate after all.

torkleyy:
Why not make `amethyst_renderer` dependent on `amethyet_assets`?

Xaeroxe:
That works.  That was a relatively straightforward fix haha :) Then sure I have no problem with this if we're allowing `amethyst_*` crates to depend on `amethst_*` crates other than `amethyst_core`

[...]

PR #438, opened 2017-10-21
[rs] 09: Flatten crate structure of amethyst_renderer

torkleyy:
Partially addresses #396


src/prelude.rs
```diff
@@ -3,5 +3,5 @@
 pub use app::{Application, ApplicationBuilder};
 pub use config::Config;
 pub use engine::Engine;
-pub use renderer::input::*;
+//pub use renderer::input::*;
```
torkleyy: I'm not yet sure what we should do with this prelude module, let's decide once [rs] PRs are merged

amethyst_renderer/src/prelude.rs
```diff
@@ -1,19 +0,0 @@
-//! Contains common types that can be glob-imported (`*`) for convenience.
```
Rhuagh: Convention seem to be the prelude containing traits,  i.e. things that are used implicitly, but not directly instantiated.

src/prelude.rs
Rhuagh: See above

amethyst_renderer/src/prelude.rs
torkleyy: I heard that, yeah, but I think that many libraries have different preludes with different usages.

amethyst_renderer/src/config.rs
```diff
@@ -4,7 +4,7 @@ use winit::WindowBuilder;
 
 /// Structure for holding the renderer configuration.
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
-pub struct Config {
+pub struct DisplayConfig {
```
Xaeroxe: Thanks for renaming this!

PR #440, opened 2017-10-21
Refactoring Camera as a Component

Rhuagh:
Render pass use Transform of Camera entity to compute the view matrix.

### DIFF
--- a/amethyst_renderer/src/pass/flat/interleaved.rs
+++ b/amethyst_renderer/src/pass/flat/interleaved.rs
@@ -4,12 +4,12 @@ use std::marker::PhantomData;
 
 use amethyst_assets::AssetStorage;
 use amethyst_core::transform::Transform;
-use cgmath::{Matrix4, One};
+use cgmath::{Matrix4, One, SquareMatrix};
 use gfx::pso::buffer::ElemStride;
 
 use rayon::iter::ParallelIterator;
 use rayon::iter::internal::UnindexedConsumer;
-use specs::{Fetch, ParJoin, ReadStorage};
+use specs::{Fetch, Join, ParJoin, ReadStorage};
 
 use super::*;
 use cam::Camera;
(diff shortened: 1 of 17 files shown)


amethyst_renderer/src/pass/flat/interleaved.rs
```diff
@@ -138,7 +140,8 @@ where
             ..
         } = self;
 
-        let camera = &camera;
+        // TODO: multiple cameras
+        let camera = (&camera, &global).join().next();
```
torkleyy: Maybe log an error in case `next().is_some()`

torkleyy: But then again I'm not sure if we should do this everywhere

minecrawler:
Hey, that's half my PR, however you cleared up one of my major pain points, so that's really awesome :D

1. If possible, could you add the ActiveCamera component we talked about over there, so multiple cameras are possible? Only always take the first ActiveCamera!
2. Shouldn't a camera have both, `Transform` and `LocalTransform`, for example in case you want the camera to follow a mesh? (I hope I got the purpose of `LocalTransform` right this time)

Rhuagh:
For 2. That's up to the user, a LocalTransform will always be converted to a Transform before arriving in the renderer, this is done by Transformsystem.

Rhuagh:
For 1. I'm still debating solutions to this problem

omni-viral:
`ActiveCamera` as optional resource is better. Otherwise changing active camera will have more boilerplate.

Rhuagh:
I think I'm gonna add `ActiveViewport` in a later PR

Rhuagh:
I will add the active camera resource for now, as optional, fallback to first camera if its not set, and fallback to identity from that

amethyst_renderer/src/pass/flat/interleaved.rs
```diff
@@ -140,8 +145,14 @@ where
             ..
         } = self;
 
-        // TODO: multiple cameras
-        let camera = (&camera, &global).join().next();
+        let camera: Option<(&Camera, &Transform)> = active
+            .and_then(|a| {
+                let cam = camera.get(a.entity);
+                let transform = global.get(a.entity);
+                cam.into_iter().zip(transform.into_iter()).next()
```
omni-viral: This is unusual but good alternative to `cam.and_then(|cam| transform.map(|transform| (cam, transform)))` 😄

Rhuagh: That one doesn't work because of borrow rules.

PR #441, opened 2017-10-22
Add UI framework

Xaeroxe:
Work done in this PR:

- Add UiTransform Component
- Add UiImage component
- Add font loading
- Add UiText component
- Add UiDraw pass


amethyst_ui/src/pass.rs
```diff
@@ -0,0 +1,237 @@
...
+use amethyst_renderer::pipe::pass::{Pass, PassApply, PassData, Supplier};
+use amethyst_renderer::Texture;
+use amethyst_renderer::Encoder;
+use amethyst_renderer::vertex::{PosTex, Position, Query, TexCoord};
+
+const VERT_SRC: &[u8] = include_bytes!("vertex.glsl");
+const FRAG_SRC: &[u8] = include_bytes!("frag.glsl");
+
+#[derive(Copy, Clone)]
+#[allow(dead_code)] // This is used by the shaders
```
torkleyy: Don't we need C layout then?

amethyst_ui/src/transform.rs
```diff
@@ -0,0 +1,23 @@
+use specs::{Component, DenseVecStorage};
+
+/// The raw pixels on screen that are populated.
+///
+/// TODO: Eventually this should be either replaced by a citrine type, or citrine may just
+/// populate it.
+pub struct UiTransform {
```
torkleyy: Maybe we should limit this to the size and take the position from `Transform`?

Xaeroxe: I feel it's probably important to keep the concepts separate as this is in pixels, whereas Transform uses unspecified units.

amethyst_ui/src/pass.rs
Xaeroxe: Yes, thank you

amethyst_ui/src/transform.rs
torkleyy: Okay, true. But if we already have that advantage, we should really do

```rust
pub enum Dimension {
    /// In pixels
    Absolute(f32),
    /// Multiplied with window width / height
    Relative(f32),
}
```

torkleyy: Also, I think `Bounds` would be more accurate.

Aceeri: I'm thinking `amethyst_ui` might just be glue between `citrine` and our renderer, although we could just merge this into `citrine` instead and move it into the amethyst repo.

Xaeroxe: @torkleyy I was mostly just trying to sidestep the design decisions we talked about on gitter before.  This structure's purpose is to be replaced, it's meant to be slightly sub-optimal.

Xaeroxe:
Example doesn't work yet, probably have to fix my broken shaders.

amethyst_ui/src/pass.rs
Xaeroxe: Added.

amethyst_renderer/src/pass/pbm/interleaved.rs
```diff
@@ -94,18 +94,10 @@ where
...
-            mesh_storage,
-            tex_storage,
-            material_defaults,
-            mesh,
-            material,
-            global,
-            light,
-        ): (
-            Option<Fetch<'a, Camera>>,
+        (active, camera, ambient, mesh_storage, tex_storage, material_defaults,
```
torkleyy: I'm not sure about this formatting, but well

Xaeroxe: That formatting was merged in from @Rhuagh 's PR, but I still have more code to write here so I'll give it a rustfmt pass again.

Rhuagh: That's weird. It shouldn't do that

torkleyy:
So what's the current status of this PR @Xaeroxe ?

Xaeroxe:
Still requires additional work, I won't have time to work on it for a few days though.

Xaeroxe:
(If anyone wants to push this to completion I gladly accept PRS on my PR)

Xaeroxe:
Z-ordering tested and works 👍 
Had to add resizing to blend targets to fix resize issue.

I'm now starting work on a text component.

Xaeroxe:
Text component is ready.  This is still super rough around the edges, font loading in particular is kind of hairy, but it does indeed work!

image

Xaeroxe:
image

Pong now displays score :D  Font loading has also been drastically simplified.

Xaeroxe:
Broken by upstream changes, fixing now.

amethyst_renderer/src/formats/texture.rs
```diff
@@ -163,7 +163,7 @@ impl PngFormat {
 }
 
 impl SimpleFormat<Texture> for PngFormat {
-    const NAME: &'static str = "JPEG";
+    const NAME: &'static str = "PNG";
```
Rhuagh: Nice catch :)

amethyst_renderer/src/pipe/pass.rs
```diff
@@ -158,11 +158,21 @@ impl<P> CompiledPass<P> {
 
     pub fn new_target(&mut self, target: &Target) {
         for effect in &mut self.effects {
+            // Distribute new no blend targets
```
Rhuagh: This comment confused me for a couple of seconds. I don't know how to improve it though.

amethyst_renderer/src/system.rs
```diff
@@ -89,14 +89,7 @@ where
...
 
-    fn do_render(
-        &mut self,
-        (mut event_handler, mut window_messages, mut screen_dimensions, data): RenderData<P>,
-    ) {
-        #[cfg(feature = "profiler")]
-        profile_scope!("render_system");
-        use std::time::Duration;
-
+    fn do_window_management(&mut self, (mut window_messages, mut screen_dimensions): WindowData) {
```
Rhuagh: Nice.

amethyst_ui/src/format.rs
```diff
@@ -0,0 +1,43 @@
...
+pub struct FontFormat;
+
+impl SimpleFormat<FontAsset> for FontFormat {
+    const NAME: &'static str = "FONT";
+    type Options = ();
+
+    fn import(&self, bytes: Vec<u8>, _: ()) -> Result<FontData, Error> {
+        FontCollection::from_bytes(bytes)
+            .into_fonts()
+            .nth(0)
```
Rhuagh: I guess only 1 font per file is supported for this iteration?

amethyst_ui/src/pass.rs
```diff
@@ -0,0 +1,278 @@
...
+                            None => continue,
+                        };
+                        let vertex_args = VertexArgs {
+                            proj_vec: proj_vec.into(),
+                            coord: [ui_transform.x, ui_transform.y],
+                            dimension: [ui_transform.width, ui_transform.height],
+                        };
+                        effect.update_constant_buffer("VertexArgs", &vertex_args, encoder);
+                        effect.data.vertex_bufs.push(vbuf);
+                        if let Some(ui_image) = ui_image.get(entity) {
```
Rhuagh: can't this be done with `if let Some(image) = ui_image.get(entity).and_then(|ui_image| tex_storage.get(&ui_image.texture)) { .. }` ?

amethyst_ui/src/pass.rs
```diff
@@ -0,0 +1,278 @@
...
+                            if let Some(image) = tex_storage.get(&ui_image.texture) {
+                                effect.data.textures.push(image.view().clone());
+                                effect.data.samplers.push(image.sampler().clone());
+                                effect.draw(mesh.slice(), encoder);
+                                effect.clear();
+                            }
+                        }
+
+                        if let Some(ui_text) = ui_text.get(entity) {
```
Rhuagh: See above.

amethyst_ui/src/vertex.glsl
```diff
@@ -0,0 +1,27 @@
...
+
+void main() {
+    vertex.position = vec4(position, 1);
+    vertex.position *= vec4(dimension, 1, 1);
+    vertex.position += vec4(coord, 0, 0);
+    vertex.position *= proj_vec;
+    vertex.position += vec4(-1, 1, 0, 0);
+    vertex.tex_coord = tex_coord;
+    gl_Position = vertex.position;
+}
```
Rhuagh: `vertex.position = vec4((position.xy * dimension.xy + coord.xy) * proj_vec.xy + vec2(-1, 1), position.z * -2, 1);` ?

examples/04_pong/main.rs
```diff
@@ -82,8 +77,20 @@ fn run() -> Result<()> {
         .with_bundle(TransformBundle::new().with_dep(&["ball_system", "paddle_system"]))?
         .with_bundle(AudioBundle::new(|music: &mut Music| music.music.next()))?
         .with_bundle(RenderBundle::new())?
-        .with_local(RenderSystem::build(pipe, Some(display_config))?);
-    Ok(game.build()?.run())
+        .with_bundle(UiBundle::new(&[]))?;
+
+    let pipe = {
```
Rhuagh: This makes me sad. I don't know how to fix it though :/

torkleyy: Basically by fixing the `ApplicationBuilder` I guess.

amethyst_renderer/src/pipe/pass.rs
Xaeroxe: How about "new targets that don't blend"?

amethyst_renderer/src/system.rs
Xaeroxe: I needed it in order to access the ScreenDimensions in a pass :P

amethyst_ui/src/format.rs
Xaeroxe: Correct, with a note that if this doesn't work for people they should complain.

amethyst_ui/src/pass.rs
Xaeroxe: Probably, it just didn't occur to me :P

amethyst_ui/src/vertex.glsl
Xaeroxe: Works, but it makes the line really long :P I tried to cut it down so the line wasn't so long.

Xaeroxe:
Since we've determined we don't want to fix pass initialization until the gfx-hal rewrite is complete

r? @Rhuagh @torkleyy

amethyst_renderer/src/formats/texture.rs
Xaeroxe: I only caught it because the string was concatenated into a very confusing error message :P

torkleyy: Which confusing error message? Is my error handling code broken?

Aceeri:
@Xaeroxe Do we want to keep the placeholder position component for now and just integrate the citrine Position one later?

torkleyy:
I'll give this a full review tomorrow.

amethyst_ui/src/pass.rs
```diff
@@ -0,0 +1,279 @@
...
+use gfx::pso::buffer::ElemStride;
+use gfx::state::ColorMask;
+use hibitset::BitSet;
+
+use rayon::iter::ParallelIterator;
+use rayon::iter::internal::UnindexedConsumer;
+use specs::{Entities, Entity, Fetch, Join, ParJoin, ReadStorage};
+
+use super::*;
+use amethyst_renderer::{Mesh, MeshHandle};
```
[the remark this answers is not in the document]
Xaeroxe: Yes

Xaeroxe:
@Aceeri Probably yes, since citrine has a little ways to go before integrating into here and this is better than the nothing we have in Amethyst right now.

amethyst_renderer/src/formats/texture.rs
Xaeroxe: @torkleyy no no, it's fine, it just told me it couldn't load a JPEG when I was trying to load a PNG, because this string was concatenated into the message.  I appreciate your vigilance, but if your asset loading produced confusing error messages for me I would have complained :)

amethyst_renderer/src/system.rs
Rhuagh: Yeah, I realised that. But it also improved the state of `RenderSystem`, so it's all good :)

amethyst_renderer/src/system.rs
```diff
@@ -164,6 +163,7 @@ where
 {
     fn run_now(&mut self, res: &'a Resources) {
         self.do_asset_loading(AssetLoadingData::<'a>::fetch(res, 0));
+        self.do_window_management(WindowData::<'a>::fetch(res, 0));
         self.do_render(RenderData::<'a, P>::fetch(res, 0));
```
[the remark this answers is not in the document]
Xaeroxe: I was just following the convention set from before, but sure I can do that.

examples/04_pong/systems/winner.rs
```diff
@@ -34,10 +37,20 @@ impl<'s> System<'s> for WinnerSystem {
             let did_hit = if ball_x <= ball.radius {
                 // Right player scored on the left side.
                 score_board.score_right += 1;
+                for (transform, text) in (&ui_transform, &mut text).join() {
+                    if "P2" == transform.id {
```
torkleyy: Entities should be used for this. While having a name for entities (which I btw. planned to implement in Specs: you would be able to just do `world.create_entity().name("Blah")` then) is nice, I don't think we should recommend it for component identification purposes.

examples/09_ui/main.rs
```diff
@@ -0,0 +1,226 @@
...
+                         PngFormat, PointLight, PosNormTex, RenderBundle, RenderSystem, Rgba,
+                         ScreenDimensions, Stage};
+use amethyst::ui::{DrawUi, TtfFormat, UiBundle, UiImage, UiText, UiTransform};
+use amethyst::winit::{Event, KeyboardInput, VirtualKeyCode, WindowEvent};
+use genmesh::{MapToVertices, Triangulate, Vertices};
+use genmesh::generators::SphereUV;
+
+const SPHERE_COLOUR: [f32; 4] = [0.0, 0.0, 1.0, 1.0]; // blue
+const AMBIENT_LIGHT_COLOUR: Rgba = Rgba(0.01, 0.01, 0.01, 1.0); // near-black
+const POINT_LIGHT_COLOUR: Rgba = Rgba(1.0, 1.0, 1.0, 1.0); // white
```
torkleyy: Just a note, I think the `Rgba` structure is redundant.

examples/09_ui/main.rs
```diff
@@ -0,0 +1,226 @@
...
+        initialise_camera(&mut engine.world);
+        let (logo, font) = {
+            let loader = engine.world.read_resource::<Loader>();
+
+            let logo = loader.load(
+                "texture/logo_transparent.png",
+                PngFormat,
+                Default::default(),
+                (),
+                &engine.world.read_resource(),
```
torkleyy: Let's not call this method twice; please introduce a binding above.

examples/09_ui/main.rs
```diff
@@ -0,0 +1,226 @@
...
+
+fn run() -> Result<(), amethyst::Error> {
+    let display_config_path = format!(
+        "{}/examples/09_ui/resources/display.ron",
+        env!("CARGO_MANIFEST_DIR")
+    );
+
+    let resources = format!("{}/examples/assets", env!("CARGO_MANIFEST_DIR"));
+
+
```
torkleyy: Please remove the extra lines.

amethyst_renderer/src/system.rs
```diff
@@ -89,14 +89,7 @@ where
...
 
-    fn do_render(
-        &mut self,
-        (mut event_handler, mut window_messages, mut screen_dimensions, data): RenderData<P>,
-    ) {
-        #[cfg(feature = "profiler")]
-        profile_scope!("render_system");
-        use std::time::Duration;
-
+    fn window_management(&mut self, (mut window_messages, mut screen_dimensions): WindowData) {
```
torkleyy: It might be worth noting that this PR changes the profiling in a way such that window management isn't included anymore. Maybe we should have an extra scope for it?

amethyst_renderer/src/system.rs
```diff
@@ -163,8 +162,9 @@ where
     P: PolyPipeline,
 {
     fn run_now(&mut self, res: &'a Resources) {
-        self.do_asset_loading(AssetLoadingData::<'a>::fetch(res, 0));
-        self.do_render(RenderData::<'a, P>::fetch(res, 0));
+        self.asset_loading(AssetLoadingData::<'a>::fetch(res, 0));
+        self.window_management(WindowData::<'a>::fetch(res, 0));
+        self.render(RenderData::<'a, P>::fetch(res, 0));
```
torkleyy: Not relevant to this addition, but the `::<'a>` is unnecessary.

amethyst_renderer/src/tex.rs
```diff
@@ -159,7 +160,7 @@ where
         let mut data = self.data.as_ref();
 
         if cfg!(feature = "opengl") {
-            let pixel_width = (self.info.format.get_total_bits() / 8) as usize;
+            let pixel_width = (self.info.format.get_total_bits() / 8) as usize / size_of::<T>();
```
torkleyy: Good catch!

amethyst_ui/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "amethyst_ui"
+version = "0.1.0"
+authors = ["Jacob Kiesel <kieseljake@gmail.com>"]
```
torkleyy: We'll need some more metadata in order to release it, you wanna do that in a follow-up?

amethyst_ui/src/format.rs
```diff
@@ -0,0 +1,52 @@
...
+/// Loads font files, supports TrueType and **some** OpenType files.
+///
+/// OpenType is a superset of TrueType, so if your OpenType file uses any features that don't
+/// exist in TrueType this will fail.  This will only load the first font contained in a file.
+/// If this is a problem for you please file an issue with Amethyst on GitHub.
+#[derive(Clone)]
+pub struct TtfFormat;
+
+impl SimpleFormat<FontAsset> for TtfFormat {
+    const NAME: &'static str = "TTF";
```
torkleyy: Maybe this should be `"TTF/OTF"`?

amethyst_ui/src/format.rs
```diff
@@ -0,0 +1,52 @@
...
+    type Options = ();
+
+    fn import(&self, bytes: Vec<u8>, _: ()) -> Result<FontData, Error> {
+        FontCollection::from_bytes(bytes)
+            .into_fonts()
+            .nth(0)
+            .map(|f| FontData(f))
+            .ok_or(Error::from_kind(ErrorKind::Format("Font parsing error")))
```
torkleyy: Should be `chain_err(|| "Font parsing error")`. Is there a way to get a more detailed error message when loading font collections?

amethyst_ui/src/pass.rs
```diff
@@ -0,0 +1,274 @@
...
+        let cached_draw_order = &cached_draw_order;
+
+        let proj_vec = vec2(
+            2. / screen_dimensions.width(),
+            -2. / screen_dimensions.height(),
+        ).extend(-2.)
+            .extend(1.);
+
+        supplier
+            .supply(bitset.par_join().map(move |_id| {
```
torkleyy: Couldn't this just use a range? Something like `(0..1)`?

amethyst_ui/src/pass.rs
```diff
@@ -0,0 +1,274 @@
...
+                move |encoder: &mut Encoder, effect: &mut Effect| for &(_z, entity) in
+                    cached_draw_order.iter()
```
torkleyy: nit: no need for `iter`

amethyst_ui/src/pass.rs
```diff
@@ -0,0 +1,274 @@
...
+                {
+                    // These are safe as we guaranteed earlier these entities are present.
+                    let ui_transform = ui_transform.get(entity).unwrap();
```
torkleyy: I'm not sure, maybe the `RestrictStorage` could work here?

amethyst_ui/src/pass.rs
```diff
@@ -0,0 +1,274 @@
...
+            -2. / screen_dimensions.height(),
+        ).extend(-2.)
+            .extend(1.);
+
+        supplier
+            .supply(bitset.par_join().map(move |_id| {
+                move |encoder: &mut Encoder, effect: &mut Effect| for &(_z, entity) in
+                    cached_draw_order.iter()
+                {
+                    // These are safe as we guaranteed earlier these entities are present.
```
torkleyy: Technically, there's no way safe code couldn't be safe. If my suggestion below doesn't work, please reword this.

amethyst_ui/src/pass.rs
```diff
@@ -0,0 +1,274 @@
...
+                            .get(entity)
+                            .and_then(|image| tex_storage.get(&image.texture))
+                        {
+                            effect.data.textures.push(image.view().clone());
+                            effect.data.samplers.push(image.sampler().clone());
+                            effect.draw(mesh.slice(), encoder);
+                            effect.clear();
+                        }
+
+                        if let Some(image) = ui_text
```
torkleyy: nit: `image` should be `text`

amethyst_ui/src/shaders/vertex.glsl
```diff
@@ -0,0 +1,27 @@
+// TODO: Needs documentation.
+
+#version 150 core
+
+layout (std140) uniform VertexArgs {
```
torkleyy: I don't know what `std140` does, could you add a comment, please?

amethyst_ui/src/pass.rs
```diff
@@ -0,0 +1,274 @@
...
+        // pass in the first iteration.
+        let mut bitset = BitSet::new();
+        bitset.add(0);
+
+        let cached_draw_order = &cached_draw_order;
+
+        let proj_vec = vec2(
+            2. / screen_dimensions.width(),
+            -2. / screen_dimensions.height(),
+        ).extend(-2.)
```
torkleyy: This is confusing to me.

amethyst_ui/src/shaders/vertex.glsl
```diff
@@ -0,0 +1,27 @@
...
+  vec4 position;
+  vec2 tex_coord;
+} vertex;
+
+void main() {
+    vertex.position = vec4(position, 1);
+    vertex.position *= vec4(dimension, 1, 1);
+    vertex.position += vec4(coord, 0, 0);
+    vertex.position *= proj_vec;
+    vertex.position += vec4(-1, 1, 0, 0);
```
torkleyy: Why aren't you just using a matrix?

amethyst_ui/src/pass.rs
```diff
@@ -0,0 +1,274 @@
...
+            .supply(bitset.par_join().map(move |_id| {
+                move |encoder: &mut Encoder, effect: &mut Effect| for &(_z, entity) in
+                    cached_draw_order.iter()
+                {
+                    // These are safe as we guaranteed earlier these entities are present.
+                    let ui_transform = ui_transform.get(entity).unwrap();
+                    if let Some(mesh) = mesh_storage.get(unit_mesh) {
+                        let vbuf = match mesh.buffer(PosTex::ATTRIBUTES) {
+                            Some(vbuf) => vbuf.clone(),
+                            None => continue,
```
torkleyy: If there's no `unit_mesh`, we can just `return`.

torkleyy: The same applies for the check above, let's return and decrease the indention.

amethyst_ui/src/text.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    /// Gets a mutable reference to the string to render.
+    ///
+    /// Avoid calling this if you don't have to, as calls to this invalidate the text render cache.
+    pub fn text_mut(&mut self) -> &mut String {
+        self.dirty = true;
+        &mut self.text
+    }
+
+    /// The RGBA color with a maximum of 1.0 and a minimum of 0.0 for each channel of the text.
```
torkleyy: Thanks for the clarification!

amethyst_ui/src/text.rs
```diff
@@ -0,0 +1,166 @@
...
+    type SystemData = (
+        ReadStorage<'a, UiTransform>,
+        WriteStorage<'a, UiText>,
+        Fetch<'a, Loader>,
+        Fetch<'a, AssetStorage<Texture>>,
+        Fetch<'a, AssetStorage<FontAsset>>,
+    );
+
+    fn run(&mut self, (transform, mut text, loader, tex_storage, font_storage): Self::SystemData) {
+        for (transform, text) in (&transform, &mut text).join() {
```
torkleyy: I'm not sure if this is a more general problem, but if a text doesn't have a transform, it will be silently ignored. No need to fix it here, just wanted to note.

torkleyy: Please filter and flat_map this, otherwise the indention is too much.

amethyst_renderer/src/system.rs
Xaeroxe: Oh yeah that is important.  Thanks!

amethyst_ui/Cargo.toml
Xaeroxe: I'd rather do it here so I don't forget.  Can you give me suggestions on what kinds of things to add?

amethyst_ui/src/format.rs
Xaeroxe: Probably yes

amethyst_ui/src/format.rs
Xaeroxe: Not with the rusttype crate sadly.  I really wish that weren't the case.

amethyst_ui/src/pass.rs
Xaeroxe: Yeah I'm not sure using extend here was the best idea, so I'll fix this.

amethyst_ui/src/pass.rs
Xaeroxe: Helps bypass the borrow checker, removing iter here causes errors that I don't know of a better way to fix.

amethyst_ui/src/pass.rs
Xaeroxe: I feel it's important in this context to indicate we're drawing an image rendered by another system, that's why I chose image.

amethyst_ui/src/shaders/vertex.glsl
Xaeroxe: This is more efficient, less multiplying by 1 and adding 0.

amethyst_ui/src/transform.rs
```diff
@@ -0,0 +1,79 @@
...
+    pub z: f32,
+    /// The width of this UI element
+    pub width: f32,
+    /// The height of this UI element
+    pub height: f32,
+    /// This function is called with this element and the new size whenever the window is resized.
+    ///
+    /// Do not try and replace this while the inner function  is being called.  Whatever you put
+    /// here would be overwritten.
+    pub resize_fn: Option<Box<FnMut(&mut UiTransform, (f32, f32)) + Send + Sync>>,
```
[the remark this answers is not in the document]
Xaeroxe: That's not a bad idea @torkleyy , thanks!

examples/09_ui/main.rs
Xaeroxe: This was copy pasted from the sphere example, should I change it there too?

examples/09_ui/main.rs
Xaeroxe: They're not the same thing, the inferred type is different between the calls.  I'll add the type.

examples/04_pong/systems/winner.rs
Xaeroxe: I guess I like the getElementById from JavaScript and I was trying to mimic that here.  It's nice being able to just keep things in your head and request for static values rather than having to store the entity, but storing the entity is also more efficient, so I'll go with that.

amethyst_renderer/src/tex.rs
Xaeroxe: Yeah it was pretty confusing when my f32 textures were panicking here, I eventually figured out it was because size_of::<f32>() was greater than 1, which hasn't been true for any of our other textures thus far miraculously.

amethyst_ui/src/pass.rs
Xaeroxe: I don't think `ParJoin` is implemented for ranges though.

amethyst_ui/src/pass.rs
Xaeroxe: Storage isn't mutable so this doesn't help.

amethyst_ui/Cargo.toml
torkleyy:
```toml
description = "Amethyst UI crate"
keywords = ["ui", "specs", "ecs", "amethyst"]
categories = ["game-engines"]

documentation = "https://www.amethyst.rs/doc/develop/doc/amethyst_ui/index.html"
homepage = "https://www.amethyst.rs/"
license = "MIT/Apache-2.0"
```

amethyst_ui/src/pass.rs
torkleyy: But I guess `ParIter` is?

amethyst_ui/src/shaders/vertex.glsl
torkleyy: Ah I see. What's the default behavior? I can't recall myself writing anything other than

```c
struct MyStruct {
    int a;
}

uniform MyStruct my_struct;
```

amethyst_ui/src/shaders/vertex.glsl
torkleyy: I'm not sure if we can measure GPU performance in terms of how often a multiplication is happening, but rather how typical and thus optimized it is for a certain situation. So I don't know what's more efficient here, it's just unconventional.

examples/09_ui/main.rs
torkleyy: No, it's not related to this PR.

examples/09_ui/main.rs
torkleyy: Huh, didn't notice that.

amethyst_ui/src/pass.rs
torkleyy: Sorry, how does it not help? Why would it need to be mutable?

amethyst_ui/Cargo.toml
Rhuagh: You can get `cargo package`, I think that fails on missing metadata

amethyst_ui/src/pass.rs
Rhuagh: Move it to the shader imho, (the extend part)

amethyst_ui/src/shaders/vertex.glsl
Rhuagh: Vec4 operations are gonna be faster than s matrix for the most part

Xaeroxe: So I did this because @omni-viral and  @Rhuagh told me the projection matrix was excessive.

amethyst_ui/src/pass.rs
Xaeroxe: As you've already noted on gitter, RestrictedStorage only works with mutable storages.

amethyst_ui/src/shaders/vertex.glsl
Rhuagh: You can even reduce it further using vec2 and swizzle operators if you want.

amethyst_ui/Cargo.toml
Xaeroxe: Done.

amethyst_renderer/src/system.rs
Xaeroxe: Done.

amethyst_ui/src/pass.rs
```diff
@@ -0,0 +1,275 @@
...
+        let tex_storage = &tex_storage;
+        let ui_image = &ui_image;
+        let ui_text = &ui_text;
+        let ui_transform = &ui_transform;
+        let unit_mesh = &unit_mesh;
+
+        // Populate and update the draw order cache.
+        // TODO: Replace all of this with code taking advantage of specs::TrackedStorage.
+        // TrackedStorage doesn't exist yet but it will in a later version of specs.
+        cached_draw_order.retain(|&(_z, entity)| {
```
Rhuagh: Is it really worth it to cache this data and not just recalculate it per frame ? The loop below is best case n^2, because position will potentially go through all items in the cache, plus a sort that's n*log(n). Wouldn't it be faster to use `binary_search_by` and do insertion sort and create the list from scratch every frame? That would be a guaranteed n*log(n).

Xaeroxe: @Rhuagh Yeah you're right this eventually turned out to be not quite what was desired.  I have a couple ideas on how to fix it though.  First is to introduce a private AtomicBool to the UiTransform indicating if it was added to the cache or not, taking our n^2 operation down to just n.

Caching the sorting is valuable imo because sorting an already sorted collection is very cheap.

Rhuagh: I agree, it's just when the cache update is more expensive than rebuilding the whole thing I get suspicious :P TrackedStorage and insertion sort on z would turn it into nlogn worst case

Xaeroxe:
r? @Rhuagh 

I've tightened up the caching a lot, it uses bitset to be much more efficient now.

Rhuagh:
Much better, thanks

Xaeroxe:
@torkleyy After adding the new and improved cache I've determined TrackedStorage would actually be a detriment here instead of an advantage as the new cache is actually faster, (mostly because it takes advantage of things we can only do with this particular setup)

Xaeroxe:
Since bitset cloning is pretty cheap I'm going to

bors r+

Xaeroxe:
@torkleyy has informed me that bit set clones are not cheap.  I'll fix this in the morning

remram44:
The `ui` example shows me a dark blue sphere with nothing else. Closing the window causes a panic "InvalidFramebufferOperation".

screen shot 2017-11-01 at 13 58 06

The `pong` example doesn't display any text either.

Xaeroxe:
Oh right, the blend targets still don't work on Mac.  I forgot about that.  I'll fix it before we release.  That's not a problem with anything UI related, it's a problem with using a rendering pass with blending, interestingly this only occurs on OSX.

Xaeroxe:
I wrote up #473 to address this.

PR #442, opened 2017-10-22
Add custom drop to `AssetStorage`

torkleyy:
As requested by @omni-viral

### DIFF
--- a/amethyst_assets/src/storage.rs
+++ b/amethyst_assets/src/storage.rs
@@ -112,9 +112,20 @@ impl<A: Asset> AssetStorage<A> {
     }
 
     /// Process finished asset data and maintain the storage.
-    pub fn process<F>(&mut self, mut f: F, errors: &Errors)
+    pub fn process<F>(&mut self, f: F, errors: &Errors)
     where
         F: FnMut(A::Data) -> Result<A, BoxedErr>,
+    {
+
+        self.process_custom_drop(f, |_| {}, errors);
+    }
+
+    /// Process finished asset data and maintain the storage.
+    /// This calls the `drop_fn` closure for assets that were removed from the storage.
+    pub fn process_custom_drop<F, D>(&mut self, mut f: F, mut drop_fn: D, errors: &Errors)
+        where
+            D: FnMut(A),
+            F: FnMut(A::Data) -> Result<A, BoxedErr>,
     {
         while let Some(processed) = self.processed.try_pop() {
             let Processed {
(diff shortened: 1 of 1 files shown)


amethyst_assets/src/storage.rs
```diff
@@ -148,7 +159,7 @@ impl<A: Asset> AssetStorage<A> {
             let old = self.handles.swap_remove(i);
```
omni-viral: You see how you have to iterate over all handles inefficiently (one line above). This is what can be optimized by feature I've requested here https://github.com/slide-rs/specs/issues/292

torkleyy: No, it's not. You just have to check the handles, it's how it works.

torkleyy: (I'll add some functionality which doesn't check that often, but the overall concept is fixed)

omni-viral: That what I was talking about. If you have custom drop in specs you can check only dropped handles

torkleyy: I think I know what you're talking about, and if that's the case, your reasoning would assume that handles are only dropped after entity deletions. This, however, is not the case. Handles don't even have to be in the world and even if, they could also be removed without having the `Entity` deleted.

omni-viral: Good point. But you can have two ways to drop `Handle`s (or another resource holders). One which works in `Drop` implementation but adds some significant overhead. And another that require access to some context (`AssetStorage` in case of `Handle`s) but has less overhead.
For instance you can have `MsQueue` to send inner data of dopped `Handle`s (if it's last reference) in `Drop` implementation. And `VecDeque` to put dropped `Handle`s of deleted entities.

torkleyy: I know there are more efficient ways, but coming back to the topic, can you agree that this is not a responsibility of Specs? And if so, do I get your approval on this PR?

omni-viral: Specs owns `Component`s when they are inside storages. And it silently drop them in `World::maintain` when their entities die. I'm sure it's responsibility of Specs to give user a way to grab those detached components instead of silent drop.

torkleyy: So what would be your plan? Control everything and force users to execute custom drop implementations for everything using or owning a handle?

omni-viral: Not force anything (until we have linear types). If user just drop component with `Handle` it will execute some non-optimal code to free resource correctly.
But `Handle<Mesh>` will be dropped with entity in 99.(9)% cases. And simple installation of custom drop function to the specs may increase performance a bit.

PR #444, opened 2017-10-22
Generifying mesh creation, so users can define their own mesh attribute combinations

Rhuagh:


Rhuagh:
How do you do that without having to handle 7 generic type parameters in multiple locations?

Rhuagh:
So, to do that is a much larger rewrite as I would have to change all mesh asset loading, which I would like to hold off on until the gfx-hal rewrite of the renderer is done, so as to avoid duplicate work. Can we accept this PR as an interim solution?

PR #445, opened 2017-10-22
Add asset hot reloading

### LINKED ISSUE #266: Asset hot reloading

torkleyy:
* Adds a new resource to the world, `Arc<rayon::ThreadPool>` (or short: `core::ThreadPool`)
* Adds `WeakHandle`s
* Automatically reloads assets (checked every second)

Fixes #266 

## Problems

* ~~If a file has been written only partially and the asset storage tries to load it, it fails :( I'm not sure if there's a check we can do~~ Fixed by using fallback

### DIFF
--- a/amethyst_assets/examples/hl.rs
+++ b/amethyst_assets/examples/hl.rs
@@ -99,10 +99,16 @@ impl<'a> System<'a> for RenderingSystem {
     type SystemData = (
         FetchMut<'a, AssetStorage<MeshAsset>>,
         Fetch<'a, Errors>,
-        Fetch<'a, Arc<ThreadPool>>, /* texture storage, transforms, .. */
+        Fetch<'a, Arc<ThreadPool>>,
+        Option<Fetch<'a, HotReloadStrategy>>,
+        /* texture storage, transforms, .. */
     );
 
-    fn run(&mut self, (mut mesh_storage, errors, pool): Self::SystemData) {
+    fn run(&mut self, (mut mesh_storage, errors, pool, strategy): Self::SystemData) {
+        use std::ops::Deref;
+
+        let strategy = strategy.as_ref().map(Deref::deref);
+
         mesh_storage.process(
             |vertex_data| {
                 // Upload vertex data to GPU and give back an asset
@@ -111,6 +117,7 @@ impl<'a> System<'a> for RenderingSystem {
             },
             &errors,
             &**pool,
+            strategy,
         );
     }
 }
--- a/amethyst_assets/examples/hl.rs
+++ b/amethyst_assets/examples/hl.rs
@@ -74,18 +74,18 @@ impl Asset for MeshAsset {
 }
 
 /// A format the mesh data could be stored with.
+#[derive(Clone)]
 struct Ron;
 
-impl Format<MeshAsset> for Ron {
+impl SimpleFormat<MeshAsset> for Ron {
     const NAME: &'static str = "RON";
 
     type Options = ();
 
-    fn import(&self, name: String, source: Arc<Source>, _: ()) -> Result<VertexData, BoxedErr> {
+    fn import(&self, bytes: Vec<u8>, _: ()) -> Result<VertexData, BoxedErr> {
         use ron::de::from_str;
         use std::str::from_utf8;
 
-        let bytes = source.load(&name)?;
         let s = from_utf8(&bytes).map_err(BoxedErr::new)?;
 
         from_str(s).map_err(BoxedErr::new)
--- a/amethyst_assets/examples/hl.rs
+++ b/amethyst_assets/examples/hl.rs
@@ -33,7 +33,8 @@ impl App {
 
         world.add_resource(Errors::new());
         world.add_resource(AssetStorage::<MeshAsset>::new());
-        world.add_resource(Loader::new(path, pool));
+        world.add_resource(Loader::new(path, pool.clone()));
+        world.add_resource(pool);
 
         App {
             dispatcher,
@@ -98,10 +99,10 @@ impl<'a> System<'a> for RenderingSystem {
     type SystemData = (
         FetchMut<'a, AssetStorage<MeshAsset>>,
         Fetch<'a, Errors>,
-                       /* texture storage, transforms, .. */
+        Fetch<'a, Arc<ThreadPool>>, /* texture storage, transforms, .. */
     );
 
-    fn run(&mut self, (mut mesh_storage, errors): Self::SystemData) {
+    fn run(&mut self, (mut mesh_storage, errors, pool): Self::SystemData) {
         mesh_storage.process(
             |vertex_data| {
                 // Upload vertex data to GPU and give back an asset
@@ -109,6 +110,7 @@ impl<'a> System<'a> for RenderingSystem {
                 Ok(MeshAsset { buffer: () })
             },
             &errors,
+            &**pool,
         );
     }
 }
--- a/amethyst_assets/examples/hl.rs
+++ b/amethyst_assets/examples/hl.rs
@@ -3,6 +3,7 @@
 #![allow(unused)]
 
 extern crate amethyst_assets;
+extern crate amethyst_core;
 extern crate rayon;
 extern crate ron;
 #[macro_use]
@@ -12,6 +13,7 @@ extern crate specs;
 use std::sync::Arc;
 
 use amethyst_assets::*;
+use amethyst_core::Time;
 use rayon::ThreadPool;
 use specs::{DenseVecStorage, Dispatcher, DispatcherBuilder, Fetch, FetchMut, System, World};
 use specs::common::Errors;
@@ -99,12 +101,13 @@ impl<'a> System<'a> for RenderingSystem {
     type SystemData = (
         FetchMut<'a, AssetStorage<MeshAsset>>,
         Fetch<'a, Errors>,
+        Fetch<'a, Time>,
         Fetch<'a, Arc<ThreadPool>>,
         Option<Fetch<'a, HotReloadStrategy>>,
         /* texture storage, transforms, .. */
     );
 
-    fn run(&mut self, (mut mesh_storage, errors, pool, strategy): Self::SystemData) {
+    fn run(&mut self, (mut mesh_storage, errors, time, pool, strategy): Self::SystemData) {
         use std::ops::Deref;
 
         let strategy = strategy.as_ref().map(Deref::deref);
@@ -116,6 +119,7 @@ impl<'a> System<'a> for RenderingSystem {
                 Ok(MeshAsset { buffer: () })
             },
             &errors,
+            time.frame_number,
             &**pool,
             strategy,
         );
--- a/amethyst_assets/src/asset.rs
+++ b/amethyst_assets/src/asset.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use specs::UnprotectedStorage;
 
-use {BoxedErr, Handle, Source};
+use {BoxedErr, Handle, Reload, SingleFile, Source};
 
 /// One of the three core traits of this crate.
 ///
@@ -38,10 +38,64 @@ pub trait Format<A: Asset>: Send + 'static {
     type Options: Send + 'static;
 
     /// Reads the given bytes and produces asset data.
+    ///
+    /// ## Reload
+    ///
+    /// The reload structure has metadata which allows the asset management
+    /// to reload assets if necessary (for hot reloading).
+    /// You should only create this if `create_reload` is `true`.
+    /// Also, the parameter is just a request, which means you can also return `None`.
     fn import(
         &self,
         name: String,
         source: Arc<Source>,
         options: Self::Options,
-    ) -> Result<A::Data, BoxedErr>;
+        create_reload: bool,
+    ) -> Result<(A::Data, Option<Box<Reload<A>>>), BoxedErr>;
+}
+
+/// This is a simplified version of `Format`, which doesn't give you as much as freedom,
+/// but in return is simpler to implement.
+/// All `SimpleFormat` types automatically implement `Format`.
+/// This format assumes that the asset name is the full path and also the only file.
+pub trait SimpleFormat<A: Asset> {
+    /// A unique identifier for this format.
+    const NAME: &'static str;
+    /// Options specific to the format, which are passed to `import`.
+    /// E.g. for textures this would be stuff like mipmap levels and
+    /// sampler info.
+    type Options: Clone + Send + Sync + 'static;
+
+    /// Produces asset data from given bytes.
+    fn import(&self, bytes: Vec<u8>, options: Self::Options) -> Result<A::Data, BoxedErr>;
+}
+
+impl<A, T> Format<A> for T
+where
+    A: Asset,
+    T: SimpleFormat<A> + Clone + Send + Sync + 'static,
+{
+    const NAME: &'static str = T::NAME;
+    type Options = T::Options;
+
+    fn import(
+        &self,
+        name: String,
+        source: Arc<Source>,
+        options: Self::Options,
+        create_reload: bool,
+    ) -> Result<(A::Data, Option<Box<Reload<A>>>), BoxedErr> {
+        if create_reload {
+            let (b, m) = source.load_with_metadata(&name)?;
+            let data = T::import(&self, b, options.clone())?;
+            let reload = SingleFile::new(self.clone(), m, options, name, source);
+
+            Ok((data, Some(Box::new(reload))))
+        } else {
+            let b = source.load(&name)?;
+            let data = T::import(&self, b, options)?;
+
+            Ok((data, None))
+        }
+    }
 }
--- a/amethyst_assets/src/asset.rs
+++ b/amethyst_assets/src/asset.rs
@@ -72,7 +72,8 @@ impl<A: Asset> FormatValue<A> {
 /// This is a simplified version of `Format`, which doesn't give you as much as freedom,
 /// but in return is simpler to implement.
 /// All `SimpleFormat` types automatically implement `Format`.
-/// This format assumes that the asset name is the full path and also the only file.
+/// This format assumes that the asset name is the full path and the asset is only
+/// contained in one file.
 pub trait SimpleFormat<A: Asset> {
     /// A unique identifier for this format.
     const NAME: &'static str;
--- a/amethyst_assets/src/asset.rs
+++ b/amethyst_assets/src/asset.rs
@@ -69,7 +69,7 @@ impl<A: Asset> FormatValue<A> {
     }
 }
 
-/// This is a simplified version of `Format`, which doesn't give you as much as freedom,
+/// This is a simplified version of `Format`, which doesn't give you as much freedom,
 /// but in return is simpler to implement.
 /// All `SimpleFormat` types automatically implement `Format`.
 /// This format assumes that the asset name is the full path and the asset is only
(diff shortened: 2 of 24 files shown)


torkleyy:
So as @Rhuagh pointed out, we need some sort of notification system, but I think it's better to do that in a follow-up.

amethyst_assets/examples/hl.rs
```diff
@@ -73,18 +74,18 @@ impl Asset for MeshAsset {
 }
 
 /// A format the mesh data could be stored with.
+#[derive(Clone)]
 struct Ron;
 
-impl Format<MeshAsset> for Ron {
+impl SimpleFormat<MeshAsset> for Ron {
```
Rhuagh: Good idea to simplify the simple formats :)

amethyst_assets/src/loader.rs
```diff
@@ -25,6 +26,7 @@ impl Loader {
     {
         Loader {
             directory: Arc::new(Directory::new(directory)),
+            hot_reload: true, // TODO: allow disabling
```
Rhuagh: Should we really default to true here? Even for the first iteration?

amethyst_renderer/src/formats/texture.rs
```diff
@@ -176,36 +172,27 @@ impl PngFormat {
...
 
-    fn import(
-        &self,
-        name: String,
-        source: Arc<Source>,
-        options: TextureMetadata,
-    ) -> Result<TextureData, BoxedErr> {
-        imagefmt::bmp::read(&mut Cursor::new(source.load(&name)?), ColFmt::RGBA)
+    fn import(&self, bytes: Vec<u8>, options: TextureMetadata) -> Result<TextureData, BoxedErr> {
+        imagefmt::bmp::read(&mut Cursor::new(bytes), ColFmt::RGBA)
```
omni-viral: It's better to read BMP file directly into gpu visible memory.
But this can be addressed later. Just put a note, please.

amethyst_assets/src/asset.rs
```diff
@@ -33,10 +33,64 @@ pub trait Format<A: Asset>: Send + 'static {
...
+    /// You should only create this if `create_reload` is `true`.
+    /// Also, the parameter is just a request, which means you can also return `None`.
     fn import(
         &self,
         name: String,
         source: Arc<Source>,
         options: Self::Options,
-    ) -> Result<A::Data, BoxedErr>;
+        create_reload: bool,
+    ) -> Result<(A::Data, Option<Box<Reload<A>>>), BoxedErr>;
```
omni-viral: This result type looks sophisticated.
Maybe it's worth to make a structure.

amethyst_assets/src/reload.rs
```diff
@@ -1,18 +1,115 @@
...
+
+    fn reload(self: Box<Self>) -> Result<(A::Data, Option<Box<Reload<A>>>), BoxedErr> {
+        let this: SingleFile<_, _> = *self;
+        let SingleFile {
+            format,
+            path,
+            source,
+            options,
+            ..
+        } = this;
```
omni-viral: Can't you just write `*self` here?

amethyst_assets/src/asset.rs
```diff
@@ -33,10 +33,64 @@ pub trait Format<A: Asset>: Send + 'static {
...
+    ///
+    /// ## Reload
+    ///
+    /// The reload structure has metadata which allows the asset management
+    /// to reload assets if necessary (for hot reloading).
+    /// You should only create this if `create_reload` is `true`.
+    /// Also, the parameter is just a request, which means you can also return `None`.
     fn import(
         &self,
         name: String,
```
omni-viral: `name` and `options` should be passed by reference to avoid unnecessary cloning.

amethyst_assets/src/reload.rs
```diff
@@ -1,18 +1,115 @@
...
+
+        format.import(path, source, options, true)
```
omni-viral: Can't it ask not to create new `Reload` object and return `self`?

amethyst_assets/src/asset.rs
omni-viral: And in `SimpleFormat` as well. Otherwise cloning will be necessary 😄

amethyst_assets/src/reload.rs
```diff
@@ -1,18 +1,115 @@
...
+use {Asset, BoxedErr, Format, Source};
+
+/// The `Reload` trait provides a method which checks if an asset needs to be reloaded.
+pub trait Reload<A: Asset>: ReloadClone<A> + Send + Sync + 'static {
+    /// Checks if a reload is necessary.
+    fn needs_reload(&self) -> bool;
+    /// Returns the asset name.
+    fn name(&self) -> String;
+    /// Returns the format name.
+    fn format(&self) -> String;
```
omni-viral: All formats has it's name as associated `&'static str`. So you can contain and return `&'static str`.

amethyst_assets/src/storage.rs
```diff
@@ -30,9 +33,11 @@ impl Allocator {
 pub struct AssetStorage<A: Asset> {
     assets: VecStorage<A>,
     bitset: BitSet,
-    handles: Vec<Handle<A>>,
+    handles: Vec<WeakHandle<A>>,
```
omni-viral: No need for using `Weak` inside `AssetStorage`
* Upgrading `Weak` has more overhead than cloning `Arc`. There is possibly infinite loop with CAS.
* You get nothing from `Weak` as you store simple `u32` and you don't have to drop it asap to free some resource. No memory get freed until you drop last `Weak`.
* And then you have to allocate new `Arc` after you finally drop `Weak`.

Using `WeakHandle` outside of `AssetStorage` may make some sense though.

amethyst_assets/src/loader.rs
torkleyy: No, I have to add configs to this; the param is just `true` to try it out.

amethyst_assets/src/asset.rs
torkleyy: I'm aware of the fact that an allocation is not desirable, but we can't point to the stack in a multi-threaded system.

amethyst_assets/src/reload.rs
torkleyy: I had this as `&'static str` first, but changed it back for an unknown reason. I'll try to change this later.

amethyst_assets/src/reload.rs
torkleyy: It seems to be an unsoundness w.r.t. the `Box<Self>` stuff.

amethyst_assets/src/reload.rs
torkleyy: No, imagine a GLTF file which points to other files; in case the GLTF file gets reloaded, new files might have been added which we also want to track.

amethyst_renderer/src/formats/texture.rs
torkleyy: Will do.

amethyst_assets/src/storage.rs
torkleyy: The biggest reason I used `WeakHandle` here is for consistency, since the `AssetStorage` does not in fact own a handle (logically). And for `reloads` it is indeed necessary.

amethyst_assets/src/asset.rs
omni-viral: I don't follow. If you use those only locally in function (or nested function calls). It's perfectly ok to pass them as references. And if you need to save them for future-use (In returned `Reload`) you clone value and store it.

amethyst_assets/src/reload.rs
omni-viral: But this is `SingleFile`

amethyst_assets/src/asset.rs
torkleyy: Please point me to the exact location where I could just pass a reference. There is exactly one point, `SimpleFormat::import`, otherwise I can't.

amethyst_assets/src/reload.rs
torkleyy: Yep.

amethyst_assets/src/reload.rs
torkleyy: So how would you do that?

amethyst_assets/src/storage.rs
```diff
@@ -112,46 +118,133 @@ impl<A: Asset> AssetStorage<A> {
...
                 }
 
                 Ok(())
             });
         }
 
-        while let Some(i) = self.handles.iter().position(Handle::is_unused) {
-            let old = self.handles.swap_remove(i);
-            let id = i as u32;
+        while let Some(i) = self.handles.iter().position(Handle::is_unique) {
```
omni-viral: You shouldn't search from start each time.
Try this.
```
let mut iter = self.handles.iter();
let mut i = 0;
while let Some(next) = iter.by_ref().position(Handle::is_unique) {
    i += next;
```

amethyst_assets/src/storage.rs
```diff
@@ -112,46 +118,133 @@ impl<A: Asset> AssetStorage<A> {
...
+            let handle = self.handles.swap_remove(i);
+            let id = handle.id();
             unsafe {
                 self.assets.remove(id);
             }
             self.bitset.remove(id);
-            self.unused_handles.push(old);
+
+            // Can't reuse old handle here, because otherwise weak handles would still be valid.
+            // TODO: maybe just store u32?
```
omni-viral: Can reuse again

amethyst_assets/src/storage.rs
```diff
@@ -112,46 +118,133 @@ impl<A: Asset> AssetStorage<A> {
...
+            self.unused_handles.push(Handle {
+                id: Arc::new(id),
+                marker: PhantomData,
+            });
+        }
+
+        // Reload every seconds
+        // TODO: more configuration
```
omni-viral: You can add more details to `TODO`.
Configure timeout.
Configure automatic or only manual with special method in `AssetStorage`.

amethyst_assets/src/storage.rs
```diff
@@ -230,10 +332,46 @@ where
     type Storage = DenseVecStorage<Self>;
 }
 
-// TODO: may change with hot reloading
-pub struct Processed<A: Asset> {
-    pub data: Result<A::Data, BoxedErr>,
-    pub format: String,
-    pub handle: Handle<A>,
-    pub name: String,
+pub enum Processed<A: Asset> {
```
omni-viral: It `pub` necessary here?

amethyst_assets/src/storage.rs
torkleyy: Sorry, I was being stupid during refactoring.

amethyst_assets/src/storage.rs
torkleyy: Sorry, I don't understand?

amethyst_assets/src/storage.rs
torkleyy: I will add such a structure in this PR.

amethyst_assets/src/storage.rs
torkleyy: Enums can't have private members.

torkleyy: Sorry, I think I misunderstood your question. Yes, `Processed` needs to be pub, because `Loader` creates such a structure. It's not exported at crate root, though.

jojolepro:
>So as @Rhuagh pointed out, we need some sort of notification system, but I think it's better to do that in a follow-up.

Events??? :P

torkleyy:
@jojolepro Not really, more like watching a file system.

amethyst_assets/src/storage.rs
omni-viral: Can you make it `pub(crate)` then?

amethyst_assets/src/storage.rs
omni-viral: Cool!

amethyst_assets/src/storage.rs
omni-viral: You can `Handle`. As it not `WeakHandle` again.

amethyst_assets/src/storage.rs
torkleyy: There's no need for that and besides, I think that syntax is horrible. I try not to use it where possible.

amethyst_assets/src/storage.rs
torkleyy: I think what you're saying is that because I changed it back to `Handle`, I can reuse the handles again?

I don't think we'd want that, because:

* Handle A gets allocated
* Handle A has a `downgraded` version in a cache
* Handle A gets dropped
* Handle A gets added to the `unused_handles` queue
* Another allocation is required; `unused_handles` gives us the "old" handle A
* The cached value we stored before (using `WeakHandle`) is now valid again

That's why I added that comment and recreated the handle. Did a make a mistake in my reasoning?

amethyst_assets/src/reload.rs
```diff
@@ -1,18 +1,203 @@
...
+    pub fn never() -> Self {
+        HotReloadStrategy {
+            inner: HotReloadStrategyInner::Never,
+        }
+    }
+
+    /// The frame after calling this, all changed assets will be reloaded.
+    pub fn trigger(&mut self) {
+        let counter = match self.inner {
+            HotReloadStrategyInner::Trigger { counter } => counter.checked_add(1).unwrap_or(0),
```
omni-viral: Just `wrapping_add` instead of `checked_add` and `unwrap_or`

amethyst_assets/src/reload.rs
```diff
@@ -1,18 +1,203 @@
...
+            _ => 0,
```
omni-viral: It shouldn't switch do another strategy.

amethyst_assets/src/reload.rs
omni-viral: Weird.

amethyst_assets/src/asset.rs
omni-viral: `name` and `options` in `Format::import` and `options` in `SimpleFormat`.
And then in `impl Reload for SingleFile` you can reuse `SingleFile` value instead of creating new one

amethyst_assets/src/storage.rs
omni-viral: Makes sense.

amethyst_assets/src/storage.rs
```diff
@@ -112,46 +122,147 @@ impl<A: Asset> AssetStorage<A> {
...
+            .unwrap_or(false)
+        {
+            self.hot_reload(pool);
+        }
+    }
+
+    fn hot_reload(&mut self, pool: &ThreadPool) {
+        let elapsed = self.last_reload.elapsed().as_secs();
+
+        if elapsed >= 1 {
```
omni-viral: You've already checked that reload is required with strategy.

amethyst_assets/src/reload.rs
torkleyy: That's what I've been searching for, thanks!

amethyst_assets/src/reload.rs
torkleyy: It probably shouldn't switch from `Every`.

amethyst_assets/src/storage.rs
torkleyy: Oops, forgot to remove the old code.

amethyst_assets/src/asset.rs
torkleyy: As I've explained, you can't do that.

amethyst_assets/src/asset.rs
```diff
@@ -38,10 +38,80 @@ pub trait Format<A: Asset>: Send + 'static {
...
+    /// Creates a `FormatValue` from only the data (setting `reload` to `None`).
+    pub fn data(data: A::Data) -> Self {
+        FormatValue { data, reload: None }
+    }
+}
+
+/// This is a simplified version of `Format`, which doesn't give you as much as freedom,
+/// but in return is simpler to implement.
+/// All `SimpleFormat` types automatically implement `Format`.
+/// This format assumes that the asset name is the full path and also the only file.
```
Xaeroxe: Minor grammar nit: Maybe this should end with "and the asset is only contained within one file."

amethyst_assets/src/storage.rs
Xaeroxe: Why?  It's very straightforward and obvious.  You don't need to look at other files to figure out it's not actually pub, and pub(crate) permits easy re-use of internal structures.

amethyst_assets/src/asset.rs
torkleyy: Makes sense, thanks.

torkleyy:
Okay, all rebased, fixed, addressed and ready for a final review ;)

torkleyy:
> I'd be overall much happier if we provided a hot reload strategy that uses file system events to trigger hot reloading.

Me too, but I don't want to implement everything in one PR.

amethyst_core/src/lib.rs
```diff
@@ -2,20 +2,25 @@
...
+extern crate rayon;
 #[macro_use]
 extern crate serde;
 extern crate shred;
 extern crate specs;
 
-#[cfg(test)]
-#[cfg_attr(test, macro_use)]
-extern crate quickcheck;
+//#[cfg(test)]
```
Rhuagh: Why is this commented out ?

amethyst_assets/src/asset.rs
```diff
@@ -38,10 +38,81 @@ pub trait Format<A: Asset>: Send + 'static {
...
+}
+
+impl<A: Asset> FormatValue<A> {
+    /// Creates a `FormatValue` from only the data (setting `reload` to `None`).
+    pub fn data(data: A::Data) -> Self {
+        FormatValue { data, reload: None }
+    }
+}
+
+/// This is a simplified version of `Format`, which doesn't give you as much as freedom,
```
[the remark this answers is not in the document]
torkleyy: Oh ;P It's funny to see how crappy the docs I wrote are.

amethyst_assets/src/reload.rs
```diff
@@ -1,18 +1,211 @@
...
+
+    /// This allows to use `trigger` for hot reloading.
+    pub fn when_triggered() -> Self {
+        HotReloadStrategy {
+            inner: HotReloadStrategyInner::Trigger { counter: 0 },
+        }
+    }
+
+    /// No periodical hot-reloading is performed.
+    /// Instead, you can `trigger` it to check for changed assets
```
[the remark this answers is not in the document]
torkleyy: It does, and it's outdated.

amethyst_core/src/lib.rs
torkleyy: There's not a single usage in the whole crate, I'm not sure what @Aceeri did here.

Rhuagh: Ah, seems he removed all the tests using it. Remove it altogether then I'd say.

amethyst_assets/src/reload.rs
```diff
@@ -54,50 +87,66 @@ impl HotReloadStrategy {
...
+        interval: u8,
+        last: Instant,
+        do_reload: bool,
+    },
+    Trigger { triggered: bool },
     Never,
 }
 
+/// System for updating `HotReloadStrategy`.
+/// **NOTE:** You have to add this after all asset processing systems.
```
Rhuagh: The `RenderSystem` which process `Mesh` and `Texture` will mess this up.

torkleyy:
Assigning @Rhuagh to merge in case of an approval.

Xaeroxe:
Yay!  No more rebasing! :P

PR #446, opened 2017-10-23
Add frame_number to Time resource

Xaeroxe:
This PR adds how many frames have been simulated to the Time resource, which may be helpful for execution order scheduling (such as what's needed in https://github.com/amethyst/amethyst/pull/445) and it also helps with timing for fixed framerate videogames.

Note: This won't overflow, if the game runs at 1000 FPS the u64 won't over flow for >500 million years.

2^64 / 1000 / 60 / 60 / 24 / 365  >500 million


Xaeroxe:
@torkleyy So the basic idea I have is this: When you trigger a hot reload, you store the frame count + 1, then when the frame count == the number you stored, you perform the hot reload.

PR #450, opened 2017-10-24
Add f32 timing variables

Xaeroxe:
This takes a bit of legwork out of a very common task in gamedev.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -2,6 +2,7 @@
 
 use std::path::Path;
 use std::sync::Arc;
+use std::time::Duration;
 
 use core::ECSBundle;
 use rayon::ThreadPool;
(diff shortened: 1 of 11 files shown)


Xaeroxe:
cc @jojolepro 

Just wanted to show you we're finally using your function :)

src/app.rs
```diff
@@ -155,8 +155,10 @@ impl<'a, 'b> Application<'a, 'b> {
         {
             let mut time = self.engine.world.write_resource::<Time>();
             time.frame_number += 1;
+            time.delta_seconds = duration_to_secs(self.time.delta_time);
```
Aceeri: I can't say I like that `time.fixed_step` and `time.fixed_seconds` can differ due to human error. Maybe we should put this behind a method to set/get the fields?

Xaeroxe:
@Aceeri This is my best attempt at controlling for mutability.  Let me know what you think.

Xaeroxe:
Squashed.

bors r+

PR #453, opened 2017-10-25
feat: Animation of properties of LocalTransform

Rhuagh:
Opening early to get feedback. Note that this PR only contains animation of properties on LocalTransform, and have no skinning support. That will come in a later PR.

Work left:

* ~Testing!!!~
* ~Documentation~
* ~Interpolation needs a rework~


amethyst_animation/src/resources.rs
```diff
@@ -0,0 +1,165 @@
...
+pub struct AnimationHierarchy {
+    pub nodes: HashMap<usize, Entity>,
+}
+
+impl Component for AnimationHierarchy {
+    type Storage = DenseVecStorage<Self>;
+}
+
+/// Defines a single animation.
+/// Defines relationships between the node index in `AnimationHierarchy` and a `Sampler` in
```
Rhuagh: Outdated docs.

amethyst_animation/src/resources.rs
```diff
@@ -0,0 +1,165 @@
...
+    Running(Instant, Duration),
+    /// Animation is paused at the accumulated duration
+    Paused(Duration),
+    /// Request termination of the animation
+    Abort,
+    /// Animation is completed
+    Done,
+}
+
+/// The actual animation data for a single attribute
```
Rhuagh: Outdated docs.

amethyst_animation/src/resources.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// Pause the animation
+    Pause,
+    /// Abort the animation, will cause the control object to be removed from the world
+    Abort,
+}
+
+/// Attaches to an entity, to control what animations are currently active
+#[derive(Clone)]
+pub struct AnimationControl {
+    /// Animation index into `AnimationStorage`
```
Rhuagh: Outdated docs.

Cargo.toml
```diff
@@ -42,6 +42,7 @@ thread_profiler = { version = "0.1", optional = true }
 [dev-dependencies]
 cgmath = "0.14"
 genmesh = "0.4"
+amethyst_animation = { path = "amethyst_animation", version = "0.1.0" }
```
torkleyy: So you decided not to reexport this package? Should we maybe do the same for the `gltf` crate then, which also doesn't seem like a "mandatory" crate?

amethyst_animation/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+authors = ["Simon Rönnberg <seamonr@gmail.com>"]
+description = "Animation support for Amethyst"
+keywords = ["game", "engine", "animation", "3d", "amethyst"]
+
+documentation = "https://www.amethyst.rs/doc/master/amethyst_animation/"
+homepage = "https://www.amethyst.rs/"
+repository = "https://github.com/amethyst/amethyst"
+
+readme = "README.md"
+license = "MIT OR Apache-2.0"
```
torkleyy: should be `MIT/Apache-2.0`

amethyst_animation/src/bundle.rs
```diff
@@ -0,0 +1,89 @@
...
+                .add(
+                    SamplerInterpolationSystem::new(),
+                    "sampler_interpolation_system",
+                    self.dep,
+                ),
+        )
+    }
+}
+
+/// Bundle for a complete animation setup including sampler interpolation and animation control
```
torkleyy: Should add comment that this automatically adds a `SamplingBundle`

amethyst_animation/src/interpolation.rs
```diff
@@ -0,0 +1,275 @@
...
+    pub fn new<I>(interp: I) -> Self
+    where
+        I: Interpolate + 'static,
+    {
+        InterpolationType::Function(Box::new(interp))
+    }
+}
+
+pub trait Interpolate: Debug + Send + Sync {
+    fn interpolate_vec(
```
torkleyy: Are you planning on replacing this?

amethyst_animation/src/interpolation.rs
```diff
@@ -0,0 +1,275 @@
...
+                catmull_rom_spline(time_index, time, times, values, normalize)
+            }
+            InterpolationType::Function(ref f) => {
+                f.interpolate_quat(time_index, time, times, values, normalize)
+            }
+        }
+    }
+}
+
+fn linear<D>(time_index: usize, time: f32, times: &Vec<f32>, values: &Vec<D>, normalize: bool) -> D
```
torkleyy: Still gotta check these functions

amethyst_animation/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+extern crate specs;
+
+pub use self::bundle::{AnimationBundle, SamplingBundle};
+pub use self::interpolation::{Interpolate, InterpolationPrimitive, InterpolationType};
+pub use self::resources::{Animation, AnimationCommand, AnimationControl, AnimationHierarchy,
+                          AnimationOutput, ControlState, EndControl, RestState, Sampler,
+                          SamplerControl, SamplerControlSet};
+pub use self::systems::{AnimationControlSystem, AnimationProcessor, SamplerInterpolationSystem,
+                        SamplerProcessor};
+pub use self::util::play_animation;
```
torkleyy: Nice, flat crate structure :+1:

amethyst_animation/src/resources.rs
```diff
@@ -0,0 +1,165 @@
...
+/// Sampler defines a single animation for a single attribute of the `LocalTransform` of the entity
+/// it is attached to.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Sampler {
+    /// Time of key frames
+    pub input: Vec<f32>,
+    /// Actual output data to interpolate
+    pub output: AnimationOutput,
+    /// How should interpolation be done
+    pub interpolation_type: InterpolationType,
```
torkleyy: Please rename this to `ty`, I think this field name is a bit long.

amethyst_animation/src/systems/control.rs
```diff
@@ -0,0 +1,329 @@
...
+        WriteStorage<'a, SamplerControlSet>,
+        ReadStorage<'a, AnimationHierarchy>,
+        ReadStorage<'a, LocalTransform>,
+    );
+
+    fn run(
+        &mut self,
+        (entities, animation_storage, sampler_storage, mut controls, mut samplers, hierarchies,
+            transforms): Self::SystemData,
+){
```
torkleyy: Odd formatting

amethyst_animation/src/systems/control.rs
```diff
@@ -0,0 +1,329 @@
...
+        let now = Instant::now();
+        let mut remove = Vec::default();
+        for (entity, control, hierarchy) in (&*entities, &mut controls, &hierarchies).join() {
+            let new_state = match animation_storage.get(&control.animation) {
```
torkleyy: Could we add a helper function for the `match` logic below? That way, we could call `animation_storage.get` and do `and_then` while keeping a flat layout.

amethyst_animation/src/systems/control.rs
```diff
@@ -0,0 +1,329 @@
...
+    transforms: &ReadStorage<LocalTransform>, // for rest state
+) -> bool {
+    // check that hierarchy is valid
+    if animation.nodes.iter().any(|&(ref node_index, _)| {
+        hierarchy.nodes.get(node_index).is_none()
+    }) {
+        return false;
+    }
+    // check that all samplers exist
+    if animation.nodes.iter().any(|&(_, ref sampler_handle)| {
```
torkleyy: Could we combine both iterators into one?

amethyst_animation/src/systems/control.rs
```diff
@@ -0,0 +1,329 @@
...
+            };
+
+            if let Some(state) = new_state {
+                control.state = state;
+            }
+        }
+    }
+}
+
+fn request_animation(
```
torkleyy: Could you add a short comment what this function does?

amethyst_animation/src/systems/control.rs
```diff
@@ -0,0 +1,329 @@
...
+) -> bool {
+    hierarchy
+        .nodes
+        .iter()
+        .flat_map(|(_, node_entity)| samplers.get(*node_entity))
+        .all(check_control_set_termination)
+}
+
+fn do_control_set_termination(control_set: &mut SamplerControlSet) {
+    match control_set.translation {
```
torkleyy: Why not `if let Some(x) = ..`?

amethyst_animation/src/systems/control.rs
```diff
@@ -0,0 +1,329 @@
...
+        Some(ref mut s) => if s.state != ControlState::Done {
+            s.state = ControlState::Abort;
+        },
+        _ => (),
+    }
+}
+
+
+fn do_control_set_pause(control_set: &mut SamplerControlSet) {
+    match control_set.translation {
```
torkleyy: Same as above

amethyst_animation/src/systems/control.rs
```diff
@@ -0,0 +1,329 @@
...
+                ControlState::Running(_, dur) => ControlState::Paused(dur),
+                _ => ControlState::Paused(Duration::from_secs(0)),
+            };
+        }
+        _ => (),
+    }
+}
+
+fn do_control_set_unpause(now: &Instant, control_set: &mut SamplerControlSet) {
+    match control_set.translation {
```
torkleyy: Same as above

amethyst_animation/src/systems/sampling.rs
```diff
@@ -0,0 +1,189 @@
...
+    type SystemData = (
+        Fetch<'a, AssetStorage<Sampler>>,
+        WriteStorage<'a, SamplerControlSet>,
+        WriteStorage<'a, LocalTransform>,
+    );
+
+    fn run(&mut self, (samplers, mut controls, mut transforms): Self::SystemData) {
+        let now = Instant::now();
+        for (control_set, transform) in (&mut controls, &mut transforms).join() {
+            match control_set.translation {
```
torkleyy: could be `if let Some(sampler) = control_set.translation.as_mut().and_then(|c| samplers.get(c.sampler))`

amethyst_animation/src/systems/sampling.rs
```diff
@@ -0,0 +1,189 @@
...
+        control.end = end.clone();
+    }
+
+    // Do sampling
+    match new_state {
+        Running(_, duration) | Paused(duration) => {
+            do_sampling(&sampler, &duration, transform);
+        }
+        Done => do_end_control(&control, transform),
+        _ => (),
```
torkleyy: The idiomatic version would be `_ => {}` I think

amethyst_animation/src/systems/sampling.rs
```diff
@@ -0,0 +1,189 @@
...
+
+fn do_end_control(control: &SamplerControl, transform: &mut LocalTransform) {
+    match control.end {
+        EndControl::Normal => match control.after {
+            RestState::Translation(tr) => transform.translation = tr,
+            RestState::Rotation(r) => transform.rotation = r,
+            RestState::Scale(s) => transform.scale = s,
+        },
+        // looping is handled during duration update
+        _ => (),
```
torkleyy: Same as above

Cargo.toml
Rhuagh: Yeah, we should.

amethyst_animation/src/interpolation.rs
Rhuagh: Yes.

Rhuagh:
After `minterpolate` has been released this will be ready. Need to get more testing done on that first though.

Rhuagh:
Ready for review.

amethyst_animation/src/resources.rs
```diff
@@ -54,6 +55,7 @@ impl Component for AnimationHierarchy {
 
 /// Defines a single animation.
 /// Defines relationships between the node index in `AnimationHierarchy` and a `Sampler` handle.
+/// If the animation only target a single node index, `AnimationHierarchy` is not required.
```
torkleyy: should be "targets" I think

amethyst_animation/src/systems/control.rs
```diff
@@ -11,7 +11,8 @@ use resources::{Animation, AnimationCommand, AnimationControl, AnimationHierarch
 /// System for setting up animations, should run before `SamplerInterpolationSystem`.
 ///
 /// Will process all active `AnimationControl` + `AnimationHierarchy`, and do processing of the
-/// animations they describe.
+/// animations they describe. If an animation only target a single node/entity, there is no need for
```
torkleyy: Same here

amethyst_animation/src/systems/control.rs
```diff
@@ -70,14 +71,36 @@ impl<'a> System<'a> for AnimationControlSystem {
     }
 }
 
+/// Check if the given animation list is for a single node. If so, we don't need an
+/// `AnimationHierarchy`.
+fn only_one_index(nodes: &Vec<(usize, Handle<Sampler>)>) -> bool {
+    if nodes.len() == 0 {
```
torkleyy: nit: should be `nodes.is_empty()`

amethyst_animation/src/systems/control.rs
```diff
@@ -70,14 +71,36 @@ impl<'a> System<'a> for AnimationControlSystem {
     }
 }
 
+/// Check if the given animation list is for a single node. If so, we don't need an
+/// `AnimationHierarchy`.
+fn only_one_index(nodes: &Vec<(usize, Handle<Sampler>)>) -> bool {
```
torkleyy: Can take slice here

Rhuagh:
I'll squash after approvals are in place.

amethyst_animation/src/resources.rs
```diff
@@ -0,0 +1,193 @@
...
+    /// What to do when sampler ends
+    pub end: EndControl,
+    /// What the transform should return to after end
+    pub after: RestState,
+}
+
+/// Sampler control set, containing optional samplers for each of the possible channels.
+///
+/// Note that the `AnimationOutput` in the `Sampler` referenced, and the `RestState` referenced in
+/// each `SamplerControl` should match the attribute channel from the set here. There is no check
```
omni-viral: Shouldn't be `SamplerControl` parametrised in this case?

amethyst_animation/src/systems/control.rs
```diff
@@ -0,0 +1,417 @@
...
+            mut controls,
+            mut samplers,
+            hierarchies,
+            transforms,
+        ) = data;
+        let now = Instant::now();
+        let mut remove = Vec::default();
+        for (entity, control) in (&*entities, &mut controls).join() {
+            if let Some(state) = animation_storage.get(&control.animation).and_then(
+                |ref animation| {
```
omni-viral: What the point of `ref` keyword here?

amethyst_animation/src/systems/sampling.rs
```diff
@@ -0,0 +1,198 @@
...
+                        while last_frame != 0. && dur_s > last_frame {
+                            dur_s -= last_frame;
+                        }
+                        let s = dur_s.floor();
+                        let nanos = dur_s - s;
+                        let next_dur = Duration::new(dur_s as u64, (nanos * 1_000_000_000.) as u32);
+                        match times {
+                            // finite amount of looping, if 0 go do done,
+                            // else reduce iteration count and keep sampling
+                            Some(i) => if i <= 1 {
```
omni-viral: Use guards to avoid too many nesting levels

amethyst_animation/src/resources.rs
```diff
@@ -0,0 +1,193 @@
...
+    type Storage = DenseVecStorage<Self>;
+}
+
+/// Defines a single animation.
+/// Defines relationships between the node index in `AnimationHierarchy` and a `Sampler` handle.
+/// If the animation only targets a single node index, `AnimationHierarchy` is not required.
+#[derive(Clone, Debug)]
+pub struct Animation {
+    /// node index -> sampler handle
+    pub nodes: Vec<(usize, Handle<Sampler>)>,
```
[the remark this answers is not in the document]
Rhuagh: No, because you can play animation on many different entities when you use prefabs.

amethyst_animation/src/systems/control.rs
Rhuagh: Remnant of old code

amethyst_animation/src/systems/sampling.rs
Rhuagh: Guards?

Rhuagh:
This is a first version, so no blending, speed cobtrol or stacking no.

amethyst_animation/src/resources.rs
Rhuagh: This is in part due to how it looked at the start. I did consider changing this note, but came to the conclusion I wanted to get something out first. I will deal with this in a later PR.

amethyst_animation/src/resources.rs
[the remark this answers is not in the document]
Rhuagh: You want to be able to share animations between many families of entities.

amethyst_animation/src/systems/sampling.rs
torkleyy: https://rustbyexample.com/flow_control/match/guard.html

amethyst_animation/src/resources.rs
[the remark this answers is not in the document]
omni-viral: Can you describe use case?

omni-viral:
@Rhuagh ok. But you need to keep those features in mind so that adding them won't require total rewrite.

amethyst_animation/src/resources.rs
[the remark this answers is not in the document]
Rhuagh: A quite common scenario for indie game development is that you have a single bone structure for many different units (think human, vulkan, klingon etc), and all units that share bone structure can share animations. This saves on time and cost of developing the game, because you don't have to produce as many animations.

So, animations are loaded once and then potentially shared between many different types of units/entities whatever. 

The second issue with the suggestion you have is one of forward lookup. Given an animation and entity to play that animation on, how do i find which child entities to add samplers to ? If the only link we have is that the child entities have a reference to the entity of the prefab they were created from, you would have to look through the entire `Parent` storage to find the entitites that should be animated, which would be far to expensive imho. The other solution would be to have the `AnimationHierarchy` contain Entity -> Entity mappings. I went with the "node index -> Entity" solution, because it makes animation asset loading much, much easier.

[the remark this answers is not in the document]
Rhuagh: Animations can also target only a part of a hierarchy, not necessarily the entire hierarchy. Example: Sword swing will probably only affect arms and upper torso.

Rhuagh:
For the most part it will go something like this:

Speed control: additional parameter on AnimationCommand, and on SamplerControl.
Animation blending: SamplerControlSet gets a weight parameter, and it becomes Many<SamplerControlSet> as Component, Many<AnimationControl> as Component. Needs support in both systems.
Animation stacking: AnimationControl get's a control type parameter to differentiate between blend and stacking, and animation control system needs support for determining what to do at animation end.

amethyst_animation/src/resources.rs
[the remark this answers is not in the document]
Rhuagh: Regardless, this is not something that can be done within the scope of this PR, because it needs prefab loading first.

Rhuagh:
@omni-viral review addressed and issues fixed, please review again.

amethyst_animation/src/resources.rs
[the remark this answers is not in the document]
omni-viral: > Example: Sword swing will probably only affect arms and upper torso.

But this system animate `LocalTransform` of entities. Will it share some parts with skeletal animation system?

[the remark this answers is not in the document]
Rhuagh: Skeletal animation is this system together with vertex skinning.

amethyst_animation/src/systems/sampling.rs
```diff
@@ -0,0 +1,202 @@
...
+                match control.end {
+                    // Do loop control
+                    EndControl::Loop(Some(i)) if i <= 1 => (Done, Some(EndControl::Normal)),
+                    EndControl::Loop(None) => (
+                        Running(now.clone(), next_duration(last_frame, duration_secs)),
+                        None,
+                    ),
+                    EndControl::Loop(Some(i)) => (
+                        Running(now.clone(), next_duration(last_frame, duration_secs)),
+                        Some(EndControl::Loop(Some(i - 1))),
```
omni-viral: You should subtract number equal to `time_elapsed / animation_duration` and put `time_elapsed % animation_duration` into `Running(_, >here<)`

amethyst_animation/src/systems/sampling.rs
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        // Done and paused will be handled during sampling
+        ref state => (state.clone(), None),
+    }
+}
+
+fn next_duration(last_frame: f32, mut dur_s: f32) -> Duration {
+    while last_frame != 0. && dur_s > last_frame {
+        dur_s -= last_frame;
```
omni-viral: Division should work faster then loop.
And you probably want to work with milliseconds in `u32` (or nanoseconds in `u64`) instead of seconds in `f32`.

amethyst_gltf/src/format/mod.rs
```diff
@@ -138,24 +143,108 @@ fn load_data(
...
+            Ok((
+                node_index,
+                Sampler {
+                    input,
+                    ty,
+                    output: AnimationOutput::Rotation(output),
+                },
+            ))
+        }
+        Weights => Err(GltfError::NotImplemented),
```
omni-viral: What are those? Skinning weights?

amethyst_animation/src/util.rs
```diff
@@ -0,0 +1,85 @@
...
+/// - `end`: action to perform when the animation has reached its end.
+pub fn play_animation(
+    controls: &mut WriteStorage<AnimationControl>,
+    animation: &Handle<Animation>,
+    entity: Entity,
+    end: EndControl,
+) {
+    if !match controls.get_mut(entity) {
+        Some(ref mut control) => {
+            if control.state.is_paused() {
```
omni-viral: If pause was requested then it will be paused on next system run.
Command should be set to `Start` unconditionally.

amethyst_animation/src/util.rs
```diff
@@ -0,0 +1,85 @@
...
+///             not be run.
+/// - `end`: action to perform when the animation has reached its end.
+pub fn play_animation(
+    controls: &mut WriteStorage<AnimationControl>,
+    animation: &Handle<Animation>,
+    entity: Entity,
+    end: EndControl,
+) {
+    if !match controls.get_mut(entity) {
+        Some(ref mut control) => {
```
omni-viral: Should check if it is the same animation.

amethyst_animation/src/util.rs
```diff
@@ -0,0 +1,85 @@
...
+///             matches the `Animation`, or only refer to a single node, else the animation will
+///             not be run.
+/// - `end`: action to perform when the animation has reached its end.
+pub fn play_animation(
+    controls: &mut WriteStorage<AnimationControl>,
+    animation: &Handle<Animation>,
+    entity: Entity,
+    end: EndControl,
+) {
+    if !match controls.get_mut(entity) {
```
omni-viral: `if !match` is poorly readable.
Use single `match` block with guards or `if let Some` if appropriate.

amethyst_animation/src/util.rs
```diff
@@ -0,0 +1,85 @@
...
+    animation: &Handle<Animation>,
+    entity: Entity,
+    end: EndControl,
+) {
+    if controls
+        .get(entity)
+        .map(|c| c.state.is_running())
+        .unwrap_or(false)
+    {
+        pause_animation(controls, entity);
```
omni-viral: What if another animation is running?

amethyst_gltf/src/format/mod.rs
Rhuagh: Morph target weights.

amethyst_animation/src/resources.rs
```diff
@@ -0,0 +1,193 @@
...
+}
+
+impl Asset for Animation {
+    type Data = Self;
+    type HandleStorage = DenseVecStorage<Handle<Self>>;
+}
+
+impl Into<Result<Animation>> for Animation {
```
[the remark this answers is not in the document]
Rhuagh: For asset processor.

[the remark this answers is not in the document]
omni-viral: You've answered only first question.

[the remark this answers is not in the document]
Rhuagh: I'm using the `Processor` from the assets crate to do asset processing for Sampler and Animation, and it has Asset::Data = Animation (it converts into itself since it doesn't need a factory to convert or such), and that requires an Into<assets::Result<Asset>> for Asset::Data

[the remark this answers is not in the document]
Rhuagh: Yes, I could do manual processing of them in my own system, just felt silly to build a new system for something that already exists.

amethyst_animation/src/util.rs
Rhuagh: Since we only support a single animation atm, it felt like an unnecessary check for now.

amethyst_animation/src/systems/sampling.rs
Rhuagh: isn't modulo on floats kind of frowned upon?

amethyst_animation/src/util.rs
Rhuagh: The problem is borrowing rules. I'd have to break it up into two `if let ..`

Rhuagh:
@omni-viral review comments addressed, please review again.

torkleyy:
Some more reviews on this would be really appreciated. I'll try to do another in-depth review on this, but one or two more wouldn't hurt.

Issue #454, opened 2017-10-25
`config.ron` should be renamed to `display_config.ron`


torkleyy:
We plan to introduce more configs, so renaming it would make sense.

PR #455, opened 2017-10-25
Create a minimal, but working "getting started" chapter

torkleyy:
Blocked on cli release and #454


Rhuagh:
Push this as soon as the cli and the config files are done?

torkleyy:
CLI is done already, but I think we may wanna wait before merging it.

PR #462, opened 2017-10-26
Upgrade and use cgmath

### LINKED ISSUE #200: Provide a clean and simple math module

### LINKED ISSUE #236: Consider exposing math types via `mint`
https://github.com/kvark/mint

Exposing `cgmath` types would introduce conflicts whenever `cgmath` updates faster than `Amethyst`. More importantly, it would alienate `nalgebra` and `euclid` users. The solution I see is to use a neutral standard for math interop, which is where `mint` comes handy. It's not a math library, so it's not competing with anything. It's not introducing any new types, so it's not even required to be a dependency.

### LINKED ISSUE #452: [Discussion] Use cgmath as our official math library?

Xaeroxe:
This upgrades us to cgmath 0.15, uses it where appropriate, and enables the mint feature for it.

You may have noticed that the renderer actually lost cgmath types in this PR, this was a necessary sacrifice to upgrade to 0.15. 0.15 is the only version released with mint support.  Our current gfx version uses cgmath 0.14 so the gfx trait implementations were not compatible with cgmath 0.15.

Closes #452
Closes #200 
Closes #236

### DIFF
--- a/amethyst_core/src/orientation.rs
+++ b/amethyst_core/src/orientation.rs
@@ -16,9 +16,9 @@ pub struct Orientation {
 impl Default for Orientation {
     fn default() -> Self {
         Self {
-            forward: [1.0, 0.0, 0.0].into(),
-            right: [0.0, -1.0, 0.0].into(),
-            up: [0.0, 0.0, 1.0].into(),
+            forward: Vector3::unit_x(),
+            right: -Vector3::unit_y(),
+            up: Vector3::unit_z(),
         }
     }
 }
--- a/amethyst_core/src/orientation.rs
+++ b/amethyst_core/src/orientation.rs
@@ -1,23 +1,24 @@
 //! Orientation of objects
 
+use cgmath::Vector3;
 
 /// Orientation struct.
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 pub struct Orientation {
     /// Forward vector [x, y, z]
-    pub forward: [f32; 3],
+    pub forward: Vector3<f32>,
     /// Right vector [x, y, z]
-    pub right: [f32; 3],
+    pub right: Vector3<f32>,
     /// Up vector [x, y, z]
-    pub up: [f32; 3],
+    pub up: Vector3<f32>,
 }
 
 impl Default for Orientation {
     fn default() -> Self {
         Self {
-            forward: [1.0, 0.0, 0.0],
-            right: [0.0, -1.0, 0.0],
-            up: [0.0, 0.0, 1.0],
+            forward: [1.0, 0.0, 0.0].into(),
+            right: [0.0, -1.0, 0.0].into(),
+            up: [0.0, 0.0, 1.0].into(),
         }
     }
 }
--- a/amethyst_core/src/transform/components/local_transform.rs
+++ b/amethyst_core/src/transform/components/local_transform.rs
@@ -1,7 +1,7 @@
 //! Local transform component.
 
-use cgmath::{Deg, EuclideanSpace, InnerSpace, Matrix3, Matrix4, Point3, Quaternion, Rotation,
-             Rotation3, Vector3};
+use cgmath::{Array, Deg, EuclideanSpace, InnerSpace, Matrix3, Matrix4, One, Point3, Quaternion, Rotation,
+             Rotation3, SquareMatrix, Vector3, Zero};
 use orientation::Orientation;
 use specs::{Component, DenseVecStorage, FlaggedStorage};
 
@@ -35,11 +35,7 @@ impl LocalTransform {
     #[inline]
     pub fn matrix(&self) -> Matrix4<f32> {
         let quat: Matrix3<f32> = Quaternion::from(self.rotation).into();
-        let scale: Matrix3<f32> = Matrix3::<f32> {
-            x: [self.scale[0], 0.0, 0.0].into(),
-            y: [0.0, self.scale[1], 0.0].into(),
-            z: [0.0, 0.0, self.scale[2]].into(),
-        };
+        let scale: Matrix3<f32> = Matrix3::from_diagonal(self.scale);
         let mut matrix: Matrix4<f32> = (&quat * scale).into();
         matrix.w = self.translation.extend(1.0f32);
         matrix
@@ -164,9 +160,9 @@ impl LocalTransform {
 impl Default for LocalTransform {
     fn default() -> Self {
         LocalTransform {
-            translation: [0.0, 0.0, 0.0].into(),
-            rotation: [1.0, 0.0, 0.0, 0.0].into(),
-            scale: [1.0, 1.0, 1.0].into(),
+            translation: Vector3::zero(),
+            rotation: Quaternion::one(),
+            scale: Vector3::from_value(1.),
         }
     }
 }
--- a/amethyst_core/src/transform/components/local_transform.rs
+++ b/amethyst_core/src/transform/components/local_transform.rs
@@ -1,17 +1,7 @@
 //! Local transform component.
 
-use cgmath::{
-    Deg,
-    EuclideanSpace,
-    InnerSpace,
-    Matrix3,
-    Matrix4,
-    Point3,
-    Quaternion,
-    Rotation,
-    Rotation3,
-    Vector3
-};
+use cgmath::{Deg, EuclideanSpace, InnerSpace, Matrix3, Matrix4, Point3, Quaternion, Rotation,
+             Rotation3, Vector3};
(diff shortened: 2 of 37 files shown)


amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -1,28 +1,38 @@
 //! Local transform component.
 
-use cgmath::{Deg, InnerSpace, Matrix3, Matrix4, Point3, Quaternion, Rotation, Rotation3, Vector3};
+use cgmath::{
```
Rhuagh: Weird formatting

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -1,28 +1,38 @@
...
 pub struct LocalTransform {
     /// Quaternion [w (scalar), x, y, z]
-    pub rotation: [f32; 4],
+    pub rotation: Quaternion<f32>,
     /// Scale vector [x, y, z]
-    pub scale: [f32; 3],
+    pub scale: Vector3<f32>,
     /// Translation/position vector [x, y, z]
-    pub translation: [f32; 3],
+    pub translation: Point3<f32>,
```
Rhuagh: This should be Vector, translation is a movement quantity, but a positional quantity

Rhuagh: And it's a relative translation from the parent and not an absolute position

Rhuagh: I think the most compelling part is that it's a relative quantity, not absolute, in `LocalTransform`

amethyst_renderer/src/light.rs
```diff
@@ -31,14 +30,14 @@ pub struct DirectionalLight {
     /// Color of the light in RGBA8 format.
     pub color: Rgba,
     /// Direction that the light is pointing.
-    pub direction: Vector3<f32>,
+    pub direction: [f32; 3],
```
torkleyy: Wasn't the goal consistency? So could we maybe just upgrade to gfx 0.17?

Xaeroxe: 0.17 isn't on crates :(

Xaeroxe:
Apparently the most recent version of rustfmt-nightly says we were wrong on all of our formatting.

Xaeroxe:
Hope ya'll are ready for some rebases :/

Xaeroxe:
Using our cgmath now requires a couple features to be enabled by the person depending on cgmath, should we pub extern cgmath in the core crate so they can easily get the right version and features?

amethyst_renderer/src/light.rs
Xaeroxe: Although I'll add a Todo for when we do upgrade our gfx

amethyst_core/src/transform/components/transform.rs
```diff
@@ -35,12 +36,14 @@ impl Component for Transform {
 
 impl Default for Transform {
     fn default() -> Self {
-        Transform([
-            [1.0, 0.0, 0.0, 0.0],
-            [0.0, 1.0, 0.0, 0.0],
-            [0.0, 0.0, 1.0, 0.0],
-            [0.0, 0.0, 0.0, 1.0],
-        ])
+        Transform(
```
Rhuagh: `Matrix4::one()`  ?

amethyst_core/src/orientation.rs
```diff
@@ -1,23 +1,24 @@
...
+    pub up: Vector3<f32>,
 }
 
 impl Default for Orientation {
     fn default() -> Self {
         Self {
-            forward: [1.0, 0.0, 0.0],
-            right: [0.0, -1.0, 0.0],
-            up: [0.0, 0.0, 1.0],
+            forward: [1.0, 0.0, 0.0].into(),
```
Rhuagh: In theory, you could use `Vector3::unit_x()` `-Vector3::unit_y()` etc, but might not be a point tbh

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -31,16 +33,16 @@ impl LocalTransform {
     /// Combined with the parent's global `Transform` component it gives
     /// the global (or world) matrix for the current entity.
     #[inline]
-    pub fn matrix(&self) -> [[f32; 4]; 4] {
+    pub fn matrix(&self) -> Matrix4<f32> {
         let quat: Matrix3<f32> = Quaternion::from(self.rotation).into();
         let scale: Matrix3<f32> = Matrix3::<f32> {
```
Rhuagh: `Matrix3::from_diagonal(self.scale)`

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -163,9 +164,9 @@ impl LocalTransform {
 impl Default for LocalTransform {
     fn default() -> Self {
         LocalTransform {
-            translation: [0.0, 0.0, 0.0],
-            rotation: [1.0, 0.0, 0.0, 0.0],
-            scale: [1.0, 1.0, 1.0],
+            translation: [0.0, 0.0, 0.0].into(),
```
Rhuagh: `Vector3::zero()`, `Quaternion::one()`, `Vector3::from_value(1.)`

amethyst_core/src/transform/systems.rs
```diff
@@ -238,8 +242,8 @@ mod tests {
         let (mut world, mut system) = transform_world();
 
         let mut transform = LocalTransform::default();
-        transform.translation = [0.0, 0.0, 0.0];
-        transform.rotation = [1.0, 0.0, 0.0, 0.0];
+        transform.translation = Vector3::zero();
+        transform.rotation = Quaternion::new(1.0, 0.0, 0.0, 0.0);
```
Rhuagh: `Quaternion::one()`

examples/03_renderable/main.rs
```diff
@@ -52,12 +52,12 @@ impl<'a> System<'a> for ExampleSystem {
...
 
-        let delta_rot = Quaternion::from_angle_z(Rad(camera_angular_velocity * time.delta_seconds()));
+        let delta_rot =
+            Quaternion::from_angle_z(Rad(camera_angular_velocity * time.delta_seconds()));
         for (_, transform) in (&camera, &mut transforms).join() {
             // rotate the camera, using the origin as a pivot point
             transform.translation = delta_rot
-                .rotate_point(Point3::from(transform.translation))
-                .into();
+                .rotate_point(Point3::from_vec(transform.translation)).to_vec();
```
Rhuagh: `.rotate_vector(transform.translation)`

examples/05_assets/main.rs
```diff
@@ -87,8 +89,8 @@ impl State for AssetsExample {
         };
 
         let mut trans = LocalTransform::default();
-        trans.translation = [-5.0, 0.0, 0.0];
-        trans.scale = [2.0, 2.0, 2.0];
+        trans.translation = Vector3::new(-5.0, 0.0, 0.0);
+        trans.scale = [2.0; 3].into();
```
Rhuagh: in theory, `Vector3::from_value(2.)`, but not necessary.

Xaeroxe:
Rebased and addressed @Rhuagh

Xaeroxe:
@torkleyy Since my current design sort of goes against the grain on our current crate structure allow me to explain myself:

cgmath is compiled with a particular version and set of features, it's of the utmost importance that that version and set of features are used consistently throughout not only our user's games but also our internal ecosystem.  Otherwise, types that seemingly should be compatible, suddenly aren't.  So, in order to guarantee synchronization and reduce the amount of times we have to specify these versions and features all amethyst crates and downstream pull in a cgmath that's re-exported from amethyst_core.  This should allow us to drastically reduce the amount of turmoil caused by future upgrades and feature changes.

torkleyy:
I couldn't spot any logical issues, it's just the implied strategy regarding our math API that I'm not so sure about.

Xaeroxe:
@torkleyy care to elaborate?

torkleyy:
Well, I think the arguments on all sides are clear, we've gone over them a couple of times. It seems we can't find any full agreement, so..

torkleyy:
This is the only way we can make progress.

torkleyy:
So why don't we let @Aceeri approve this?

torkleyy:
Doesn't seem like @Aceeri is going to review this in the near future, so I have to merge this now.

bors r+

PR #464, opened 2017-10-28
Improve error handling for assets

torkleyy:
* Introduces error-chain for assets and core crate

## Example output

```
error: Failed to load asset with name "mesh/teapoto.obj"
caused by: Format "WAVEFRONT_OBJ" could not load asset
caused by: Failed to load bytes from source
caused by: Failed to fetch metadata for "/home/thomas/Workspace/amethyst/examples/assets/mesh/teapoto.obj"
caused by: No such file or directory (os error 2)
note: to handle the error, use a `Progress` other than `()`
```

### DIFF
--- a/amethyst_assets/examples/hl.rs
+++ b/amethyst_assets/examples/hl.rs
@@ -89,7 +89,7 @@ impl SimpleFormat<MeshAsset> for Ron {
         use ron::de::from_str;
         use std::str::from_utf8;
 
-        let s = from_utf8(&bytes).chain_err(|| "Bytes are invalid UTF-8")?;
+        let s = from_utf8(&bytes)?;
 
         from_str(s).chain_err(|| "Failed to decode mesh file")
     }
--- a/amethyst_assets/examples/hl.rs
+++ b/amethyst_assets/examples/hl.rs
@@ -85,13 +85,13 @@ impl SimpleFormat<MeshAsset> for Ron {
 
     type Options = ();
 
-    fn import(&self, bytes: Vec<u8>, _: ()) -> Result<VertexData, BoxedErr> {
+    fn import(&self, bytes: Vec<u8>, _: ()) -> Result<VertexData> {
         use ron::de::from_str;
         use std::str::from_utf8;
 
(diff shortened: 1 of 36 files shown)


amethyst_assets/examples/hl.rs
```diff
@@ -85,13 +85,13 @@ impl SimpleFormat<MeshAsset> for Ron {
 
     type Options = ();
 
-    fn import(&self, bytes: Vec<u8>, _: ()) -> Result<VertexData, BoxedErr> {
+    fn import(&self, bytes: Vec<u8>, _: ()) -> Result<VertexData> {
         use ron::de::from_str;
         use std::str::from_utf8;
 
-        let s = from_utf8(&bytes).map_err(BoxedErr::new)?;
+        let s = from_utf8(&bytes).chain_err(|| "Bytes are invalid UTF-8")?;
```
omni-viral: You can create `foreign_link` in `error_chain` and then just use `?` here.
It will help to avoid boxing error in common cases.

omni-viral: Same in many other places

torkleyy: 1) This isn't "boxing errors"
2) We can't know in advance which errors can occur in asset loading, can we? I don't think it makes any sense to link to some random error types in `amethyst_assets::error`.
3) I like the `chain_err` since you can easily include more information

omni-viral: 1. Original error become boxed
2. Of course we can't know all errors in advance. But we can guess that some will be there. Like `std::io::Error`. But maybe you're right that we should link random errors.
3. Here you include the same information that is already in original error. Two lines below you do better by providing additional information about error context.

torkleyy: Having the error only boxed has other implications than this technique. Regarding `io::Error`: I wouldn't create a link for that, since it does not include file names. That's why I prefer using `chain_err` because you can include more information. Maybe we can make an exception for UTF-8 errors by adding `Source::load_string` which would justify that?

PR #476, opened 2017-11-01
fix(renderer): Reducing max cap for point lights to reduce shader reflection time on OSX.

### LINKED ISSUE #472: Long boot time on macOS
When you launch any amethyst driven program on any macOS system, the boot time (with the white screen) is very long compared to other OS, it's displayed more than 5 seconds.

Rhuagh:
Closes #472.

### DIFF
--- a/amethyst_renderer/src/pass/shaders/fragment/pbm.glsl
+++ b/amethyst_renderer/src/pass/shaders/fragment/pbm.glsl
@@ -17,7 +17,7 @@ struct PointLight {
 };
 
 layout (std140) uniform PointLights {
-    PointLight plight[512];
+    PointLight plight[128];
 };
 
 struct DirectionalLight {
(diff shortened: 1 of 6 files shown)


amethyst_renderer/src/pass/shaders/fragment/pbm.glsl
```diff
@@ -17,7 +17,7 @@ struct PointLight {
 };
 
 layout (std140) uniform PointLights {
-    PointLight plight[512];
+    PointLight plight[128];
```
torkleyy: Would also be interesting to control this with some template substitution.

Xaeroxe:
Hang on before we merge this, there might be a better way: https://www.khronos.org/opengl/wiki/Shader_Storage_Buffer_Object

Rhuagh:
Not sure gfx support that and the limit will disappear when we go to hal.

Xaeroxe:
:+1: merge away then

Rhuagh:
And I'd say, the scenes that need > 128 active point lights will be kind of rare :P

Issue #473, opened 2017-11-01
Targets without depth test display nothing on Mac


Xaeroxe:
Targets without a depth test don't work on Mac, this makes the UI pass useless on OSX.  We should fix this before we release.

omni-viral:
Haven't you tested it with depth target?

Xaeroxe:
@omni-viral As in while using a blend target with a depth buffer?  Yes I have, why?

omni-viral:
It works fine if you specify some depth testing in `with_blend_target`.

Xaeroxe:
Right, I remember now.  So it only fails when you have no depth test on a blend target.

Xaeroxe:
@Rhuagh helped narrow down the problem, so I've updated the issue title and comment.

PR #477, opened 2017-11-02
Add dummy depth buffer on mac

Xaeroxe:
It ain't pretty, but it's the best workaround I've got.

Fixes https://github.com/amethyst/amethyst/issues/473


amethyst_renderer/src/pipe/effect/mod.rs
```diff
@@ -248,6 +257,15 @@ impl<'a> EffectBuilder<'a> {
                 Stencil::default(),
             ));
         }
+        // OSX doesn't seem to work without a depth test, so here's a workaround.
```
Aceeri: Might be better for future refactoring to just use a single method for these two

PR #482, opened 2017-11-04
Add bundle for things we use internally, that can be stored in World

Rhuagh:


Rhuagh:
Failed inside gl-generator crate on a gl api call. I'd say hey again.

Rhuagh:
Try...

Xaeroxe:
We might need to lock the crate responsible to an earlier version and inform the author that they broke SEMVER.  I don't think this is going to work.

Xaeroxe:
I restarted our develop branch to see if we can reproduce without this PR: https://travis-ci.org/amethyst/amethyst/builds/297157029

Xaeroxe:
If you don't get to this in about 12 hours I'll fix it tonight when I get back to a Windows PC

Rhuagh:
Need to lock khronos_api on version 1.0

Rhuagh:
They've been informed they broke the gfx world

Xaeroxe:
@Rhuagh since this isn't our direct dependency I'm not sure that's the right syntax, might want to look into patch dependencies?  Anyways we'll see if this builds or not.

Rhuagh:
It builds on my computer now, it didn't before.

PR #484, opened 2017-11-05
Locking khronos_api at 1.0.1 because 1.1.0 is broken

Rhuagh:
This issue also exists in master and our released version, so this is for fixing those.
Note we need to release a patch version after this has been merged into master.

PR #486, opened 2017-11-05
Revert "Locking khronos_api at 1.0.1 because 1.1.0 is broken"

Xaeroxe:
Reverts amethyst/amethyst#484

This is no longer necessary as the problem crates have been yanked.


Xaeroxe:
Override

PR #489, opened 2017-11-08
UI text input

Xaeroxe:
This is a first implementation that's missing a couple features such as multi-line text and
double click to highlight a word.

This also rewrites pieces of the renderer so that rendering can be done in the main thread without requiring execution in the thread pool.

### DIFF
--- a/amethyst_renderer/src/pipe/pass.rs
+++ b/amethyst_renderer/src/pipe/pass.rs
@@ -1,128 +1,38 @@
 //! Types for constructing render passes.
 
-#![allow(missing_docs)]
-use std::marker::PhantomData;
-
-use rayon::iter::ParallelIterator;
-use rayon::iter::internal::UnindexedConsumer;
-use rayon_core;
 use specs::SystemData;
 
 use error::Result;
 use pipe::{Effect, NewEffect, Target};
 use types::{Encoder, Factory};
 
-/// Is used to pass different `Encoder` and `Effect` into closure in different threads
-pub struct Supplier<'a> {
-    encoders: *mut [Encoder],
-    effects: *mut [Effect],
-    pd: PhantomData<&'a ()>,
-}
-
-impl<'a> Supplier<'a> {
-    /// Create `Supplier` by passing enough `Encoder`s and `Effect`s
-    /// The number is equal to thread count in `ThreadPool`
-    fn new(encoders: &'a mut [Encoder], effects: &'a mut [Effect]) -> Self {
-        Supplier {
-            encoders: encoders,
-            effects: effects,
-            pd: PhantomData,
-        }
-    }
-
-    fn index(&self) -> usize {
-        rayon_core::current_thread_index().expect("Should be called from ThreadPool")
-    }
-
-    /// Dispense mutable references to `Encoder` and `Effect` for slice
-    /// Different threads gets different pair
-    /// unsafe due to ability to call mulitple times
-    /// causing it to return multiple mutable references to the same `Encoder` and `Effect`
-    /// `Apply` use this function once in each thread and
-    /// drops references before calling it again
-    unsafe fn get(&self) -> (&mut Encoder, &mut Effect) {
-        let slice = &mut *self.encoders;
-        let count = slice.len();
-        let encoder = slice.get_mut(self.index()).expect(&format!(
-            "Not enough objects. Index: {}, Supplier count: {}",
-            self.index(),
-            count
-        ));
-        let slice = &mut *self.effects;
-        let count = slice.len();
-        let effect = slice.get_mut(self.index()).expect(&format!(
-            "Not enough objects. Index: {}, Supplier count: {}",
-            self.index(),
-            count
-        ));
-        (encoder, effect)
-    }
-}
-
-pub struct Apply<'a, I> {
-    inner: I,
-    supplier: Supplier<'a>,
-}
-
-impl<'a, I> ParallelIterator for Apply<'a, I>
-where
-    I: ParallelIterator,
-    I::Item: FnOnce(&mut Encoder, &mut Effect),
-{
-    type Item = ();
-
-    fn drive_unindexed<C>(self, consumer: C) -> C::Result
-    where
-        C: UnindexedConsumer<()>,
-    {
-        let Apply { inner, supplier } = self;
-        inner
-            .map(move |f| {
-                let (encoder, effect) = unsafe { supplier.get() };
-                effect.clear();
-                f(encoder, effect);
-            })
-            .drive_unindexed(consumer)
-    }
-}
-
-impl<'a> Supplier<'a> {
-    pub fn supply<I, F>(self, iter: I) -> Apply<'a, I>
-    where
-        I: ParallelIterator<Item = F>,
-        F: FnOnce(&mut Encoder, &mut Effect) + Send,
-    {
-        Apply {
-            inner: iter,
-            supplier: self,
-        }
-    }
-}
-
-unsafe impl<'a> Send for Supplier<'a> {}
-unsafe impl<'a> Sync for Supplier<'a> {}
-
-
-pub trait PassApply<'a> {
-    type Apply: ParallelIterator<Item = ()>;
-}
-
+/// Used to fetch data from the game world for rendering in your pass.
 pub trait PassData<'a> {
+    /// The data itself.
     type Data: SystemData<'a> + Send;
 }
 
-pub trait Pass: for<'a> PassApply<'a> + for<'a> PassData<'a> + Send + Sync {
+/// Structures implementing this provide a renderer pass.
+pub trait Pass: for<'a> PassData<'a> {
+    /// The pass is given an opportunity to compile shaders and store them in an `Effect`
+    /// which is then passed to the pass in `apply`.
     fn compile(&self, effect: NewEffect) -> Result<Effect>;
+    /// Called whenever the renderer is ready to apply the pass.  Feed commands into the
+    /// encoder here.
     fn apply<'a, 'b: 'a>(
         &'a mut self,
-        supplier: Supplier<'a>,
+        encoder: &mut Encoder,
+        effect: &mut Effect,
+        factory: Factory,
         data: <Self as PassData<'b>>::Data,
-    ) -> <Self as PassApply<'a>>::Apply;
+    );
 }
 
+/// A compiled pass.  These are created and managed by the `Renderer`.  You should never need to
+/// use this.
 #[derive(Clone, Debug)]
 pub struct CompiledPass<P> {
-    effects: Vec<Effect>,
+    effect: Effect,
     inner: P,
 }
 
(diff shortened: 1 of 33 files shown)


amethyst_ui/src/focused.rs
```diff
@@ -0,0 +1,7 @@
+use specs::Entity;
+
+/// This resource with id 0 stores the currently focused UI element.
```
torkleyy: I don't think we need to describe the id anywhere. IDs are for scripting, there's no reason to use it for anything else.

Xaeroxe:
All of our rendering passes currently have much improved performance when executing serially rather than in parallel.   In addition if you had more passes than cores there was actually a chance this would draw out of order.  So that bug was fixed in this PR.  If a pass needs parallel execution it should be able to retrieve our `ThreadPool` from the `specs::World`.  However one remaining problem with this approach is how does a parallel pass knows when to flush its encoders?  Right now there isn't a correct way for the pass to do this.  One possible solution for this is to flush the encoder after every pass rather than at the end of all passes, which would work but also means potentially inferior batching.  I'll have to play around with the approach and see if we lose too much FPS from doing that.

Xaeroxe:
Things that are now working in text editing:
- Ctrl + shift + arrow key to select words
- Using the left and right arrow keys without modifiers when text is highlighted
- Home and End keys work correctly
- Clipboard works correctly

Things that aren't working
- Anything mouse related
- Up and down arrow keys
- Tabbing between fields

Xaeroxe:
I also learned this doesn't handle multi line text very well, because I made the assumption early on that glyphs == graphemes.  While the two concepts are very similar graphemes such as a new line, are not glyphs.

torkleyy:
I think tabbing between fields is something that should be implemented by another PR.

jojolepro:
Along with "focused" and "focus-able" components.

Xaeroxe:
@torkleyy @jojolepro these aren't even the hard part of this PR, seriously those pieces are easy and I basically already have it down how I'm going to handle them.  Focusing components is already built on the data side I just haven't coded the input piece yet.

Honestly the thing that's got my head turning is multi-line text.  That's going to be a bit complicated with the way this is built currently.

Xaeroxe:
Also, you know how when entering text into a single line field if the text is wider than the input box the box will auto-advance the rendered text to the cursor?  This doesn't do that yet, and it really should.  I think it'll be as as simple as using the glyph widths and advancing by those when needed, but it just hasn't been coded yet.

amethyst_ui/src/pass.rs
```diff
@@ -297,6 +293,16 @@ impl Pass for DrawUi {
                 }
                 // Build text sections.
                 let editing = editing.get(entity);
+                let rendered_string = if ui_text.password {
+                    // Build a string composed of black dot characters.
+                    let mut ret = String::with_capacity(ui_text.text.len());
+                    for _grapheme in ui_text.text.graphemes(true) {
```
torkleyy: Could just do `ui_text.text.graphemes(true).map(|_| 'u{2022}').collect()`

amethyst_ui/src/pass.rs
```diff
@@ -297,6 +293,16 @@ impl Pass for DrawUi {
...
+                        ret.push('\u{2022}');
+                    }
+                    ret
+                } else {
+                    ui_text.text.clone()
```
torkleyy: We don't need to clone; just return a reference above

torkleyy:
I'm building a small Amethyst demo atm, it would be nice to have some text fields there. When do you plan on merging it?

Xaeroxe:
@torkleyy do you need multi-line text? I can probably finish this much more quickly if I cut that

torkleyy:
I don't need multiline input, no. I'd like to finish the demo on Wednesday, the input fields would just be nice, not necessary.

amethyst_ui/src/pass.rs
Xaeroxe: This has the disadvantage that the string isn't allocated with the correct capacity, thus potentially causing more allocations as the string gets resized.

Xaeroxe:
There are two features missing from this that I'd really like to get in.  Clicking on fields to focus and/or highlight them doesn't work, and if your text is wider than your box it just won't render.  

The first feature I'm not completely sure what to do with, as we need double clicks for that.  Winit doesn't emit double clicks, I'd like it to but it's non-trivial to make it emit double clicks because a double click event was considered out of scope for X11, so you have to retrieve it from elsewhere on the system.  How you fetch that changes from machine to machine.  We could just ignore the system double click time setting and just use 500 ms everywhere, but that feels sub-optimal.  However that said it might be better than just doing nothing.  I have not researched this very thoroughly so if anyone has any corrections to offer please tell me.

Getting the second feature in is likely going to require changes to gfx_glyph so that the text will still continue to render if it's too big, just only in the box allocated to it.  We needed maximum text length anyways, so I added that early as a stop-gap to help avoid triggering this bug.

Xaeroxe:
So while this is still missing double click to highlight a word, and multi-line text it is otherwise ready and in my opinion a satisfactory first implementation.  So I'm removing the WIP.

Ready for review!

Xaeroxe:
r? @torkleyy 

This is ready.

amethyst_renderer/src/pipe/pass.rs
```diff
@@ -1,128 +1,38 @@
...
-
-unsafe impl<'a> Send for Supplier<'a> {}
-unsafe impl<'a> Sync for Supplier<'a> {}
-
-
-pub trait PassApply<'a> {
-    type Apply: ParallelIterator<Item = ()>;
-}
-
+/// Used to fetch data from the game world for rendering in your pass.
```
Rhuagh: `the pass.` maybe ?

amethyst_renderer/src/pipe/pass.rs
```diff
@@ -1,128 +1,38 @@
...
-        supplier: Supplier<'a>,
+        encoder: &mut Encoder,
+        effect: &mut Effect,
+        factory: Factory,
         data: <Self as PassData<'b>>::Data,
-    ) -> <Self as PassApply<'a>>::Apply;
+    );
 }
 
+/// A compiled pass.  These are created and managed by the `Renderer`.  You should never need to
```
Rhuagh: `This should not be used directly outside the renderer` ?

examples/renderable/main.rs
```diff
@@ -347,18 +374,28 @@ fn run() -> Result<(), Error> {
...
-        Stage::with_backbuffer()
-            .clear_target([0.0, 0.0, 0.0, 1.0], 1.0)
-            .with_pass(DrawShaded::<PosNormTex>::new()),
-    );
 
-    let mut game = Application::build(resources_directory, Example)?
+
+    let game = Application::build(resources_directory, Example)?
         .with::<ExampleSystem>(ExampleSystem, "example_system", &[])
+        .with_frame_limit(FrameRateLimitStrategy::Unlimited, 0)
```
Rhuagh: Remove this maybe?

amethyst_ui/src/text.rs
```diff
@@ -1,27 +1,42 @@
...
+    /// The string rendered by this.
+    pub text: String,
+    /// The height of a line of text in pixels.
+    pub font_size: f32,
+    /// The color of the rendered text, using a range of 0.0 to 1.0 per channel.
+    pub color: [f32; 4],
+    /// The font used for rendering.
+    pub font: FontHandle,
+    /// If true this will be rendered as dots instead of the text.
+    pub password: bool,
```
Rhuagh: Not sure `password` is the correct field name for this. `hidden` maybe? Or maybe an enum for how to render the text (Normal, Hidden, Dotted) or some such ? Not sure, just an idea.

examples/renderable/main.rs
Xaeroxe: I was using this to benchmark the example's performance so I could see how text rendering impacted the FPS.  I decided to leave it because it's probably nice to show off the frame rate anyways on an example called `renderable`.  What do you think?

Xaeroxe:
travis-ci failure: `oci runtime error: exec failed: container_linux.go:265: starting container process caused "could not create session key: disk quota exceeded`

Nice job travis.  I'll wait for them to go run out and buy a hard drive I guess.

examples/renderable/main.rs
Rhuagh: As long as peoples computers don't burn to cinders I'm fine with it :P

Rhuagh:
I think they run on EC2, so it's more like "gives Amazon a larger heap of money"

amethyst_ui/src/pass.rs
```diff
@@ -209,87 +199,373 @@ impl<'a> ParallelIterator for DrawUiApply<'a> {
...
+                };
 
-                    if let Some(image) = ui_text
-                        .get(entity)
-                        .and_then(|ref ui_text| ui_text.texture.as_ref())
-                        .and_then(|texture| tex_storage.get(texture))
+                // Render background highlight
+                let brush = &mut self.glyph_brushes
+                    .get_mut(&ui_text.brush_id.unwrap())
+                    .unwrap()
```
Aceeri: Are all these unwraps safe?

amethyst_ui/src/text.rs
```diff
@@ -35,133 +50,682 @@ impl UiText {
...
+                                    focused_text.cached_glyphs.iter().last()
+                                {
+                                    if (last_glyph.position().x - mouse_x).abs()
+                                        > ((last_glyph.position().x
+                                            + last_glyph
+                                                .unpositioned()
+                                                .h_metrics()
+                                                .advance_width)
+                                            - mouse_x)
+                                            .abs()
```
Aceeri: This is some weird formatting...

Xaeroxe: rustfmt-nightly 0.2.16 official.  But yeah it does look odd.

amethyst_ui/src/pass.rs
Xaeroxe: yes they are, because the code beforehand proved the unwrap safe.  The only reason option values are used here is because the structures can't be populated at initialization.

amethyst_ui/src/lib.rs
```diff
@@ -25,9 +31,13 @@ mod text;
...
 pub use self::format::{FontAsset, FontHandle, OtfFormat, TtfFormat};
 pub use self::image::UiImage;
 pub use self::pass::DrawUi;
 pub use self::resize::{ResizeSystem, UiResize};
-pub use self::text::{UiText, UiTextRenderer};
+pub use self::text::{TextEditing, UiSystem, UiText};
 pub use self::transform::UiTransform;
+
+/// How many times the cursor blinks per second while editing text.
+const CURSOR_BLINK_RATE: f32 = 2.0;
```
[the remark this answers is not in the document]
Xaeroxe: When I asked if this should be hardcoded no one seemed to care to make it configurable, so a resource might be overkill.

amethyst_ui/src/pass.rs
```diff
@@ -118,79 +148,39 @@ impl Pass for DrawUi {
...
-        let mesh_storage = &mesh_storage;
-        let tex_storage = &tex_storage;
-        let ui_image = &ui_image;
-        let ui_text = &ui_text;
-        let ui_transform = &ui_transform;
-        let unit_mesh = &unit_mesh;
-
+    ) {
         // Populate and update the draw order cache.
         {
```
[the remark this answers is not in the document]
Xaeroxe: yeah I thought about that, it turns out the overhead from those storage types actually ends up being worse than this setup, but only because this setup is very specifically optimized for tracking changes to one field.

examples/renderable/main.rs
Xaeroxe: Well, mine didn't :D

amethyst_ui/src/text.rs
Xaeroxe: What would be the difference between Hidden and Dotted?

Rhuagh: Hidden = nothing, Dotted= the current behavior for password

amethyst_ui/src/lib.rs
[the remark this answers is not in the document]
Rhuagh: Not a resource imho, possibly a configurable system parameter at some point.

amethyst_ui/src/text.rs
Xaeroxe: What would be the use case for hidden in this context? I know sometimes that's used for password entry in console programs but I haven't seen anywhere else use it.

amethyst_ui/src/bundle.rs
```diff
@@ -34,15 +29,17 @@ impl<'a, 'b> ECSBundle<'a, 'b> for UiBundle {
...
         world.add_resource(AssetStorage::<FontAsset>::new());
+        world.add_resource(UiFocused { entity: None });
         let reader = world
             .read_resource::<EventChannel<Event>>()
             .register_reader();
         Ok(
             builder
-                .add(UiTextRenderer, "ui_text", self.deps)
+                .add(UiSystem::new(reader.clone()), "ui_system", &[])
                 .add(Processor::<FontAsset>::new(), "font_processor", &[])
```
torkleyy: It would be better if the processor was before the ui system

amethyst_ui/src/lib.rs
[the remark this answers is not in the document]
torkleyy: It should be configurable, but let's leave it for now.

amethyst_ui/src/pass.rs
```diff
@@ -36,18 +41,40 @@ struct CachedDrawOrder {
...
     mesh_handle: MeshHandle,
     cached_draw_order: CachedDrawOrder,
+    cached_color_textures: HashMap<KeyColor, TextureHandle>,
+    glyph_brushes: HashMap<
+        u32,
+        (
+            GlyphBrush<'static, Resources, Factory>,
+            WeakHandle<FontAsset>,
+        ),
+    >,
```
torkleyy: Ugh, a typedef would probably be better here.

amethyst_ui/src/pass.rs
```diff
@@ -209,87 +199,373 @@ impl<'a> ParallelIterator for DrawUiApply<'a> {
...
-                {
-                    // This won't panic as we guaranteed earlier these entities are present.
-                    let ui_transform = ui_transform.get(entity).unwrap();
-                    let mesh = match mesh_storage.get(unit_mesh) {
-                        Some(mesh) => mesh,
-                        None => return,
-                    };
-                    let vbuf = match mesh.buffer(PosTex::ATTRIBUTES) {
-                        Some(vbuf) => vbuf.clone(),
+        let vbuf = match mesh.buffer(PosTex::ATTRIBUTES) {
```
torkleyy: In a few weeks we can use `?` for this :)

amethyst_ui/src/pass.rs
Xaeroxe: It's 100% internal and this is the only place it's written out.  Would you still have me use a typedef for this?

torkleyy: It just doesn't look very nice in code, that's all.

Xaeroxe:
I've rebased, addressed review comments, and have two approvals.  This is also over 24 hours old so I'm going to go ahead and merge this.

@Rhuagh if we want to change the field name still we can do that in a follow-up PR.

bors r+

torkleyy:
> This is also over 24 hours old so I'm going to go ahead and merge this.

While that's true, it wasn't ready for review for 24 hours. That said, I don't have any concerns, so it's fine by me.

torkleyy:
It seems you made a commit after bors merged your branch. Are you sure it gets included?

Xaeroxe:
@torkleyy Yes the commits are being included correctly.  The timestamps were a little off because my local machine disagrees with Github on what time it is by a few minutes but they correct commits are there.

I'll make sure to time it from the review opening period in the future, that's a good note.

PR #491, opened 2017-11-15
Useful time additions.

jojolepro:
Added:
-Delta real time
-Delta real time seconds
-Absolute time
-Absoulte real time
-Time multiplier
Fixed:
-FPSCounter needs the real-time delta time to show the correct FPS value

### DIFF
--- a/amethyst_core/src/timing.rs
+++ b/amethyst_core/src/timing.rs
@@ -9,6 +9,10 @@ pub struct Time {
     delta_seconds: f32,
     /// Time elapsed since the last frame.
     delta_time: Duration,
+    /// Time elapsed since the last frame in seconds ignoring the time speed multiplier.
+    delta_real_seconds: f32,
+    /// Time elapsed since the last frame ignoring the time speed multiplier.
+    delta_real_time: Duration,
     /// Rate at which `State::fixed_update` is called in seconds.
     fixed_seconds: f32,
     /// Rate at which `State::fixed_update` is called.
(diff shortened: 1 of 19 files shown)


jojolepro:
Please be aware that any audio playing will NOT be affected by the time multiplier. Affecting audio should be the choice of the user, since it makes sense for rhythm games and slow motion cutscenes, but doesn't for small noises like hit sounds or noises that don't stretch well.

amethyst_core/src/timing.rs
```diff
@@ -50,24 +70,58 @@ impl Time {
...
     }
 
     /// Sets both `delta_time` and `delta_seconds` based on the duration given.
     ///
     /// This should only be called by the engine.  Bad things might happen if you call this in
     /// your game.
     pub fn set_delta_time(&mut self, time: Duration) {
-        self.delta_seconds = duration_to_secs(time);
-        self.delta_time = time;
+        self.delta_seconds = duration_to_secs(time) * self.time_multiplier;
```
Rhuagh: I would prefer if we don't change the semantics of an established field in this way, please change so the old fields retain their functionality, and add the time scale to new fields instead.

Xaeroxe:
Oh so this will collide with my changes #489 .  That's fine, I'll fix my PR to use this.

amethyst_core/src/timing.rs
jojolepro: I personally think it makes more sense to have time affected by the multiplier by default. It won't affect ANY game nor the engine. To affect a game, the user needs to set the speed multiplier himself. To affect the engine... well I already fixed the places where it was using delta_something and it was supposed to use real time.
The cases where you want to have real time make more sense into the engine, as it is what "controls" the simulation. End-users won't even notice the difference :)

Xaeroxe: @Rhuagh I'm not sure if a `scaled_delta_time` is the right thing here, as this kind of functionality can make it really easy to quickly introduce pausing functionality to a game by setting the time scale to 0.  So I'd actually prefer we encourage usage of a scaled delta time by default with the default scaling set to 1.

Unless the end user asks to scale the time, this doesn't change the use of the field.

Xaeroxe:
Should we permit the user to set a negative time scale?

jojolepro:
We can, but there will be some weird behaviours. For example, if you go forward, the physic system will make you go back (input * delta time). It wouldn't be a "replay" going backward but more like a "everything is inverted". Gravity would go up :D

Xaeroxe:
While it would probably be hilarious it'd also be a source of bugs for both users and engine developers who (reasonably) anticipate that the delta_time value will never be negative.  Imagine a health recharge process actually causing health to go down because of a negative delta_time.

I think we probably shouldn't permit that :P

jojolepro:
I just tested using a negative number. Any number I use will make my game run backward really fast (even -0.1) and if feels like x10.

Xaeroxe:
@jojolepro a time scale value of infinity or NaN would also cause really bizarre things.

jojolepro:
That's because Duration.nanos being a u32

Xaeroxe:
My PR against this PR: https://github.com/jojolepro/amethyst/pull/2

amethyst_core/src/timing.rs
Rhuagh: I can accept that.

amethyst_core/src/timing.rs
```diff
@@ -17,25 +21,41 @@ pub struct Time {
     pub last_fixed_update: Instant,
     /// The total number of frames that have been played in this session.
     frame_number: u64,
+    ///Time elapsed since game start, ignoring the speed multipler.
+    absolute_real_time: Duration,
+    ///Time elapsed since game start, taking the speed multiplier into account.
+    absolute_time: Duration,
+    ///Time multiplier. Affects returned delta_seconds, delta_time and absolute_time.
+    time_multiplier: f32,
```
Xaeroxe: We should call this and the accompanying functions `time_scale`

jojolepro:
I'll rename time_multiplier to time_scale according to suggestion somewhere tomorrow (or today if I'm bored)

jojolepro:
I may or may not have forgot to do that today. I'll check if I can do that in around 2 hours, otherwise it will be for tomorrow. Sorry for delay :/

Xaeroxe:
thanks!

bors r+

jojolepro:
:tada:

PR #493, opened 2017-11-21
Add rustc version to init output

Xaeroxe:
With the memory layout changing in rustc soon it occurred to me that the compiler version and branch used might be critical for debugging issues in the future, so I added rustc version info to the initialization text.  Some example output looks like this:

```
Initializing Amethyst...
Version:
Platform: x86_64-pc-windows-msvc
Amethyst git commit: afa10de66d5b1e6613ed7290aa123639b76fe0fa
Rustc version: 1.21.0 Stable
Rustc git commit: 3b72af97e42989b2fe104d8edbaee123cdf7c58f
```

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -307,17 +307,16 @@ impl<'a, 'b, T> ApplicationBuilder<'a, 'b, T> {
 
     pub fn new<P: AsRef<Path>>(path: P, initial_state: T) -> Result<Self> {
         use bundle::AppBundle;
-        use rustc_version;
+        use rustc_version_runtime;
 
         println!("Initializing Amethyst...");
         println!("Version: {}", vergen::semver());
         println!("Platform: {}", vergen::target());
         println!("Amethyst git commit: {}", vergen::sha());
-        if let Ok(rustc_meta) = rustc_version::version_meta() {
-            println!("Rustc version: {} {:?}", rustc_meta.semver, rustc_meta.channel);
-            if let Some(hash) = rustc_meta.commit_hash {
-                println!("Rustc git commit: {}", hash);
-            }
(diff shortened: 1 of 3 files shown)


src/app.rs
```diff
@@ -307,11 +307,18 @@ impl<'a, 'b, T> ApplicationBuilder<'a, 'b, T> {
...
         println!("Version: {}", vergen::semver());
         println!("Platform: {}", vergen::target());
-        println!("Git commit: {}", vergen::sha());
+        println!("Amethyst git commit: {}", vergen::sha());
+        if let Ok(rustc_meta) = rustc_version::version_meta() {
+            println!("Rustc version: {} {:?}", rustc_meta.semver, rustc_meta.channel);
+            if let Some(hash) = rustc_meta.commit_hash {
+                println!("Rustc git commit: {}", hash);
+            }
+        }
```
minecrawler: Under which conditions can `rustc_version::version_meta()` fail? Would it make sense to write the error on `Err`? Something like

> Rustc version: Error: unknown pre-release tag

Xaeroxe: This will fail if it wasn't able to parse the version output from rustc during the build process.  It seems unlikely this would happen so I didn't code anything for it.  If I turn out to be wrong I'll add a failure check here.

Xaeroxe:
Do not merge this yet, I just discovered I'm using this incorrectly.  This fetches the rustc available on the computer currently running the game, not the computer that compiled the game.  I'll restructure this to work similar to our vergen setup.

Xaeroxe:
Fixed.

Xaeroxe:
Assigned to @Rhuagh for r+ following approval.

PR #494, opened 2017-11-22
Use higher precison for absolute timing

Xaeroxe:
After doing some research I determined f32 just doesn't provide enough precision for this.  It starts breaking down and causing problems in less than a week of playtime.  While it's unusual for people to play for multiple days straight, I have observed players simply not closing their games and leaving the executable running for in some cases months at a time.  An f64 will last a great many years by contrast, and won't have this problem in a reasonable amount of time.

### DIFF
--- a/amethyst_core/src/timing.rs
+++ b/amethyst_core/src/timing.rs
@@ -76,8 +76,8 @@ impl Time {
     }
 
     /// Gets the time since the start of the game as seconds, taking into account the speed multiplier.
-    pub fn absolute_time_seconds(&self) -> f32 {
-        duration_to_secs(self.absolute_time)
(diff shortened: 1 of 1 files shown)


Xaeroxe:
While I agree with you that doesn't change the fact that people do it :(

Rhuagh:
A game client will usually not run for that long, but we want to use amethyst for the game server side aswell? That _will_ run for weeks at a time most likely.

torkleyy:
Double precision is the default for timing stuff anyways, so I don't see a reason for not merging this change.

Xaeroxe:
bors r+

We have enough approvals and this change is pretty small.

PR #497, opened 2017-11-27
Add OSX to Travis

Xaeroxe:
Travis seems to have significantly improved their OSX infrastructure, so I'd like to turn this on again.


Xaeroxe:
Uh disregard this.  I was looking at the travis backlog on the weekend while everyone was asleep.  They still have problems.

PR #505, opened 2017-12-07
[WIP] Texture animation

### LINKED ISSUE #510: Loaded sprites don't have transparencies
Hi, I've created a simple Amethyst example, based on the nightly build, to demonstrate loading a sprite from the assets and displaying it. My intent is to add it to the list of examples in the repository.

The example is here: [trsoluti/sprite_game.rs](https://gist.github.com/trsoluti/9cc7e1fb41635cb9ded6d240cc47869d).

However, the transparent parts of the logo show up as white instead of the background colour.

Am I doing something incorrectly? Or is there an issue in the renderer?

Xaeroxe:
This is a work in progress PR with only data structures at the moment.  Processing and generation of these data structures will come later, possibly this weekend. (Written as of 2017-12-07)

My plan for the future is to have the existing passes process these animation values and use them in a vertex shader to modify texture coordinates.  This could also be done in a fragment shader.  I'm not really picky.

Fixes #510 

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/505)

### DIFF
--- a/amethyst_core/src/timing.rs
+++ b/amethyst_core/src/timing.rs
@@ -378,3 +378,8 @@ pub fn duration_to_nanos(duration: Duration) -> u64 {
 pub fn nanos_to_duration(nanos: u64) -> Duration {
     Duration::new(nanos / 1_000_000_000, (nanos % 1_000_000_000) as u32)
 }
+
+/// Multiplies a given Duration by the float provided.
+pub fn duration_f32_mul(duration: Duration, float: f32) -> Duration {
+    Duration::new((duration.as_secs() as f32 * float) as u64, (duration.subsec_nanos() as f32 * float) as u32)
+}
--- a/amethyst_renderer/src/tex_animation.rs
+++ b/amethyst_renderer/src/tex_animation.rs
@@ -1,0 +1,133 @@
+use std::time::Duration;
+
+use amethyst_assets::{Asset, AssetStorage, Handle};
+use fnv::FnvHashMap as HashMap;
+use specs::VecStorage;
+
+
+/// An asset handle to sprite sheet metadata.
+pub type SpriteSheetDataHandle = Handle<SpriteSheetData>;
+
+/// Meta data for a sprite sheet texture.  Does not contain a texture, only a description of how
+/// a texture can be animated.
+#[derive(Clone, Debug)]
+pub struct SpriteSheetData {
+    /// A list of frames in this spritesheet.
+    pub frames: Vec<Frame>,
+    /// A collection of animations, the first layer contains a list of "animations" and the second
+    /// layer contains a list of frame indices within the animation.
+    pub animations: Vec<Vec<usize>>,
+    /// A mapping between string names and indexes into the first layer of the animations member.
+    /// This should only be used when switching animation by string name.
+    pub animation_mapping: HashMap<String, usize>,
+}
+
+impl Asset for SpriteSheetData {
+    type Data = Self;
+    type HandleStorage = VecStorage<Handle<Self>>;
+}
+
+impl Into<Result<SpriteSheetData, ()>> for SpriteSheetData {
+    fn into(self) -> Result<SpriteSheetData, ()> {
+        Ok(self)
+    }
+}
+
+/// A component describing the current state of animation for a sprite sheet.
+#[derive(Clone, Debug)]
+pub struct SpriteSheetAnimation {
+    /// The multiplier for playback speed.  0 is paused, 1 is normal speed.
+    /// Supports negative values as well.
+    pub playback_speed: f32,
+    /// The SpriteSheetData we are animating.
+    sprite_sheet_data: SpriteSheetDataHandle,
+    /// The animation currently playing, this is an index into the first layer of the animations
+    /// in sprite_sheet_data.
+    current_animation: usize,
+    /// The current frame in the current animation.  This is an index into the second layer of
+    /// the animations in sprite_sheet_data.
+    current_frame: usize,
+    /// How long the current frame has been played for.
+    frame_timer: Duration,
+}
+
+impl SpriteSheetAnimation {
+    /// Create a new SpriteSheetAnimation component.
+    ///
+    /// ## Parameters
+    ///
+    /// - `playback_speed`: A multiplier that makes an animation go faster or slower.
+    /// - `sprite_sheet_data`: The sprite sheet data to be used.
+    /// - `sheet_storage`: The storage that `sprite_sheet_data` is from.
+    /// - `starting_animation`: The named case-sensitive animation from 'sprite_sheet_data' that
+    /// this should play initially.
+    ///
+    /// ## Returns
+    /// If the `starting_animation` wasn't found then this will return `None`.  This could be due
+    /// to one of a few things:
+    ///
+    /// - The `sprite_sheet_data` hadn't loaded yet.  Handles aren't valid in their first frame of
+    /// existence.
+    /// - An animation with the name from `starting_animation` wasn't found in `sprite_sheet_data`.
+    /// - The `sprite_sheet_data` handle wasn't made with the `sheet_storage` AssetStorage.
+    pub fn new(playback_speed: f32, sprite_sheet_data: SpriteSheetDataHandle, sheet_storage: &AssetStorage<SpriteSheetData>, starting_animation: &str) -> Option<SpriteSheetAnimation> {
+        sheet_storage.get(&sprite_sheet_data).and_then(|data| data.animation_mapping.get(starting_animation).map(|i| *i)).map(|current_animation| {
+            SpriteSheetAnimation {
+                playback_speed,
+                sprite_sheet_data,
+                current_animation,
+                current_frame: 0,
+                frame_timer: Duration::from_secs(0),
+            }
+        })
+
+    }
+
+    /// Sets the animation to the named case-sensitive animation from the `SpriteSheetData`.
+    ///
+    /// ## Returns
+    ///
+    /// If the name provided can't be found in the `SpriteSheetData` for this component then this
+    /// will return `Err` otherwise it will return `Ok`.
+    pub fn set_animation(&mut self, sheet_storage: &AssetStorage<SpriteSheetData>, animation: &str) -> Result<(), ()> {
+        let animation = sheet_storage.get(&self.sprite_sheet_data).and_then(|data| data.animation_mapping.get(animation).map(|i| *i));
+        match animation {
+            Some(animation) => {
+                self.current_animation = animation;
+                self.current_frame = 0;
+                self.frame_timer = Duration::from_secs(0);
+                Ok(())
+            }
+            None => {
+                Err(())
+            }
+        }
+    }
+
+    /// Retrieve frame information for the current frame.
+    ///
+    /// ## Returns
+    ///
+    /// Most of the time this should return `Some`.  If it doesn't then either the
+    /// `SpriteSheetDataHandle` used in initialization or the `sheet_storage` in this function are
+    /// invalid.  Because handles take at least one frame to become valid after being loaded the
+    /// handle is the most likely point of failure.
+    pub fn current_frame(&self, sheet_storage: &AssetStorage<SpriteSheetData>) -> Option<Frame> {
+        sheet_storage.get(&self.sprite_sheet_data).map(|data| data.frames[data.animations[self.current_animation][self.current_frame]].clone())
+    }
+}
+
+/// A description of a frame in a spritesheet.
+#[derive(Clone, Debug)]
+pub struct Frame {
+    /// Normalized x coordinate, 0 is the left side and 1 is the right side
+    pub x: f32,
+    /// Normalized y coordinate, 0 is the top and 1 is the bottom
+    pub y: f32,
+    /// Normalized width, 0 is no length and 1 is the full width of the texture
+    pub width: f32,
+    /// Normalized height, 0 is no height and 1 is the full height of the texture
+    pub height: f32,
+    /// The duration this frame should be played for.
+    pub duration: Duration,
+}
--- a/amethyst_renderer/src/tex_animation.rs
+++ b/amethyst_renderer/src/tex_animation.rs
@@ -1,5 +1,5 @@
-use std::fmt::{Display, Error as FmtError, Formatter};
 use std::error::Error;
+use std::fmt::{Display, Error as FmtError, Formatter};
 
 use amethyst_assets::{Asset, AssetStorage, Handle};
 use amethyst_core::Time;
@@ -7,7 +7,6 @@ use fnv::FnvHashMap as HashMap;
 use specs::{Fetch, Join, System, VecStorage, WriteStorage};
 
 use MaterialAnimation;
-
 
 /// An asset handle to sprite sheet metadata.
 pub type SpriteSheetDataHandle = Handle<SpriteSheetData>;
@@ -49,12 +48,10 @@ pub enum SetAnimationError {
 
 impl Display for SetAnimationError {
     fn fmt(&self, f: &mut Formatter) -> Result<(), FmtError> {
-        f.write_str(
-            match self {
-                &SetAnimationError::AnimationNotFound => "Animation was not found.",
-                &SetAnimationError::HandleInvalid => "SpriteSheetData Handle given is invalid."
-            }
-        )
+        f.write_str(match self {
+            &SetAnimationError::AnimationNotFound => "Animation was not found.",
+            &SetAnimationError::HandleInvalid => "SpriteSheetData Handle given is invalid.",
+        })
     }
 }
 
@@ -62,7 +59,7 @@ impl Error for SetAnimationError {
     fn description(&self) -> &str {
         match self {
             &SetAnimationError::AnimationNotFound => "Animation was not found.",
-            &SetAnimationError::HandleInvalid => "SpriteSheetData Handle given is invalid."
+            &SetAnimationError::HandleInvalid => "SpriteSheetData Handle given is invalid.",
         }
     }
 
@@ -111,17 +108,22 @@ impl SpriteSheetAnimation {
     /// existence.
     /// - An animation with the name from `starting_animation` wasn't found in `sprite_sheet_data`.
     /// - The `sprite_sheet_data` handle wasn't made with the `sheet_storage` AssetStorage.
-    pub fn new(playback_speed: f32, sprite_sheet_data: SpriteSheetDataHandle, sheet_storage: &AssetStorage<SpriteSheetData>, starting_animation: &str) -> Option<SpriteSheetAnimation> {
-        sheet_storage.get(&sprite_sheet_data).and_then(|data| data.animation_mapping.get(starting_animation).map(|i| *i)).map(|current_animation| {
-            SpriteSheetAnimation {
+    pub fn new(
+        playback_speed: f32,
+        sprite_sheet_data: SpriteSheetDataHandle,
+        sheet_storage: &AssetStorage<SpriteSheetData>,
+        starting_animation: &str,
+    ) -> Option<SpriteSheetAnimation> {
+        sheet_storage
+            .get(&sprite_sheet_data)
+            .and_then(|data| data.animation_mapping.get(starting_animation).map(|i| *i))
+            .map(|current_animation| SpriteSheetAnimation {
                 playback_speed,
                 sprite_sheet_data,
                 current_animation,
                 current_frame: 0,
                 frame_timer: 0.0,
-            }
-        })
-
+            })
     }
 
     /// Sets the animation to the named case-sensitive animation from the `SpriteSheetData`.
@@ -130,8 +132,17 @@ impl SpriteSheetAnimation {
     ///
     /// If the name provided can't be found in the `SpriteSheetData` for this component then this
     /// will return `Err` otherwise it will return `Ok`.
-    pub fn set_animation(&mut self, sheet_storage: &AssetStorage<SpriteSheetData>, animation: &str) -> Result<(), SetAnimationError> {
-        let animation = *sheet_storage.get(&self.sprite_sheet_data).ok_or(SetAnimationError::HandleInvalid)?.animation_mapping.get(animation).ok_or(SetAnimationError::AnimationNotFound)?;
+    pub fn set_animation(
+        &mut self,
+        sheet_storage: &AssetStorage<SpriteSheetData>,
+        animation: &str,
+    ) -> Result<(), SetAnimationError> {
+        let animation = *sheet_storage
+            .get(&self.sprite_sheet_data)
+            .ok_or(SetAnimationError::HandleInvalid)?
+            .animation_mapping
+            .get(animation)
+            .ok_or(SetAnimationError::AnimationNotFound)?;
         self.current_animation = animation;
         self.current_frame = 0;
         self.frame_timer = 0.0;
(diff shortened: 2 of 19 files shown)


omni-viral:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[amethyst_renderer/src/tex_animation.rs, line 11 at r1](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L-uMqqH6tZN3ymHLwmk:-L-uMqqH6tZN3ymHLwml:b-7m5yci) ([raw file](https://github.com/amethyst/amethyst/blob/bfbf0706c91aeeefdf3c0a5eaad91ccaf2fe91cb/amethyst_renderer/src/tex_animation.rs#L11)):*
> ```Rust
> /// Meta data for a sprite sheet texture.  Does not contain a texture, only a description of how
> /// a texture can be animated.
> #[derive(Clone, Debug)]
> ```

I guess you want to implement `Serialize` and `Deserialize` here.

---

*[amethyst_renderer/src/tex_animation.rs, line 49 at r1](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L-uNIzyQ-j8aCJgfezQ:-L-uNIzyQ-j8aCJgfezR:b-l8c1vn) ([raw file](https://github.com/amethyst/amethyst/blob/bfbf0706c91aeeefdf3c0a5eaad91ccaf2fe91cb/amethyst_renderer/src/tex_animation.rs#L49)):*
> ```Rust
>     /// How long the current frame has been played for.
>     frame_timer: f32,
> }
> ```

You also need to specify what to do when animation is finished. Loop, switch to another, stop.

---

*[amethyst_renderer/src/tex_animation.rs, line 63 at r1](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L-uLaypMOsIg_Sek7EU:-L-uLaypMOsIg_Sek7EV:b-p2ta64) ([raw file](https://github.com/amethyst/amethyst/blob/bfbf0706c91aeeefdf3c0a5eaad91ccaf2fe91cb/amethyst_renderer/src/tex_animation.rs#L63)):*
> ```Rust
>     pub height: f32,
>     /// The duration this frame should be played for in seconds.
>     pub duration: f32,
> ```

We should prefer to use `Duration` for durations.
Maybe we want to use our own `Duration` for efficiency. But `f32` leaves you with a choice how to interpretate it.

---

*[amethyst_renderer/src/mtl/mod.rs, line 13 at r1](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L-uKyq7WegD5O7KqYXA:-L-uKyq7WegD5O7KqYXB:b-5v46p1) ([raw file](https://github.com/amethyst/amethyst/blob/bfbf0706c91aeeefdf3c0a5eaad91ccaf2fe91cb/amethyst_renderer/src/mtl/mod.rs#L13)):*
> ```Rust
>     /// Diffuse map.
>     pub albedo: TextureHandle,
>     /// Optional animation for albedo
> ```

IMO animation isn't part of the material. I'd put it to the separate component.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/amethyst/amethyst/505)*

Xaeroxe:
*[amethyst_renderer/src/tex_animation.rs, line 63 at r1](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L-uLaypMOsIg_Sek7EU:-L-wCqT12mqdGlpN3Dkv:bb14ho8) ([raw file](https://github.com/amethyst/amethyst/blob/bfbf0706c91aeeefdf3c0a5eaad91ccaf2fe91cb/amethyst_renderer/src/tex_animation.rs#L63)):*
<details><summary><i>Previously, omni-viral (Zakarum) wrote…</i></summary><blockquote>

We should prefer to use `Duration` for durations.
Maybe we want to use our own `Duration` for efficiency. But `f32` leaves you with a choice how to interpretate it.
</blockquote></details>

Not sure I agree with you here.  The documentation states this is in seconds so there is no ambiguity in the correct way to interpret it.  Using an f32 allows for more efficient calculations later on.  We could create our own Duration type holding just an f32 but I think that would be silly, because variable types don't need to contain semantic information, that's what the variable name is for.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/amethyst/amethyst/505)*

omni-viral:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[amethyst_renderer/src/tex_animation.rs, line 63 at r1](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L-uLaypMOsIg_Sek7EU:-L0-UGfd5CshK6JkiAX7:b-cqk8at) ([raw file](https://github.com/amethyst/amethyst/blob/bfbf0706c91aeeefdf3c0a5eaad91ccaf2fe91cb/amethyst_renderer/src/tex_animation.rs#L63)):*
<details><summary><i>Previously, Xaeroxe (Jacob Kiesel) wrote…</i></summary><blockquote>

Not sure I agree with you here.  The documentation states this is in seconds so there is no ambiguity in the correct way to interpret it.  Using an f32 allows for more efficient calculations later on.  We could create our own Duration type holding just an f32 but I think that would be silly, because variable types don't need to contain semantic information, that's what the variable name is for.
</blockquote></details>

Creating "newtypes" with single type inside to restrict usage and define semantics is what many people do, is what "newtypes" were created for.
Also I prefer u64 for time measurement

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/amethyst/amethyst/505)*

Xaeroxe:
Review status: 0 of 36 files reviewed at latest revision, 4 unresolved discussions.

---

*[amethyst_renderer/src/tex_animation.rs, line 11 at r1](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L-uMqqH6tZN3ymHLwmk:-L0_mSOlbXqcJxL2b_iD:bi8b2vn) ([raw file](https://github.com/amethyst/amethyst/blob/bfbf0706c91aeeefdf3c0a5eaad91ccaf2fe91cb/amethyst_renderer/src/tex_animation.rs#L11)):*
<details><summary><i>Previously, omni-viral (Zakarum) wrote…</i></summary><blockquote>

I guess you want to implement `Serialize` and `Deserialize` here.
</blockquote></details>

I don't think that's necessary if this is an Asset right?

---

*[amethyst_renderer/src/tex_animation.rs, line 49 at r1](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L-uNIzyQ-j8aCJgfezQ:-L0_m_fS5wGNYoAuCZeA:btwq8tt) ([raw file](https://github.com/amethyst/amethyst/blob/bfbf0706c91aeeefdf3c0a5eaad91ccaf2fe91cb/amethyst_renderer/src/tex_animation.rs#L49)):*
<details><summary><i>Previously, omni-viral (Zakarum) wrote…</i></summary><blockquote>

You also need to specify what to do when animation is finished. Loop, switch to another, stop.
</blockquote></details>

First implementation is to just loop.  I'm also planning on sending out events when an animation finishes via shrev so that the end user can implement their own logic for this.

---

*[amethyst_renderer/src/tex_animation.rs, line 63 at r1](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L-uLaypMOsIg_Sek7EU:-L0_mplSh84AbPPXHlfA:blpz7ij) ([raw file](https://github.com/amethyst/amethyst/blob/bfbf0706c91aeeefdf3c0a5eaad91ccaf2fe91cb/amethyst_renderer/src/tex_animation.rs#L63)):*
<details><summary><i>Previously, omni-viral (Zakarum) wrote…</i></summary><blockquote>

Creating "newtypes" with single type inside to restrict usage and define semantics is what many people do, is what "newtypes" were created for.
Also I prefer u64 for time measurement
</blockquote></details>

This now uses a `std::time::Duration`.  After thinking about it I determined the f32 doesn't actually help that much with efficiency.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/amethyst/amethyst/505)*

Xaeroxe:
*[amethyst_renderer/src/mtl/mod.rs, line 13 at r1](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L-uKyq7WegD5O7KqYXA:-L0_n1fdPZV2mI8cZwkf:b-896fix) ([raw file](https://github.com/amethyst/amethyst/blob/bfbf0706c91aeeefdf3c0a5eaad91ccaf2fe91cb/amethyst_renderer/src/mtl/mod.rs#L13)):*
<details><summary><i>Previously, omni-viral (Zakarum) wrote…</i></summary><blockquote>

IMO animation isn't part of the material. I'd put it to the separate component.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/amethyst/amethyst/505)*

torkleyy:
Your work looks great, but one problem I see is that you created a lot of formatting changes. Given that we have @omni-viral writing a renderer rewrite, I don't think pushing formatting changes is a good idea currently.

---

Reviewed 1 of 47 files at r2.
Review status: 1 of 47 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[rustfmt.toml, line 6 at r2](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L0zFRMRigjWUFB8dqnU:-L0zFRMRigjWUFB8dqnV:btiskop) ([raw file](https://github.com/amethyst/amethyst/blob/2a1a1e60472991db6c14acc4960030ad896b8abc/rustfmt.toml#L6)):*
> ```TOML
> reorder_imports_in_group = true
> attributes_on_same_line_as_variant = false
> attributes_on_same_line_as_field = false
> ```

Why did you remove these?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/amethyst/amethyst/505#-:-L0zFc1oT34zzSJGlVZV:b-4w2ei1)*

Xaeroxe:
Alright, I'll go ahead and remove the formatting changes.

---

Review status: 1 of 47 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[rustfmt.toml, line 6 at r2](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L0zFRMRigjWUFB8dqnU:-L0zGdRVauIaPhXbH8Gq:b-dyxvke) ([raw file](https://github.com/amethyst/amethyst/blob/2a1a1e60472991db6c14acc4960030ad896b8abc/rustfmt.toml#L6)):*

Why did you remove these?
</blockquote></details>

Latest version of rustc didn't recognize them as valid options.  Some of them were renamed, merged into other options, or just removed because they never worked in the first place.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/amethyst/amethyst/505#-:-L0zG_gotgMXCQscTqV7:b-wb1hpj)*

torkleyy:
Phew, this was a lot to review. I really like it, thank you @Xaeroxe!

---

Reviewed 46 of 47 files at r2.
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[amethyst_renderer/src/tex_animation.rs, line 18 at r2](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L1EM0ltfrflchcWNzSo:-L1EM0ltfrflchcWNzSp:b6knxax) ([raw file](https://github.com/amethyst/amethyst/blob/3250ba541ef657a6c02e64c603086b3d2d0c764b/amethyst_renderer/src/tex_animation.rs#L18)):*
> ```Rust
>     /// A collection of animations, the first layer contains a list of "animations" and the second
>     /// layer contains a list of frame indices within the animation.
>     pub animations: Vec<Vec<usize>>,
> ```

Please don't nest `Vec`s like this. I'd suggest you change this to `Vec<usize>` and add another `Vec` which just stores the number of frames for a given animation index.

---

*[amethyst_renderer/src/tex_animation.rs, line 19 at r2](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L1EL5ksyDSbA-b4tXHA:-L1EL5ksyDSbA-b4tXHB:b-yji61t) ([raw file](https://github.com/amethyst/amethyst/blob/3250ba541ef657a6c02e64c603086b3d2d0c764b/amethyst_renderer/src/tex_animation.rs#L19)):*
> ```Rust
>     /// layer contains a list of frame indices within the animation.
>     pub animations: Vec<Vec<usize>>,
>     /// A mapping between string names and indexes into the first layer of the animations member.
> ```

"indexes"

---

*[amethyst_renderer/src/tex_animation.rs, line 66 at r2](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L1EMTsjKAHF7gcYxqBg:-L1EMTsjKAHF7gcYxqBh:bjsdt95) ([raw file](https://github.com/amethyst/amethyst/blob/3250ba541ef657a6c02e64c603086b3d2d0c764b/amethyst_renderer/src/tex_animation.rs#L66)):*
> ```Rust
>     /// ## Returns
>     /// If the `starting_animation` wasn't found then this will return `None`.  This could be due
>     /// to one of a few things:
> ```

Then it should better use a `Result`.

---

*[amethyst_renderer/src/tex_animation.rs, line 100 at r2](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L1EMa3vISm0UPCxaXjD:-L1EMa3vISm0UPCxaXjE:bp10v3o) ([raw file](https://github.com/amethyst/amethyst/blob/3250ba541ef657a6c02e64c603086b3d2d0c764b/amethyst_renderer/src/tex_animation.rs#L100)):*
> ```Rust
>         sheet_storage: &AssetStorage<SpriteSheetData>,
>         animation: &str,
>     ) -> Result<(), ()> {
> ```

Just a remainder, this needs an error type (I assume you're aware of this since this is a WIP).

---

*[amethyst_renderer/src/tex_animation.rs, line 104 at r2](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L1EN0i42ExnFtMVQa39:-L1EN0i42ExnFtMVQa3A:b-6vsx80) ([raw file](https://github.com/amethyst/amethyst/blob/3250ba541ef657a6c02e64c603086b3d2d0c764b/amethyst_renderer/src/tex_animation.rs#L104)):*
> ```Rust
>             .get(&self.sprite_sheet_data)
>             .and_then(|data| data.animation_mapping.get(animation).map(|i| *i));
>         match animation {
> ```

You can match against `<above expression>.ok_or(())?`

---

*[amethyst_renderer/src/mtl/mod.rs, line 34 at r2](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L1EKq9aRFYTqMY9Sigk:-L1EKq9aRFYTqMY9Sigl:b-1dhyi8) ([raw file](https://github.com/amethyst/amethyst/blob/3250ba541ef657a6c02e64c603086b3d2d0c764b/amethyst_renderer/src/mtl/mod.rs#L34)):*
> ```Rust
> /// create a `Material` that has the spritesheets attached in the textures, then provide how
> /// those spritesheets are to be animated from this structure.
> pub struct MaterialAnimation {
> ```

Please derive `Default` for this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/amethyst/amethyst/505#-:-L1ENO-d-andwBXyaIZ7:b-lrv33l)*

Rhuagh:
Reviewed 43 of 47 files at r2.
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/amethyst/amethyst/505)*

Xaeroxe:
Review status: 41 of 46 files reviewed at latest revision, 10 unresolved discussions.

---

*[amethyst_renderer/src/tex_animation.rs, line 18 at r2](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L1EM0ltfrflchcWNzSo:-L22YjQPTTzB8-WQJUss:b-omeilc) ([raw file](https://github.com/amethyst/amethyst/blob/3250ba541ef657a6c02e64c603086b3d2d0c764b/amethyst_renderer/src/tex_animation.rs#L18)):*

Please don't nest `Vec`s like this. I'd suggest you change this to `Vec<usize>` and add another `Vec` which just stores the number of frames for a given animation index.
</blockquote></details>

That makes the assumption the frames are always sequentially laid out, which while a lot of the time that's true, is not guaranteed, particularly if animations are sharing frames.

---

*[amethyst_renderer/src/tex_animation.rs, line 100 at r2](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L1EMa3vISm0UPCxaXjD:-L22YzBjAcgeYPBRxsI9:b-wq32am) ([raw file](https://github.com/amethyst/amethyst/blob/3250ba541ef657a6c02e64c603086b3d2d0c764b/amethyst_renderer/src/tex_animation.rs#L100)):*

Just a remainder, this needs an error type (I assume you're aware of this since this is a WIP).
</blockquote></details>

I'll make zero sized type for this then

---

*[amethyst_renderer/src/tex_animation.rs, line 104 at r2](https://reviewable.io:443/reviews/amethyst/amethyst/505#-L1EN0i42ExnFtMVQa39:-L22ZBKxMur09YQw-6Hy:bj2e61r) ([raw file](https://github.com/amethyst/amethyst/blob/3250ba541ef657a6c02e64c603086b3d2d0c764b/amethyst_renderer/src/tex_animation.rs#L104)):*

You can match against `<above expression>.ok_or(())?`
</blockquote></details>

Ah yes, thank you!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/amethyst/amethyst/505)*

Rhuagh:
Reviewed 5 of 6 files at r3.
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/amethyst/amethyst/505)*

Xaeroxe:
Bad robot

amethyst_core/src/timing.rs
```diff
@@ -378,3 +378,8 @@ pub fn duration_to_nanos(duration: Duration) -> u64 {
 pub fn nanos_to_duration(nanos: u64) -> Duration {
     Duration::new(nanos / 1_000_000_000, (nanos % 1_000_000_000) as u32)
 }
+
+/// Multiplies a given Duration by the float provided.
+pub fn duration_f32_mul(duration: Duration, float: f32) -> Duration {
+    Duration::new((duration.as_secs() as f32 * float) as u64, (duration.subsec_nanos() as f32 * float) as u32)
```
omni-viral: That's why I don't like `Duration`.
Can we consider to replace it with our own type?

amethyst_renderer/src/tex_animation.rs
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    /// Sets the animation to the named case-sensitive animation from the `SpriteSheetData`.
+    ///
+    /// ## Returns
+    ///
+    /// If the name provided can't be found in the `SpriteSheetData` for this component then this
+    /// will return `Err` otherwise it will return `Ok`.
+    pub fn set_animation(&mut self, sheet_storage: &AssetStorage<SpriteSheetData>, animation: &str) -> Result<(), ()> {
+        let animation = sheet_storage.get(&self.sprite_sheet_data).and_then(|data| data.animation_mapping.get(animation).map(|i| *i));
```
omni-viral: Returned error should be either "Invalid handle" and "Animation not found".

amethyst_renderer/src/tex_animation.rs
```diff
@@ -0,0 +1,178 @@
...
+    /// `SpriteSheetDataHandle` used in initialization or the `sheet_storage` in this function are
+    /// invalid.  Because handles take at least one frame to become valid after being loaded the
+    /// handle is the most likely point of failure.
+    pub fn current_frame(&self, sheet_storage: &AssetStorage<SpriteSheetData>) -> Option<Frame> {
+        sheet_storage.get(&self.sprite_sheet_data).map(|data| data.frames[data.animations[self.current_animation][self.current_frame]].clone())
+    }
+
+    /// Advance the current animation by the time given.  Normally this will be called by the
+    /// engine by default using the regularly running game clock.  This function is made public so
+    /// that you can manually advance the animation if you require.
```
omni-viral: How user can disable automatic animation?
The system is added by bundle.
What if user wants only some entities to be manually animated?

amethyst_renderer/src/tex_animation.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub fn advance(&mut self, time: Duration, sheet_storage: &AssetStorage<SpriteSheetData>) -> Result<(), ()> {
+        use amethyst_core::timing::duration_f32_mul;
+        self.frame_timer += duration_f32_mul(time, self.playback_speed);
```
omni-viral: Shouldn't you wait with this before success one line below?

amethyst_renderer/src/tex_animation.rs
```diff
@@ -0,0 +1,178 @@
...
+        let mut frame_time = self.current_frame(sheet_storage).ok_or(())?.duration;
+        while self.frame_timer >= frame_time {
+            self.current_frame += 1;
+            let max_frame = sheet_storage.get(&self.sprite_sheet_data).map(|data| data.animations[self.current_animation].len() - 1).ok_or(())?;
```
omni-viral: Don't see any chance for failure here. Should be `unwrap`.
Or get `&SpriteSheetData` at the first line and then reuse it.

amethyst_renderer/src/tex_animation.rs
```diff
@@ -0,0 +1,178 @@
...
+            if self.current_frame > max_frame {
+                self.current_frame = 0;
```
omni-viral: So currently animation looping is the only option.
Just writing it so we wont forget to add other options 😄

mtsr:
Objectives (quoting @Xaeroxe on gitter):
- [x]     Create a data structure that can store animation data
- [x]     Make passes display the appropriate section of the SpriteSheet based on this data structure
- [x]     Create a system that can automatically advance through the frames once their stored duration has elapsed.
- [x]     Add the system to the renderer bundle (trivial, one line)
- [ ]     Build passes that can render entities with Transform transparently. The UI pass can render transparently, and all of the other passes render entities with mesh and transform, so we mostly just need to combine these two concepts and make a few tweaks.
- [ ]     @torkleyy has also requested I see what I can do about reducing duplication of pass code. Not quite sure what I want to do about that yet, but since the PR is high enough priority I'm somewhat comfortable putting that off for now.
- [ ]     A data importer for aseprite JSON export was on my priority list. I'm OK omitting this for a first implementation.

amethyst_core/src/timing.rs
jojolepro: You can always use secs f64 if you don't want to use Duration.

Xaeroxe: I'm leaving this function since I've already written it, but I'm not using it in this PR anymore.

amethyst_renderer/src/tex_animation.rs
Xaeroxe: Not built into the current implementation, but I'll give it some thought.

amethyst_renderer/src/tex_animation.rs
```diff
@@ -36,6 +37,43 @@ impl Into<Result<SpriteSheetData, ()>> for SpriteSheetData {
     }
 }
 
+/// Error returned by the `set_animation` function on the `SpriteSheetAnimation`
+#[derive(Copy, Clone, Debug)]
+pub enum SetAnimationError {
```
torkleyy: You can use `error-chain` for this, that would give us a back trace in case of an error.

amethyst_renderer/src/tex_animation.rs
```diff
@@ -36,6 +37,43 @@ impl Into<Result<SpriteSheetData, ()>> for SpriteSheetData {
...
+    /// their first frame of existence.
+    HandleInvalid,
+    /// The named animation was not found in the underlying `SpriteSheetData`.
+    AnimationNotFound,
+}
+
+impl Display for SetAnimationError {
+    fn fmt(&self, f: &mut Formatter) -> Result<(), FmtError> {
+        f.write_str(
+            match self {
```
torkleyy: According to clippy, you should dereference `self`.

amethyst_core/src/timing.rs
omni-viral: @jojolepro how many seconds I can store in `f64` until error become more than epsilon?
I'm strongly against using floats where error may become observable.

Xaeroxe: Relax, an f32 has much higher precision than we need in this circumstance.  You only really need an f64 when storing very large values, which we definitely are not.  Most of these values are going to be below 1.0.

amethyst_renderer/src/tex_animation.rs
```diff
@@ -0,0 +1,251 @@
...
+    /// This should only be used when switching animation by string name.
+    pub animation_mapping: HashMap<String, usize>,
+}
+
+impl Asset for SpriteSheetData {
+    type Data = Self;
+    type HandleStorage = VecStorage<Handle<Self>>;
+}
+
+impl Into<Result<SpriteSheetData, ()>> for SpriteSheetData {
```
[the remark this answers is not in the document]
Xaeroxe: So I realize now I did this wrong, but it was an attempt at making the type compatible with [this](https://docs.rs/amethyst_assets/0.2.0/amethyst_assets/struct.Processor.html#impl-System%3C'a%3E)

amethyst_renderer/src/tex_animation.rs
```diff
@@ -0,0 +1,251 @@
...
+
+    /// Advance the current animation by the time given.  Normally this will be called by the
+    /// engine by default using the regularly running game clock.  This function is made public so
+    /// that you can manually advance the animation if you require.
+    pub fn advance(
+        &mut self,
+        time: FloatDuration,
+        sheet_storage: &AssetStorage<SpriteSheetData>,
+    ) -> Result<(), ()> {
+        let mut frame_time = self.current_frame(sheet_storage).ok_or(())?.duration;
```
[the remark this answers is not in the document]
Xaeroxe: Yeah it does, I just wanted to return a result here.

Rhuagh:
So, with the recent additions to the core animation system this should be quite easy to add now.
The needed parts goes as follows:

* `SpriteSheetData` can be reduced to just a list of `Frame`
* `Frame` will no longer need the duration, possibly rename to `Sprite` ?
* `MaterialAnimation` can have `Option<usize>` for all attributes, and also needs a "rest pose index" per attribute (so the animation systems know what to use if there are no animations running), possibly rename to something like `MaterialSpriteRef` 
* Define `MaterialChannel` that contains all the attributes that should be possible to animate on `MaterialAnimation`
* Implement `AnimationSampling` for `MaterialAnimation`, primitive will be `usize`
* The vertex shader updates need to stay
* The passes should just do sprite sheet data lookup using the indices from `MaterialAnimation` and hand over to the shaders

Xaeroxe:
@Rhuagh Those names sound fine to me.  I was planning on doing this myself if you have something else you'd rather be doing.

Rhuagh:
I've already started, I'll push the branch tonight and you can take it from there

Xaeroxe:
Oh thanks!

Xaeroxe:
This has been largely eclipsed by work from @Rhuagh so this PR is going to be closed as we work on bringing his work into the repo.

Issue #522, opened 2018-01-02
Document ScreenDimensions better


Xaeroxe:
This resource is confusing people, so I'm creating this issue as a reminder

OvermindDL1:
Especially document what happens if it is not an integral number stored in it, like say `984.286` or something.  Though in my opinion a type should *only* represent valid values, and if something like the prior is an invalid value then the type should be corrected to prevent that.

Xaeroxe:
The interface to `ScreenDimensions` prevents invalid values being stored, it specifically requests integers.  However it does also return floats because most of the time when these are needed they're going to be converted to float values anyways.

PR #524, opened 2018-01-05
Improve documentation for ScreenDimensions

Xaeroxe:
This pull request also additionally makes it more obvious that the first frame screen size is inaccurate.

Fixes https://github.com/amethyst/amethyst/issues/522

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/524)

### DIFF
--- a/amethyst_renderer/src/resources.rs
+++ b/amethyst_renderer/src/resources.rs
@@ -68,11 +68,17 @@ impl ScreenDimensions {
     }
 
     /// Returns the current width of the window.
+    ///
+    /// This is returned as a float for user convenience, as this is typically used with other
+    /// float values.  This will only ever be a positive integer though.
     pub fn width(&self) -> f32 {
         self.w
     }
 
     /// Returns the current height of the window.
+    ///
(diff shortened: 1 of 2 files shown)


amethyst_renderer/src/resources.rs
```diff
@@ -68,11 +68,17 @@ impl ScreenDimensions {
     }
 
     /// Returns the current width of the window.
+    ///
+    /// This is returned as a float for user convenience, as this is typically used with other
+    /// float values.  This will only ever be a positive integer though.
```
torkleyy: That ain't true, you're setting it to 0,0 at the beginning.

Xaeroxe: Ah, good catch, thanks.

torkleyy:
I don't see how this PR makes that more obvious, but it can be merged as is.

Xaeroxe:
100, 100 could actually be a valid screen size, 0, 0 is obviously incorrect.  It's a small change but it's aimed at reducing user confusion regarding this resource.

torkleyy:
Ah, I see, thanks for clarifying.

Issue #541, opened 2018-01-13
The future of the `Bundle` API


torkleyy:
I think the main problem of the current API is that it is ugly to access the world while building the `Application`. Also, we should be able to add `RunNow` stuff to `DispatcherBuilder`.

cc @Xaeroxe @Rhuagh

Xaeroxe:
I'm going to call this resolved.  @torkleyy re-open if you disagree with me.

torkleyy:
I think we reached like the near time goal, so yeah closing is okay. In the future we might need more flexibility e.g. allow to add resources but not systems.

Rhuagh:
I have ideas, we can discuss at some point.

Issue #536, opened 2018-01-12
UI Example Failures to Run on Mac OS X


Hittherhod:
## Metadata
Initializing Amethyst...
Version: v0.5.1-421-ge8829ba
Platform: x86_64-apple-darwin
Amethyst git commit: e8829ba76fd29eb804faeaaa44dbe3c341bc2c39
Rustc version: 1.24.0-nightly Nightly
Rustc git commit: 687d3d15ba726dbb1ac6b85223ebe0e98c6820cc

## Additional Information
Dimensions 0 0
I added a print in the `initialise_camera()` function to print out the height and width, and it is returning 0. I see in the `amethyst_renderer/src/bundle.rs` file that the ScreenDimensions are initially set to 0. The pong example shows the same behavior (ScreenDimensions are 0), but it doesn't use the ScreenDimensions are the input to the camera so it's not a big deal.

I see that `amethyst_renderer/src/systems.rs` updates the screen dimensions, but it doesn't look like it has been called by the time we initialize the camera.

Xaeroxe:
Thanks for filing an issue!  I've known about this I just hadn't made an issue for it yet.  Now I have something to put in a PR that fixes this!

Our fix to this is likely going to involve restructuring the Amethyst initialization process, as it has a difficult time communicating with the Specs world during initialization.  That causes compromises such as this.

PR #542, opened 2018-01-14
Add `world` function to AppBuilder

Xaeroxe:
Fixes https://github.com/amethyst/amethyst/issues/541
Fixes https://github.com/amethyst/amethyst/issues/536

Adds a function called `world` to the ApplicationBuilder which permits access to the world during the procession of the builder.

Additionally adds the world to the arguments for `RenderSystem::build` so that the ScreenDimensions can be initialized properly.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/542)


examples/material/main.rs
```diff
@@ -131,9 +131,13 @@ fn run() -> Result<(), amethyst::Error> {
             .clear_target([0.0, 0.0, 0.0, 1.0], 1.0)
             .with_pass(DrawPbm::<PosNormTangTex>::new()),
     );
+    let mut renderer = None;
```
torkleyy: Sorry, but this code doesn't look convenient.

torkleyy:
I'm just thinking, you will always have a renderer added to the Application (except for the server, but we can address that later) so wouldn't it make sense to just add a convenient `with_renderer` method?

book/src/pong_tutorial/pong_tutorial_02.md
```diff
@@ -347,35 +347,43 @@ fn on_start(&mut self, world: &mut World) {
...
 ```rust,ignore
-let mut game = Application::build("./", Pong)?
-    .with_bundle(TransformBundle::new())? //Add this bundle
+let mut renderer = None;
+let game = Application::build("./", Pong)?
+    .with_bundle(TransformBundle::new())?
     .with_bundle(RenderBundle::new())?
-    .with_local(RenderSystem::build(pipe, Some(config))?)
-    .build()?;
+    .world(|world| {
```
omni-viral: You can add `RenderSystem` directly to the world inside the closure.

torkleyy:
Didn't know we need a `World` for the pipeline building, but if that's the case it could be just like `world` except for the closure's return type (that would be the render system or the pipeline).

book/src/pong_tutorial/pong_tutorial_02.md
Xaeroxe: Except you don't add systems to the world.  You add it to the `DispatcherBuilder`.

torkleyy: Add a system to the world?

omni-viral: Ah. Yeah. Forget about it 😜

omni-viral: Can you make it so the closure would return `RenderSystem` and the it would be added to the `Application`?

torkleyy: My suggestion, see below :)

Xaeroxe:
Makes sense, I'll work on implementing `with_renderer` soon.

I still kind of like the `world` function as it's a nice way to access the `world` during initialization if you need to for some reason or another, but it's not strictly required.

torkleyy:
Hold on, if the dispatcher accepts `RunNow` we can just add the render system to the bundle.

torkleyy:
If you like the `World` function you could replace it with a more general solution: a `with` method that gives you a reference to the `Application`. The world field is public anyway.

Xaeroxe:
@torkleyy That's actually not doable, as it would have to be the `ApplicationBuilder`.

On second thought the method doesn't add much, and with the changes to shred that have just been published I don't really care for any aspect of my solution here, so I'm going to close this and open a new one.

PR #543, opened 2018-01-15
fix: Only skip invalid meshes, instead of aborting the pass completely.

Rhuagh:
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/543)

### DIFF
--- a/amethyst_renderer/src/pass/pbm/separate.rs
+++ b/amethyst_renderer/src/pass/pbm/separate.rs
@@ -174,7 +174,7 @@ impl Pass for DrawPbmSeparate {
                 .map(|&(_, ref trans)| [trans.0[3][0], trans.0[3][1], trans.0[3][2]])
                 .unwrap_or([0.0; 3]),
         );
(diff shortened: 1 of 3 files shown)


amethyst_renderer/src/pass/pbm/separate.rs
```diff
@@ -188,7 +188,7 @@ impl Pass for DrawPbmSeparate {
             {
                 match mesh.buffer(attrs) {
                     Some(vbuf) => effect.data.vertex_bufs.push(vbuf.clone()),
-                    None => return,
+                    None => continue 'drawable,
```
omni-viral: `effect.data.vertex_bufs`  must be cleared.

Rhuagh:
Fixed.

omni-viral:
@torkleyy I wouldn't spend much time to fix minor problems in old renderer.

Rhuagh:
Yeah, there's some code duplication, but no good way of removing that without introducing about the same amount of duplication just trying to send data to functions.

Xaeroxe:
Overriding

Issue #555, opened 2018-01-31
Limit maximum number of playing sound effects


Xaeroxe:
Currently there are no limitations on the amount of sfx being played, this could be very problematic if a scenario arises that causes a sound effect to play far too frequently.  We should limit how many sfx can play at a time.

PR #556, opened 2018-01-31
Improve docs and add max sound limit

Xaeroxe:
Also removes play.rs from the repo as it was empty anyways.

Fixes https://github.com/amethyst/amethyst/issues/555

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/556)

### DIFF
--- a/amethyst_audio/src/output.rs
+++ b/amethyst_audio/src/output.rs
@@ -3,19 +3,36 @@
 // We have to use types from this to provide an output iterator type.
 use std::fmt::{Debug, Formatter, Result as FmtResult};
 use std::io::Cursor;
+use std::sync::atomic::{AtomicIsize, Ordering, ATOMIC_ISIZE_INIT};
 
 use cpal::{default_endpoint, endpoints};
 use cpal::EndpointsIterator;
 use rodio::{Decoder, Endpoint, Sink, Source as RSource};
 
 use DecoderError;
+use end_signal::EndSignalSource;
 use source::Source;
+
+// These are isize values because due to thread interactions it is possible, however unlikely, that
+// SOUNDS_PLAYING may be temporarily < 0.  If this happens, it should be resolved very quickly by
+// the threads completing their instructions.
+pub(crate) const MAX_SOUNDS_PLAYING: isize = 300;
+pub(crate) static SOUNDS_PLAYING: AtomicIsize = ATOMIC_ISIZE_INIT;
 
 /// A speaker(s) through which audio can be played.
 ///
-/// By convention, the default output is stored as a resouce in the `World`.
+/// By convention, the default output is stored as a resource in the `World`.
+#[derive(Clone)]
 pub struct Output {
     pub(crate) endpoint: Endpoint,
+}
+
+impl Eq for Output {}
+
+impl PartialEq for Output {
+    fn eq(&self, other: &Self) -> bool {
+        self.endpoint == other.endpoint
+    }
 }
 
 impl Output {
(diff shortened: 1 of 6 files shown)


Xaeroxe:
Initial testing shows this working.  It only attempts to play 300 sounds simultaneously. (attempts is used very loosely here, it eventually degrades into occasional clicks from the sheer load placed on the audio engine)

When sounds stop they're also dropped off the main count.

Xaeroxe:
Playing 300 sounds at a time actually works reasonably well in release mode.

As has been typical in the past, release mode rodio turns out to be really hard to break.

amethyst_audio/src/output.rs
```diff
@@ -75,11 +67,17 @@ impl Output {
     ) -> Result<(), DecoderError> {
         let sink = Sink::new(&self.endpoint);
         for _ in 0..n {
-            sink.append(
-                Decoder::new(Cursor::new(source.clone()))
-                    .map_err(|_| DecoderError)?
-                    .amplify(volume),
-            );
+            if SOUNDS_PLAYING.load(Ordering::Relaxed) < MAX_SOUNDS_PLAYING {
```
torkleyy: `Relaxed` is not enough here.

Xaeroxe:
The biggest reason to prevent this is to keep the audio engine from falling horribly behind and never recovering until the game is restarted.  This puts a cap on how far behind the audio engine can fall.  So at least if the extreme noise maker stops the game will eventually recover normal sound.

amethyst_audio/src/output.rs
Xaeroxe: So if I'm being completely honest I'm a little fuzzy on which ordering you should be using and for what reason you should be using it.  Relaxed seems fine here though, as it's not the end of the world if we're playing 302 sounds instead of 300.

amethyst_audio/src/systems/audio.rs
```diff
@@ -120,15 +80,22 @@ impl<'a> System<'a> for AudioSystem {
...
-                        let sink = SpatialSink::new(
-                            &listener.output.endpoint,
-                            emitter_position,
-                            left_ear_position,
-                            right_ear_position,
-                        );
-                        let atomic_bool = Arc::new(AtomicBool::new(false));
-                        sink.append(EndSignalSource::new(source, atomic_bool.clone()));
-                        audio_emitter.sinks.push((sink, atomic_bool));
+                        if SOUNDS_PLAYING.load(Ordering::Relaxed) < MAX_SOUNDS_PLAYING {
```
jojolepro: Ordering::Relaxed
You also use it in output.rs.
Wouldn't it be better to have it configurable with a way to ensure that they are both the same value, in case someone changes the Ordering in only one place?

amethyst_audio/src/output.rs
```diff
@@ -75,11 +67,17 @@ impl Output {
     ) -> Result<(), DecoderError> {
```
jojolepro: Shouldn't this be 3 functions instead of 2?
play // actual play logic. takes the stuff you might reuse as parameter
play_once // obvious
play_multiple // calls play n times

Its weird design wise to have play_n_times(1) and it also makes you create a int for nothing (for index).

Also, play_multiple VS play_n_times naming.

torkleyy:
If it falls behind, can't we just let it skip buffers? That strategy would be more flexible wrt setup & performance.

Xaeroxe:
Can you elaborate on "skip buffers"?  Do you mean "stop the playing sound"?  If you do, then that approach would require more setup than this one, and be worse performance wise, as you'd have to check the SOUNDS_PLAYING value for every sample, which is about 700x more frequent than what I'm doing right now.  It's easier to stop a sound playing in the first place than it is to stop a sound that's in progress.

I feel like flexibility isn't that much of a concern here, as I've already given plenty.  If your game has more than 300 sfx playing simultaneously you've got some serious sound design issues that need to be addressed.  Some engines stop you at 255 sounds.

torkleyy:
What exactly will happen if we play too many sounds? I'm not 100% sure I understand the problem here.

jojolepro:
You'll slow down the audio thread until it sounds like a pile of trash, and new sounds will always be playing late until the thread catches up.

Xaeroxe:
It's like filling a sink faster than the sink can be drained.

torkleyy:
But in that case wouldn't this be a bigger problem on slower computers? So the limit may be too high in some cases..

Xaeroxe:
I agree, the intent isn't to stop people from shooting themselves in the foot with this, the intent is more to allow them to recover once they do.  It's to keep the pain from this from getting too bad.

That being said, I could try and make it more dynamic so it adjusts to the device being played on.

Xaeroxe:
Relevant gitter conversation: https://gitter.im/amethyst/general?at=5a7499cfb3c4a0d376d1c05d

Pinging @OvermindDL1 can you weigh in here since you were the one who originally argued in favor of having a limit on SFX?

Xaeroxe:
Content aside from rate limiting in this PR isn't being questioned, so the rate limiting has been removed for now and I'm going to go ahead and merge this as it doesn't currently change anything of substance.

bors r+

jojolepro:
@Xaeroxe I feel like you ignored my review?

Xaeroxe:
Your review was regarding existing structure and naming, so I left it as is because I just wanted to get my cleanup merged.  I'll respond to it now.

jojolepro:
Technically my review was more cleanup to do :P

amethyst_audio/src/output.rs
Xaeroxe: That structure doesn't work here because just calling play multiple times won't get the desired result.  play_n_times doesn't play all the sounds at once, it queues them into a sink which will then play them sequentially.  I don't understand what your distinction would be between play and play_once though.

ints on the stack are super duper cheap, so calling it with 1 as the parameter is just a way to reduce code duplication, mainly so that changes done in one place aren't reflected in another.

Xaeroxe: Sorry disregard the first paragraph there, it just clicked what you mean.

Xaeroxe: I'm not quite sure if I favor that approach because it requires me to add another function and interface for little gain imo.

Xaeroxe:
@jojolepro The empty file was deleted by this PR

amethyst_audio/src/output.rs
jojolepro: it'd be private

PR #565, opened 2018-02-09
[WIP] Try to get rendering working on Android

torkleyy:
I fixed all the compile errors and crashes.. and in return Android now gives me a black screen. I'm not quite sure how to debug it :(

cc @ozkriff Maybe you have an idea since you're building your games for Android?

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/565)

### DIFF
--- a/amethyst_renderer/src/pass/shaders/vertex/basic.glsl
+++ b/amethyst_renderer/src/pass/shaders/vertex/basic.glsl
@@ -1,6 +1,4 @@
-// TODO: Needs documentation.
-
-#version 150 core
+#version 300 es
 
 layout (std140) uniform VertexArgs {
     uniform mat4 proj;
@@ -13,17 +11,15 @@ in vec3 normal;
 in vec3 tangent;
 in vec2 tex_coord;
 
-out VertexData {
-    vec4 position;
-    vec3 normal;
-    vec3 tangent;
-    vec2 tex_coord;
-} vertex;
+out vec4 vertex_position;
+out vec3 vertex_normal;
+out vec3 vertex_tangent;
+out vec2 vertex_tex_coord;
 
 void main() {
-    vertex.position = model * vec4(position, 1.0);
-    vertex.normal = mat3(model) * normal;
-    vertex.tangent = mat3(model) * tangent;
-    vertex.tex_coord = tex_coord;
-    gl_Position = proj * view * vertex.position;
+    vertex_position = model * vec4(vertex_position.xyz, 1.0);
+    vertex_normal = mat3(model) * vertex_normal;
+    vertex_tangent = mat3(model) * vertex_tangent;
+    vertex_tex_coord = vertex_tex_coord;
+    gl_Position = proj * view * vertex_position;
 }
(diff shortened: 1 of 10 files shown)


anderejd:
I've had some fun with black screens on android. My solution was spamming my code with glGetError for every single opengl call. Should those calls have been there from the start? Possibly. There are most likely better ways to deal with that though :)
Have you tried using the android emulator?
Did a quick googling and found this: https://github.com/google/gapid

Xaeroxe:
Impressive!!  I didn't know we were even this close.  I'll look into helping out here once I finish the specs PR and my animation PR.

amethyst_renderer/src/renderer.rs
```diff
@@ -275,13 +275,13 @@ fn init_backend(wb: WindowBuilder, el: &EventsLoop, config: &DisplayConfig) -> R
...
 fn init_backend(wb: WindowBuilder, el: &EventsLoop, config: &DisplayConfig) -> Result<Backend> {
     use gfx_window_glutin as win;
-    use glutin::{self, GlProfile, GlRequest};
+    use glutin::{self, Api, GlProfile, GlRequest};
 
     let ctx = glutin::ContextBuilder::new()
         .with_multisampling(config.multisampling)
         .with_vsync(config.vsync)
         .with_gl_profile(GlProfile::Core)
-        .with_gl(GlRequest::Latest);
```
Rhuagh: This should be a separate backend imho.

ametrine/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
```
Rhuagh: This looks like an example, maybe put it under the examples instead of a free standing crate?

src/app.rs
```diff
@@ -315,7 +315,7 @@ impl<'a, 'b, T> ApplicationBuilder<'a, 'b, T> {
                     message
                 ))
             })
-            .level(LevelFilter::Debug)
+            .level(LevelFilter::Warn)
```
Rhuagh: We should make this controllable from the commandline (you can ignore it in this PR though).

amethyst_renderer/src/pass/shaders/vertex/basic.glsl
```diff
@@ -1,6 +1,4 @@
-// TODO: Needs documentation.
-
-#version 150 core
+#version 300 es
```
Rhuagh: Will this still work on a standard desktop?

torkleyy:
These changes are not intended to be merged like this, I was just editing the code until it worked :P

Rhuagh:
Ok :)

ametrine/Cargo.toml
torkleyy: I'm not sure if it can be an example, depends on whether `cargo-apk` supports that.

torkleyy:
Okay, I've tried to debug the whole thing, but didn't find something yet. Here's the gfxtrace if somebody wants to inspect it. I'm currently only getting some black lines :(

[ametrine_20180210_1107.gfxtrace.zip](https://github.com/amethyst/amethyst/files/1713249/ametrine_20180210_1107.gfxtrace.zip)

torkleyy:
This will be done for hal instead.

PR #579, opened 2018-02-20
Update to rayon 1.0

cuviper:
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/579)


torkleyy:
The only concern I have is that Specs uses `ParallelIterator` from an older version of Rayon, thus we will link to two versions of Rayon now. This  may work with `rayon_core`, but I'm not sure if we shouldn't wait on Specs 0.11.

Xaeroxe:
I think we should wait for specs 0.11.

cuviper:
Would you like a rayon-1.0 PR on specs too?  Or is that waiting on other factors?

Xaeroxe:
@cuviper It's actually already happened, we're just waiting on other factors to get a specs 0.11 release out.

I think I've been the one holding that up mostly because I haven't finished an open PR on there.  Woops.  I'll work on that soon.

cuviper:
Oh, is that somewhere besides master?  I still see 0.8.2:
https://github.com/slide-rs/specs/blob/master/Cargo.toml#L30

Xaeroxe:
Oh hm, we updated it in shred, which is one of the dependencies for this.  I didn't realize specs had an explicit rayon dependency.

Then yeah we could use a PR to specs too.

torkleyy:
It's already merged IIRC, so I guess I forgot that.

jojolepro:
If it doesn't cause any issues currently, why just not include it? Code seems to run just fine.

Xaeroxe:
@jojolepro The current state of the code might result in a confusing API.  That's the only reason to not merge this right now.

cuviper:
See slide-rs/specs#352, but shred and hibitset need new versions published.

amethyst_renderer/Cargo.toml
```diff
@@ -36,7 +36,7 @@ gfx_core = { version = "0.8", features = ["serialize"] }
 gfx_macros = "0.2"
 imagefmt = "4.0"
 log = "0.4"
-rayon = "0.8"
+rayon = "1.0"
```
omni-viral: Need to bump version of `hetseq` to `0.2` here so it also depend on `rayon` `1.0`.

cuviper: I can, but note the `hetseq/rayon` feature is not enabled anyway.

omni-viral: Indeed 😄 Then you may skip this request

Xaeroxe:
@torkleyy Remember to merge this once we get specs up to date.

torkleyy:
I think this was implemented in the Specs update, right?

Rhuagh:
Yes

PR #595, opened 2018-03-06
refactor: Make passes work on Visibility rather than Transparent directly

Rhuagh:


amethyst_renderer/src/pass/flat/interleaved.rs
```diff
@@ -99,48 +98,69 @@ where
...
-                Some(global),
-                &[V::QUERIED_ATTRIBUTES],
-                &TEXTURES,
-            );
-        }
-
-        for entity in &back_to_front.entities {
-            if let Some(mesh) = mesh.get(*entity) {
+        match visibility {
+            None => for (mesh, material, global) in (&mesh, &material, &global).join() {
```
jojolepro: Format difference with line 126.

amethyst_renderer/src/pass/flat/separate.rs
```diff
@@ -122,51 +121,77 @@ impl Pass for DrawFlatSeparate {
...
-                Some(global),
-                &ATTRIBUTES,
-                &TEXTURES,
-            );
-        }
-
-        for entity in &back_to_front.entities {
-            if let Some(mesh) = mesh.get(*entity) {
+        match visibility {
+            None => for (entity, mesh, material, global) in
```
jojolepro: format difference with line 152

amethyst_renderer/src/pass/pbm/interleaved.rs
```diff
@@ -103,51 +102,72 @@ where
...
-                Some(global),
-                &[V::QUERIED_ATTRIBUTES],
-                &TEXTURES,
-            );
-        }
-
-        for entity in &back_to_front.entities {
-            if let Some(mesh) = mesh.get(*entity) {
+        match visibility {
+            None => for (mesh, material, global) in (&mesh, &material, &global).join() {
```
jojolepro: that line is formatted wrong in all passes basically

amethyst_renderer/src/visibility.rs
```diff
@@ -0,0 +1,109 @@
...
+                .join()
+                .map(|(entity, global)| (entity, global.0.transform_point(origin)))
+                .map(|(entity, centroid)| Internals {
+                    entity,
+                    transparent: transparent.get(entity).is_some(),
+                    centroid,
+                    camera_distance: centroid.distance2(camera_centroid),
+                    from_camera: centroid - camera_centroid,
+                })
+                .filter(|c| c.from_camera.dot(camera_forward) > 0.), // filter entities behind the camera
```
jojolepro: Doesn't take into account the camera FOV.
If its > 180, you can see behind the camera.
If its < 180, you might not be culling things that could be culled.

amethyst_renderer/src/pass/pbm/interleaved.rs
Rhuagh: Nope, that's what rustfmt wants it to be, since it's a single block.

amethyst_renderer/src/visibility.rs
Rhuagh: This is known, to handle all those cases requires a much larger amount of math, which basically requires bounding volumes etc, which I don't want to add here. This is a system for fast prototyping, and will with 99% likelyhood not be used in larger games.

Rhuagh: Note that the `Visibility` structure can handle those since it has no concept of all that, it's up to the culling system to do all those things. Note that amethyst-rhusics is likely to have a better implementation for this, since there we can do real frustum culling based on the bounding volumes.

amethyst_renderer/src/pass/pbm/interleaved.rs
```diff
@@ -103,51 +102,72 @@ where
...
+                for (mesh, material, global, _) in
+                    (&mesh, &material, &global, &visibility.visible_unordered).join()
+                {
+                    draw_mesh(
+                        encoder,
+                        effect,
+                        false,
+                        mesh_storage.get(mesh),
+                        None,
+                        &*tex_storage,
```
torkleyy: Are these derefs becessary? I don't think so.

Rhuagh: I couldn't find a different way without sending Fetch etc

Rhuagh:
Bors r=torkleyy

PR #599, opened 2018-03-08
[WIP] refactor: Breaking game data out from Application

Rhuagh:
Putting this out there to get feedback because it's a pretty controversial change. 

Only the core stuff is changed, none of the examples work yet.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/599)

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -13,9 +13,9 @@ use shred::Resource;
 use shrev::{EventChannel, ReaderId};
 #[cfg(feature = "profiler")]
 use thread_profiler::{register_thread_with_profiler, write_profile};
-use winit::{Event, WindowEvent};
+use winit::Event;
 
-use assets::{Asset, Loader, Source};
+use assets::{Loader, Source};
 use core::frame_limiter::{FrameLimiter, FrameRateLimitConfig, FrameRateLimitStrategy};
 use core::timing::{Stopwatch, Time};
 use ecs::{Component, Dispatcher, DispatcherBuilder, System, World};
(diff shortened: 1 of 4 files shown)


torkleyy:
So how is this going to work? We'll have to provide every functionality outside of the ECS, too? I thought Application was meant to be a simple wrapper, not the most customizable one.

Rhuagh:
No, I believe we should keep using ecs, this is just to make it easier for states to share multiple dispatchers and not have to store them in the state (enum states for example are annoying to work with internal data imho).

torkleyy:
Okay, :+1: for not requiring the dispatcher in there. Though, I don't quite understand the motivation for removing the `world` field.

src/app.rs
```diff
@@ -333,15 +273,160 @@ impl<'a, 'b, T> ApplicationBuilder<'a, 'b, T> {
...
+}
+
+impl<'a, 'b> DefaultGameData<'a, 'b> {
+    /// Build
+    pub fn build<P: AsRef<Path>>(path: P) -> Result<DefaultGameDataBuilder<'a, 'b>> {
+        DefaultGameDataBuilder::new(path)
+    }
+
+    /// Get events
+    pub fn get_window_events<I>(&mut self) -> Vec<Event> {
```
torkleyy: I'd be very much in favor of calling this `poll_window_events` since `get_` makes the impression of a simple getter.

Rhuagh:
I didn't want to increase  the number of parameters to the state too much.

src/app.rs
Rhuagh: That's a better name yeah

jojolepro:
?

Rhuagh:
I will redo this pr, but I need some other architectural changes done first

PR #591, opened 2018-03-04
[Review] UI Layouts

jojolepro:
Adds layouting systems.
Anchor based layouting: You define which anchor you want, that what offset in pixel you want from there.
e.g TopRight + x: 100, y: 75

Stretching: Stretch the size of a UiTransform to the size of his parents on one or two axes. Uses the screen size if no parent is available.

Stretch margins: Creates space on the sides of something stretched.

See examples/ui for a usage example.

Also, fixes text not being visible if the z isn't between [-1,1]. (Thanks @Xaeroxe )

Once the review process is done, I'll squash the commits.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/591)

PR #601, opened 2018-03-09
[BugFix] Entity Recreation + Parenting

### LINKED ISSUE #600: [Bug] Entity recreation causes Parenting systems to fail.
How to replicate:
Create an entity.
Delete the entity a few frames later.
Re-create the same entity at the same index.

Known affected Systems:
-UiParentSystem #591 [Fixed]
-TransformSystem [Pending #601]

Impact:
As those systems think that it is the same entity, when it is not, they do not execute changes that should be applied to those newly created entities.

If you know of a System using a Bitset to track entities by their index, it is most likely affected by this too.

jojolepro:
Fixes #600

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/601)

### DIFF
--- a/amethyst_core/src/transform/systems.rs
+++ b/amethyst_core/src/transform/systems.rs
@@ -1,7 +1,6 @@
 //! Scene graph system and types
 
 use fnv::{FnvHashMap as HashMap, FnvHashSet as HashSet};
-use hibitset::BitSet;
 use specs::{Entities, Entity, Join, System, WriteStorage};
 use transform::{GlobalTransform, Parent, Transform};
 
@@ -15,8 +14,8 @@ pub struct TransformSystem {
     /// with parents.
     sorted: Vec<Entity>,
 
-    init: BitSet,
-    frame_init: BitSet,
+    init: Vec<Entity>,
+    frame_init: Vec<Entity>,
 
     dead: HashSet<Entity>,
     remove_parent: Vec<Entity>,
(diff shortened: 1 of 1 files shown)


Aceeri:
I don't think frame init needs to be changed. You can't have 2 entities of the same generation in the same loop.

amethyst_core/src/transform/systems.rs
```diff
@@ -70,11 +71,18 @@ impl<'a> System<'a> for TransformSystem {
 
         {
             // Checks for entities with a modified local transform or a modified parent, but isn't initialized yet.
-            let filter = locals.open().0 & parents.open().0 & !&self.init; // has a local, parent, and isn't initialized.
-            for (entity, _) in (&*entities, &filter).join() {
+            let filter = locals.open().0 & parents.open().0; // has a local, parent, and isn't initialized.
+            let mut to_add = Vec::<Entity>::new();
+            for (entity, _) in (&*entities, &filter)
+                .join()
+                .filter(|&(e, _)| !self.init.contains(&e))
```
torkleyy: This is pretty bad for performance :(

jojolepro: O(n) RIP :/

jojolepro:
Which data structure would be best for this one? HashSet?

torkleyy:
The bitshet should stay and get updated by FlaggedStorage.

jojolepro:
Will be fixed when the new specs releases. This will also include the performance fix for the ui parenting system. See the ui layout PR for more info. #591

jojolepro:
Being worked on by something else to fix multiple bugs at once.

PR #605, opened 2018-03-24
refactor: Reexport shred and specs through the core crate

Rhuagh:
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/605)


torkleyy:
I'll abstain, I'm not a fan of reexporting everything but then I know how ugly dealing with multiple versions of the same crate can be. I honestly don't know what's the best way to go here.

jojolepro:
I approved since we use specs almost everywhere, and since we are going to make a ton of crates, having a single re-export reduces the maintenance cost.

Rhuagh:
I debated this with myself for over a week before I created this pr. My motivation is the same as @jojolepro, you can't use anything in amethyst without specs and shred, so reexporting them makes sense

torkleyy:
Yeah, I tend to approve, but I'll leave it up to @omni-viral, @Aceeri and @Xaeroxe to decide.

Rhuagh:
bors r=Xaeroxe,jojolepro

PR #610, opened 2018-03-27
Add a variant to animation::EndControl

JadeSnail:
Add a Stay variant to EndControl, which make the property being animated on stay at the last sampled state at the end of the animation.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/610)

### DIFF
--- a/amethyst_animation/src/resources.rs
+++ b/amethyst_animation/src/resources.rs
@@ -194,6 +194,8 @@ pub enum EndControl {
     Loop(Option<u32>),
     /// When duration of sampler/animation is reached, go back to rest state
     Normal,
+    /// When duration of sampler/animation is reached, do nothing: stay at the last sampled state
+    Stay,
 }
 
 /// Control a single active sampler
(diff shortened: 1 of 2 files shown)


Rhuagh:
Try doing interpolate with a time value beyond the last input value, that should do it.

Rhuagh:
The basic issue is that the output data is dependent in the interpolation method. As an example, for spline interpolation the last value in the output data will be a tangent, not a usable value

JadeSnail:
Thanks for the help, I'm not sure what you mean when you say to sample with a value beyond the last input. I've changed it to run the interpolating function on the last input value.

Rhuagh:
The interpolation will return the last valid value of the input value is larger than the last input

Rhuagh:
If..

JadeSnail:
Do you want me to change last_frame to last_frame + 1. ? 
Reading the doc for the interpolation function for common interpolation methods, as far as I can tell, both should behave the same.
I'm also not sure that your assumption is always correct : when you interpolate with a custom function beyond the last input, you have no guarantees : https://docs.rs/minterpolate/0.2.2/src/minterpolate/lib.rs.html#122-137

torkleyy:
bors r=Rhuagh,torkleyy

PR #612, opened 2018-03-28
Remove unwrap in amethyst::animation

JadeSnail:
I was getting panics when adding, starting and then aborting an animation all before the animation control has a chance to run. This removes them.

Do tell me if such small pull requests are more annoying than useful.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/612)


jojolepro:
> Do tell me if such small pull requests are more annoying than useful.

I'm happy with that, as I can check them quickly when I'm short on time.

PR #613, opened 2018-03-29
Add UiButton and UiButtonBuilder

Hittherhod:
It's not showing up properly on my Mac, but I'd like to get a chance to iterate over the API and defaults.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/613)


amethyst_ui/src/button.rs
```diff
@@ -0,0 +1,158 @@
...
+    pub text: Entity,
+    /// Represents the background of the image. Defaults to a grey rectangle.
+    pub image: Entity,
+}
+
+impl<'a, 'b> UiButtonBuilder<'a, 'b> {
+    /// Construct a new UiButtonBuilder.
+    /// This allows easy use of default values for text and button appearance and allows the user
+    /// to easily set other UI-related options.
+    pub fn new(name: &'a str, world: &'b mut World) -> Self {
```
jojolepro: I think that having the text is an essential parameter here.

amethyst_ui/src/button.rs
```diff
@@ -0,0 +1,158 @@
...
+            let font = loader.load(
+                "font/square.ttf",
+                TtfFormat,
+                Default::default(),
+                (),
+                &world.read_resource::<AssetStorage<FontAsset>>(),
+            );
+            let text = UiText::new(font, "".to_string(), [0.0, 0.0, 0.0, 1.0], 32.0);
+            let grey = loader.load_from_data(
+                [0.82, 0.83, 0.83, 1.0].into(),
```
jojolepro: Why the 0.82 and 0.83?

Hittherhod:
Review status: 0 of 11 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[amethyst_ui/src/button.rs, line 33 at r1](https://beta.reviewable.io/reviews/amethyst/amethyst/613#-L8mC2iKvsu0XNeFpoHi-r1-33:-L8mdAoWCNR1PAz_fXEu:bebrt1w) ([raw file](https://github.com/amethyst/amethyst/blob/06dfc44cf0868242180f7a7220be177a44fdfd5a/amethyst_ui/src/button.rs#L33)):*
<details><summary><i>Previously, jojolepro (Joël Lupien) wrote…</i></summary><blockquote>

I think that having the text is an essential parameter here.
</blockquote></details>

Will do

---

*[amethyst_ui/src/button.rs, line 46 at r1](https://beta.reviewable.io/reviews/amethyst/amethyst/613#-L8mC2iKvsu0XNeFpoHi-r1-46:-L8md3Y_3-kcXh0qBiQJ:b-xbmsy5) ([raw file](https://github.com/amethyst/amethyst/blob/06dfc44cf0868242180f7a7220be177a44fdfd5a/amethyst_ui/src/button.rs#L46)):*
<details><summary><i>Previously, jojolepro (Joël Lupien) wrote…</i></summary><blockquote>

Why the 0.82 and 0.83?
</blockquote></details>

That was just a RGB color I found that is light gray

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/amethyst/amethyst/613)*

torkleyy:
@jojolepro Why exactly do you think such an `ToEntities` trait is useful?

jojolepro:
When you have multiple menu and want to change between them, you need to remove the first menu before adding the second. (maybe the multiple World idea was good, after all). To easily do so, you can create all your ui inside a vec like so
```
fn create_ui(....) -> Vec<Entity>{
  vec![
    UiButton::new(.......).to_entities(),
    ...
  ]
}

fn change_menu(){
  for e in old_entities{
    entities.delete(e);
  }
}
```

torkleyy:
If they all have one ui node as parent, wouldn't that be better?

jojolepro:
Technically I think there will always be a root entity and the rest as childs, so it could return only a single entity.

Hittherhod:
Could you walk the tree to get all the entities you want?

torkleyy:
If not, that functionality should be added.

jojolepro:
We can't really walk from parent to child without iterating over all entities.

Hittherhod:
Would you already know a fairly local parent (like, say, the menu) that you could use as the root of your search?

jojolepro:
Well actually, you just made me realize we could use a "root" ui entity to delete a menu all at once.

When you delete an entity, all the childs are deleted automatically by the parent/layout system, so I would guess that by using UiButton::new(...).with_parent(root_menu) and then deleting the root_menu entity, the problem is fixed, and we don't need the ToEntities trait.

Hittherhod:
Okay. I'll delete the `ToEntities` trait.

amethyst_ui/src/button.rs
```diff
@@ -0,0 +1,158 @@
...
+    pub text: Entity,
+    /// Represents the background of the image. Defaults to a grey rectangle.
+    pub image: Entity,
+}
+
+impl<'a, 'b> UiButtonBuilder<'a, 'b> {
+    /// Construct a new UiButtonBuilder.
+    /// This allows easy use of default values for text and button appearance and allows the user
+    /// to easily set other UI-related options.
+    pub fn new<S: ToString>(name: &'a str, text: S, world: &'b mut World) -> Self {
```
torkleyy: Can we pass the `World` to `build` instead of `new`?

examples/ui/main.rs
```diff
@@ -176,48 +176,28 @@ impl State for Example {
...
+                    9,
+            ))
+            .with_image(UiImage {
+                texture: green.clone(),
+            })
+            .with_anchored(Anchored::new(Anchor::TopMiddle))
+            .with_parent(Parent {
+                entity: background.clone(),
             })
             .build();
```
jojolepro: Could you add another button showing a less heavy syntax?
e.g: UiButtonBuilder::new("btn2", "Simple Button").with_font(font.clone()).with_position(50.0,50.0).build(&mut world);

Rhuagh:
Basically, my concern is that we should make it possible to use the builder without access to the full world (for creation from inside systems)

Hittherhod:
@Rhuagh Could you explain more about LazyUpdate? I've never seen it in use, so I'm not sure how to work with it.

Rhuagh:
https://github.com/amethyst/amethyst-rhusics/blob/ffbd88bb3650f65026d3cb6664a8475d535dddb4/examples/boxes/emission.rs#L79

Rhuagh:
There's an example

Hittherhod:
Review status: 0 of 12 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[amethyst_ui/src/button.rs, line 33 at r4](https://beta.reviewable.io/reviews/amethyst/amethyst/613#-L8mC2iKvsu0XNeFpoHi-r4-33:-L95UUM6BUHKEPgDwAAG:b-896fix) ([raw file](https://github.com/amethyst/amethyst/blob/f4e8ce9b0ff5da3e6d3dd72c01da86adc4120497/amethyst_ui/src/button.rs#L33)):*

Can we pass the `World` to `build` instead of `new`?
</blockquote></details>

Done.

---

*[examples/ui/main.rs, line 203 at r4](https://beta.reviewable.io/reviews/amethyst/amethyst/613#-L8mC2iKvsu0XNeFpoHh-r4-203:-L95UW5e5E7wzbUp46__:b-896fix) ([raw file](https://github.com/amethyst/amethyst/blob/f4e8ce9b0ff5da3e6d3dd72c01da86adc4120497/examples/ui/main.rs#L203)):*
<details><summary><i>Previously, jojolepro (Joël Lupien) wrote…</i></summary><blockquote>

Could you add another button showing a less heavy syntax?
e.g: UiButtonBuilder::new("btn2", "Simple Button").with_font(font.clone()).with_position(50.0,50.0).build(&mut world);
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/amethyst/amethyst/613)*

amethyst_ui/src/button.rs
```diff
@@ -0,0 +1,224 @@
...
+const DEFAULT_Y: f32 = 0.0;
+const DEFAULT_Z: f32 = -1.0;
+const DEFAULT_WIDTH: f32 = 128.0;
+const DEFAULT_HEIGHT: f32 = 64.0;
+const DEFAULT_TAB_ORDER: i32 = 9;
+const DEFAULT_BKGD_COLOR: [f32; 4] = [0.82, 0.83, 0.83, 1.0];
+const DEFAULT_TXT_COLOR: [f32; 4] = [0.0, 0.0, 0.0, 1.0];
+const DEFAULT_FONT_NAME: &'static str = "font/square.ttf";
+
+/// Container that wraps the resources we need to initialize button defaults
```
Rhuagh: #[derive(SystemData)]

amethyst_ui/Cargo.toml
```diff
@@ -27,4 +27,4 @@ shrev = "0.8.1"
 specs = "0.10"
 unicode-normalization = "0.1"
 unicode-segmentation = "1.2"
-winit = "0.10"
+winit = "0.11"
```
Rhuagh: shred-derive = "0.3"

amethyst_ui/src/lib.rs
```diff
@@ -22,17 +22,19 @@ extern crate unicode_segmentation;
 extern crate winit;
```
Rhuagh: #[macro_use]
extern crate shred_derive;

amethyst_ui/src/button.rs
```diff
@@ -0,0 +1,224 @@
...
+pub struct UiButtonResources<'a> {
+    loader: Fetch<'a, Loader>,
+    font_asset: Fetch<'a, AssetStorage<FontAsset>>,
+    texture_asset: Fetch<'a, AssetStorage<Texture>>,
+}
+
+impl <'a> UiButtonResources<'a> {
+    /// Grab the resources we need from the world. 
+    pub fn from_world(world: &World) -> UiButtonResources {
+        UiButtonResources {
```
Rhuagh: Self::fetch(&world.res, 0)

amethyst_ui/src/button.rs
```diff
@@ -0,0 +1,224 @@
...
+        }
+        if let Some(stretched) = self.stretched.take() {
+            image_builder = image_builder.with(stretched);
+        }
+
+        image_builder.build()
+    }
+
+    /// Create the UiButton based on provided configuration parameters.
+    pub fn build(mut self, world: &mut World) -> UiButton {
```
Rhuagh: Make another of the "Resources" structs, that contain the storages required for creating a button (WriteStorage<'a, UiImage> etc), and make the build function take that resources struct. Make the resources struct contain `Entities<'a>` so you can create entities.

amethyst_ui/src/button.rs
Hittherhod: What does `SystemData` do?

amethyst_ui/src/button.rs
Hittherhod: How do I create replicate the current `World` centric functionality with Storages and Entities? Does this new Resources struct also need to derive SystemData?

amethyst_ui/src/lib.rs
Hittherhod: Done. I needed to add shred as well to make the compiler be quiet.

amethyst_ui/src/button.rs
Rhuagh: Makes it so you can use the struct directly in a system.

Hittherhod:
@Rhuagh I've added changes to use the WriteStorage resources structure. How would this work with a LazyUpdate?

amethyst_ui/src/button.rs
```diff
@@ -185,48 +204,40 @@ impl<'a> UiButtonBuilder<'a> {
...
             image: image_entity,
         }
     }
+    /// Create the UiButton based on provided configuration parameters.
+    pub fn build_from_world(self, world: &World) -> UiButton {
+        self.build(UiButtonBuilderResources::from_world(world))
+    }
 
     /// Lazily build the UiButton. Need to call `World::maintain` to have the values actually added.
     pub fn lazy_build(mut self, res: UiButtonLazyResources) -> UiButton {
```
Rhuagh: You can remove this, it's no longer needed, and LazyUpdate might disappear soon (or atleast be redesigned since), since it has issues.

Hittherhod:
@Rhuagh Any feedback on the changes? I've removed the LazyUpdate structure. I want to add an example of having the button text update if you click on the button, but my click events aren't registering (the "you clicked on this element" isn't getting printed). @jojolepro Could this be another Mac specific issue?

jojolepro:
@Hittherhod You are not adding the MouseReactive component to the button's image.
You are not reading the UiEvent s either.

Hittherhod:
Where do I read the `UiEvents`? I was trying to modify the UiEventHandlerSystem in main.rs, but it's not pushed because I didn't want to add a broken commit.

jojolepro:
Actually, the example is in fact reading for events :P my bad
https://github.com/jojolepro/amethyst/blob/develop/examples/ui/main.rs#L295

You are still missing MouseReactive tho

Hittherhod:
So, I added the `MouseReactive` and I see HoverStart and HoverStop events, but I'm not seeing click events.

jojolepro:
I am
:works_on_my_pc: :)

Hittherhod:
Okay. I'll refrain from adding that because I can't test it, but it'd be a nice to have for another PR.

Rhuagh:
We're waiting for a bugfix in winit that triggers the retina bugs.

torkleyy:
Doesn't seem to be fixed really. Could you also try to clean up the commit history?

torkleyy:
Ah the build is just failing because of Travis.

Hittherhod:
Sure. I’ll try to squash things to just include the major fixes.

Hittherhod:
@Rhuagh I tried to clean it up so that it only reports the significant updates rather than every little change.

amethyst_ui/src/button.rs
```diff
@@ -0,0 +1,237 @@
...
+#[derive(SystemData)]
+pub struct UiButtonResources<'a> {
+    font_asset: Fetch<'a, AssetStorage<FontAsset>>,
+    loader: Fetch<'a, Loader>,
+    texture_asset: Fetch<'a, AssetStorage<Texture>>,
+}
+
+/// Container for all the resources the builder needs to make a new UiButton.
+#[derive(SystemData)]
+struct UiButtonBuilderResources<'a> {
```
jojolepro: Resources is a confusing name, could you use something like "parameters"?

Hittherhod:
Is there some way to make this merge?

Rhuagh:
bors r=torkleyy,Rhuagh

Rhuagh:
bors r=torkleyy,Rhuagh

jojolepro:
Congratz!

Issue #622, opened 2018-04-04
Windows: amethyst_assets (font) loader does not work with mixed slashes


azriel91:
Hiya, I found a bug when loading fonts on Windows. If I instantiate the Amethyst application with an assets directory path like `"assets"`, and load a Font asset using "relative/forward_slashed/path/to/font.ttf", it works.

However, if I instantiate it with a canonicalized path, like:

```rust
// pseudo code
let assets_dir = env::current_exe()?
    .parent()?
    .canonicalize()?
    .join("assets");
let application = Application::new(assets_dir, /* .. */) // .. bundles etc.
```

Then when I go `loader.load("forward_slashed/path/to/font.ttf", /* .. */);`, it fails the load the path.
If I use `loader.load("back_slashed\\path\\to\\font.ttf", /* .. */);`, it works.

The actual log output I get is as follows:

```
[amethyst_assets::loader][DEBUG] "ui::Font": Loading asset "font/source-code-pro-2.030R-ro-1.050R-it/TTF/SourceCodePro-Regular.ttf" with format "TTF/OTF" from source "[default source]" (handle id: Handle { id: 0, marker: PhantomData })
[amethyst_assets::loader][DEBUG] "ui::Font": Loading asset "font/source-code-pro-2.030R-ro-1.050R-it/TTF/SourceCodePro-Bold.ttf" with format "TTF/OTF" from source "[default source]" (handle id: Handle { id: 1, marker: PhantomData })
[amethyst_assets::loader][DEBUG] "ui::Font": Loading asset "font/source-code-pro-2.030R-ro-1.050R-it/TTF/SourceCodePro-It.ttf" with format "TTF/OTF" from source "[default source]" (handle id: Handle { id: 2, marker: PhantomData })
[amethyst_assets::loader][DEBUG] "ui::Font": Loading asset "font/source-code-pro-2.030R-ro-1.050R-it/TTF/SourceCodePro-BoldIt.ttf" with format "TTF/OTF" from source "[default source]" (handle id: Handle { id: 3, marker: PhantomData })
[amethyst_assets::storage][ERROR] "ui::Font": Asset "font/source-code-pro-2.030R-ro-1.050R-it/TTF/SourceCodePro-Regular.ttf" (handle id: Handle { id: 0, marker: PhantomData }) could not be loaded: Failed to load asset with name "font/source-code-pro-2.030R-ro-1.050R-it/TTF/SourceCodePro-Regular.ttf"
[...]

PR #623, opened 2018-04-04
Construct `PathBuf` from asset path before pushing asset location.

azriel91:
This tolerates forward slashed asset paths when the asset location was
constructed with a path containing backslashes (i.e. canonicalized
Windows paths).

Issue #622

Checklist:

* [x] Actually fix the issue + write tests for it.
* [x] cargo build --all-targets
* [x] cargo test --all
* [x] cargo run --example pong

By the way, the `.pre-commit.sh` script doesn't exist, I suppose it must've been removed at some point.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/623)


azriel91:
Hold on, wrapping the call with `PathBuf::from(..)` didn't fix it. Shall find something that does

azriel91:
Not sure if you like that fix, it's not _that_ elegant, but I couldn't think of a better way.

amethyst_assets/src/source/dir.rs
```diff
@@ -27,7 +27,7 @@ impl Directory {
 
     fn path(&self, s_path: &str) -> PathBuf {
         let mut path = self.loc.clone();
-        path.push(s_path);
+        path.push(Path::new(s_path).iter().collect::<PathBuf>());
```
Rhuagh: Doesn't `PathBuf::from(s_path)` work?

Rhuagh: Ah, you already answered that.

torkleyy: Can we do the following?

```rust
path.extend(Path::new(s_path).iter());
```

That would save us some allocations.

torkleyy:
@azriel91 Memory allocation can happen multiple times actually. For one, `collect` will push element for element and the size isn't known beforehand. So multiple allocations for `collect` + potentially one for the `push`. By using the source I posted above we don't allocate any extra buffers.

azriel91:
Woosh, the more you know. Pushed =D

---







*Comments from [Reviewable](https://beta.reviewable.io/reviews/amethyst/amethyst/623#-:-L9P5S6L50CLR66Q3tvs:b4x0p8v)*

torkleyy:
Okay, here you go.

bors r+

jojolepro:
I'm late, but there's a missing new line https://github.com/amethyst/amethyst/pull/623/files#diff-6f5ca58e4d4ca434ece88e3f7cfedac1

Issue #583, opened 2018-02-26
Using transparencies requires the user to link in other crates


trsoluti:
The with_transparency() method on passes takes three arguments:

```
        mask: ColorMask,
        blend: Blend,
        depth: Option<DepthMode>
```
The types and possible values for these are private within Amethyst, meaning the user has to link in gfx and gfx_core just to get a program using with_transparency() to compile.

I recommend the following:

1) re-export the appropriate types and constants from gfx and gfx_core (`ColorMask`, `Blend`, `DepthMode`, and `gfx::preset::blend::*`).
2) add another method `with_standard_transparency` which takes no arguments and calls `with_transparencies(ColorMask::all(), ALPHA, None)` (these are the arguments the UI pass uses).

This should provide flexibility and also ease of use.

Another option is to make transparencies the default, and to add another method `without_transparency` for people who want a faster pass.

Rhuagh:
I'd like to avoid the last part, simply because the vast majority of meshes will be opaque. The rest sound good to me.

PR #624, opened 2018-04-07
Discussion: Example/sprite render

azriel91:
Example of how to render sprites.

image

Sprite loading code was based off @trsoluti's https://github.com/trsoluti/space_shooter, and sprites are from <https://opengameart.org/content/bat-32x32> (creative commons 0 license &mdash; "you can use it for free").

I'm not sure if this example should be merged, or whether the sprite loading code should be cleaned up and put into the engine itself, and we make the example use the engine's types.

---

Other things I encountered, which probably will fall under their own issues/PRs:

* Related to #583, I needed to use a `Blend`, and there's nice defaults in `gfx::preset::blend`. However the consuming crate needs to depend on `gfx` themselves to get this.
* `BlendValue` from `gfx_core` isn't exported, so I couldn't copy paste the `ALPHA` `Blend`:
* I don't understand these things, but I had to do them to make the example work. Probably fixable with documentation / education (haha, please educate me):
    - add a `GlobalTransform` component to the entity
    - make sure there's no `with_visibility_sorting` call on the `RenderBundle` (I added it as part of "I don't know what's going on, let's just call functions that I found in semi related PRs")
    - make sure there's a `with_transparency` call on the `DrawFlat` pass.
* How come the rendered sprites are brighter than the actual image's pixels? Probably another learning exercise

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/624)


jojolepro:
If you want to, I think this could fit nicely has an engine feature. It would need some cleaning up and a couple more options (horizontal vs vertical first, size of each sprite).

>add a GlobalTransform component to the entity

Indicates the position of an entity, after calculating it according to his parent's position (if any).
It contains a matrix which is used for the draw pass too.

>make sure there's no with_visibility_sorting call on the RenderBundle

I'm not sure why that would cause a problem.

>make sure there's a with_transparency call on the DrawFlat pass

That tells the pass to calculate transparent pixels, allowing to see behind the sprites where they have transparent pixels.

Rhuagh:
You do need visibility sorting, without it layering won't be correct

Rhuagh:
@Xaeroxe don't we already have partial support for spritesheets?

examples/sprite_render/main.rs
```diff
@@ -0,0 +1,166 @@
...
+            // This combines multiple `Transform`ations.
+            // You need to `use amethyst::core::cgmath::Transform`;
+            sprite_transform.concat_self(&common_transform);
+
+            world
+                .create_entity()
+                .with(sprite_sheet.sprite_meshes[i].clone())
+                .with(sprite_sheet.image.clone())
+                .with(sprite_transform)
+                // This is needed to make anything show up. I'm not sure why.
```
Xaeroxe: It's necessary because the engine uses this to compute and store the rendered position.

Now you know why so you can remove this comment :P

azriel91:
Closing - superseded by #638.

PR #627, opened 2018-04-08
Fix critical bug in asset handle deletion

### LINKED ISSUE #618: Wrong assets get freed
Ref https://github.com/thepowersgang/fury3clone/blob/master/src/main.rs#L687-L704

When adding multiple entities that use the same model (and material), both `DrawShadedSeparate` and `DrawFlatSeparate` panic with a None unwrap. However, when the original handles to the models and materials are forgotten (using mem::forget of the containing Vec in the above code), the render acts as expect (drawing all models in the world).

torkleyy:
Fixes #618

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/627)

### DIFF
--- a/amethyst_assets/src/storage.rs
+++ b/amethyst_assets/src/storage.rs
@@ -243,6 +243,8 @@ impl<A: Asset> AssetStorage<A> {
         let mut skip = 0;
         while let Some(i) = self.handles.iter().skip(skip).position(Handle::is_unique) {
             count += 1;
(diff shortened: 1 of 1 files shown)


torkleyy:
Looking at this now, I can't believe I wrote this..

azriel91:
A developer once said, "if you can look back at your previous code without saying 'I can't believe I wrote this', then you haven't progressed." 😛

Issue #629, opened 2018-04-09
Document the animation crate


azriel91:
I was trying to understand how the animation crate worked, and this was my attempt at understanding it.

Excerpt from gitter chat:

> Is this correct:
> 
> * [`Sampler`](https://www.amethyst.rs/doc/develop/doc/amethyst_animation/struct.Sampler.html):
> 
>     - `input`: Distinct "points in time" of the animation called "key frames".
>     - `output`: Given `input` has 10 key frames, the "output" should contain 10 `T` values which are "correct" values for the animation during those key frames.
>     - `function`: How to calculate an output value that lies between two key frames.
> 
> * `Channel`: (there's no concrete _type_ for this besides the associated type in [`AnimationSampling`](https://www.amethyst.rs/doc/develop/doc/amethyst_animation/trait.AnimationSampling.html), but the concept is there)
> 
>     A grouping of animatable attributes that are related (and usually affect each other).
> 
>     e.g. [`TransformChannel`](https://www.amethyst.rs/doc/develop/doc/amethyst_animation/enum.TransformChannel.html) has `Translate`, `Rotate`, and `Scale` because depending on the order of applying those operations, the final output will be different.
> 
>     And the following are possible channels:
> 
>     - `MaterialChannel`: different changes to the underlying textures (e.g. the albedo, etc.) affect the final rendered colour
>     - `TransparencyChannel`: How visible a texture is rendered (maybe this can be merged into the `MaterialChannel`?)
> 
> * [`AnimationSampling`](https://www.amethyst.rs/doc/develop/doc/amethyst_animation/trait.AnimationSampling.html):
> 
>     Trait to bridge the sampler function to the component (that contains the in-game values as input to the sampler function).
> 
>     What is [`BlendMethod`](https://www.amethyst.rs/doc/develop/doc/amethyst_animation/enum.BlendMethod.html) used for? Need to look at code to figure it out.
> 
> * [`Animation`](https://www.amethyst.rs/doc/develop/doc/amethyst_animation/struct.Animation.html):
> 
>     A grouping of individual parts &mdash; "nodes" &mdash; of the animation. Each node operates on one of the `Channel`s.
> 
>     What is the `usize` in the first element of the tuple? The [animation example](https://github.com/amethyst/amethyst/blob/develop/examples/animation/main.rs#L252-L259) just uses 0 for all of them.
> 
>     I have no idea what the `AnimationHierarchy` is, I got lost in the words.

`JadeSnail`:

[...]

PR #631, opened 2018-04-10
Docs/629/animation

azriel91:
Added documentation for animation crate (#629).

There were some bits that didn't seem to fit in the crate documentation, so I put it under `book`, but it feels "halfway" since there isn't a complete tutorial.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/631)


amethyst_animation/src/resources.rs
```diff
@@ -42,10 +45,25 @@ where
     T: InterpolationPrimitive,
 {
     /// Time of key frames
+    ///
+    /// A simple example of this for animations that are defined with 4 evenly spaced key frames is
+    /// `vec![0., 1., 2., 3.]`.
     pub input: Vec<f32>,
     /// Actual output data to interpolate
+    ///
+    /// For `input` of size `i`, the `output` is size `n * i`, where `n` is the number of values for
```
Rhuagh: This is not completely correct, the size of the output can be anything the function requires, so there's no specific limit on this size.

Rhuagh: A summary of the predefined functions would be:

* Linear - i
* SphericalLinear - i
* CubicSpline - i * 3
* CatmullRomSpline - i + 2
* Step - i

amethyst_animation/src/lib.rs
```diff
@@ -1,3 +1,50 @@
...
+//! `(Channel, Sampler)` pair is enough to define one part the animation, and a
+//! `Vec<(Channel, Sampler)>` defines the whole animation.
+//!
+//! In a more complex situation, an object in game may be made up of multiple entities. Say you have
+//! a dragon monster, that is defined by a skinned mesh that has a skeleton with 10 joints. Each
+//! joint will then be an `Entity`. Our animation definition holds the samplers to run for the whole
+//! object. To animate each of the entities of this complex object, we need a way to link the
+//! sampler to the each of the entities.
+//!
+//! Animation definitions are persistent and can be stored on disk. Entities however, are not. To
```
jojolepro: Entities are not, but they will soon

book/src/animation/definition.md
```diff
@@ -0,0 +1,13 @@
+# Definition
+
+Animations can be defined for objects made of a single entity, or complex objects made up of multiple entities.
+
+Right now we do not have a tutorial for defining an animation from scratch, but take a look at the following resources to get an idea of how to write one:
```
jojolepro: We don't have a tutorial for animations. Do you mind opening an issue for that, and reference this line so we don't forget to change it when we do?

amethyst_animation/src/lib.rs
Rhuagh: No, the Entity itself cannot be stored. Markers can.

book/src/animation/definition.md
azriel91: Certainly :slightly_smiling_face: #637

PR #633, opened 2018-04-10
READMEs extended with examples

MrMinimal:
To make it easier for new users to get into the project, this PR adds to the existing README files sections about compiling the examples as well as screenshots of the expected results.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/633)


torkleyy:
Thanks @MrMinimal!

Would you mind squashing all the commits to a single one?

README.md
```diff
@@ -67,24 +67,39 @@ does not aim to be API-compatible with it in any way. Some goals include:
...
+If you are compiling on Linux make sure to install the following dependencies:
+
+### Ubuntu
+```
+$ sudo apt install libasound2-dev libx11-xcb-dev
+```
+
+### Fedora
+```
+$ sudo yum install alsa-lib-devel libx11-xcb-dev
```
torkleyy: I don't think libx11-xcb-dev exists on Fedora.

MrMinimal:
@torkleyy Not at all! Done.

README.md
MrMinimal: That was in the Usage section of the README before. Is it only relevant for Ubuntu or does Fedora need it as well, just with a different package name?

torkleyy: I think for Fedora it's all bundled in `libx11-devel`, but last time I used Fedora.. it's a long time ago, I can't say for sure.

Xaeroxe:
@Rhuagh I don't think so personally.  They don't provide a ton over the pong example and are mostly just there to alert us when documentation needs to be updated.

Issue #628, opened 2018-04-08
Warn about unnecessary asset loading


torkleyy:
We could just entirely skip loading an asset when nobody is holding a handle to it.

# Known drawbacks

* when benchmarking asset loading you need to hold the handle, otherwise it doesn't load anything

# Unresolved questions

* should we log this?
* is there anything wrong with just calling `Tracker::success` in such a case?

# Alternatives

* only log a warning, keep loading the asset as usual

Xaeroxe:
I actually think this particular optimization doesn't help a ton.  It complicates the system rules, which makes performance less intuitive, and it only helps in scenarios where users are loading things that they shouldn't be, which shouldn't be a common scenario.  If the user doesn't want that particular asset to load then they simply shouldn't load it.

torkleyy:
I think I have an alternative here. How about we just `warn!` in such a case and continue as usual? Would that be a better solution in your eyes?

zoonage:
Looking into this

jojolepro:
Off topic, but can we actually unload assets?

jojolepro:
Found my answer while reading through my notifications :)

zoonage:
@torkleyy could you clarify what the issue is if the handle is unique?Even if only one existed then it is still a valid handle to an asset, so why would we not want to load it?

I made some changes based off of the the task, however I think I have the wrong end of the stick with this `Handle` thing https://github.com/zoonage/amethyst/commit/b083eded0f2ce1519b2a4b7a3af991b2d310bbf1.

zoonage:
I've put in PR #642 for this (and some profling code that got left out of my last issue) however I've got absolutely no idea how to actually test this change

zoonage:
Are there any docs on how asset loading works? My current understanding is something like

* `World` adds a `Asset` source via the `Loader`
* The `Loader` queries the `AssetStorage`
* The `Loader` runs a task in the `threadpool` to read the requested `Asset` from disk
* The `Loader` returns a `Handle` that corresponds to the asset
* When the `Asset` is queried for there is no guarantee that it has been read yet, but this is visible by some sort of `tracker` that is associated with the `Handle`
* `process_custom_drop` runs in a background thread to update the status of the loading of assets and trigger hot reloads of assets and update the details of the tracker

From this I'm guessing that `process_custom_drop` is the place where we want this check?

zoonage:
For future reference @torkleyy  replied in Gitter

```


[...]

Issue #503, opened 2017-12-04
Add thread profiler to all subcrates


Aceeri:
Profiling was broken when we moved the code into subcrates, so we need to add the subcrate features to the top level Cargo.toml

Hittherhod:
Which subcrates use it? The latest develop branch doesn't seem to have any subcrates referencing the thread_profiler.

zoonage:
Looks like this was done in commit 07cf8b45f2551eee7beb5f645fbd16b5acac4380, should this be closed?

torkleyy:
I think this issue is about two things:

* enable the `profiler` feature for the subcrates if the `profiler` feature of the main crate is enabled
* add profiling code to the subcrates

Both are not fixed as far as I can see.

Rhuagh:
That sounds about right to me yeah. All systems should have profiling atleast. For the asset crate I'd guess all format calls should be profiled.

zoonage:
@Rhuagh could you clarify what you mean by 'all format calls should be profiled'?

Going from the examples it looks like it would be up to the program using the library to implement the profiling for the functions in the `Format` trait.

Rhuagh:
https://github.com/amethyst/amethyst/blob/a78139f8612ffe1aea7e0eb53f4389b43d7b62e9/amethyst_assets/src/loader.rs#L142

The import here should be profiled, so we can tell how fast/slow individual formats are.

zoonage:
@Rhuagh I've added this into the PR.

On a generic note should we standardise the names of each profile? Currently they seem to be named sensibly, but arbitrarily. Maybe we could use a naming scheme like `<crate_name>::<path>::<to>::<function>::<comment>` i.e. `amethyst_assets::source::dir::Directory::load::load_asset`, this would make debugging a lot easier as it would tell you where to look, however these are quite lengthy.

jojolepro:
Looks like a good idea, but if we change the directory of a file we'll forget to change the profile name.
What about amethyst_assets:load_asset

zoonage:
I've logged an issue against thread_profiler to get some comments https://github.com/glennw/thread_profiler/issues/12

zoonage:
I've put in a PR to fix glennw/thread_profiler#12 after that this issue is just adding the profiling code to our sub-crates

Rhuagh:
Because you don't want profiling in a live game, because it consumes cpu cycles.

zoonage:
[...]

PR #642, opened 2018-04-14
Unecessary asset loading

zoonage:
__STORAGE__ 
- Warn on trying to load an asset with nothing except the storage holding a handle #628.

__ANIMATION__ 
- Adds some profiling code to animation crate that wasn't submitted during #503.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/642)

### DIFF
--- a/amethyst_assets/src/storage.rs
+++ b/amethyst_assets/src/storage.rs
@@ -147,18 +147,7 @@ impl<A: Asset> AssetStorage<A> {
                     handle,
                     name,
                     tracker,
-                } => {
-
-                    // Add a warning if a handle is unique (i.e. asset does not
-                    // need to be loaded as it is not used by anything)
-                    // https://github.com/amethyst/amethyst/issues/628
-                    if handle.is_unique() {
-                        warn!(
-                            "Loading unecessary asset. Handle {} is unique ",
-                            handle.id()
-                        );
-                    }
-                    
+                } => {                    
                     let (asset, reload_obj) = match data.map(|FormatValue { data, reload }| {
                         (data, reload)
                     }).and_then(|(d, rel)| f(d).map(|a| (a, rel)))
@@ -189,6 +178,16 @@ impl<A: Asset> AssetStorage<A> {
                         }
                     };
 
+                    // Add a warning if a handle is unique (i.e. asset does not
+                    // need to be loaded as it is not used by anything)
+                    // https://github.com/amethyst/amethyst/issues/628
+                    if handle.is_unique() {
+                        warn!(
+                            "Loading unecessary asset. Handle {} is unique ",
+                            handle.id()
+                        );
+                    }
+                    
                     let id = handle.id();
                     bitset.add(id);
                     handles.push(handle.clone());


amethyst_assets/src/storage.rs
```diff
@@ -93,6 +95,14 @@ impl<A: Asset> AssetStorage<A> {
     /// Get an asset from a given asset handle.
     pub fn get(&self, handle: &Handle<A>) -> Option<&A> {
         if self.bitset.contains(handle.id()) {
+            // Add a warning if a handle is unique (i.e. asset does not
+            // need to be loaded as it is not used by anything) 
+            // https://github.com/amethyst/amethyst/issues/628
+            if handle.is_unique() {
```
torkleyy: Err this condition will always be false.

zoonage:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[amethyst_assets/src/storage.rs, line 101 at r1](https://beta.reviewable.io/reviews/amethyst/amethyst/642#-LA2dN12buLzsUtKL8Ow-r1-101:-LACYlCG06PkWOfsrRrF:b-49qly) ([raw file](https://github.com/amethyst/amethyst/blob/7b62e11c9a14f884a3708d30921f1ad1586fbe6e/amethyst_assets/src/storage.rs#L101)):*

Err this condition will always be false.
</blockquote></details>

Is that because the `is_unqieu()` method will always be false or because it is in the function? Either way I'm not sure I understand why it would always be false, could you explain?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/amethyst/amethyst/642)*

torkleyy:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[amethyst_assets/src/storage.rs, line 101 at r1](https://beta.reviewable.io/reviews/amethyst/amethyst/642#-LA2dN12buLzsUtKL8Ow-r1-101:-LAJS4Vn4qCVluuM8D1O:b2rl09c) ([raw file](https://github.com/amethyst/amethyst/blob/7b62e11c9a14f884a3708d30921f1ad1586fbe6e/amethyst_assets/src/storage.rs#L101)):*
<details><summary><i>Previously, zoonage (Aaron George) wrote…</i></summary><blockquote>

Is that because the `is_unqieu()` method will always be false or because it is in the function? Either way I'm not sure I understand why it would always be false, could you explain?
</blockquote></details>

Sorry, I was in a hurry when I reviewed. `is_unique` cannot return `true` because there is one handle stored in the `AssetStorage` and another one was passed `AssetStorage::get` (this method). Thus, the handle has at least two refs.

This check needs to go into the processing, when the asset got received from the `Loader`. I added another comment to point you to the right place.

---

*[amethyst_assets/src/storage.rs, line 194 at r1](https://beta.reviewable.io/reviews/amethyst/amethyst/642#-LAJSeHa8SwscnFkN3au:-LAJSeHa8SwscnFkN3av:b-8q8xo1) ([raw file](https://github.com/amethyst/amethyst/blob/7b62e11c9a14f884a3708d30921f1ad1586fbe6e/amethyst_assets/src/storage.rs#L194)):*
> ```Rust
>                         }
>                     };
> ```

The check can go about here.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/amethyst/amethyst/642)*

jojolepro:
Are you using the latest version of rustfmt-preview?

zoonage:
This was done with 1.26 nightly, just updated to 1.27

zoonage:
Review status: 0 of 65 files reviewed at latest revision, 2 unresolved discussions.

---

*[amethyst_assets/src/storage.rs, line 194 at r1](https://beta.reviewable.io/reviews/amethyst/amethyst/642#-LAJSeHa8SwscnFkN3au:-LAJZEuf7nImR-4VcBJm:b-x0ktkf) ([raw file](https://github.com/amethyst/amethyst/blob/7b62e11c9a14f884a3708d30921f1ad1586fbe6e/amethyst_assets/src/storage.rs#L194)):*

The check can go about here.
</blockquote></details>

Why do we only check for this in `NewAssets`? Wouldn't assets that are being `HotReload` need to be checked as well?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/amethyst/amethyst/642)*

Rhuagh:
We use the stable preview

amethyst_assets/src/asset.rs
```diff
@@ -3,6 +3,12 @@ use std::sync::Arc;
 use amethyst_core::specs::UnprotectedStorage;
 #[cfg(feature = "profiler")]
 #[macro_use]
+#[cfg(feature = "profiler")]
+#[macro_use]
+#[cfg(feature = "profiler")]
+#[macro_use]
+#[cfg(feature = "profiler")]
+#[macro_use]
```
Xaeroxe: Yerr this is almost certainly a mistake

zoonage:
re-run this with the stable toolchain

---

Review status: 0 of 93 files reviewed at latest revision, 3 unresolved discussions.

---

*[amethyst_assets/src/asset.rs, line 11 at r3](https://beta.reviewable.io/reviews/amethyst/amethyst/642#-LAJZk0PneZuPbqqg6C--r3-11:-LAJaKE63_1oIjq7uqtC:bc5j2tg) ([raw file](https://github.com/amethyst/amethyst/blob/dea793684fda874b3e55a61454561019a930ee84/amethyst_assets/src/asset.rs#L11)):*
<details><summary><i>Previously, Xaeroxe (Jacob Kiesel) wrote…</i></summary><blockquote>

Yerr this is almost certainly a mistake
</blockquote></details>

it most definitely is, I've removed these

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/amethyst/amethyst/642#-:-LAJaGnIB9NX3XfZyaiI:bgy3yxh)*

zoonage:
To clean this up I'll reset this branch back to develop and re-add the handle change so that we  can do this without rust fmt and the additional profiling changes (which I question the value of after I did more work on the profiling crate)

amethyst_assets/src/storage.rs
```diff
@@ -148,6 +148,8 @@ impl<A: Asset> AssetStorage<A> {
                     name,
                     tracker,
                 } => {
+
```
torkleyy: Please remove the whitespace and squash the commits, then this is ready to get merged :+1:

amethyst_assets/src/storage.rs
```diff
@@ -147,7 +147,7 @@ impl<A: Asset> AssetStorage<A> {
                     handle,
                     name,
                     tracker,
-                } => {
+                } => {
```
jojolepro: what is this?

torkleyy: extra whitespace supply for hard times ^^ nah, it'll be cleaned up with the next cargo fmt

torkleyy: extra whitespace supply for hard times ^^ nah, it'll be cleaned up with the next cargo fmt

torkleyy: The formatting bot would still be nice - maybe just a bot that creates PRs if there are changes.

PR #646, opened 2018-04-15
Cfg controlling mutli-threading to make web build again

Moxinilian:
This PR introduces a new rustc cfg `no_threading` in order to make amethyst build again for web browsers.
It is meant to signify amethyst must not take advantage of multi-threading facilities.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/646)

### DIFF
--- a/amethyst_core/src/lib.rs
+++ b/amethyst_core/src/lib.rs
@@ -15,6 +15,9 @@ extern crate serde;
 #[cfg(feature = "profiler")]
 extern crate thread_profiler;
 
+#[cfg(all(target_os = "emscripten", not(no_threading)))]
+compile_error!("the cfg flag \"no_threading\" is required when building for emscripten");
+
 //#[cfg(test)]
 //extern crate quickcheck;
 
--- a/src/app.rs
+++ b/src/app.rs
@@ -827,6 +827,7 @@ impl<'a, 'b, T> ApplicationBuilder<'a, 'b, T> {
         #[cfg(feature = "profiler")]
         profile_scope!("new");
 
+        #[cfg(not(no_threading))]
         let pool = self.world.read_resource::<Arc<ThreadPool>>().clone();
         let reader_id = self.world
             .write_resource::<EventChannel<Event>>()
@@ -837,7 +838,10 @@ impl<'a, 'b, T> ApplicationBuilder<'a, 'b, T> {
             // config: self.config,
             states: StateMachine::new(self.initial_state),
             events_reader_id: reader_id,
+            #[cfg(not(no_threading))]
             dispatcher: self.disp_builder.with_pool(pool).build(),
+            #[cfg(no_threading)]
+            dispatcher: self.disp_builder.build(),
             ignore_window_close: self.ignore_window_close,
         })
     }


Moxinilian:
If you just use parallel, there is no way to force amethyst on desktop to not use multi-threading.
If you don't plan on using single-threading, then you can just entirely ignore `serial` and it will work just fine. This simply is a feature in case this use case would ever show up.

torkleyy:
Hm, but who'd want to restrict it to one thread? And well if you want that you can still disable the default features. What you're doing here is emulating flavors, which doesn't quite work together with features' additive nature.

Moxinilian:
Honestly I don't really now.
I can remove it real quick, it doesn't matter much.

Moxinilian:
It is necessary if you want to have parallel by default on desktop.

torkleyy:
You can use conditional cargo default features (target.''.features I
think). Btw, you should not use cfgs in build.rs as compilation platform !=
target platform.

Moxinilian:
Are you sure about that? I looked for it but I couldn't find anything, and the only cargo issue I found abotu it implied it was not implemented yet.
Also, about cfg, this is interested because I did test multiple scenarios and it worked every single time.

build.rs
```diff
@@ -7,4 +7,16 @@ fn main() {
...
+    use std::env;
+    if let Ok(target) = env::var("TARGET") {
+        match target.as_ref() {
+            "asmjs-unknown-emscripten" => { },
+            "wasm32-unknown-emscripten" => { },
+            "wasm32-unknown-unknown" => { },
+            _ => println!("cargo:rustc-cfg=parallel"),
+        }
+    } else {
+        panic!("Could not identify the building target (missing TARGET env variable, are you sure you are running this from cargo?)");
```
jojolepro: If someone was running it without cargo, it shouldn't panic either.

build.rs
```diff
@@ -7,4 +7,16 @@ fn main() {
             err
         );
     }
+
+    use std::env;
+    if let Ok(target) = env::var("TARGET") {
+        match target.as_ref() {
+            "asmjs-unknown-emscripten" => { },
```
jojolepro: There's a syntax to avoid duplicating the => { },

build.rs
```diff
@@ -7,4 +7,16 @@ fn main() {
...
+            "wasm32-unknown-emscripten" => { },
+            "wasm32-unknown-unknown" => { },
+            _ => println!("cargo:rustc-cfg=parallel"),
```
jojolepro: I don't understand what this file does, except printing?

stenverbois: https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script

Moxinilian:
@torkleyy If we have parallel be simply a default feature, we miss the point entirely as this is meant to have building targetting the web actually work
@jojolepro cargo takes arguments returned by the build script from stdout

build.rs
Moxinilian: I felt the "|" syntax was inelegant. What syntax exactly do you have in mind?

build.rs
Moxinilian: This panic was meant for someone trying to run build.rs outside of cargo, which has no reason to happen except if it's a mistake from the user.

build.rs
jojolepro: Example: https://github.com/tomaka/cpal/blob/master/src/alsa/mod.rs#L94

build.rs
jojolepro: That's fine like this then ;)

Xaeroxe:
Hold on, I think with the code as is the Dispatcher will still use a thread pool it will just make its own instead of using ours.  Perhaps I misunderstand, but wasn't the issue that WASM cannot multi-thread?

Moxinilian:
yes but this PR is only about building

torkleyy:
@Xaeroxe More PRs will be required for wasm support, this is the
architectural start.

Regarding the code:
Please make sure it's called `no_threading` and print an error if it's not
enabled on wasm (using the error macro).

Xaeroxe:
@Moxinilian and I are no longer sure this project is worthwhile and would like to revert this PR.

Per https://github.com/kripken/emscripten/wiki/Pthreads-with-WebAssembly emscripten is going to be getting multi-threading support soon.  If this happens then there's no point in trying to move Amethyst to an optional single threaded architecture.

Moxinilian:
As a matter of fact, emscripten technically supports correctly pthreads on asmjs, but using facilities (temporarily?) disabled by browsers because of the Spectre attacks.

Xaeroxe:
@torkleyy @jojolepro can you weigh in on this?

jojolepro:
Its 7 lines of code, keep them for now and remove them once emscripten does support multi-threading. If we ever want to support a platform that ONLY works in single-thread, we'll be kinda forced to have this anyway (even if it only fixes 1% of the issue)

Issue #547, opened 2018-01-23
Resizing to expand window causes garbage to appear on Mac OS X


Hittherhod:
Discovered this while troubleshooting #544. If you expand the screen, it adds garbage to the windows. This happens on all window examples, not just with the UI, so I believe it is a separate issue. 

## Metadata
```
[amethyst::app][INFO] Version: v0.5.1-438-gd3b7d51
[amethyst::app][INFO] Platform: x86_64-apple-darwin
[amethyst::app][INFO] Amethyst git commit: d3b7d51f41e3eff2faea08acd14a4f04550d9364
[amethyst::app][INFO] Rustc version: 1.25.0-nightly Nightly
[amethyst::app][INFO] Rustc git commit: ae920dcc98c9b18b38aac03367f7f1cd6dce7d2d
```

## Screenshots
<img width="1125" alt="screen shot 2018-01-23 at 16 30 37" src="https://user-images.githubusercontent.com/1400222/35301902-d8d42908-005a-11e8-8511-110c0fbc00c1.png">

Hittherhod:
These seems related to https://github.com/tomaka/winit/issues/399

Rhuagh:
Looks very likely to be the issue yes. Retina displays usually don't have logical == physical pixels.

PR #648, opened 2018-04-16
fix: Update winit+glutin+gfx_window_*

### LINKED ISSUE #544: Pong Example Player 1 (Left Hand) ScoreText Does Not Appear
I'm using the Pong Example to learn some of the components of amethyst, but I'm confused as to why the left-side player (player 1)'s score isn't showing up. It has something to do with the coordinate system, which has me confused. The arena is configured to be 100 x 100, which is why I assume the constant "100" appears in pong.rs in the translation closures. However, the first closure evaluates to a negative number (because the ScreenDimensions are 0 at the time the `initialise_paddles()` is called #536). 

I've played around with the numbers, and it seems like the camera projects such that the visible x-coordinate values are 0-1000. I'm not super sure of the math behind it though, so I can't figure out a equation that uses knowledge of arena properly.

Rhuagh:
Fixes #544. 

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/648)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ rustc_version_runtime = "0.1"
 shred = "0.5"
 shrev = "0.8"
 specs = "0.10"
-winit = "0.11.3"
+winit = "0.12"
 
 thread_profiler = { version = "0.1", optional = true }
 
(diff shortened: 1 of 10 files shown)


jojolepro:
I'd like a confirmation from @Hittherhod that this fixes #547

Hittherhod:
@Rhuagh @jojolepro Nope. I'm still seeing issues upon resizing the pong example.
screen shot 2018-04-16 at 20 49 21

Rhuagh:
So it's not fixing that issue then, got it.

PR #652, opened 2018-04-18
Add temp seizure warning

Xaeroxe:
It's silly but until we get this fixed I feel an obligation.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/652)

### DIFF
--- a/examples/ui/main.rs
+++ b/examples/ui/main.rs
@@ -295,9 +295,16 @@ fn run() -> Result<(), amethyst::Error> {
 }
 
 fn main() {
-    if let Err(e) = run() {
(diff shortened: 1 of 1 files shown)


torkleyy:
Err why is that?

Xaeroxe:
@torkleyy the background color constantly flickers between red and white on my machine.

Can't reproduce it for @jojolepro but I can.  @jojolepro thinks it might be an issue with the parenting system but we're not sure.

torkleyy:
Would be nice to reproduce it, not that it's a platform specific bug.

jojolepro:
I doubt its the parenting system that causes the background issue.

PR #656, opened 2018-04-21
UI Event overlap support

jojolepro:
When you have two ui elements where the one under (z axis) is MouseReactive and the one on top is (or is not), the one on top should get the event checks and the one under should be ignored. Obviously, which one is picked depends on where you clicked.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/656)

### DIFF
--- a/amethyst_ui/src/event.rs
+++ b/amethyst_ui/src/event.rs
@@ -55,8 +55,8 @@ impl Component for MouseReactive {
 /// The generic types A and B represent the A and B generic parameter of the InputHandler<A,B>.
 pub struct UiMouseSystem<A, B> {
     was_down: bool,
-    old_pos: (f32, f32),
     click_started_on: Option<Entity>,
+    last_target: Option<Entity>,
     _marker1: PhantomData<A>,
     _marker2: PhantomData<B>,
 }
@@ -66,8 +66,8 @@ impl<A, B> UiMouseSystem<A, B> {
     pub fn new() -> Self {
         UiMouseSystem {
             was_down: false,
-            old_pos: (0.0, 0.0),
             click_started_on: None,
+            last_target: None,
             _marker1: PhantomData,
             _marker2: PhantomData,
         }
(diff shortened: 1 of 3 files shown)


jojolepro:
Just realized that something is broken, and fixing it proved to be more difficult than I thought.
I'll fix it tomorrow, time to :zzz:

jojolepro:
Ready for review :ok_hand:

amethyst_ui/src/event.rs
```diff
@@ -133,3 +147,28 @@ where
         self.was_down = down;
     }
 }
+
+fn targeted(
```
Xaeroxe: Not required to merge, but I think it should be possible to implement this as a `fold()` over the iterator rather than collecting and sorting a `Vec`.  Very small performance improvement though, which is why this is a note rather than a requested change.

jojolepro: https://github.com/rust-lang/rust/issues/38882

jojolepro:
If we want to support things like round ui elements and we want pixel perfect checks, we'll either need to raycast or to have pretty big algorithms.

amethyst_ui/src/event.rs
Xaeroxe: https://github.com/jojolepro/amethyst/pull/5

jojolepro:
Will squash once there's a second review. Thanks @Xaeroxe for the fold implementation ;)

Rhuagh:
Needs update after specs pr

amethyst_ui/src/event.rs
```diff
@@ -133,3 +138,31 @@ where
...
+    pos: (f32, f32),
+    transforms: JoinIter<(
```
Rhuagh: If you don't want to use the internal specs types here, you could make it something like:

```
fn targeted<'a, I>(pos: (f32, f32), transforms: I, react: &ReadStorage<MouseReactive>) 
where 
    I: Iterator<Item=(Entity, &'a UiTransform)> + 'a,
{
    ..
}
```

Rhuagh: That way you don't need `MaskedStorage` etc.

Xaeroxe: Fixed to use this in my rebase.

jojolepro:
Check the tests section. Its wasn't finished and I forgot it.

Xaeroxe:
I've fixed the tests

PR #659, opened 2018-04-21
feat: Implement traits needed to use Transform in rhusics

Rhuagh:
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/659)

### DIFF
--- a/amethyst_core/src/transform/components/local_transform.rs
+++ b/amethyst_core/src/transform/components/local_transform.rs
@@ -1,8 +1,8 @@
 //! Local transform component.
 
-use cgmath::{Array, Deg, ElementWise, EuclideanSpace, InnerSpace, Matrix3, Matrix4, One, Point3,
-             Quaternion, Rotation, Rotation3, SquareMatrix, Transform as CgTransform, Vector3,
-             Zero};
+use cgmath::{Array, Basis2, Deg, ElementWise, EuclideanSpace, Euler, InnerSpace, Matrix3, Matrix4,
+             One, Point2, Point3, Quaternion, Rotation, Rotation2, Rotation3, SquareMatrix,
+             Transform as CgTransform, Vector2, Vector3, Zero};
 use orientation::Orientation;
 use specs::{Component, DenseVecStorage, FlaggedStorage};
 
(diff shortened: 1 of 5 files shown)


Rhuagh:
If anyone have ideas on how to avoid this, feel free to shout.

Xaeroxe:
I'm of the opinion that things like this and `amethyst_animation` should be unconditionally compiled into the engine.  The use cases are common enough I don't feel like this is something we should have to "turn on" as end users.  All of the game engines I've used have this stuff included and on by default, sometimes it's not even possible to turn it off.

illis:
As an experiment I did try creating a new-type based on Transform from within amethyst-rhusics. Got stuck on figuring out how to pull the Point2 & Point3 traits in to the new type. I'm pretty new to rust thou, so maybe someone more knowledgeable might be able to figure it out.  

Attempted fix:
https://github.com/illis/amethyst-rhusics/tree/fix/use-amethyst-transform-as-pose

Rhuagh:
That's a separate discussion, I don't think an implementation for a rhusics specific trait should be inside the CORE crate if it doesn't have to.

jojolepro:
core as in amethyst::core or as in github.com/amethyst/amethyst ?
I'm talking about the later.

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -252,3 +264,60 @@ impl CgTransform<Point3<f32>> for Transform {
         }
     }
 }
+
+impl CgTransform<Point2<f32>> for Transform {
+    fn one() -> Self {
+        Default::default()
+    }
+
+    fn look_at(_eye: Point2<f32>, _center: Point2<f32>, _up: Vector2<f32>) -> Self {
```
jojolepro: Wouldn't that just be a rotation?
Also I'd prefer if _center was called target or something, if this method could be made to rotate.
I guess you'd need a forward axis too.

Also, can't you use compile_error! for that?

Example code:
```
float rot_z = Mathf.Atan2(diff.y, diff.x) * Mathf.Rad2Deg;
transform.rotation = Quaternion.Euler(0f, 0f, rot_z - 90);
```

jojolepro: What I'm saying is that you can compute a look_at for 2d. You just need to set the up vector to the perpendicular of the direction vector.

Rhuagh:
I agree with @torkleyy here, I'm fine with having an integration crate inside the amethyst organisation, but I don't think we need to put it in the same repository. Tbh I think our repository is too big as it is.

amethyst_core/src/transform/components/local_transform.rs
Rhuagh: What I'm getting at is that it makes very little sense to have a look_at for 2D. Yes, you could make it just apply a rotation based on the up vector. I just felt like it wasn't really gonna be used by anyone, since there are better functions that apply rotation directly that are much more obvious.

jojolepro:
So you are against having a amethyst_physics crate that is conditionally compiled?

Rhuagh:
What's the motivation for having all crates in a single repository?

torkleyy:
If we keep adding more and more crates to this repo, the build times will increase more and more. There's nothing complex about adding `amethyst_rhusics` to your `Cargo.toml`.

jojolepro:
That's what I'm saying. Just a simple wrapper like this one: https://github.com/jojolepro/amethyst/tree/locale/amethyst_locale/src
makes it easier to integrate stuff together.

Also, if I want to use a fork of amethyst for my game, I also have to fork amethyst-rhusics to change the cargo.toml file to avoid incompatibilities between amethyst versions.

Rhuagh:
And it's something that we can easily add to the cli aswell, which is way more appropriate imho.

Rhuagh:
I've worked with these behemoth like projects before, where everything is bundled together, and let's just say, none of them are alive today.

jojolepro:
I give up x)

jojolepro:
Any comments before the merge?
@Xaeroxe ?

Xaeroxe:
None, go ahead

jojolepro:
bors r=Jojolepro,Xaeroxe,torkleyy

jojolepro:
bors retry

Rhuagh:
Bors r+

Rhuagh:
@illis i just realised the newtype idea won't work, because it's still a new type wrt to `specs::World`, so rhusics and amethyst will still lookup different types, and hence need a sync between the components.

illis:
@Rhuagh Bugger, I completely overlooked that. Wonder if there is a way to make it auto down/up cast, as the base struct hasn't changed at all.

Rhuagh:
No, not without doubling the bounds on the systems in rhusics, which I'm not willing to do

PR #660, opened 2018-04-22
Refactor local transform

derekdreery:
I've refactored the `LocalTransform` to be more like what I would expect. I don't know if this change is correct, but hopefully if there is criticizm of it then that will help me understand the function of LocalTransform in amethyst.


amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -21,124 +20,186 @@ pub struct Transform {
...
+    pub fn look_at(&mut self, up: Vector3<f32>, position: Point3<f32>) -> &mut Self {
+        self.rotation = Quaternion::look_at(position - Point3::from_vec(self.translation), up);
         self
     }
 
     /// Returns the local object matrix for the transform.
     ///
     /// Combined with the parent's `GlobalTransform` component it gives
     /// the global (or world) matrix for the current entity.
+    // this is a hot function
```
jojolepro: First letter uppercase please x)
Also it should either be /// or it should be inside of the function //.

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -21,124 +20,186 @@ pub struct Transform {
...
-        }
+    pub fn move_local(&mut self, translation: Vector3<f32>) -> &mut Self {
+        self.translation += self.rotation * translation;
+        self
+    }
 
-        let delta = Quaternion::from(self.rotation) * axis.normalize() * amount;
+    /// Move a distance along an axis.
+    ///
+    /// Will not move in the case where the axis is zero, for any distance.
```
jojolepro: It will

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -21,124 +20,186 @@ pub struct Transform {
...
+        }
         self
     }
 
     /// Move forward relative to current position and orientation.
-    pub fn move_forward(&mut self, orientation: &Orientation, amount: f32) -> &mut Self {
-        self.move_local(orientation.forward.into(), amount)
+    #[inline]
+    pub fn move_forward(&mut self, amount: f32) -> &mut Self {
+        // sign is reversed because z comes towards us
```
jojolepro: I thought it was the inverse?
Anyway, after this change the fly_camera forward would be inverted.

jojolepro:
Looks pretty good. It will be way easier without Orientation in the way.

derekdreery:
Can always add Orientation back in if it is useful for something.

amethyst_core/src/transform/components/local_transform.rs
derekdreery: I matched the OpenGL orientation (in the left-hand right-hand sense). I should check the shaders to see which orientation we use. If you run the fly camera example now it should all work -> if it doesn't that's a bug in my PR.

jojolepro:
is there a way to get the forward/right/up vectors?

derekdreery:
So assume the default view is the following square (it is for GL, it's flipped in y axis for vulkan). The up of the object can be find by rotating the 'up' of the original system (`(0, 1, 0)` in GL) using the transform rotation. forward is `(0, 0, 1)` and right is `(1, 0, 0)`.

However I need to work out the coordinate system that amethyst uses or I might have the y and z back to front.
            y
┏━━━━━┯━━━━━┓
┃          1          ┃
┃         │         ┃
┃-1  ─│──1┃x
┃         │         ┃
┃         -1          ┃
┗━━━━━┷━━━━━┛

derekdreery:
@jojolepro I added a method for getting the information you requested, since it probably reduces cognitive load.

JadeSnail:
I disagree with the removal of Orientation in the _global and _local pitch/roll/yaw methods.

The point is to allow the user to work with a different Orientation. The default orientation you chose might make sense for some obscure computer graphics reason but I don't know about it, and I should'nt need to know about it.

I like to work with the {forward: [0,1,0], right: [1,0,0], up: [0,0,1]} Orientation. With the current methods, I can just write down this orientation somewhere, and then call pitch/yaw/roll with it. With the proposed methods, I'm forced to understand the nasty Orientation you're working with, and it isn't even written anywhere.

derekdreery:
@JadeSnail Can you explain how the `Orientation` works? Is it to specify what coordinate system you use? If so, what does the current pitch_local method mean, for example? Does Orientation need to be a full linear operation, or could it be a small enum (XYZ, XZY, etc.) I think there would be 6 ^ 2 variants, including negation.

JadeSnail:
For pitch/roll/yaw to make sense, you have to specify three orthogonal directions. It might be good to also require that the third is the vector product of the first two. That's what I think an Orientation should be.

derekdreery:
So I guess what I was thinking wast that if you want to pitch/yaw/roll you are either going to do it w.r.t. the global axes, or the local model-specific axes. Are there times where you would want to rotate around another axis?

derekdreery:
But when you load a 3D model, you get coordinates for the vertices. These numbers are relative to an origin of some kind.

derekdreery:
Do you have some example code I can look at using the existing system, so I understand how it's used?

JadeSnail:
Let's say I've got a plane in 3D space. No transform has been applied to it, and it looks just like a plane usually looks, that is the «top» of the plane is in the z direction.

Now I want to yaw this plane. As what yawing usually refers to for plane. The usual yaw.

Well with what you propose, calling either yaw_global, or yaw_local doesn't yaw the plane.

What does yaw the plane is calling the current yaw_global, or yaw_local, with the {forward: [0,1,0], right: [1,0,0], up: [0,0,1]} Orientation.

Xaeroxe:
{forward: [0,1,0], right: [1,0,0], up: [0,0,1]}

Then this orientation should be a constant.

Xaeroxe:
i.e. not provided in a function argument, but implied.

derekdreery:
Could you have a TransformedTransform, or an OrientedTransform that works like this?

JadeSnail:
> {forward: [0,1,0], right: [1,0,0], up: [0,0,1]}
>Then this orientation should be a constant.

I'm not sure that's a good idea: what if someone likes to use a different orientation, such as {forward: [1,0,0], right: [0,-1,0], up: [0,0,1]}. Depending on wether the plane in my example has his wings along the x axis or the y-axis, to roll the plane you have to call the rolll method with one or the other of these orientations.
It can also make sense to work with the proposed default Orientation  forward: -Vector3::unit_z(),right: Vector3::unit_x(),up: Vector3::unit_y(). Altough slightly less intuitive, working with it might be more efficient in some cases since it's the one used under the hood.

> EDIT alternatively just have a parent transform that sets the coordinate system how you want it.

I think I like this idea. Maybe have a Reorient component to which you provide your Orienation and that you add as a parent.

It precludes you from working with two different Orientations at the same time, but what kind of madman would do that.

minecrawler:
Thanks for the refactoring. I am really fighting with the math, so improvements are always good :) As the one who introduced `Orientation` in the first place, I did so, because:

- I didn't like the default orientation used under the hood, so I wanted to make it configurable
- the idea was to have one `Orientation` as a resource in the world, which then could be changed and used
- However I did not want to use that resource in the methods internally, as there might be the need to do calculations with a different orientation at some point for some obscure game feature (mirror worlds etc.). Imho, passing the orientation does not necessarily make the usage a lot more complex, once you wrap your head around the idea that you need to define up, left and right to do useful transformations.

I think, using a parent transform, which sets the orientation, might make the creation of entities more painful, but the translation easier in most cases, but imho overall does not make stuff a lot simpler. Your opinion might be different, though. It might affect performance and memory usage though, if we set the orientation per entity. It would use more memory, but be more performant.

One idea I would like to pitch in is: Maybe we could create a helper instead, which does automatically use the `Orientation` resource, so if you have no use-case for varying orientations, you might just use the helper and not worry about orientation anymore.

torkleyy:
I think a constant is enough for most cases, there's no need to configure
it at runtime.

Xaeroxe:
@minecrawler What didn't you like about the default orientation?  Would the vulkan coordinate system satisfy you?

derekdreery:
So I've been reading some and I know some more now.

Swapping the direction of any dimensions is easy - you just negate that value in the projection matrix.

Swapping the position of 2 dimensions is a pretty simple re-arranging of values and negations. It should certainly not be done as a general linear transformation as this is a LOT of operations. The questions is: if someone wants to use a coordinate system that's not the default vulkan one (x goes right, y goes down, z goes away), where should they specify it. I honestly think most people will use this coordinate system, so it's not worth adding complexity to the basic transform. OTOH people who want to work in a different coordinate system should be able to.

I think that the coordinate system used should be restricted to reordering and negating axes, since this is compact (an enum with 36 values, smaller than a pointer) and quick to apply. It should definitely not be a general linear transformation (9 * 32bit = 9 * 4 byte = 36 byte!), this is a lot of extra data on an entity and will kill the cache. When you pitch/yaw/roll you do it related to either the orientation of the object (local) or its parent (global). If you want to do something else there should be a utility to make this easy, but that does not impact the component.

Xaeroxe:
# My proposed compromise

I still think any configurability is excessive and overkill for this.  However I would not be opposed to a resource being used for this, which is initialized by default with the typical Vulkan system we've been discussing.

This way no one has to configure this, they can just use the defaults, but if for some reason you really need that to be different then you can write code to modify this resource.

Rather than having such a fat enum though I'd prefer it if the resource was just an instance of the `Orientation` structure we have now.

# Explaining why I feel the way I do.

Now what I really want to know is **why** people feel such flexibility is necessary.  What Orientation would you personally use as the default?  Are there mathematical branches you can show us that just don't work with a particular orientation?  I hear a lot of "well maybe someone else will want different", and my response to that is two questions

- Why would they want something different beyond mere arbitrary personal preference?  I want mathematical justification.
- Why couldn't they just live with it?  What can't they do that configurability here would enable them to?

Personally I feel like configurability here is big failure to follow the KISS principle.

jojolepro:
https://math.stackexchange.com/questions/1817242/why-do-we-need-basis-in-linear-algebra
I think having parenting defeats the arguments of this though.. ^

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -35,110 +34,179 @@ impl Transform {
     /// the global (or world) matrix for the current entity.
```
jojolepro: First letter uppercase please

amethyst_core/src/transform/components/local_transform.rs
jojolepro: :+1:

Xaeroxe:
Right now it doesn't. I approved it because I like the approach anyways, but I'd approve it again if we moved to the compromise

derekdreery:
Would the Orientation be a per-entity component (and only be applied when that component is present) or a resource applied to all entities with a transform? What about global transforms? When should the orientation be applied?

Xaeroxe:
I'd make it a global resource, anything else is too much overhead imo

jojolepro:
I personally won't use orientation at all so I don't really care.
Orientation would be global, as having a parenting system is equivalent to the benefits having orientation per entity provides.

derekdreery:
@jojolepro @Xaeroxe got you.

derekdreery:
I need to think about how this applies to all the math.

derekdreery:
Ok so I have looked at how amethyst generates the projection matrix. It just uses cgmath, that copies [glFrustrum] for perspective, and [glOrtho] for orthographic, which use the convention of GL where it reverses the z direction. So, in essence the current coordinate system is: x -> right, y -> up, z -> towards.

I think I'd recommend keeping this coordinate system for now, with the option to request a different one that I will add to this PR. It will be interesting to see whether people use this system with a shim, or the new Vulkan one. I think we should see what happens.

[glFrustrum]: https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glFrustum.xml
[glOrtho]: https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glOrtho.xml

JadeSnail:
I think this version is a lot harder to understand than the previous one. I'd guess it'd be harder to use too, but since it isn't clear what it should be used for, it's hard to tell.

I don't think the new found robustness warrants the code complexity and the lack of flexibiliy.

I don't think the amethyst user should have to manually deal with Matrix4 but rather only with the Transform components and methods. Therefore I think you should only be able to get a Quaternion, and not a Matrix4 from an Orientation (which is only possible for a subset of your proposed Orientations).

Here's an example of what the Orientation could be used for: I want to make my player look in a direction. The issue is that setting its Transform rotation to Quaternion.look_at(direction, up) doesn't work. It only works if in my player model, the player looks at Orientation::default().forward. Somehow it would be nice if I could set my chosen Orientation (which defines in what direction my player model is looking) somewhere and have look_at (or some variant) work like expected. This should only be possible if you can get a Quaternion from an Orientation.

Another natural use case for a more flexible Orientation is to keep an Orientation struct (computed from the Transform.rotation) with my player at all times. It allows easy access to its forward/right/up vectors which are constantly used.

Xaeroxe:
Alright yeah my compromise was evidently not the greatest idea.

@derekdreery we can revert the last commit on this PR.  Sorry for all the trouble, I'll try and think my ideas through a bit better in the future.

> It only works if in my player model, the player looks at Orientation::default().forward.

This is actually pretty standard fair in game dev in my experience.  Usually the model is rotated rather than messing up all the coordinate math in the game.  That's what I'd rather do personally.

One thing we could do in Amethyst is on the mesh itself we could have a Quaternion that rotates the **display** of the model but doesn't actually mess with the transform of the entity.

I'm of the opinion that would be the more correct approach to resolving this problem, because then we don't need a special Orientation based on which entity we're transforming.

derekdreery:
For what you are suggesting, you would not use this orientation. A typical user who wanted a custom orientation (axes labelling) would just create one and add it as a resource. Then the graphics engine would apply that orientation to everything. If you want something that's not global you would have a parent transform.

derekdreery:
I need to read this properly when I get home - I'll delete this comment and write another then

jojolepro:
I don't think having a mismatch between the graphics and the actual positioning is a good idea.

Xaeroxe:
@jojolepro can you elaborate on why you don't think that's a good idea?  Keep in mind what I'm proposing is just a rotation, to allow the programmer to fix poorly oriented art.

jojolepro:
If that's only what this is about, then the programmer can just rotate the model after adding it to the scene, or you can just properly export it from the 3d editor.

>One thing we could do in Amethyst is on the mesh itself we could have a Quaternion that rotates the display of the model but doesn't actually mess with the transform of the entity.

If you do that, mesh colliders aren't going to be the same as what's visible.

Xaeroxe:
> rotate the model after adding it to the scene

I'm trying to avoid this because it means code needs special casing for transformations.

> you can just properly export it from the 3d editor.

I agree that would be ideal.  This functionality would be exclusively for less than ideal situations.

However you do bring up a good point about the colliders.  So perhaps we should just leave it, and say "That's not a problem with our setup, it's a problem with yours, you fix it."

Rhuagh:
I think it should be the game developer or artist to make sure all assets have the same coordinate system

jojolepro:
You could make per-entity coordinate systems, but you'd need to actually change the transform and take it into account in all calculations. It would be confusing as hell and prone to errors when, for example, translating all entities in a separate system, but some of them actually translate the other way.

Xaeroxe:
@torkleyy Can you weigh in here?  I'd be happy with the transform system if we reverted the last commit then merged this.  If you agree with me can you 👍 this, and if you disagree 👎 this?

If we can get a consensus from at least 4 core team members I say we should merge this.

torkleyy:
I preferred the PR before the last commit, yes. I'm not sure though what this PR is improving though. Could you revert the commit @derekdreery and explain the motivation?

Xaeroxe:
So the biggest reason I like it is we get rid of the `Orientation` parameter and make it implicit to

{forward: [0,1,0], right: [1,0,0], up: [0,0,1]}

Specifying that orientation for every transformation feels redundant, and to me it also feels like the wrong solution to whatever problem we're trying to solve.  Some of the proposed problems seem to come down to
- Art is misaligned (fix your art)
- I don't like the default coordinate system Amethyst is using (Then let's fix that by discussing and possibly changing the coordinate system.)

Additionally requiring this parameter makes the purpose of it unclear.  I for one originally thought that the orientation needed to be computed from the rotation every time a transform function was called.  I can see now that wasn't the intended usage.

The whole API seems to be much more clear to me as it follows the conventions I'm used to much better.

derekdreery:
I'm perfectly happy to work in OpenGL coordinate space.

Xaeroxe:
@derekdreery oh eh it seems I was unclear, we don't need to remove `Orientation` entirely, as I quite like your `orientation()` function.  I just didn't care for the idea of using `Orientation` as in input parameter.

derekdreery:
I'm just dumping some work here while I sort out my rebasing.
```
    /// Makes the entity point towards `position`.
    ///
    /// `up` says which direction the entity should be 'rolled' to once it is pointing at
    /// `position`. If `up` is parallel to the direction the entity is looking, the result will be
    /// garbage.
    ///
    /// This function only works with respect to the coordinate system of its parent, so when used
    /// with an object that's not a sibling it will not do what you expect.
    ///
    /// # Examples
    ///
    /// ```
    /// # use amethyst_core::transform::components::Transform;
    /// # use amethyst_core::cgmath::{Quaternion, One, Vector3, Point3, Matrix3};
    /// let mut t = Transform::default();
    /// // No rotation by default
    /// assert_eq!(t.rotation, Quaternion::one());
    /// // look up with up pointing backwards
    /// t.look_at(Point3::new(0.0, 1.0, 0.0), Vector3::new(0.0, 0.0, 1.0));
    /// // our rotation should match the arc from straight ahead to straight up
    /// let rotation = Quaternion::from_arc(
    ///     Vector3::new(0.0, 0.0, -1.0),
    ///     Vector3::new(0.0, 1.0, 0.0),
    ///     None);
    /// assert_eq!(t.rotation, rotation, "{:?}, {:?}", Matrix3::from(t.rotation),
    /// Matrix3::from(rotation));
    /// ```
    #[inline]
    pub fn look_at(&mut self, position: Point3<f32>, up: Vector3<f32>) -> &mut Self {
        self.rotation = Quaternion::look_at(Point3::from_vec(self.translation) - position, up);
        debug_assert!(self.rotation.s.is_finite()
                      && self.rotation.v.x.is_finite()
                      && self.rotation.v.y.is_finite()
                      && self.rotation.v.z.is_finite());
        self
    }
```

derekdreery:
I've squashed everything, but I need to add more examples and docs before it's merge ready.

Xaeroxe:
@derekdreery haha yeah I was trying to sort out these merge conflicts myself because I felt bad about yanking your chain around.  You beat me to it.  Nice job!

derekdreery:
> {forward: [0,0,1], right: [1,0,0], up: [0,1,0]}
thought it was {forward: [0,0,-1]}

Xaeroxe:
forward being positive made sense to me as it was consistent with the other two, but tbh I don't really care.

amethyst_core/src/transform/components/local_transform.rs
Rhuagh: Not fixed.

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -35,110 +63,179 @@ impl Transform {
...
+    /// Move a distance along an axis.
+    ///
+    /// It will not move in the case where the axis is zero, for any distance.
     #[inline]
-    pub fn move_local(&mut self, axis: Vector3<f32>, amount: f32) -> &mut Self {
-        if axis.magnitude2() == 0.0 {
-            return self;
+    pub fn move_along_global(&mut self, direction: Vector3<f32>, distance: f32) -> &mut Self {
+        if !ulps_eq!(direction, Zero::zero()) {
+            self.translation += direction.normalize() * distance;
```
Rhuagh: You can use `direction.normalize_to(distance)`

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -35,110 +63,179 @@ impl Transform {
...
-        let delta = Quaternion::from(self.rotation) * axis.normalize() * amount;
-
-        self.translation = self.translation + delta;
+    /// Move a distance along an axis.
+    ///
+    /// It will not move in the case where the axis is zero, for any distance.
+    #[inline]
+    pub fn move_along_local(&mut self, direction: Vector3<f32>, distance: f32) -> &mut Self {
+        if !ulps_eq!(direction, Zero::zero()) {
+            self.translation += self.rotation * direction.normalize() * distance;
```
Rhuagh: See above.

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -148,33 +245,11 @@ impl Transform {
     }
 
     /// Set the rotation using Euler x, y, z.
-    pub fn set_rotation<D: Into<Deg<f32>>>(&mut self, x: D, y: D, z: D) -> &mut Self {
-        let rotation = Quaternion::from_angle_x(x.into()) * Quaternion::from_angle_y(y.into())
-            * Quaternion::from_angle_z(z.into());
-
-        self.rotation = rotation.into();
+    pub fn set_rotation<A: Into<Rad<f32>>>(&mut self, x: A, y: A, z: A) -> &mut Self {
```
Rhuagh: Should we take `cgmath::Euler` instead to type enforce that it should be Euler angles?

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -148,33 +245,11 @@ impl Transform {
...
+        self.rotation = Quaternion::from_angle_x(x) * Quaternion::from_angle_y(y)
+            * Quaternion::from_angle_z(z);
         self
     }
-
-    /// Calculate the view matrix from the given data.
-    pub fn to_view_matrix(&self, orientation: &Orientation) -> Matrix4<f32> {
```
Rhuagh: I think we should provide a `view_matrix()` function. It's the inverse of the normal matrix, but should be possible to calculate directly without doing the inverse explicitly.

derekdreery:
Ideally I'd like to add examples to all methods, but I'll do that in a separate PR, just want this one to land now.

amethyst_core/src/transform/components/local_transform.rs
derekdreery: I don't know what the correct answer is here. Do you still get the 'expected' result if you use angles from outside the expected ranges?

amethyst_core/src/transform/components/local_transform.rs
derekdreery: Ok

Rhuagh: You can defer this point to a followup PR.

amethyst_core/src/transform/components/local_transform.rs
Rhuagh: The problem with the current solution is that it suffers from gimbal lock, because the rotations are applied in sequence. Using the Euler conversion formula in cgmath is likely to not suffer from this problem. See https://docs.rs/cgmath/0.16.1/src/cgmath/quaternion.rs.html#280.

derekdreery: You've convinced me :)

derekdreery:
Can anyone help with the multiplication implementation?

Rhuagh:
Look at concat in the impl cgmath::Transform

derekdreery:
@Rhuagh I used that initially, but the matrices didn't match in my test.

Rhuagh:
Make sure the order is correct

JadeSnail:
As far as I can tell, you cannot implement a multiplication of Transforms.

Currently the amethyst transform can only represent a subset of all affine transformations, and that subset isn't closed under multiplication.

derekdreery:
I think this is worth doing -> multiplying 2 transforms should be much faster than multiplying the corrsponding matrices.

derekdreery:
@JadeSnail That would scupper my plans :P. Can you give a counterexample?

Rhuagh:
Sure you can implement multiplication for Transforms, they might not behave as you expect if you do non uniform scaling, but multiplying transforms is exactly what a scene graph does

Rhuagh:
A transform in amethyst is just a different way of representing a Matrix4

Rhuagh:
Check the mul implementation of cgmath::Decomposed

JadeSnail:
As you say, any multiplication implementation will only work for Transforms with uniform scaling, which means that all the 3 components of the Transform.scale are equal.

Rhuagh:
I'm in favor of only allowing uniform scaling on `Transform`. If people really need it they have `GlobalTransform`

Rhuagh:
If we're removing the non-uniform scaling we should consider just using `Decomposed` directly.

derekdreery:
I'll implement this to see what it looks like. I can always revert it.

JadeSnail:
> I'm in favor of only allowing uniform scaling on Transform. If people really need it they have GlobalTransform

Using GlobalTransform instead of Transform precludes you from using the Parent system, doesn't it ?

I don't really care either way.

Rhuagh:
It does, yeah, but I'm ok with that. I don't think non-uniform scaling is used very often, so it should be fine<tm>.

derekdreery:
> For example, you can't animate a lightsaber lighting up, or anything that's growing in only one dimension.

I'm no expert in this, but for me I don't think that's the way to animate a lightsaber. I would do volume rendering and just change the dimensions of the box it's in over time.

If it turned out that non-uniform scaling was useful for something, then could it be a separate component, applied after all other transforms, so it is independent of them?

derekdreery:
Changing to using `Decomposed` allows some neatening up. Wondering what people think?

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -1,310 +1,447 @@
...
 
     fn concat(&self, other: &Self) -> Self {
-        Self {
-            scale: self.scale.mul_element_wise(other.scale),
-            rotation: self.rotation * other.rotation,
-            translation: self.rotation
-                .rotate_vector(other.translation.mul_element_wise(self.scale))
-                + self.translation,
+        <Self as CgTransform<Point3<f32>>>::concat(self, other)
+            /*
```
Xaeroxe: Please remove commented out code

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -1,310 +1,447 @@
...
+            scale: self.0.scale * other.0.scale,
+            rotation: self.0.rot * other.0.rot,
+            translation: self.0.rot.rotate_vector(other.0.disp * self.0.scale) + self.disp,
         }
+        */
     }
 
     fn inverse_transform(&self) -> Option<Self> {
+        <Self as CgTransform<Point3<f32>>>::inverse_transform(self)
+            /*
```
Xaeroxe: Same here

Xaeroxe:
Side note: Couldn't multiplication of non-uniform scaling simply be implemented as

```
parent.x * child.x
parent.y * child.y
parent.z * child.z
```

That's how I'd expect that to behave at least.

amethyst_core/src/transform/components/local_transform.rs
derekdreery: oops

Rhuagh:
Doesn't work because scaling affect both rotation and translation

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -1,310 +1,447 @@
...
-    /// Scale vector [x, y, z]
-    pub scale: Vector3<f32>,
-    /// Translation/position vector [x, y, z]
-    pub translation: Vector3<f32>,
-}
+///
+/// The transforms are preformed in this order: scale, then rotation, then translation,
+/// or in maths `world_vertex = translation * rotation * scale * model_vertex`
+#[derive(Copy, Clone, Debug)]
+pub struct Transform(Decomposed<Vector3<f32>, Quaternion<f32>>);
```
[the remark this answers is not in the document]
derekdreery: There's definitely an argument for this. Is repr(transparent) stable yet?

Xaeroxe:
@Rhuagh I can imagine that maybe we could create the unscaled transformation then multiply the diff on each axis by the scaled amount, but my limited mathematical experience probably means I'm way off trajectory.

amethyst_core/src/transform/components/local_transform.rs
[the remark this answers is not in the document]
derekdreery: Or maybe impl Deref?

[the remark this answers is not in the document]
Xaeroxe: The only problem here is that with this approach our `Component` impl would break the orphan rule.

Rhuagh:
The problem is the parenting system really, it would be fine if it was a single transformation, but when you apply potentially a large amount of combines it becomes incredibly complex.

amethyst_core/src/transform/components/local_transform.rs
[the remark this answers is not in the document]
Xaeroxe: @derekdreery Using `Deref` as a form of inheritance is typically considered an anti-pattern, but other than that I can't really give a solid reason why we shouldn't do this.

Rhuagh:
What I meant was `pub type Transform = Decomposed<.. >`

derekdreery:
@Rhuagh can you add methods to a type alias?

Rhuagh:
With a trait yes

derekdreery:
I'll finish this version, then do one with an alias and then both will be in the history.

Xaeroxe:
This still breaks the orphan rule.  A type alias won't change that.

derekdreery:
Right I've fixed all the exmaples, so when it (maybe) passes CI, there's just the philosophical questions to answer.

Xaeroxe:
http://smallcultfollowing.com/babysteps/blog/2015/01/14/little-orphan-impls/

This may prevent us from implementing component for the transform.

Rhuagh:
Ah, right, Component would be a problem, crap

Rhuagh:
So we can't use Decomposed directly, we need to wrap it in a newtype.

Rhuagh:
And if that's the case I'm not sure how much value we actually get out of `Decomposed` to begin with :/

Rhuagh:
Btw, why do we need `Mul` implemented for the transform ?

Rhuagh:
@JadeSnail yes, there are only 9 parameters in the amethyst Transform, but that's because it's decomposed, there's no loss of information compared to a matrix. A transformation matrix has a certain redundancy of information that can be derived from the decomposed representation.

Rhuagh:
And the parenting system computes matrices for the combined transforms, so the end result should be accurate.

derekdreery:
Reverted to old transform.

JadeSnail:
To be clear.

@Rhuagh When you start composing Transforms, which the parenting system does, you can get any affine transformation of the space (what you refer to as «matrix»: an affine transformation is actually given by a 3by3 matrix and a vector in R^3, these two things are then usually put in a 4by4 matrix, where the last line is (0,0,0,1))

As I said, an affine transformation has 12 parameters, which means that not every affine transformation can be represented by a Transform.

Nonetheless, what the parenting system does is cast the Transforms as affine transformations (using the to matrix4 method) and then multiply these affine transformations. The resulting affine transformation need not be representable by a Transform (though it is, if the two multiplied transforms had uniform scaling).

Rhuagh:
True. Which is why I asked specifically why we'd want a `Mul` for `Transform`. I don't see a real use case for it.

JadeSnail:
Me neither !

derekdreery:
I've removed Mul, since it's not really multiply any more (if you use non-uniform scaling).

Rhuagh:
LGTM!

derekdreery:
This is ready for review

Rhuagh:
bors r=Xaeroxe,Rhuagh,jojolepro

PR #661, opened 2018-04-23
Refactor Z order evaluation to use text cache more effectively

Xaeroxe:
This PR changes UI rendering to use OpenGL depth evaluation instead of the order of operations for drawing.

This has a few benefits
- Sorting the entities is no longer necessary
- Text cache from gfx_glyph is much more efficient because we can batch the draw calls for text now.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/661)

### DIFF
--- a/amethyst_ui/Cargo.toml
+++ b/amethyst_ui/Cargo.toml
@@ -22,6 +22,7 @@ fnv = "1"
 gfx = { version = "0.17", features = ["serialize"] }
 gfx_glyph = "0.9.0"
 hibitset = "0.3"
+log = "0.4"
 rusttype = "0.4"
 shred = "0.5"
 shred-derive = "0.3"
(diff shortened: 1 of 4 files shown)


Xaeroxe:
DO NOT MERGE

Initial testing shows this may have not been a zero impact change.  Requires further testing to make sure the rendered UI is correct.

Xaeroxe:
I'm no longer certain if this approach is viable or not as it messes with the transparency in rendering.  I'll keep playing with it though.

Rhuagh:
Transparent surfaces must be drawn ordered, there is no way around that problem.

Xaeroxe:
I have other ideas for how to improve cache usage but you're right @Rhuagh this approach isn't viable.

PR #671, opened 2018-04-25
Add separate demo for mulisampling

MrMinimal:
Because the multisampling setting can break the demos on somem systems, this creates a demo dedicated to this feature.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/671)


Xaeroxe:
I'm mostly curious what system was broken by that.

Rhuagh:
I think this is a case where we should investigate if we can handle this more gracefully inside the engine, and downgrade the multisampling if the target device don't support it.

examples/sphere_multisample/main.rs
```diff
@@ -0,0 +1,159 @@
+//! Displays a shaded sphere to the user.
```
jojolepro: shaded sphere with multisampling enabled

examples/sphere_multisample/resources/display_config.ron
```diff
@@ -0,0 +1,10 @@
+(
+  dimensions: None,
+  max_dimensions: None,
+  min_dimensions: None,
+  fullscreen: false,
+  multisampling: 1,
+  title: "Sphere example",
```
jojolepro: Wrong title

examples/sphere_multisample/resources/display_config.ron
```diff
@@ -0,0 +1,10 @@
+(
+  dimensions: None,
+  max_dimensions: None,
+  min_dimensions: None,
+  fullscreen: false,
+  multisampling: 1,
```
jojolepro: Can you set the multisampling higher? I think we can have values of 2,4,8,16

jojolepro:
Thanks for that! :)

MrMinimal:
@jojolepro Changes made! Thanks for the review

derekdreery:
I agree about the graceful downgrade. I'm currently having a similar problem running pong because `cpal` is panicking. It should return an error so that the example can warn and then run without sound.

Rhuagh:
Seems like the suggested solution is to not write directly to the framebuffer, but write to a render target with multisampling defined and then just blit to the main framebuffer at the end.

I suggest we defer that change until the new renderer is in place.

Rhuagh:
As a side note, Hal does not support multisampling on backbuffers, so the only way to get support there is using the method I described above.

Rhuagh:
Bors r+

PR #674, opened 2018-04-26
Upgrade to specs alpha4

Xaeroxe:
alpha4 is being treated as a patch version, so this will need to merge as soon as it compiles on CI.  Until we do this every build of the develop branch is going to fail.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/674)


Rhuagh:
Aha, I'll need to fix my incoming pr then because I did this in there :)

Xaeroxe:
@Rhuagh woops, sorry :P

torkleyy:
Make sure to lock the version with =0.11-alpha4 so it doesn't break again.

PR #679, opened 2018-04-28
fix: Background music did not work because of a incorrectly setup Sys…

Rhuagh:
…temData

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/679)

### DIFF
--- a/amethyst_animation/src/systems/control.rs
+++ b/amethyst_animation/src/systems/control.rs
@@ -3,14 +3,14 @@ use std::marker;
 use std::time::Duration;
 
 use amethyst_assets::{AssetStorage, Handle};
-use amethyst_core::specs::prelude::{Component, Entities, Entity, Join, Read, ReadStorage, Resources,
-                                    System, SystemData, WriteStorage};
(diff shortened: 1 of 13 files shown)


derekdreery:
FWIW this patch fixes the background music issue on my machine (linux) too.

Rhuagh:
@torkleyy review addressed.

Rhuagh:
Yep, I think some of us are still using rustfmt-nightly. We should be using stable rust fmt preview.

jojolepro:
Currently using
>rustup component add rustfmt-preview

Rhuagh:
Bors r=Xaeroxe,jojolepro

Rhuagh:
I think it's some of our newer contributors :)

derekdreery:
Could rustfmt be put onto the CI? I think there is a switch that makes it issue non-zero exit code if it would have made changes.

jojolepro:
We already had a discussion on it and it was mostly a "no", because it makes it harder to create PR and to get them merged.

PR #682, opened 2018-04-29
Proposal to fix CI breakage on nightly

derekdreery:
This patch adds a script in src/bin/ci.rs that runs all the required tests. It's good because it provides a single place to edit how tests are run, as well as making it easier to have custom logic in the testing.

It currently just runs the tests that were being run and dumps the output if it fails (and exits with non-zero exit status).

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/682)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -43,5 +43,4 @@ build: false
 
 # Generate documentation, compile the engine, run tests.
 test_script:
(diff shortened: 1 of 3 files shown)


Rhuagh:
The change in cargo has been reverted, so I'm closing this.

PR #684, opened 2018-04-30
Use contains() where appropriate

Xaeroxe:
A new function was added to specs, this PR uses that function.

Low priority, but also pretty easy to review.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/684)

### DIFF
--- a/amethyst_animation/src/util.rs
+++ b/amethyst_animation/src/util.rs
@@ -20,11 +20,7 @@ where
     I: Send + Sync + 'static,
     T: AnimationSampling,
 {
-    // TODO: replace with entry API when specs is released
-    if let None = controls.get(entity) {
-        controls.insert(entity, AnimationControlSet::default());
-    }
-    controls.get_mut(entity).unwrap()
+    controls.entry(entity).unwrap().or_insert_with(AnimationControlSet::default)
 }
 
 /// Sampler primitive
--- a/amethyst_animation/src/resources.rs
+++ b/amethyst_animation/src/resources.rs
@@ -183,7 +183,7 @@ where
         F: Fn(Entity) -> Option<T>,
     {
         for entity in self.nodes.values() {
-            if states.get(*entity).is_none() {
+            if !states.contains(*entity) {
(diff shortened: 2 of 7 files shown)


Xaeroxe:
@jojolepro these were the only places my regex search found.

Rhuagh:
It's not as common as you'd think because most places need access to the data, so they do matching instead.

jojolepro:
https://github.com/amethyst/amethyst/blob/develop/amethyst_ui/src/layout.rs#L173
https://github.com/amethyst/amethyst/blob/develop/amethyst_ui/src/layout.rs#L316

Check for ".get("

Xaeroxe:
Yeah I was in the middle of fixing those as you wrote that comment.  It just occurred to me I didn't check for `is_none()`

amethyst_animation/src/util.rs
```diff
@@ -20,11 +20,7 @@ where
     I: Send + Sync + 'static,
     T: AnimationSampling,
 {
-    // TODO: replace with entry API when specs is released
-    if let None = controls.get(entity) {
-        controls.insert(entity, AnimationControlSet::default());
-    }
-    controls.get_mut(entity).unwrap()
```
Xaeroxe: @Rhuagh There's a tiny bug here.  This will panic if you call it with a dead entity.  I've preserved the bug as I don't want this PR to have a controversial change in it, but this function could panic.

Rhuagh: Ah, good call. Please file an issue for it so I don't forget.

torkleyy: Should be using Entry API anyway.

Xaeroxe: I've updated it to do that, but the problem is I can't make this behave correctly without altering the return type of the function, and I'm declaring that to be out of scope for this PR.

Rhuagh: I guess we need to return a result or option here ?

PR #690, opened 2018-05-01
Move from error-chain to failure

derekdreery:
I've been messing about moving from error-chain to failure, what would be involved, and how the error types might be streamlined.

It might be that ppl don't wanna make this change, so I've just done 2 modules to start with to show how it might happen.

One thing that `failure` stresses is that error kinds should be fairly simple, in particular they should impl `Eq` if possible, so that they can be compared easily. They are just giving the current context: more specific information will be in the `cause()`.

Having Traits like SystemBundle ask for `failure::Error` seems right to me because amethyst doesn't need to inspect the type of this error - it just needs to know that there was an error, so it can for example use an old version of an asset, but still report the cause chain if necessary.

`failure` uses trait objects and indirection, so maybe isn't so good for the happy path, but I don't think any errors I have seen so far are on a happy path, and amethyst makes heavy use of trait objects anyway.

In the end, the `causes` of the error form a nice chain like they do in `error-chain`. A backtrace is also generated at the first point a `Fail` is created, so for example if a user uses failure internally and writes some code that fails, you get a backtrace from the original site in their code. If not, your backtrace will point to the cause farthest back in your code.

What do people think?

Ideally, using failure should allow errors like the following:

```
    Error: there was an error in the amethyst engine
caused by: An asset hot-reload failed.
caused by: Could not convert asset "mesh1" of type "MESH" from byte array.
caused by: Failed to parse .obj file line 32: unexpected character, expected ....
```

This would be very useful for quickly identifying and fixing the source of errors, with contextual information about what caused the error. Sometimes this contextual information has useful stuff not in a stack trace, and would require a debugger to see otherwise.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/690)


derekdreery:
I'll carry on while I wait for input from the others :).

Rhuagh:
btw, feel free to improve the SystemBundle with an associated error type if you wish.

derekdreery:
I did have a look at that, but I wasn't sure if it added any useful functionality to justify the complexity.

First-off I have assumed that all errors are exceptional, so we aren't worried about performance too much when we have one. The failure crate probably isn't the way to go for errors that happen often and/or on hot paths.

So where you have user-implemented traits, what does amethyst want to do with them? My assumption is "not very much". It may

 1. *Recover from it without really knowing anything about it.* - Unless we bound the error, we don't know much about its type, and I don't think we need to. For example, if the asset hot reloader errors, we just use the old asset. We don't need to know anything about *why* it errored, other than being able to print any cause chain to the user (which this patch does).

 2. *Just pass the error through.* - The user should either handle errors before they are returned by amethyst's traits, or downcast to them back in own code after amethyst has wrapped the error. Again we don't need to work with the error type at all.

The only other reason to use a cocrete (generic) type would be to avoid trait objects, but since error cause chains are always trait objects, we're not losing anything.

On the other hand just returning a `failure::Error` is really easy from an ergonomic point of view. Anything that implements `std::error::Error + Send + Sync + 'static` (99% of errors in my experience) can just be returned with `?` in the user's code. The failure crate has helpers to wrap anything that isn't `Send` or `Sync` (basically by wrapping it in an Arc - not efficient but it works.)

Rhuagh:
Your call :)

derekdreery:
@torkleyy yeah, and it's pretty similar to failure. failure just does its stuff without macros, and forces 'static + Send + Sync so that you can move errors around easily and downcast them to their concrete type.

Rhuagh:
@derekdreery any status update on this?

derekdreery:
I'm just learning the rendering engine so I can choose sensible error types. I'm also away from my computer for a few days. :)

derekdreery:
Done renderer - this is quite a big PR :P

torkleyy:
Wow thanks, have to review this later.

derekdreery:
Still have to do some of the submodules - not ready for merge yet.

derekdreery:
I'm gonna close this so that CI doesn't run all the time - I'll reopen when I've finished.

derekdreery:
Right, I've squashed most of the commits and rebased against latest, so should be ready for review!

I'd like people to check that error types are correct. They should not try to contain the exact cause of error, but give context to it within amethyst. The chain of errors (where each error is wrapped in more context) gives a stacktrace-like view of the error bubbling up from the root cause to the level where it is handled (persumably top level, though it doesn't have to be).

Also, please check there are no errors on happy paths. `Result<T, failure::Error>` is very small (if I remember correctly) because all error information is behind a pointer, that is never allocated if we are `Ok`. However, when you hit an error you suffer a performance penalty. You make the trade-off that better error information is worth more than crashing slightly quicker.

I'd also like it if people tried it out and saw what errors were like. They should make it really easy to see where an error occured, and the causes.

torkleyy:
Reviewed 8 of 70 files at r1, 4 of 12 files at r2.
Review status: 12 of 71 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[amethyst_animation/src/bundle.rs, line 5 at r2](https://reviewable.io/reviews/amethyst/amethyst/690#-LCoWYF-Dsxj_8gip2Gb:-LCoWYF-Dsxj_8gip2Gc:b-q9cpnc) ([raw file](https://github.com/amethyst/amethyst/blob/c73494cc342956714554f7c1f7a804336408e923/amethyst_animation/src/bundle.rs#L5)):*
> ```Rust
> 
> use amethyst_core::specs::prelude::{Component, DispatcherBuilder};
> use amethyst_core::{Result, SystemBundle};
> ```

This should still export a `Result`.

---

*[amethyst_assets/src/progress.rs, line 161 at r2](https://reviewable.io/reviews/amethyst/amethyst/690#-LCoX5y76XBmYb7GUwSN:-LCoX5y76XBmYb7GUwSO:b2ip4up) ([raw file](https://github.com/amethyst/amethyst/blob/113999fa8f540d68c46c2201b9d419e23b400c8e/amethyst_assets/src/progress.rs#L161)):*
> ```Rust
>     fn fail(self: Box<Self>, e: Error) {
>         error!("error: {}", e);
>         while let Some(e) = e.cause() {
> ```

Does this work? I think this will print the same cause over and over.

---

*[amethyst_assets/src/reload.rs, line 192 at r2](https://reviewable.io/reviews/amethyst/amethyst/690#-LCoXFfv9MjwIWZIy-Do:-LCoXFfv9MjwIWZIy-Dp:b-vcrgme) ([raw file](https://github.com/amethyst/amethyst/blob/113999fa8f540d68c46c2201b9d419e23b400c8e/amethyst_assets/src/reload.rs#L192)):*
> ```Rust
> 
> /// What is this for?
> pub trait ReloadClone<A> {
> ```

It's a helper trait for cloning trait objects of `Reload`.

---

*[amethyst_assets/src/source/dir.rs, line 36 at r2](https://reviewable.io/reviews/amethyst/amethyst/690#-LCoXfxT2VSPz15GPApA:-LCoXfxT2VSPz15GPApB:b-fp3kyp) ([raw file](https://github.com/amethyst/amethyst/blob/113999fa8f540d68c46c2201b9d419e23b400c8e/amethyst_assets/src/source/dir.rs#L36)):*
> ```Rust
> 
> impl Source for Directory {
>     fn modified(&self, path: &str) -> Result<u64, Error> {
> ```

Hmm, this makes it hard to match on the actual errors. This would be very important here.

---

*[amethyst_assets/src/source/dir.rs, line 42 at r2](https://reviewable.io/reviews/amethyst/amethyst/690#-LCoXpwJEkQIKBBlaI1K:-LCoXpwJEkQIKBBlaI1L:bxqpnhk) ([raw file](https://github.com/amethyst/amethyst/blob/113999fa8f540d68c46c2201b9d419e23b400c8e/amethyst_assets/src/source/dir.rs#L42)):*
> ```Rust
>         let path = self.path(path);
> 
>         Ok(fs::metadata(&path)?
> ```

While admittedly the strings didn't add any type safety, they could tell us what exactly failed.

---


*Comments from [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/690)*

derekdreery:
Review status: 12 of 71 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[amethyst_animation/src/bundle.rs, line 5 at r2](https://reviewable.io/reviews/amethyst/amethyst/690#-LCoWYF-Dsxj_8gip2Gb:-LCs_Ggq2w6wLkxCwrHj:b-sade5m) ([raw file](https://github.com/amethyst/amethyst/blob/c73494cc342956714554f7c1f7a804336408e923/amethyst_animation/src/bundle.rs#L5)):*

This should still export a `Result`.
</blockquote></details>

`amethyst_core` still exports a result, I'm just not using it here.

---

*[amethyst_assets/src/progress.rs, line 161 at r2](https://reviewable.io/reviews/amethyst/amethyst/690#-LCoX5y76XBmYb7GUwSN:-LCsZN2sDkxbzNpaCh05:buzrn9b) ([raw file](https://github.com/amethyst/amethyst/blob/113999fa8f540d68c46c2201b9d419e23b400c8e/amethyst_assets/src/progress.rs#L161)):*

Does this work? I think this will print the same cause over and over.
</blockquote></details>

Yeah this is the canonical way of descending a pointer chan, afaik.

---

*[amethyst_assets/src/source/dir.rs, line 36 at r2](https://reviewable.io/reviews/amethyst/amethyst/690#-LCoXfxT2VSPz15GPApA:-LCs_Paq2WkXVocvO6_J:bsw2km2) ([raw file](https://github.com/amethyst/amethyst/blob/113999fa8f540d68c46c2201b9d419e23b400c8e/amethyst_assets/src/source/dir.rs#L36)):*

Hmm, this makes it hard to match on the actual errors. This would be very important here.
</blockquote></details>

Ok we need to decide if we need to use associated error types in these traits. Let's talk about it in the comments.

---

*[amethyst_assets/src/source/dir.rs, line 42 at r2](https://reviewable.io/reviews/amethyst/amethyst/690#-LCoXpwJEkQIKBBlaI1K:-LCs_h-L-Z1sK1ocNnjm:b-j2zmoy) ([raw file](https://github.com/amethyst/amethyst/blob/113999fa8f540d68c46c2201b9d419e23b400c8e/amethyst_assets/src/source/dir.rs#L42)):*

While admittedly the strings didn't add any type safety, they could tell us what exactly failed.
</blockquote></details>

In that case maybe we need to add an ErrorKind to give the context of the error here. The underlying cause is still passed through.

---


*Comments from [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/690)*

torkleyy:
> Ok we need to decide if we need to use associated error types in these traits. Let's talk about it in the comments.

I simply think there should be a static `ErrorKind` for errors when fetching data from asset sources. Something like

* time out
* doesn't exist
* insufficient permissions
* unsupported operation (for when we can't get metadata)

derekdreery:
> I simply think there should be a static ErrorKind for errors when fetching data from asset sources. 

So do you think that the user should choose the most appropriate, and then optionally supply more info using a `cause`?

torkleyy:
Uh honestly I didn't think about adding a cause for it. The information that the asset under path `/foo/bar` is not available should be enough IMHO.

derekdreery:
@torkleyy the cause will just be the underlying error (can be essentially anything that implements std::error::Error) so you'll get *permission denied* or *resource does not exist* or whatever with basically no extra work.

derekdreery:
Right hopefully it should build now, so we have a solid base to make any necessary changes.

derekdreery:
Here are all the places where failure::Error is used

`Format::import @ amethyst_asset/src/asset.rsL61`
`SimpleFormat::import @ amethyst_asset/src/asset.rsL97`
`Source::modified @ amethyst_asset/src/source/mod.rsL15`
`Source::load @ amethyst_asset/src/source/mod.rsL20`
`Reload::reload @ amethyst_asset/src/reload.rsL188`

and `impl`s of these.

Should any of these be assoicated types?

torkleyy:
Didn't we have generics / associated types for those before? I think they should have them. But please wait before working on it, there might be other opinions.

derekdreery:
There weren't any associated types before I started, they had hard-coded types.

Rhuagh:
All the formats return `amethyst_assets::Error` currently.

derekdreery:
@torkleyy I posted a comment earlier about why I didn't use associated types for errors for my first draft.

derekdreery:
Ok I think there's enough of a consensus for me to investigate associated error types for formats.

derekdreery:
A minor annoyance: how to pick an error type for formats that cannot error (like all the audio formats). I want to use `!` but don't think it's stable yet.

Rhuagh:
() ?

derekdreery:
I've put a trait bound of `failure::Fail` on the error type. I could remove
this but it makes chaining errors more useful/ergonomic.

derekdreery:
I've added a commit to change to associated error types on the `Format` trait. What do ppl think.

amethyst_assets/src/progress.rs
```diff
@@ -157,7 +158,9 @@ impl Tracker for () {
     fn success(self: Box<Self>) {}
     fn fail(self: Box<Self>, e: Error) {
         error!("error: {}", e);
-        e.iter().skip(1).for_each(|e| error!("caused by: {}", e));
+        while let Some(e) = e.cause() {
```
Xaeroxe: I'm not sure this is correct, it doesn't recurse down the cause chain it just seems to print the same cause over and over again.

torkleyy: Yeah I commented the same.

torkleyy:
Reviewed 2 of 70 files at r1, 1 of 12 files at r2, 4 of 59 files at r3, 9 of 17 files at r4.
Review status: 19 of 90 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[amethyst_assets/examples/ll.rs, line 46 at r4](https://reviewable.io/reviews/amethyst/amethyst/690#-LDCyYwR9uiwMroOSpCE:-LDCyYwS1hFoRcYfTg8Y:b-soiqaz) ([raw file](https://github.com/amethyst/amethyst/blob/a897b52e12ad09116a83a799c37e453768a4c8c8/amethyst_assets/examples/ll.rs#L46)):*
> ```Rust
>     ) -> StdResult<FormatValue<DummyAsset>, Self::Error> {
>         let data = source.load(&name).unwrap();
>         let dummy = from_utf8(data.as_slice()).unwrap().to_owned();
> ```

Not happy about the unwrap here.

---

*[amethyst_assets/src/asset.rs, line 46 at r4](https://reviewable.io/reviews/amethyst/amethyst/690#-LDCyfAa77Tz58S-CHNZ:-LDCyfAa77Tz58S-CHN_:b-szhuze) ([raw file](https://github.com/amethyst/amethyst/blob/a897b52e12ad09116a83a799c37e453768a4c8c8/amethyst_assets/src/asset.rs#L46)):*
> ```Rust
>     /// The import operation may fail with this associated Error type.
>     ///
>     /// If your import cannot fail consider using the never type `!` TODO when
> ```

Should not hint to unstable features.

---

*[amethyst_assets/src/asset.rs, line 98 at r4](https://reviewable.io/reviews/amethyst/amethyst/690#-LDCykcT9EnQ33k0zDmC:-LDCykcT9EnQ33k0zDmD:buo9nmt) ([raw file](https://github.com/amethyst/amethyst/blob/a897b52e12ad09116a83a799c37e453768a4c8c8/amethyst_assets/src/asset.rs#L98)):*
> ```Rust
>     /// The import operation may fail with this associated Error type.
>     ///
>     /// If your import cannot fail consider using the never type `!` TODO when
> ```

Same here

---

*[amethyst_assets/src/reload.rs, line 192 at r2](https://reviewable.io/reviews/amethyst/amethyst/690#-LCoXFfv9MjwIWZIy-Do:-LDCz5zI0oT7dOKrL4SA:b-p0yj88) ([raw file](https://github.com/amethyst/amethyst/blob/113999fa8f540d68c46c2201b9d419e23b400c8e/amethyst_assets/src/reload.rs#L192)):*

It's a helper trait for cloning trait objects of `Reload`.
</blockquote></details>

TODO: remove this comment

---

*[amethyst_assets/src/source/dir.rs, line 36 at r4](https://reviewable.io/reviews/amethyst/amethyst/690#-LDCzLBECGBMxBHXVm_U:-LDCzLBECGBMxBHXVm_V:b-xk0rsd) ([raw file](https://github.com/amethyst/amethyst/blob/a897b52e12ad09116a83a799c37e453768a4c8c8/amethyst_assets/src/source/dir.rs#L36)):*
> ```Rust
> 
> impl Source for Directory {
>     fn modified(&self, path: &str) -> Result<u64, Error> {
> ```

Please help me here, didn't we agree this should be using an associated type? Sorry if I misunderstood something here.

---


*Comments from [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/690)*

Rhuagh:
Reviewed 18 of 70 files at r1, 4 of 12 files at r2, 41 of 59 files at r3, 9 of 17 files at r4.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[amethyst_assets/src/error.rs, line 60 at r4](https://reviewable.io/reviews/amethyst/amethyst/690#-LDDM3bFDLj2QhPrReY4:-LDDM3bFDLj2QhPrReY5:b-3muguj) ([raw file](https://github.com/amethyst/amethyst/blob/a897b52e12ad09116a83a799c37e453768a4c8c8/amethyst_assets/src/error.rs#L60)):*
> ```Rust
>     #[fail(display = "A custom asset dropper returned an error.")]
>     DropAsset,
>     ///// Returned if a source could not retrieve something.
> ```

Can we remove the unused variants?

---

*[amethyst_assets/src/source/dir.rs, line 36 at r4](https://reviewable.io/reviews/amethyst/amethyst/690#-LDCzLBECGBMxBHXVm_U:-LDDTt4i0jq4ABizvTvO:b-8nnvn) ([raw file](https://github.com/amethyst/amethyst/blob/a897b52e12ad09116a83a799c37e453768a4c8c8/amethyst_assets/src/source/dir.rs#L36)):*

Please help me here, didn't we agree this should be using an associated type? Sorry if I misunderstood something here.
</blockquote></details>

I believe that was for `Format` ? But I think `Source` should also be associated type.

---

*[amethyst_gltf/src/error.rs, line 49 at r4](https://reviewable.io/reviews/amethyst/amethyst/690#-LDDOnFT6odxgF7BOQb5:-LDDOnFT6odxgF7BOQb6:b-4dls6a) ([raw file](https://github.com/amethyst/amethyst/blob/a897b52e12ad09116a83a799c37e453768a4c8c8/amethyst_gltf/src/error.rs#L49)):*
> ```Rust
>     MalformedJson,
>     /// The .gltf data is invalid.
>     // TODO I can't get this to work #[derivative(Hash="ignore")]
> ```

Is this yours or this outdated information from the old Errors?

---

*[amethyst_gltf/src/lib.rs, line 142 at r4](https://reviewable.io/reviews/amethyst/amethyst/690#-LDDP7dvDQrfItvXv-U_:-LDDP7dvDQrfItvXv-Ua:bca47c0) ([raw file](https://github.com/amethyst/amethyst/blob/a897b52e12ad09116a83a799c37e453768a4c8c8/amethyst_gltf/src/lib.rs#L142)):*
> ```Rust
> }
> 
> impl<E> Into<StdResult<GltfSceneAsset, E>> for GltfSceneAsset {
> ```

Why the generic type here ?

---


*Comments from [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/690)*

amethyst_assets/src/progress.rs
derekdreery: Oops thankyou I'll fix, sorry i thought it was correct.

derekdreery:
Thanks for the reviews everyone! Once I'm home I'll go through them all :)

derekdreery:
Review status: 86 of 90 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[amethyst_assets/examples/ll.rs, line 46 at r4](https://reviewable.io/reviews/amethyst/amethyst/690#-LDCyYwR9uiwMroOSpCE:-LDQdy4C1YX9FIeTyDyO:byjr3r0) ([raw file](https://github.com/amethyst/amethyst/blob/a897b52e12ad09116a83a799c37e453768a4c8c8/amethyst_assets/examples/ll.rs#L46)):*

Not happy about the unwrap here.
</blockquote></details>

I'm cheating here so I didn't have to implement a special error type. I don't mind doing it though.

---

*[amethyst_assets/src/source/dir.rs, line 36 at r4](https://reviewable.io/reviews/amethyst/amethyst/690#-LDCzLBECGBMxBHXVm_U:-LDQe93J2AgmRdnamStt:bh7wa7t) ([raw file](https://github.com/amethyst/amethyst/blob/a897b52e12ad09116a83a799c37e453768a4c8c8/amethyst_assets/src/source/dir.rs#L36)):*
<details><summary><i>Previously, Rhuagh (Simon Rönnberg) wrote…</i></summary><blockquote>

I believe that was for `Format` ? But I think `Source` should also be associated type.
</blockquote></details>

I'll use an assoicated type for `Source` as well.

---

*[amethyst_gltf/src/error.rs, line 49 at r4](https://reviewable.io/reviews/amethyst/amethyst/690#-LDDOnFT6odxgF7BOQb5:-LDQde7PAKvmkcAVlomR:b-6wm8ww) ([raw file](https://github.com/amethyst/amethyst/blob/a897b52e12ad09116a83a799c37e453768a4c8c8/amethyst_gltf/src/error.rs#L49)):*
<details><summary><i>Previously, Rhuagh (Simon Rönnberg) wrote…</i></summary><blockquote>

Is this yours or this outdated information from the old Errors?
</blockquote></details>

It's mine, I was trying to implement the standard functionality (Ord, Hash etc.), but couldn't here because validation::Error doesn't implement these Traits. I decided that they probably weren't that important and so just didn't implement them.

---

*[amethyst_gltf/src/lib.rs, line 142 at r4](https://reviewable.io/reviews/amethyst/amethyst/690#-LDDP7dvDQrfItvXv-U_:-LDQeIpD-wxpkakK3WYr:b-hmbs8d) ([raw file](https://github.com/amethyst/amethyst/blob/a897b52e12ad09116a83a799c37e453768a4c8c8/amethyst_gltf/src/lib.rs#L142)):*
<details><summary><i>Previously, Rhuagh (Simon Rönnberg) wrote…</i></summary><blockquote>

Why the generic type here ?
</blockquote></details>

I wanted something that worked with all errors. Do you think there may be coherence issues?

---


*Comments from [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/690)*

derekdreery:
TODO:

 - [ ] Source associated type(s)
 - [ ] Tests
 - [ ] Pass CI

derekdreery:
@Rhuagh how do you think the associated type should work on `Source`. There are 3 methods:
 1. `modified` for getting last modified time
 2. `load` for (re)loading data from source
 3. `load_with_metadata` for doing both

What I'd like to do is have associated errors for each function, and then a new `Kind` enum for `load_with_metadata` that is either `Metadata` or `Load`, not sure if this is overkill though

derekdreery:
Also, what kind of tests would everyone like to see?

derekdreery:
@torkleyy if you check the most recent commit I've added a test to make sure error messages are correct. Is this the kind of thing you meant, or do you mean something more involved?

Rhuagh:
@derekdreery i'm fine with using failure::Error for `Source` for now.

Rhuagh:
Review status: 80 of 90 files reviewed at latest revision, 8 unresolved discussions.

---

*[amethyst_assets/src/source/dir.rs, line 36 at r4](https://reviewable.io/reviews/amethyst/amethyst/690#-LDCzLBECGBMxBHXVm_U:-LDTS7jJ3ewf5b_btRR3:bqknpgb) ([raw file](https://github.com/amethyst/amethyst/blob/a897b52e12ad09116a83a799c37e453768a4c8c8/amethyst_assets/src/source/dir.rs#L36)):*
<details><summary><i>Previously, derekdreery (Richard Dodd (dodj)) wrote…</i></summary><blockquote>

I'll use an assoicated type for `Source` as well.
</blockquote></details>

Let's defer that.

---

*[amethyst_gltf/src/lib.rs, line 142 at r4](https://reviewable.io/reviews/amethyst/amethyst/690#-LDDP7dvDQrfItvXv-U_:-LDTSIeL7Sq7dOqLMqC_:bemkkt7) ([raw file](https://github.com/amethyst/amethyst/blob/a897b52e12ad09116a83a799c37e453768a4c8c8/amethyst_gltf/src/lib.rs#L142)):*
<details><summary><i>Previously, derekdreery (Richard Dodd (dodj)) wrote…</i></summary><blockquote>

I wanted something that worked with all errors. Do you think there may be coherence issues?
</blockquote></details>

Na, was just curious.

---


*Comments from [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/690)*

Rhuagh:
Reviewed 3 of 4 files at r5, 3 of 6 files at r6, 4 of 4 files at r7.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/690)*

derekdreery:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[amethyst_assets/src/asset.rs, line 46 at r4](https://reviewable.io/reviews/amethyst/amethyst/690#-LDCyfAa77Tz58S-CHNZ:-LDXNZJz64M3E6hFSK3R:b-896fix) ([raw file](https://github.com/amethyst/amethyst/blob/a897b52e12ad09116a83a799c37e453768a4c8c8/amethyst_assets/src/asset.rs#L46)):*

Should not hint to unstable features.
</blockquote></details>

Done.

---

*[amethyst_assets/src/asset.rs, line 98 at r4](https://reviewable.io/reviews/amethyst/amethyst/690#-LDCykcT9EnQ33k0zDmC:-LDXN_P6DLT8TCtYBWf8:b-896fix) ([raw file](https://github.com/amethyst/amethyst/blob/a897b52e12ad09116a83a799c37e453768a4c8c8/amethyst_assets/src/asset.rs#L98)):*

Same here
</blockquote></details>

Done.

---

*[amethyst_assets/src/error.rs, line 60 at r4](https://reviewable.io/reviews/amethyst/amethyst/690#-LDDM3bFDLj2QhPrReY4:-LDXNaNSDTHRC9YSotMW:b-896fix) ([raw file](https://github.com/amethyst/amethyst/blob/a897b52e12ad09116a83a799c37e453768a4c8c8/amethyst_assets/src/error.rs#L60)):*
<details><summary><i>Previously, Rhuagh (Simon Rönnberg) wrote…</i></summary><blockquote>

Can we remove the unused variants?
</blockquote></details>

Done.

---

*[amethyst_assets/src/reload.rs, line 192 at r2](https://reviewable.io/reviews/amethyst/amethyst/690#-LCoXFfv9MjwIWZIy-Do:-LDXNbdkCAO99Ub6Q6mx:b-896fix) ([raw file](https://github.com/amethyst/amethyst/blob/113999fa8f540d68c46c2201b9d419e23b400c8e/amethyst_assets/src/reload.rs#L192)):*

TODO: remove this comment
</blockquote></details>

Done.

---

*[amethyst_assets/src/source/dir.rs, line 36 at r4](https://reviewable.io/reviews/amethyst/amethyst/690#-LDCzLBECGBMxBHXVm_U:-LDXNd1UA_C59BllQKff:b-896fix) ([raw file](https://github.com/amethyst/amethyst/blob/a897b52e12ad09116a83a799c37e453768a4c8c8/amethyst_assets/src/source/dir.rs#L36)):*
<details><summary><i>Previously, Rhuagh (Simon Rönnberg) wrote…</i></summary><blockquote>

Let's defer that.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/690)*

derekdreery:
I cant' tick off the last comment because it crashes Reviewable :(

Rhuagh:
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/690)*

torkleyy:
Dismissed @Xaeroxe from a discussion.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[amethyst_audio/src/output.rs, line 54 at r7](https://reviewable.io/reviews/amethyst/amethyst/690#-LDXjYd-0o6Ygt4EUVCs:-LDXjYd-0o6Ygt4EUVCt:bvmi246) ([raw file](https://github.com/amethyst/amethyst/blob/44ec52471b8f9d32402b533310d9d4e32f59dca5/amethyst_audio/src/output.rs#L54)):*
> ```Rust
>     ///
>     /// This will return an Error if the loaded audio file in source could not be decoded.
>     pub fn try_play_n_times(&self, source: &Source, volume: f32, n: u16) -> Result<(), Error> {
> ```

Why can't we return the specific error here?

---

*[amethyst_audio/src/sink.rs, line 27 at r7](https://reviewable.io/reviews/amethyst/amethyst/690#-LDXjguL4YWfUshgijV6:-LDXjguL4YWfUshgijV7:b68q291) ([raw file](https://github.com/amethyst/amethyst/blob/44ec52471b8f9d32402b533310d9d4e32f59dca5/amethyst_audio/src/sink.rs#L27)):*
> ```Rust
>         self.sink
>             .append(Decoder::new(Cursor::new(source.clone()))
>                 .map_err(|e| e.context(ErrorKind::Decoder))?);
> ```

Meh, I think I understand, you want to chain them instead of wrapping.. I'm somehow not 100% happy with this :(

---

*[amethyst_audio/src/source.rs, line 31 at r7](https://reviewable.io/reviews/amethyst/amethyst/690#-LDXjtDf0dq4Qg1oha1v:-LDXjtDf0dq4Qg1oha1w:btbjnwb) ([raw file](https://github.com/amethyst/amethyst/blob/44ec52471b8f9d32402b533310d9d4e32f59dca5/amethyst_audio/src/source.rs#L31)):*
> ```Rust
> }
> 
> impl Into<Result<Source, Error>> for AudioData {
> ```

Shouldn't this be `Void`?

---

*[amethyst_audio/src/components/audio_emitter.rs, line 31 at r7](https://reviewable.io/reviews/amethyst/amethyst/690#-LDXjyVW1ZS0f_7sap1s:-LDXjyVW1ZS0f_7sap1t:b-l2dy7s) ([raw file](https://github.com/amethyst/amethyst/blob/44ec52471b8f9d32402b533310d9d4e32f59dca5/amethyst_audio/src/components/audio_emitter.rs#L31)):*
> ```Rust
> 
>     /// Plays an audio source from this emitter.
>     pub fn play(&mut self, source: &Source) -> Result<(), Error> {
> ```

Same here with the specific error

---

*[amethyst_config/src/error.rs, line 13 at r7](https://reviewable.io/reviews/amethyst/amethyst/690#-LDXk85M4KhZlse8L8sT:-LDXk85M4KhZlse8L8sU:b-q36qj0) ([raw file](https://github.com/amethyst/amethyst/blob/44ec52471b8f9d32402b533310d9d4e32f59dca5/amethyst_config/src/error.rs#L13)):*
> ```Rust
> /// The assets subsystem error type
> #[derive(Debug)]
> pub struct Error {
> ```

I don't see any advantage in creating distinct error types just to later throw away the extra type safety for the user. Sorry, can you explain what's the point of this pattern?

---


*Comments from [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/690)*

torkleyy:
My keyboard shortcuts for Reviewable don't work anymore which makes me kind of slow.. anybody else experiencing this? It works like for a few files and then it just doesn't do anything anymore.

derekdreery:
@torkleyy I don't think we should rush to merging this PR. I want to really discuss the error structure - failure gives you choices, for example you can wrap errors rather than chaining (although you lose things like Eq, Clone implementations for your ErrorKind). I also want to make sure I'm using the right error kinds, maybe some should be grouped or split I don't know. In any case I'm happy to keep rebasing until we are happy with all of this patch.

Rhuagh:
I think we should decide fairly soon how we want this, push this PR, and then iterate the design.

Xaeroxe:
@derekdreery we don't really have much choice other than to merge large PRs like this quickly because otherwise they rapidly go out of date.  This PR for example has 19 conflicting files.  If we still feel additional discussion is necessary it should happen on an issue or on Discord/gitter before the PR is opened.

derekdreery:
I'll rebase this and try to continue with the patch. I'm just applying for
jobs atm.

torkleyy:
Wow, it's overwhelming how many changes we had in a few days. @derekdreery We should really merge this once you rebased, because it's a lot of work to keep it up to date with the code base.

jojolepro:
@derekdreery Status?

derekdreery:
I've had some other work I've had to do.

In some ways I'm tempted to close, because really I believe that each function that can fail should have its own error type, to describe the ways that *that* error can fail. (I think this was mentioned earlier in the comments). The advantage of that approach is also that it could be incremental, and could be split up into lots of jobs rather than being one massive PR. Also, there's discussion on how to move the failure crate forward, in [failure#209]. They are talking about agreeing on a new Error trait for stdlib (called something different like Fail) with a signature like

```rust
pub trait Fail: Debug + Display + Send + Sync + 'static {
    pub fn cause(&self) -> &Fail {
        None
    }

    pub fn backtrace(&self) -> &Backtrace {
        None
    }
}
```

Then different crates can iterate on designing macros to generate this impl. They won't rush this though because exactly what `Backtrace` is isn't fully clear yet, and if a trait is going in libstd it has to be spot on (we saw what happens when this isn't the case with `Error`).

So sad as I am I would say close. If I have a change of heart I'll rebase and reopen.

[failure#209]: https://github.com/rust-lang-nursery/failure/issues/209

Xaeroxe:
@derekdreery, thanks for your work on this, however when the time comes to revamp our error structure it would probably be easier to re-write this due to how many conflicts there will be.  With that in mind and acting on the author recommendation I'm going to close this.

PR #691, opened 2018-05-01
refactor: Make game data pluggable

### LINKED ISSUE #576: [RFC] Alternate root GameData structure
Per request from @Xaeroxe this RFC was inspired by recent discussions on Gitter.

As more use cases for Amethyst begin to explore the virtues and limitations of
the current data architecture some hurdles are beginning to present themselves for
both optimization and flexibility. The current ECS design forces a specific way of
thinking that works for a lot of designs but requires use of anti-pattern solutions
in others. When examining how one would implement a dense voxel map idiomatically in
Amethyst general consensus is that each voxel should be an entity within
the system. While very simple this is not an optimal structure for chunked densely 
packed Voxel data accessed by range query. Thus Voxel information would be 
relegated to an outside resource with array data and a second ecs world containing
Voxel definitions. The access pattern of this second ecs is random and differs 
greatly from the sequential access pattern of the primary ecs iterator.

This proposal suggests removing the global specs World structure and replacing it
with a structure containing just the resource portion of the specs world structure.
Each state would be handed this new resource structure instead of the global ecs
and when building system pipelines a Vector of ResourceID(TypeID, GUID) would be
provided. This would still match the current behavior of providing a "global" ecs
to each system but would also allow supplemental resources to be provided to a 
system while allowing background threads to dynamically load and save new resources
or build new pipelines.

For System specialization this allows for specialized or multiple data structures
to be provided to a system in an idiomatic way. An example of this is using 2 
separate ECS structures for entity data, a chunked voxel structure and ecs, 
or eschew the ecs entirely in favor of an alternate system.

Rhuagh:
Remove Dispatcher from Application, and add a pluggable generic type.
Add a direct replacement for the old game data, featuring only a Dispatcher.
Make the states take a new StateData type, containing references to World and pluggable game data.
Move responsibility for doing system dispatch to the states.

Opening this early to get feedback.

Resolves #576 

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/691)


jojolepro:
I'll need a more complex example to really get the feel of it.
Maybe an example with:
-Global systems
renderer,input,ui,assets
-StateLoading
waits for assets to be all loaded, shows a ui specific to this state (loading ui), which get removed automatically when changing state (because when you switch you lose this World)
-StateGame
simple ui showing the loaded asset

The prototype looks good ;)

Rhuagh:
The World is not replaced, it's just the Dispatcher that's been replaced with a generic type.

Rhuagh:
I was thinking of maybe adding a loading state to the renderable example, since some of those assets take a while to load in debug mode.

Rhuagh:
This is basically my stab at #576.

Because we use `World` for a lot of things internally in `Application` and a lot of the core, my proposal is basically to keep the core `World` in place, and simply abstract away the `Dispatcher`, and any "extra" data that the user might want to keep centrally. The naming of the different types could use improvements, but the core concept is solid I believe.

Xaeroxe:
I think it might be a good idea to add a chapter to the book called *Advanced Game Architecture* that shows how to use dispatchers in a custom `GameData` structure.

Rhuagh:
I was planning on creating an example atleast once I get some feedback on the basic design :)

jojolepro:
I'll re-review it soon.

amethyst_animation/src/util.rs
```diff
@@ -20,7 +20,10 @@ where
     I: Send + Sync + 'static,
     T: AnimationSampling,
 {
-    controls.entry(entity).unwrap().or_insert_with(AnimationControlSet::default)
+    controls
+        .entry(entity)
+        .unwrap()
```
torkleyy: This should not unwrap but return the error.

src/game_data.rs
```diff
@@ -0,0 +1,240 @@
+use std::sync::Arc;
```
torkleyy: Missing top-level docs

amethyst_animation/src/util.rs
Rhuagh: That's a known (and reported issue), yes, will fix that in a separate PR as it changes the API.

examples/game_data/main.rs
```diff
@@ -0,0 +1,583 @@
...
+            },
+            _ => Trans::None,
+        }
+    }
+
+    fn update(&mut self, data: StateData<GameData>) -> Trans<GameData<'a, 'b>> {
+        data.data.update(&data.world, true);
+        match self.progress.as_ref().unwrap().complete() {
+            Completion::Failed => {
+                println!("Failed loading assets");
```
jojolepro: printerr! or error!

examples/game_data/main.rs
```diff
@@ -0,0 +1,583 @@
...
+    );
+
+    let display_config = DisplayConfig::load(display_config_path);
+    let pipeline_builder = Pipeline::build().with_stage(
+        Stage::with_backbuffer()
+            .clear_target([0.0, 0.0, 0.0, 1.0], 1.0)
+            .with_pass(DrawShaded::<PosNormTex>::new())
+            .with_pass(DrawUi::new()),
+    );
+    let game_data = GameDataBuilder::default()
```
jojolepro: Games don't only have a base and a running state. There should be a way to provide multiple systems that are specific to each of the states. Those systems should be added during the state start function, to allow for better logic locality.

examples/game_data/resources/display_config.ron
```diff
@@ -0,0 +1,7 @@
+(
+  fullscreen: false,
+  multisampling: 0,
+  title: "Renderable Example",
```
jojolepro: Renderable again

examples/game_data/main.rs
Rhuagh: Ofc not, but I'm not about to write a whole game in a single example, and this example is for demonstrating a custom game data implementation, not teach people how to structure a game code wise.

amethyst_ui/src/layout.rs
```diff
@@ -206,10 +206,8 @@ impl<'a> System<'a> for UiParentSystem {
         ReadExpect<'a, ScreenDimensions>,
         ReadExpect<'a, ParentHierarchy>,
     );
-    fn run(
-        &mut self,
-        (entities, mut locals, parents, anchors, stretches, screen_dim, hierarchy): Self::SystemData,
-){
+    fn run(&mut self, data: Self::SystemData) {
+        let (entities, mut locals, parents, anchors, stretches, screen_dim, hierarchy) = data;
```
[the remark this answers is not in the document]
Rhuagh: Formatting. The previous code complained about too long row every time I ran rustfmt, and I got tired of it.

examples/game_data/main.rs
```diff
@@ -0,0 +1,583 @@
...
+                         MeshHandle, ObjFormat, Pipeline, PngFormat, PointLight, PosNormTex,
+                         Projection, RenderBundle, Rgba, Stage, VirtualKeyCode, WindowEvent};
+use amethyst::ui::{Anchor, Anchored, DrawUi, FontHandle, TtfFormat, UiBundle, UiText, UiTransform};
+use amethyst::utils::fps_counter::{FPSCounter, FPSCounterBundle};
+
+use game_data::{GameData, GameDataBuilder};
+
+mod game_data;
+
+struct DemoState {
```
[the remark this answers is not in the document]
Rhuagh: The renderable example is the only one where loading is slow enough to actually warrant a Loading state, that's why I chose it.

Rhuagh:
It's a generic type without bounds, you can put anything you like in it. Data would be in the `World` though, so there's no need to "share" that. Personally, I would put state specific dispatchers in the state, and not in a central location.

jojolepro:
So GameData is global? If so I'm not sure what you can do with it that you couldn't do using Resource.

Xaeroxe:
@jojolepro The main thrust of the PR is to permit people to access data in a way that isn't via `specs`.  So this would be for people who don't even want a `Resource`.

jojolepro:
1) What would be the use case for that?
2) Is it worth the added complexity?

Xaeroxe:
@Rhuagh mentioned https://github.com/amethyst/amethyst/issues/576 in a comment on this PR (Maybe we should mention it in the opening PR description)

If we go to that issue some helpful context is present.  Specifically this quote:

> As more use cases for Amethyst begin to explore the virtues and limitations of
the current data architecture some hurdles are beginning to present themselves for
both optimization and flexibility. The current ECS design forces a specific way of
thinking that works for a lot of designs but requires use of anti-pattern solutions
in others. When examining how one would implement a dense voxel map idiomatically in
Amethyst general consensus is that each voxel should be an entity within
the system. While very simple this is not an optimal structure for chunked densely
packed Voxel data accessed by range query. Thus Voxel information would be
relegated to an outside resource with array data and a second ecs world containing
Voxel definitions. The access pattern of this second ecs is random and differs
greatly from the sequential access pattern of the primary ecs iterator.

jojolepro:
And how does Resource not solve this? My current plan for a voxel world involves having it as a resource and having Systems use that resource to create entities (optimised) that will end up getting rendered.

Xaeroxe:
Yeah a `Resource` could be used for this.  That means that to me there's one remaining major reason to merge this PR, and that is to facilitate types that aren't `Any + Send + Sync`.

Rhuagh:
You can't store systems in a resource.

Xaeroxe:
@Rhuagh is correct due to this: https://docs.rs/shred/0.7.0-alpha5/shred/struct.Dispatcher.html#impl-Send

jojolepro:
That is a good argument. ;)
Okay then.

examples/game_data/main.rs
jojolepro: printerr! then?

examples/game_data/main.rs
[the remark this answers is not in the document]
jojolepro: You could keep the .obj loading, which is the slow part, but remove the lights. If not then its fine.

amethyst_ui/src/layout.rs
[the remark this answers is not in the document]
jojolepro: :+1:

examples/game_data/main.rs
Rhuagh: printerr? Not a part of core rust from what I know ?

jojolepro: Ya sorry its "eprintln!" actually :P

Rhuagh:
Re-review if you can please.

torkleyy:
Reviewed 3 of 8 files at r1, 5 of 10 files at r2, 19 of 21 files at r3, 3 of 3 files at r4, 1 of 2 files at r6, 13 of 20 files at r7.
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/691)*

jojolepro:
I thought cfg!(target_os = "ios") didn't work? 
anyway
Waiting for label change to merge. @Rhuagh

Rhuagh:
I still need to add a chapter to the book, and I'm currently focusing on getting specs out.

Rhuagh:
Rebased, squashed, book chapter added. @Xaeroxe feel free to take over.

jojolepro:
outside most of the day, I'll review tonight if I can, or tomorow

Xaeroxe:
err actually nevermind, I think this might fail on CI due to some new examples.

jojolepro:
Will review now

Rhuagh:
bors r=Xaeroxe,torkleyy,jojolepro

Rhuagh:
bors r=Xaeroxe,torkleyy,jojolepro

PR #696, opened 2018-05-06
Expose the length of `MaterialTextureSet`.

azriel91:
Allows consumers to query how many textures are in the set.

Maybe I should ask this first: is it a good idea to query the length, and then add textures just by incrementing the index?

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/696)

### DIFF
--- a/amethyst_animation/src/material.rs
+++ b/amethyst_animation/src/material.rs
@@ -41,6 +41,10 @@ impl MaterialTextureSet {
         }
     }
 
+    pub fn len(&self) -> usize {
+        self.textures.len()
+    }
+
     pub fn clear(&mut self) {
         self.textures.clear();
         self.texture_inverse.clear();


Rhuagh:
The texture set is not length based so I don't really know what the use would be? The index is user supplied and just an id, not an offset in an array.

azriel91:
My use case is, I'm loading many sprite sheets across different "objects" in my game.

* Each object has multiple sprite sheets
* Different objects do not share sprite sheets
* Material offset animation uses a zero-based texture id (i.e. the animation configuration specifies, use sprite sheet N, sprite number M)
* When sprite sheet N is loaded, I store it in the `MaterialTextureSet` at "`offset` + N", and tell the animation sampler to use that, instead of just N

I could store the offset externally, and every time I load an object, update `offset` to be `offset + N`, or I could just read it from the `MaterialTextureSet` if it were present.

Reason for wanting to read it on the fly instead of storing it externally is, I've got different *object types*, each having their own loader, and that index offset needs to be known to each loader when it loads an object. Since my loading is done sequentially, the number of textures in the `MaterialTextureSet` is an authoritative source for "find me a unique id", and saves me the trouble of having either a parent "I load everything" wrapper or a "next texture ID tracker" resource

Rhuagh:
I guess. I would make an address resolution resource personally, so I can replicate.

PR #697, opened 2018-05-07
Re-use vec allocation for event compression

Xaeroxe:
If event compression was causing any noticeable  performance impact it'll be gone after merging this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/697)

### DIFF
--- a/amethyst_renderer/src/system.rs
+++ b/amethyst_renderer/src/system.rs
@@ -71,7 +71,7 @@ where
             pipe,
             renderer,
             cached_size,
-            event_vec: Vec::new(),
+            event_vec: Vec::with_capacity(20),
         }
     }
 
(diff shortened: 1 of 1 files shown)


amethyst_renderer/src/system.rs
```diff
@@ -68,6 +71,7 @@ where
             pipe,
             renderer,
             cached_size,
+            event_vec: Vec::new(),
```
jojolepro: Since you know you are going to receive events, you could pre-initialize it with a decent size. e.g 50? That would reduce load of the first few frames when the buffer gets resized.

amethyst_renderer/src/system.rs
```diff
@@ -132,13 +136,11 @@ where
 
     fn render(&mut self, (mut event_handler, data): RenderData<P>) {
         self.renderer.draw(&mut self.pipe, data);
-
-        let mut events: Vec<Event> = Vec::new();
+        let events = &mut self.event_vec;
```
torkleyy: You may want to clear it before using :)

amethyst_renderer/src/system.rs
Xaeroxe: I went with a more conservative 20 events, but that was a good note.

Issue #593, opened 2018-03-06
[amethyst-controls] Add arc-ball camera


Candunc:
The currently implemented fly camera is good for 'birds eye view' movement around the scene, however I was wondering if a sort of third person 'arc-ball camera' could be added to amethyst. The [kiss3d crate](https://github.com/sebcrozet/kiss3d/) has a good description of how it operates: 

> An arc-ball camera is a camera rotating around a fixed point (the focus point) and always looking at it.

@jojolepro, I understand you implemented the fly camera two weeks ago with your amethyst_controls crate, would it be possible to add this sort of camera as well?

jojolepro:
Its definitely something that I would like.
Would you like to do the implementation/PR?
If you don't I can do it eventually.

Candunc:
I could look into it, however until I get started other people are free to do the implementation.

jojolepro:
If someone takes this over, please write a comment here to prevent duplicate work. Thanks!

jojolepro:
Here is an example of a first person camera, for reference: https://github.com/amethyst/amethyst/blob/develop/amethyst_controls/src/systems.rs#L103

mnivoliez:
Hello, I will *try* to answer this issue.
From a mathematical point of view, a just do rotation in the target referencial (the thing we look at). Therefore, the arc ball component should have a "target" property indicating what the camera is looking for. It seems to me that the calcul is essentially the same as @jojolepro reference some minutes ago https://github.com/amethyst/amethyst/blob/develop/amethyst_controls/src/systems.rs#L103.

jojolepro:
Thanks a lot @mnivoliez

PR #700, opened 2018-05-08
[WIP] Feature/#593 add arc ball camera

mnivoliez:
Here is my work on issue #593 .
For now, I have set up a system and component checking distance between the camera and the target, the free rotation system should take care of the rotation itself.

There is at this point no test, I'll do that as soon as possible.
Also, I do not know if it's the case but in the FPS/TPS case, the camera should be child to the player and evolve in player space (in the case of TPS, we should also check for other collidable object).

### DIFF
--- a/amethyst_controls/src/components.rs
+++ b/amethyst_controls/src/components.rs
@@ -1,5 +1,5 @@
-use amethyst_core::specs::prelude::Component;
-use amethyst_core::specs::storage::NullStorage;
+use amethyst_core::specs::prelude::{Component, Entity};
+use amethyst_core::specs::storage::{NullStorage, HashMapStorage};
 
 /// Add this to a camera if you want it to be a fly camera.
 /// You need to add the FlyControlBundle or the required systems for it to work.
@@ -8,4 +8,15 @@ pub struct FlyControlTag;
 
 impl Component for FlyControlTag {
     type Storage = NullStorage<FlyControlTag>;
+}
+
+pub struct ArcBallCameraTag {
+    pub target: Entity,
+}
+
+impl Component for ArcBallCameraTag {
+    // we can use HashMapStorage here because, according to the specs doc, this storage should be
+    // use when the component is use with few entity, I think there will rarely more than one
+    // camera
+    type Storage = HashMapStorage<ArcBallCameraTag>;
 }
--- a/amethyst_controls/src/components.rs
+++ b/amethyst_controls/src/components.rs
@@ -12,6 +12,7 @@ impl Component for FlyControlTag {
 
 pub struct ArcBallCameraTag {
     pub target: Entity,
+    pub distance: f32,
 }
 
 impl Component for ArcBallCameraTag {
--- a/amethyst_controls/src/systems.rs
+++ b/amethyst_controls/src/systems.rs
@@ -1,4 +1,4 @@
-use amethyst_core::cgmath::{Deg, Vector3};
+use amethyst_core::cgmath::{Deg, InnerSpace, Vector3};
 use amethyst_core::specs::prelude::{Join, Read, ReadExpect, ReadStorage, Resources, System, Write,
                                     WriteStorage};
 use amethyst_core::timing::Time;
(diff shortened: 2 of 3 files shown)


amethyst_controls/src/systems.rs
```diff
@@ -80,6 +80,32 @@ where
...
+    fn run(&mut self, (mut transforms, tags): Self::SystemData) {
+        let mut position = None;
+        for (transform, arc_ball_camera_tag) in (&transforms, &tags).join() {
+            if let Some(target_transform) = transforms.get(arc_ball_camera_tag.target) {
+                let target_to_cam = transform.translation - target_transform.translation;
+                let new_target_to_cam = target_to_cam.normalize() * arc_ball_camera_tag.distance;
+                position = Some(target_transform.translation + new_target_to_cam);
+            }
+        }
+        if let Some(new_pos) = position {
```
torkleyy: This won't work for multiple arc ball cameras, I wonder if we should support that.

amethyst_controls/src/systems.rs
```diff
@@ -80,6 +80,32 @@ where
...
+pub struct ArcBallMovementSystem;
+
+impl<'a> System<'a> for ArcBallMovementSystem {
+    type SystemData = (
+        WriteStorage<'a, Transform>,
+        ReadStorage<'a, ArcBallCameraTag>,
+    );
+
+    fn run(&mut self, (mut transforms, tags): Self::SystemData) {
+        let mut position = None;
```
torkleyy: There's a slightly better solution that also resolves my concern below: you can use `RestrictedStorage` to iterate over the transforms and then set the position in the same iteration of the loop. I'll come back to you and explain in more detail later.

mnivoliez: I though that it would be only one arcball camera (I got no reference about a game having several arcball camera other that split screen game). The exemple is planned, but at 1h30 am, it was difficult to explain to my partner I was still working on code ^^'
The restricted storage could be something.

amethyst_controls/src/systems.rs
jojolepro: You should only be controlling a single camera at once normally. If you want multi controller support input events will need to have a tracker of which controller/mouse does what.

examples/arc_ball_camera/README.md
```diff
@@ -0,0 +1,2 @@
+# Fly Camera Example
```
torkleyy: Rather than a dedicated readme it would actually be more important to add it to the overview.

amethyst_controls/src/components.rs
```diff
@@ -9,3 +9,16 @@ pub struct FlyControlTag;
 impl Component for FlyControlTag {
     type Storage = NullStorage<FlyControlTag>;
 }
+
+#[derive(Debug)]
+pub struct ArcBallCameraTag {
```
torkleyy: Could you also derive `Clone` for this component?

amethyst_controls/src/components.rs
```diff
@@ -9,3 +9,19 @@ pub struct FlyControlTag;
 impl Component for FlyControlTag {
     type Storage = NullStorage<FlyControlTag>;
 }
+
+/// Add this to a camera to which you have already add the FlyControlTag to add a arc ball
```
MrMinimal: Rephrase

amethyst_controls/src/components.rs
```diff
@@ -9,3 +9,19 @@ pub struct FlyControlTag;
...
+/// behaviour to the camera.
+/// Please not that this component require the ArcBallControlSystem to work.
```
MrMinimal: note, requires

amethyst_controls/src/components.rs
```diff
@@ -9,3 +9,19 @@ pub struct FlyControlTag;
...
+#[derive(Debug, Clone)]
+pub struct ArcBallControlTag {
+    pub target: Entity,
+    pub distance: f32,
+}
+
+impl Component for ArcBallControlTag {
+    // we can use HashMapStorage here because, according to the specs doc, this storage should be
+    // use when the component is use with few entity, I think there will rarely more than one
```
MrMinimal: used

amethyst_controls/src/systems.rs
```diff
@@ -80,6 +82,35 @@ where
     }
 }
 
+/// The system that manages the arc ball movement;
+/// In essence, the system will allign the camera with its target while keeping the distance to it
+/// and while keeping the orientation of the camera.
+/// To modify the orientation of the camera in according with the mouse input, please use the
```
MrMinimal: in accordance with

torkleyy:
@mnivoliez Please rebase this PR onto `develop`, it has two conflicts.

amethyst_controls/src/components.rs
MrMinimal: I'd go with "To add an arc ball behaviour, add this to a camera which already has the FlyControlTag added."

Rhuagh:
If the PR is ready for merge, please change the title.

Rhuagh:
@jojolepro please merge if you're ok with this.

PR #702, opened 2018-05-09
Use derivative default better

Xaeroxe:
The redundancy of these bounds in the docs was bothering me.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/702)

### DIFF
--- a/amethyst_input/src/input_handler.rs
+++ b/amethyst_input/src/input_handler.rs
@@ -17,7 +17,7 @@ use super::event::InputEvent::*;
 /// For example, if a key is pressed on the keyboard, this struct will record
 /// that the key is pressed until it is released again.
 #[derive(Derivative)]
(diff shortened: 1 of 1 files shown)


Rhuagh:
bors r=Rhuagh,torkleyy

amethyst_input/src/input_handler.rs
```diff
@@ -17,7 +17,7 @@ use super::event::InputEvent::*;
 /// For example, if a key is pressed on the keyboard, this struct will record
 /// that the key is pressed until it is released again.
 #[derive(Derivative)]
-#[derivative(Default(bound = "AX: Hash + Eq, AC: Hash + Eq"))]
+#[derivative(Default(bound = ""))]
```
jojolepro: why did you keep bound?

Rhuagh: Because otherwise it would require the full bounds of the actual type.

PR #706, opened 2018-05-10
Add issue templates

torkleyy:
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/706)

### DIFF
--- a/docs/ISSUE_TEMPLATE/bug_report.md
+++ b/docs/ISSUE_TEMPLATE/bug_report.md
@@ -1,22 +1,26 @@
+<!--
+
 Thanks for filing a bug! We provide an issue template for you to make sure
 all the important details are presented upfront and we don't have to ask
 for everything individually.
 
-There are multiple places where you can tick an answer. For that, just put an "x" between the brackets.
+There are multiple places where you can provide one or multiple answers.
+For that, just uncomment the points and fill in the missing information.
 
-* [x] This is how a ticked answer looks
+<!-- * This is what a commented (invisible) point looks like (without the "\") --\>
+* This one got uncommented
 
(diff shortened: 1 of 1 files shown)


docs/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,68 @@
+Thanks for filing a bug! We provide an issue template for you to make sure
```
OvermindDL1: If you surround this block with `<!-- ` and `-->` then even if they accidentally leave it in then it will not be visible anyway.  :-)

stenverbois:
Using checkboxes in this way results in the same issue I originally brought up in #349. Github will render all issues as having some sort of progress, while they are not used that way. It makes it harder to keep track of issues that actually have a checklist of things to complete. Maybe this happens less often than with PRs, but its something to keep in mind anyway.

torkleyy:
Ah right, I forgot about that. What about commenting the bullet points with html comments? So the contributor just needs to uncomment it.

stenverbois:
That could certainly work. Another approach is just to have some comments explaining what kind of information is expected in each section. But I don't manage the issues so you and the other members will know best what's easiest to review.

docs/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,68 @@
...
+### Reproduction steps
+
+1. Launch the project with `cargo run`
+2. [etc]
+
+### Screenshot (if appropriate)
+
+
+
+## Source of the bug
```
jojolepro: + I suspect the bug is caused by something related to the MODULE_NAME module

PR #709, opened 2018-05-10
fix: Update shred/specs/etc after SystemData refactor

Rhuagh:
---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/709)


Rhuagh:
Yeah, but they're all safe, because all those entities are created just before.

Rhuagh:
It's for the situations that actually use entities more than once...

PR #712, opened 2018-05-13
Make main functions return Results

vitvakatu:
Rust 1.26 supports returning Results from `main()`. Good time to get rid of `run()` functions!

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/712)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,15 @@ rust:
   - beta
   - nightly
 
-# Allow for occasional compiler breakage on nightly Rust.
 matrix:
+  # Allow for occasional compiler breakage on nightly Rust.
   allow_failures:
     - rust: nightly
+  # Check on Rust 1.23.
+  include:
+    - rust: 1.23
+      script:
+        - cargo check --tests
 
 # Build the Bors staging and trying branches, plus PRs to develop, master, and
 # the latest release branch.
@@ -48,7 +53,7 @@ script: |
   # Build and test without profiler
   cargo test --all -v &&
   # Build and test with profiler
-  cargo test --all --features profiler -v 
+  cargo test --all --features profiler -v
 
 # Push notifications to `amethyst/general` and `amethyst/engine` Gitter chats.
 notifications:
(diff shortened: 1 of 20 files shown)


torkleyy:
I think using new-ish features in the examples is fine, but the project
should build with older versions of Rust, too. We should ensure (using CI)
that the build works for 1.23.

vitvakatu:
Well, it will not work on Rust 1.23 (and any other stable prior to 1.26). Personally I don't think it's a problem, as we support stable toolchain.

torkleyy:
As I said, only the build has to work on 1.23, the examples may use the
latest version.

Xaeroxe:
@torkleyy Rust is still a budding language with lots of valuable features to come.  rustc itself is still in active development, it'd be a shame to lock ourselves down so soon imo.  Why 1.23 specifically?

torkleyy:
Latest version - 3, that's the minimum I chose for me personally. So every six weeks we can bump that version.

jojolepro:
My opinion is that we can safely follow rust stable latest version. Reason: Its "stable". Nothing is going to break except if you don't rustup update.

torkleyy:
> except if you don't rustup update.

That's kind of the point here. I don't want to break anyone's build because a new patch uses a very new feature.

Rhuagh:
I agree with @Xaeroxe here, 1 version behind seems fine to me.

torkleyy:
Also fine for me

jojolepro:
I guess

jojolepro:
Status: Postponed until rust stable version 1.27 is released.

omni-viral:
If we will release once in 6 weeks as well than we can follow stable Rust immediately in development.
Use 1.26 features now and release when 1.27 out.

omni-viral:
@torkleyy if urgent we can release without waiting since we still use only stable. And after few weeks most people have updated compiler anyway.

Xaeroxe:
@vitvakatu Seems PR #691 hit this pretty hard.  Can you rebase it?

PR #715, opened 2018-05-14
Use glsl-layout

omni-viral:
I try to use `glsl-layout` in closer to "real world" applications before releasing it.

This should work everywhere except macos because apple failed to implement opengl correctly.
I'm trying to find least ugly solution.

P.S. I'll release `glsl-layout` before merging and replace git with version in `Cargo.toml`

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/715)

### DIFF
--- a/amethyst_renderer/src/pass/shaded_util.rs
+++ b/amethyst_renderer/src/pass/shaded_util.rs
@@ -22,6 +22,7 @@ pub(crate) struct FragmentArgs {
 pub(crate) struct PointLightPod {
     position: vec3,
     color: vec3,
+    pad: float,
     intensity: float,
 }
 
@@ -44,8 +45,9 @@ pub(crate) fn set_light_args(
             if let Light::Point(ref light) = *light {
                 Some(PointLightPod {
                     position: light.center.into(),
-                    color: light.color.into(),
+                    color: [100.0, 0.0, 0.0].into(),
                     intensity: light.intensity,
+                    pad: 0.0,
                 }.std140())
             } else {
                 None
--- a/amethyst_renderer/src/pass/shaded_util.rs
+++ b/amethyst_renderer/src/pass/shaded_util.rs
@@ -2,7 +2,8 @@ use std::mem;
 
 use amethyst_core::GlobalTransform;
 use amethyst_core::specs::prelude::{Join, ReadStorage};
-use gfx::traits::Pod;
+
+use glsl_layout::*;
 
 use cam::Camera;
 use light::Light;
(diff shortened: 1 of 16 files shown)


Moxinilian:
Does it work with OpenGL ES then?
I'd assume as it doesn't work on macOS, it probably won't work on iOS OpenGL ES, will it?

Rhuagh:
OpenGL 2.0+ or OpenGL ES 3.0+ works with std140, so it should be fine.

omni-viral:
The exact problem is this:
```glsl
struct Foo {
  vec3 x;
  float y;
};

layout(std140) uniform Foos {
  Foo foos[1];
};

foos[0].y; // access bytes from 12 to 16.
int i = 0;
foos[i].y; // access bytes from 0 to 4 on mac and 12 to 16 on other platforms. wat.jpg
```

omni-viral:
The workaround to it is to align all members to 16 manually in structures that are used in arrays.

amethyst_renderer/src/pass/shaders/fragment/shaded.glsl
```diff
@@ -64,16 +65,15 @@ void main() {
     vec4 ecolor = texture(emission, tex_coords(vertex.tex_coord, emission_offset.u_offset, emission_offset.v_offset));
     vec3 lighting = vec3(0.0);
     vec3 normal = normalize(vertex.normal);
-    for (uint i = 0u; i < point_light_count; i++) {
+    for (uint i = 0u; i < point_light_count && i < 1u; i++) {
```
Rhuagh: Only 1 light? :)

omni-viral: Oh. How does this sneaky debugging change got here 😄

amethyst_renderer/src/pass/shaders/fragment/shaded.glsl
```diff
@@ -1,28 +1,26 @@
...
 
-struct PointLight {
-    vec4 position;
-    vec4 color;
+layout (std140) struct PointLight {
+    vec3 position;
+    vec3 color;
+    float pad; // Workaround for bug in mac's implementation of opengl (loads garbage when accessing members of structures in arrays with dynamic indices).
     float intensity;
-    float radius;
```
Rhuagh: Where did radius and smoothness go?

Rhuagh:
Bors r+

Rhuagh:
Bors r+

Rhuagh:
Bors r-

Rhuagh:
Right, the library crate needs to be released

amethyst_renderer/src/pass/shaded_util.rs
```diff
@@ -2,88 +2,79 @@ use std::mem;
...
-    let point_lights: Vec<PointLightPod> = light
+    let point_lights: Vec<_> = light
         .join()
         .filter_map(|light| {
             if let Light::Point(ref light) = *light {
                 Some(PointLightPod {
-                    position: pad(light.center.into()),
-                    color: pad(light.color.into()),
+                    position: light.center.into(),
+                    color: [100.0, 0.0, 0.0].into(),
```
jojolepro: Just doing a really really quick review.

Why is the color hardcoded?

amethyst_renderer/src/pass/shaded_util.rs
```diff
@@ -2,88 +2,79 @@ use std::mem;
...
-    let directional_lights: Vec<DirectionalLightPod> = light
+    let directional_lights: Vec<_> = light
         .join()
         .filter_map(|light| {
             if let Light::Directional(ref light) = *light {
                 Some(DirectionalLightPod {
-                    color: pad(light.color.into()),
-                    direction: pad(light.direction.into()),
-                })
+                    color: light.color.into(),
```
jojolepro: and here not hardcoded

amethyst_renderer/src/pass/shaded_util.rs
omni-viral: Debugging stuff. Nice catch.

PR #718, opened 2018-05-16
Fixes Transform's to_matrix method

JadeSnail:
Per Transform's documentation : `The transforms are preformed in this order: scale, then rotation, then translation.`

Previously, the rotation was performed before the scaling part.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/718)


amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -70,37 +70,14 @@ impl Transform {
...
         let mat = Matrix4 {
-            x,
-            y,
-            z,
+            x: (quat.x * self.scale.x).extend(0.),
+            y: (quat.y * self.scale.y).extend(0.),
+            z: (quat.z * self.scale.z).extend(0.),
             w: self.translation.extend(1.0),
         };
         mat
```
Xaeroxe: Doesn't seem to be much point to declaring a variable then immediately returning it.  Why not just return the value as it's being declared?

Xaeroxe:
bors retry

PR #748, opened 2018-05-25
Switch to sip hash for Bindings

Xaeroxe:
FNV isn't great for large keys, and since we recommend strings by default for this I figure sip hash is probably better.  It's a good compromise between various key sizes.

I also nabbed a few unused imports in the examples.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/748)


Xaeroxe:
I'm not really willing to put in the time to benchmark this, when I could be spending it in other areas of Amethyst, so I'll just close this.

PR #749, opened 2018-05-27
fix: Log an error if a pass tries to render a mesh with incompatible vertex buffers

Rhuagh:
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/749)


Rhuagh:
@torkleyy @Alex-Addy

Rhuagh:
Yes, which is why we need to add tagging to the passes so we can target specific meshes for specific passes.

Alex-Addy:
I can't speak to any other concerns, but this error message would certainly have helped me. It is still confusing looking but it would have definitely pointed me in the right direction.

PR #752, opened 2018-05-29
Upgrade winit, glutin, gfx_glyph, and gfx_window_glutin

### LINKED ISSUE #729: Periodic freezing on Linux
For reasons yet to be determined sometimes Amethyst on Linux ends up in a dead lock.  We believe it has to do with interactions between the windowing libraries and the graphics libraries, but we're not sure why.

Xaeroxe:
~~Should fix #729 , needs further testing to confirm that though.~~

Fixes #729 

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/752)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,7 +62,7 @@ fern = "0.5"
 log = "0.4"
 rayon = "1.0.1"
 rustc_version_runtime = "0.1"
-winit = "0.13.1"
+winit = "0.15"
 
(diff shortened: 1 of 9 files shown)


Xaeroxe:
Per @Alex-Addy this does fix #729

Xaeroxe:
Ok arc ball broke because I was using the wrong winit in amethyst_controls.  Thanks @JadeSnail for pointing this out!

PR #759, opened 2018-05-31
feat: Add utilities for dealing with common inputs

Rhuagh:
After #757.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/759)

PR #758, opened 2018-05-31
fix: Remove individual example READMEs, and only describe them in the…

Rhuagh:
… main examples README

PR #760, opened 2018-05-31
Preparation for Prefab loading

Rhuagh:
Restructures a few types, adds serde implementations for a bunch of core components.
Refactors the return type from `Processor` systems, so initialisation can be deferred as needed.

After ~~#757~~ ~~#758~~ ~~#759~~.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/760)

### DIFF
--- a/amethyst_assets/src/storage.rs
+++ b/amethyst_assets/src/storage.rs
@@ -1,6 +1,6 @@
 use std::marker::PhantomData;
 use std::sync::atomic::{AtomicUsize, Ordering};
-use std::sync::{Arc, Weak};
+use std::sync::{Arc, Mutex, Weak};
 
 use amethyst_core::specs::prelude::{Component, Read, ReadExpect, System, VecStorage, Write};
 use amethyst_core::specs::storage::UnprotectedStorage;
@@ -37,6 +37,18 @@ pub struct AssetStorage<A: Asset> {
     pub(crate) processed: Arc<MsQueue<Processed<A>>>,
     reloads: Vec<(WeakHandle<A>, Box<Reload<A>>)>,
     unused_handles: MsQueue<Handle<A>>,
+    requeue: Mutex<Vec<Processed<A>>>,
+}
+
+/// Returned by processor systems, describes the loading state of the asset.
+pub enum ProcessingState<A>
+where
+    A: Asset,
+{
+    /// Asset is not fully loaded yet, need to wait longer
+    Loading(A::Data),
+    /// Asset have finished loading, can now be inserted into storage and tracker notified
+    Loaded(A),
 }
 
 impl<A: Asset> AssetStorage<A> {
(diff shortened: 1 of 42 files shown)


amethyst_assets/src/storage.rs
```diff
@@ -37,6 +37,18 @@ pub struct AssetStorage<A: Asset> {
     pub(crate) processed: Arc<MsQueue<Processed<A>>>,
     reloads: Vec<(WeakHandle<A>, Box<Reload<A>>)>,
     unused_handles: MsQueue<Handle<A>>,
+    //requeue: Vec<Processed<A>>,
```
Rhuagh: If anyone know how I can make this work with Send + Sync, feel free to shoot me a question. The basics are that I want to save the allocation for assets that need reprocessing. Right now it just allocates a new Vec every frame, which is stupid. The problem is that adding the above makes `AssetStorage` non Send Sync. Should I just unsafe impl Send + Sync, since I know that `AssetStorage` can only be accessed mutably at one location?

Rhuagh:
R? @Xaeroxe @torkleyy @jojolepro

amethyst_assets/src/storage.rs
torkleyy: No, the asset storage is too complex to get such an impl. I'm not able to review this currently.

Rhuagh: Do you have an alternative? I'd like to avoid doing an allocation every frame. I know for a fact that the `Vec` will only be used from a single location which is not called in parallel, so how do I tell rust not to worry about it ?

Rhuagh: Wrapping it in `Mutex` works, I can use `get_mut` and it won't be locked, and the compiler accepts `AssetStorage` as `Send + Sync` still.

amethyst_assets/src/storage.rs
```diff
@@ -130,120 +142,145 @@ impl<A: Asset> AssetStorage<A> {
...
+                        old_reload,
+                    } => {
+                        let (asset, reload_obj) = match data.map(|FormatValue { data, reload }| {
+                            (data, reload)
+                        }).and_then(|(d, rel)| f(d).map(|a| (a, rel)))
+                            .chain_err(|| ErrorKind::Asset(name.clone()))
+                        {
+                            Ok((ProcessingState::Loaded(x), r)) => (x, r),
+                            Ok(..) => {
+                                continue; // FIXME
```
torkleyy: This is unresolved

amethyst_assets/src/storage.rs
```diff
@@ -130,120 +142,145 @@ impl<A: Asset> AssetStorage<A> {
...
+                    reloads.push((handle.downgrade(), reload_obj));
                 }
-            };
+            }
 
-            // Add the reload obj if it is `Some`.
-            if let Some(reload_obj) = reload_obj {
-                reloads.push((handle.downgrade(), reload_obj));
+            for p in requeue.drain(..) {
+                self.processed.push(p);
```
torkleyy: I don't understand why you first push the elements to `requeue` just to `drain` them here..

Rhuagh: Because if I push them in msqueue I'll create an infinite loop

Rhuagh:
Basically : attempt to process all assets, if they aren't ready (because of a lower tier asset not having been loaded), requeue it for processing next frame. It uses a scratch vec because requeing directly would cause it to be reprocessed in the same frame over and over again.

Xaeroxe:
@Rhuagh I'm happy to r+ this if you'll rebase it real quick.  This one's been open long enough.

PR #770, opened 2018-06-04
feat: Add bundle for arc ball camera

Rhuagh:
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/770)

### DIFF
--- a/amethyst_controls/src/bundles.rs
+++ b/amethyst_controls/src/bundles.rs
@@ -35,11 +35,13 @@ impl<A, B> FlyControlBundle<A, B> {
             _marker: PhantomData,
         }
     }
+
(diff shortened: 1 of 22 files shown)


amethyst_controls/src/bundles.rs
```diff
@@ -76,3 +78,52 @@ where
         Ok(())
     }
 }
+
+/// The bundle that creates a flying movement system.
+/// Note: Will not actually create a moving entity. It will only register the needed resources and systems.
```
Xaeroxe: We should provide a small example of how to make said entity

Rhuagh: There's the full example I guess, would probably require quite a few lines of code to document it in the docs.

PR #774, opened 2018-06-06
Refactor UI to be more user friendly

Xaeroxe:
`ScaleMode` works now yay 😁 

This PR makes us more courteous towards the `UiTransform` local values.  We never overwrite the local values anymore. and it removes the conflict between modifications to global value and local values by preventing the end user from writing to the global values.  Global values should only be used to prevent re-calculation anyways.

`UiResize` has been refactored to be more user friendly and also more helpful in general.  Now we don't need to execute the internal closure on the transform before attaching it, and neither do we need to box the closure ourselves.

I also folded anchoring and stretching into the `UiTransform` itself.  Behavior without these components was unspecified before.

I updated some colors in the UI example as I didn't like the red, and since I can no longer see screen flickering I removed the seizure warning.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/774)

### DIFF
--- a/amethyst_ui/src/layout.rs
+++ b/amethyst_ui/src/layout.rs
@@ -5,8 +5,7 @@ use amethyst_core::specs::prelude::{BitSet, Entities, InsertedFlag, Join, Modifi
 use amethyst_core::{HierarchyEvent, Parent, ParentHierarchy};
 use amethyst_renderer::ScreenDimensions;
 
-/// Unused, will be implemented in a future PR.
-/// Indicated if the position and margins should be calculated in pixel or
+/// Indicates if the position and margins should be calculated in pixel or
 /// relative to their parent size.
 #[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize)]
 pub enum ScaleMode {
@@ -87,7 +86,7 @@ pub enum Stretch {
 /// It does almost the same as the `TransformSystem`, but with some differences,
 /// like `UiTransform` alignment and stretching.
 #[derive(Default)]
-pub struct UiParentSystem {
+pub struct UiTransformSystem {
     transform_modified: BitSet,
 
     inserted_transform_id: Option<ReaderId<InsertedFlag>>,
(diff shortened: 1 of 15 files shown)


Xaeroxe:
@Rhuagh I'd love to but `Stretched` has margins in it.

Rhuagh:
Ah, then add `Stretched` directly inside `UiTransform` in this PR :)

Xaeroxe:
Although now that you got me thinking about it in 2/3 of the cases half of the margins aren't used, so these margins probably should go in the enum.

Rhuagh:
Sure. Btw, I opened another PR on UI, if you want you can pull that into this PR instead, so we don't have to fight conflictomania.

amethyst_ui/src/layout.rs
```diff
@@ -91,262 +58,215 @@ impl Anchored {
...
+        /// The margin length for the height
+        y_margin: f32,
+    },
 }
 
 /// Manages the `Parent` component on entities having `UiTransform`
 /// It does almost the same as the `TransformSystem`, but with some differences,
 /// like `UiTransform` alignment and stretching.
 #[derive(Default)]
 pub struct UiParentSystem {
```
Rhuagh: `UiTransformSystem` maybe instead? `UiParentSystem` is a bit vague imho.

amethyst_ui/src/transform.rs
```diff
@@ -115,6 +132,27 @@ impl UiTransform {
         self.opaque = false;
         self
     }
+
+    /// Adds stretching to this ui element so it can fill its parent.
+    pub fn with_stretching(mut self, stretch: Stretch) -> Self {
```
Rhuagh: `with_stretch` ?

Might wanna add `with_anchor` also

Xaeroxe: I felt `with_anchor` was unnecessary because the field is public and I ask for the anchor in the constructor.  I'll fix the naming though.

Rhuagh:
Could we add a changelog entry here? :)

Xaeroxe:
Rebased and added a changelog.

Xaeroxe:
Well I certainly did.  I've been reviewing and testing my own code here pretty aggressively so I'm going to

bors r+

PR #776, opened 2018-06-07
Don't log warning if logger already set

randomPoison:
Currently, amethyst logs a warning if the user already registered a global logger before calling `ApplicationBuilder::new`. In practice, this isn't really an issue that the user needs to be warned about, it just means that the user chose to setup a different logging library than the one that amethyst provides by default. As such, this PR removes the warning so as not to confuse/worry users, while still allowing them to setup their own logging system if they so choose. I have put in a message at a lower verbosity so that there's still some record that the default logger was overridden.

I had originally [discussed in gitter](https://gitter.im/amethyst/engine?at=5b183d4be26c847ac8c1ec36) adding an option for not setting up the default logger at all, but it occurs to me that there's no need to explicitly disable the default logger, since the log crate already has a mechanism for safely rejecting redundant loggers. Automatically disabling the default logger if the user sets up their own is probably the cleanest, most ergonomic API amethyst could provide.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/776)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,6 +68,7 @@ thread_profiler = { version = "0.1", optional = true }
 
 [dev-dependencies]
 amethyst_gltf = { path = "amethyst_gltf", version = "0.2" }
+env_logger = "0.5.10"
 genmesh = "0.5"
 ron = "0.2"
 serde = "1.0"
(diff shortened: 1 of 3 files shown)


randomPoison:
Yeah, good point. I forgot this hadn't been merged yet, so I can add the docs to this PR instead of opening a new one.

randomPoison:
Alright, I've updated with documentation on the default logging functionality and instructions on how to override it if desired. This is ready for re-review.

Rhuagh:
Once #778 is pushed through, could you add a changelog entry for this PR ?

randomPoison:
Will do!

Rhuagh:
If you add a changelog entry, we can merge this.

PR #785, opened 2018-06-13
feat: Add shape generator prefabs

Rhuagh:


amethyst_renderer/src/formats/mod.rs
```diff
@@ -9,34 +9,63 @@ use amethyst_assets::{AssetPrefab, Format, PrefabData, PrefabError, ProgressCoun
...
 
 mod mesh;
 mod mtl;
 mod texture;
 
+/// Internal mesh loading
+///
+/// ### Type parameters:
+///
+/// `V`: Vertex format to use for generated `Mesh`es, must to be one of:
```
Xaeroxe: "must to be one of" doesn't quite sound right, should probably read "must be one of"

amethyst_renderer/src/formats/mod.rs
```diff
@@ -9,34 +9,63 @@ use amethyst_assets::{AssetPrefab, Format, PrefabData, PrefabError, ProgressCoun
...
+    Asset(AssetPrefab<Mesh, M>),
+    /// Generate a Mesh from basic type
+    Shape(ShapePrefab<V>),
+}
+
 /// `PrefabData` for loading graphics, ie `Mesh` + `Material`
 ///
 /// ### Type parameters:
 ///
+/// `V`: Vertex format to use for generated `Mesh`es, must to be one of:
```
Xaeroxe: Same as above

amethyst_renderer/src/formats/mod.rs
```diff
@@ -9,34 +9,63 @@ use amethyst_assets::{AssetPrefab, Format, PrefabData, PrefabError, ProgressCoun
...
+///     * `Vec<PosTex>`
+///     * `Vec<PosNormTex>`
+///     * `Vec<PosNormTangTex>`
+///     * `ComboMeshCreator`
 /// - `M`: `Format` to use for loading `Mesh`es from file
 /// - `T`: `Format` to use for loading `Texture`s from file
 #[derive(Deserialize, Serialize)]
-pub struct GraphicsPrefab<M, T>
+pub struct GraphicsPrefab<V, M = ObjFormat, T = TextureFormat>
```
Xaeroxe: I'd like to do something similar with the font format for the font prefab since we only support one format.

Rhuagh: Right, yeah. There's a `FontFormat` that contain the fonts we have formats for already, we should default to that

Rhuagh: The only sad part is you can't use type defaults in typedefs :/

Rhuagh: I can add that to the todo for the ui PR.

amethyst_renderer/src/shape.rs
```diff
@@ -1,13 +1,75 @@
...
+        ReadExpect<'a, Loader>,
+        WriteStorage<'a, Handle<Mesh>>,
+        Read<'a, AssetStorage<Mesh>>,
+    );
+    type Result = ();
+
+    fn load_prefab(
+        &self,
+        entity: Entity,
+        system_data: &mut Self::SystemData,
```
torkleyy: Please destructure here, makes the code below hard to read otherwise.

amethyst_renderer/src/shape.rs
```diff
@@ -1,13 +1,75 @@
...
+        system_data
+            .1
+            .insert(entity, self.handle.as_ref().unwrap().clone())
+            .map(|_| ())
+    }
+
+    fn trigger_sub_loading(
+        &mut self,
+        progress: &mut ProgressCounter,
+        system_data: &mut <Self as PrefabData>::SystemData,
```
torkleyy: Same as above.

examples/assets/prefab/sphere.ron
```diff
@@ -0,0 +1,50 @@
+#![enable(implicit_some)]
+Prefab (
+    entities: [
+        (
+            data: (
```
torkleyy: IMO we need field names here, the `None` values are not self-documenting otherwise (which I think is a useful thing).

Rhuagh: This is already fixed in a followup PR.

Rhuagh: With the followup PR, you don't need to add the None fields at all, you only have to add the fields that you actually need.

amethyst_renderer/src/shape.rs
```diff
@@ -1,13 +1,76 @@
...
 use {ComboMeshCreator, Mesh, MeshData, MeshHandle, Normal, PosNormTangTex, PosNormTex, PosTex,
      Position, Separate, Tangent, TexCoord};
 
+// TODO: enable Circle when PR is released in genmesh
+
+/// Prefab for generating `Mesh` from basic shapes
+///
+/// ### Type parameters:
+///
+/// `V`: Vertex format to use, must to be one of:
```
Xaeroxe: Similar phrasing here as before

PR #786, opened 2018-06-13
feat: Specialised UI prefab format

Rhuagh:
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/786)

### DIFF
--- a/amethyst_ui/src/prefab.rs
+++ b/amethyst_ui/src/prefab.rs
@@ -1,8 +1,12 @@
-use amethyst_assets::{AssetPrefab, Format, PrefabData, PrefabError, ProgressCounter};
-use amethyst_core::specs::prelude::{Entity, WriteStorage};
-use amethyst_renderer::{Texture, TextureMetadata, TexturePrefab};
+use amethyst_assets::{AssetPrefab, AssetStorage, Format, Handle, Loader, Prefab, PrefabData,
+                      PrefabError, PrefabLoaderSystem, Progress, ProgressCounter,
+                      Result as AssetResult, ResultExt, SimpleFormat};
+use amethyst_core::specs::prelude::{Entities, Entity, Read, ReadExpect, Write, WriteStorage};
+use amethyst_renderer::{Texture, TextureFormat, TextureMetadata, TexturePrefab};
+use serde::Deserialize;
 
-use {Anchor, FontAsset, MouseReactive, Stretch, TextEditing, UiImage, UiText, UiTransform};
+use {Anchor, FontAsset, FontFormat, MouseReactive, Stretch, TextEditing, UiFocused, UiImage,
+     UiText, UiTransform};
 
 /// Loadable `UiTransform` data
 #[derive(Debug, Clone, Deserialize, Serialize)]
@@ -144,7 +148,7 @@ impl<'a> PrefabData<'a> for UiTransformBuilder {
 ///
 /// - `F`: `Format` used for loading `FontAsset`
 #[derive(Deserialize, Serialize)]
-pub struct UiTextBuilder<F>
+pub struct UiTextBuilder<F = FontFormat>
 where
     F: Format<FontAsset, Options = ()>,
 {
@@ -176,6 +180,8 @@ pub struct TextEditingPrefab {
     pub selected_background_color: [f32; 4],
     /// Use block cursor instead of line cursor
     pub use_block_cursor: bool,
+    /// Set this text field as focused
+    pub focused: bool,
 }
 
 impl Default for TextEditingPrefab {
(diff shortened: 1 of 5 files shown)


amethyst_ui/src/prefab.rs
```diff
@@ -220,6 +228,9 @@ where
                     editing.use_block_cursor,
                 ),
             )?;
+            if editing.focused && system_data.3.entity.is_none() {
+                system_data.3.entity = Some(entity);
```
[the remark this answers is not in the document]
Rhuagh: Sure.

amethyst_ui/src/prefab.rs
```diff
@@ -176,6 +180,8 @@ pub struct TextEditingPrefab {
     pub selected_background_color: [f32; 4],
     /// Use block cursor instead of line cursor
     pub use_block_cursor: bool,
+    /// Set this text field as focused, will not be obeyed if a field is already focused
+    pub focused: bool,
```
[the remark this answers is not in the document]
Rhuagh: I debated that, I went with the safer route for now.

amethyst_ui/src/prefab.rs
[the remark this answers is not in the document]
Rhuagh: I'll fix this tonight sometime

amethyst_ui/src/prefab.rs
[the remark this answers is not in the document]
Xaeroxe: Well we'd want this true for something like a menu popup right?  It's supposed to get in your face and be ready to use.  Stealing focus redirects text input and there's almost always going to be something focused, even if the user isn't currently interacting with it.

amethyst_ui/src/prefab.rs
```diff
@@ -282,3 +293,242 @@ where
...
+    /// Container component, have a transform, an optional background image, and children
+    Container(
+        UiTransformBuilder,
+        Option<UiImageBuilder<I>>,
+        Vec<UiWidget<I, F>>,
+    ),
+    /// Image component, have a transform and an image
+    Image(UiTransformBuilder, UiImageBuilder<I>),
+    /// Text component, have a transform and text
+    Text(UiTransformBuilder, UiTextBuilder<F>),
```
torkleyy: Can we make these a struct-like variant? Looking at the RON file, the tuple style is a bit confusing at first. I generally think that tuples should be avoided in serialization except it's something that fits well (a coordinate for example, but nothing that's nested).

amethyst_ui/src/prefab.rs
```diff
@@ -282,3 +293,242 @@ where
...
+/// Load `UiPrefab` from `ron` file.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct UiFormat;
+
+impl<I, F> SimpleFormat<UiPrefab<I, F>> for UiFormat
+where
+    I: Format<Texture, Options = TextureMetadata> + Sync + for<'a> Deserialize<'a>,
+    F: Format<FontAsset, Options = ()> + Sync + for<'a> Deserialize<'a>,
+{
+    const NAME: &'static str = "Ui";
```
torkleyy: Correct me if I'm wrong, but I think we used a path for this so far.

Rhuagh: No, the others have single word names.

amethyst_ui/src/prefab.rs
Rhuagh: I don't know how to make the separate variants then ?

Rhuagh: Unless you mean that the data inside the variant should be a struct?

torkleyy: Just

```
Image {
    transform: ..,
    image: ..,
}
```

as far as I remember. So instead of a tuple variant, just a variant with fields.

Rhuagh: I will make it so

amethyst_ui/src/prefab.rs
Rhuagh: `const NAME: &'static str = "OGG";` as an example

examples/renderable/main.rs
```diff
@@ -446,7 +439,12 @@ fn main() -> Result<(), Error> {
     );
 
     let game_data = GameDataBuilder::default()
-        .with::<ExampleSystem>(ExampleSystem, "example_system", &[])
+        .with(
+            UiLoaderSystem::<TextureFormat, FontFormat>::default(),
+            "",
```
torkleyy: Any reason why you didn't give this an identifier?

Rhuagh: Not really, just didn't have anything that needed to depend on it.

Rhuagh:
The UiFormat is a simpler way of describing a UI setup than doing it with the generic prefabs. It's also an example of how to implement a custom prefab format.

PR #777, opened 2018-06-07
feat: Add shape generation

Rhuagh:
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/777)

PR #787, opened 2018-06-13
Add beginner-friendly sprite rendering utilities

Moxinilian:
This PR provides utilities to abstract away meshes and materials to beginners wanting to quickly attach sprites and textures to entity.

It provides a structure implementing ``SystemData`` to be requested in the context of systems. This struct exposes two methods: ``add(entity: Entity, sprite: Sprite, texture: TextureHandle, texture_size: (f32, f32))`` to attach a mesh for the corresponding sprite and spritesheet to the entity. The TextureHandle points to the spritesheet's texture. The texture size must be provided as the TextureHandle might not be loaded already, however I also added a ``texture.size()`` method on the ``Texture`` struct to easily get the size of a loaded texture.
``add_multiple(entities: Vec<Entity>, sprite: Sprite, texture: TextureHandle, texture_size: (f32, f32))`` does the same but for all entities in the vector using the same mesh, useful to save memory when you have a lot of identical sprites, in a tilemap for example.

Additionally, using ``WithSpriteRender`` implements on the ``EntityBuilder`` a method ``builder.with_sprite(sprite, texture, texture_size)`` that does the same with shorter syntax.

If your only goal is to display and move sprites around, this now can be done in a couple lines of code.
(I will squash the commits as soon as the review process ends in case I need to make other changes)

**Please tell me if the behavior for empty vectors in the ``multiple`` method is what you would expect, and if ``Error::Arguments`` is okay to have (I feel like it's too generic).**
I was tempted to make it do nothing if the Vec is empty, but receiving an empty Vec might also be a hint that something went wrong so one could prefer an error. I really am not sure.

**TL;DR**
```rust
use amethyst::renderer::sprite::SpriteRenderData;

world.exec(|mut data: SpriteRenderData| {
    data.add(entity, &sprite, texture, (64.0, 64.0)) // spritesheet size
}).unwrap();
```
```rust
use amethyst::renderer::sprite::SpriteRenderData;

world.exec(|mut data: SpriteRenderData| {
    data.add_multiple(vec![entity1, entity2], &sprite, texture, (64.0, 64.0))
}).unwrap();
```
```rust
use amethyst::renderer::sprite::WithSpriteRender;

world
    .create_entity()
    .with(GlobalTransform::default())
    .with(transform)
    .with_sprite(&sprite, texture, (64.0, 64.0))? // spritesheet size
    .build();
```

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/787)

### DIFF
--- a/examples/assets/mesh/LICENSE_MESHES.txt
+++ b/examples/assets/mesh/LICENSE_MESHES.txt
@@ -1,3 +1,7 @@
+teapot.obj
+==========
+
+
 This is the Mitsuba version of 'Utah Teapot' by Benedikt Bitterli, downloaded from https://benedikt-bitterli.me/resources/
 
 The original file may be obtained here: https://benedikt-bitterli.me/resources
@@ -8,3 +12,15 @@ attribution. However, crediting the artist is encouraged.
 
 For more information about the license, please see
 https://creativecommons.org/publicdomain/zero/1.0/
+
+
+
+puffy.gltf and puffy.bin
+========================
+
+Model and animation by Thomas Herzog. The source material can be found here: https://gitlab.com/karroffel/puffy-assets
+
+This work is licensed under a Creative Commons Attribution-NonCommercial 4.0 International License.
+
+For more information about the license, please see
+https://creativecommons.org/licenses/by-nc/4.0/


Moxinilian:
Oh, and it also just occured to me that I forgot to change the mesh to have its anchor be in the center.
I'm gonna commit this.

amethyst_renderer/src/sprite.rs
```diff
@@ -61,6 +68,145 @@ impl From<[f32; 4]> for Sprite {
...
+    fn with_sprite_render(
+        self,
+        sprite: Sprite,
+        texture: TextureHandle,
+        texture_size: (f32, f32),
+    ) -> Result<Self>;
+}
+
+impl<'a> WithSpriteRender for EntityBuilder<'a> {
+    fn with_sprite_render(
```
torkleyy: Isn't `with_sprite` enough? Also, please add an `add_sprite` variant.

Moxinilian: Well I assumed doing that would blur the line between the ``Sprite`` struct and the actual rendering process. Do you think it matters?

Moxinilian: Also, I believe that in the SpriteRenderData, the methods should be ``add`` and ``add_multiple`` instead.

Moxinilian:
@Xaeroxe feel free to review the updated pong example

Moxinilian:
Ah yes, but not tonight then :p

Rhuagh:
`Error::Arguments` is a bit too generic yes.

amethyst_renderer/src/sprite.rs
```diff
@@ -61,6 +68,143 @@ impl From<[f32; 4]> for Sprite {
...
+    pub fn add_sprite_render_multiple(
+        &mut self,
+        entities: Vec<Entity>,
+        sprite: &Sprite,
+        texture: TextureHandle,
+        texture_size: (f32, f32),
+    ) -> Result<()> {
+        let len = entities.len();
+        if len == 0 {
+            Err(Error::NoEntitySpecified)
```
torkleyy: Don't think this needs an error.

amethyst_renderer/src/tex.rs
```diff
@@ -46,6 +46,12 @@ impl Texture {
     pub fn view(&self) -> &RawShaderResourceView {
         &self.view
     }
+
+    /// Returns the texture's dimensions ``(width, height)``
+    pub fn size(&self) -> (usize, usize) {
+        let (w, h, _, _) = self.texture.get_info().kind.get_dimensions();
```
torkleyy: I'm not entirely sure about this, doesn't this invoke a driver function?

msiglreith: nop, not on the gfx-rs side

torkleyy: Okay, great thanks.

Xaeroxe: Yeah @msiglreith is correct, this merely traverses structures in RAM.  **Some** computation could be saved by calling this less frequently but this absolutely falls in the category of premature optimization imo.

Xaeroxe:
@torkleyy Yeah I was going to take care of that once this was merged.  I don't really want to ask @Moxinilian to work more on this.

torkleyy:
But why not move this into another example? I don't see the point in making things out of date just to correct them later on. Also, the current pong example does show useful concepts, as I think we agreed on already.

amethyst_renderer/src/tex.rs
Moxinilian: My first approach was to store the size in the Texture struct, but I cancelled that approach because it felt useless. I can redo it if you want me to, it's not hard.

Moxinilian:
As I said yesterday, I'm about to start reordering and expanding the pong example so that it covers sprite rendering. Does that solve your issue?

torkleyy:
It's great you are going to work on the pong tutorial! Still I'd like a low level example kept in the repo, which at the same time would fulfill the requirement that the book is up to date with our example.

Moxinilian:
I guess I can rebrand my testing playground then.

termhn:
I tend to agree with @torkleyy. I think we could either make a new version or just a completely new tutorial with this rather than changing the current Pong tutorial. We still want there to be an example of the fundamental concepts. And I'm not sure that it's good to teach people upfront a model which is kind of counter-intuitive to the way things "actually" happen in Amethyst.

Xaeroxe:
> kind of counter-intuitive to the way things "actually" happen in Amethyst.

Can you elaborate on this?  What are we presenting in this PR vs what "actually" happens?

This doesn't seem contradictory to me personally.

Moxinilian:
The point of such utility is to allow the user to not bother about how it actually works. An user wanting to make a 2D game probably does not care about mesh, they should focus on game design.

termhn:
Hm, I guess counter-intuitive might not be the best way to put it, I think the main thing is that one `.with_XXX` is actually adding multiple components and doing it transparently, which could be misleading to a beginner with Amethyst. I don't think this is really a problem, per-say, except in the context of using it as a learning tool. As long as this is explicitly stated/explained in the tutorial, it would be fine... and of course, the whole point is to let people who don't care about the functionality they're losing to do things easier.

Moxinilian:
Yes, it should never be assumed that ``with``-like syntax will add only one component.

Moxinilian:
What is the state of this PR? I have updated the pong tutorial, should I include it here or does it belong to another PR?

examples/pong/pong.rs
```diff
@@ -120,22 +123,51 @@ fn initialise_paddles(world: &mut World) {
...
 
-    // Create the mesh, material and translation.
-    let mesh = create_mesh(world, generate_circle_vertices(BALL_RADIUS, 16));
-    let material = create_colour_material(world, BALL_COLOUR);
+    // Create the translation.
     let mut local_transform = Transform::default();
     local_transform.translation = Vector3::new(ARENA_WIDTH / 2.0, ARENA_HEIGHT / 2.0, 0.0);
 
+    // Create the sprite for the ball.
+    let sprite = Sprite {
```
jojolepro: I'm really confused here.

amethyst_renderer/src/sprite.rs
```diff
@@ -61,6 +68,141 @@ impl From<[f32; 4]> for Sprite {
...
+        &mut self,
+        sprite: &Sprite,
+        texture: TextureHandle,
+        size: (f32, f32),
+    ) -> (MeshHandle, Material) {
+        use vertex::PosTex;
+        let half_width = (sprite.right - sprite.left) * 0.5;
+        let half_height = (sprite.bottom - sprite.top) * 0.5;
+
+        let vertices: Vec<PosTex> = vec![
```
jojolepro: We need to standardize this at some point. We could use genmesh or create our own crate containing those meshes that we re-use all the time.

amethyst_renderer/src/sprite.rs
```diff
@@ -61,6 +68,141 @@ impl From<[f32; 4]> for Sprite {
...
+pub trait WithSpriteRender
+where
+    Self: Sized,
+{
+    /// Adds a mesh and a material to the entity being built corresponding to the sprite and texture given.
+    fn with_sprite(
+        self,
+        sprite: &Sprite,
+        texture: TextureHandle,
+        texture_size: (f32, f32),
```
jojolepro: There isn't any way to delay the texture_size evaluation (where it is needed) until we check if the texture is loaded in the storage?

examples/pong/main.rs
```diff
@@ -23,15 +23,15 @@ use bundle::PongBundle;
...
+const PADDLE_HEIGHT: f32 = 16.0;
+const PADDLE_WIDTH: f32 = 4.0;
 const PADDLE_VELOCITY: f32 = 75.0;
-const PADDLE_COLOUR: [f32; 4] = [0.0, 0.0, 1.0, 1.0];
 
 const BALL_VELOCITY_X: f32 = 75.0;
 const BALL_VELOCITY_Y: f32 = 50.0;
-const BALL_RADIUS: f32 = 2.5;
-const BALL_COLOUR: [f32; 4] = [1.0, 0.0, 0.0, 1.0];
+const BALL_RADIUS: f32 = 2.0;
```
jojolepro: :+1:

examples/pong/pong.rs
```diff
@@ -18,9 +20,22 @@ impl<'a, 'b> State<GameData<'a, 'b>> for Pong {
         let StateData { world, .. } = data;
         use audio::initialise_audio;
 
+        // Load the spritesheet necessary to render the graphics.
+        let spritesheet = {
```
jojolepro: world.exec possible?

Moxinilian: Why?

examples/pong/pong.rs
Moxinilian: From my updated pong tutorial:
> We now need to define what part of the spritesheet we want to render.
> To do that, we need to create a `Sprite`, which is a fancy name to call a rectangle on
> the sprite sheet, before the creation of the entities.

amethyst_renderer/src/sprite.rs
Moxinilian: I couldn't find any unfortunately. All of them include having too much runtime. In case the user does not know the size of the texture at compile time, they should load the texture, get its size thanks to the method I added, and pass it to ``with_sprite``. It's not that bad.

examples/pong/pong.rs
jojolepro: I was talking about the top left right down coordinates. Sorry I didn't ping the correct line.

examples/pong/pong.rs
jojolepro: 1) Cleaner
2) Torkley will ask you to add it, like he did on my pr ;)

examples/pong/pong.rs
Moxinilian: Ah yes, just give a look at a sprite sheet and it'll be much clearer.
Maybe I could have it be simpler. Any idea?

amethyst_renderer/src/sprite.rs
Moxinilian: Maybe we could have SpriteMesh asset that once loaded can be unwrapped and added to the entity. But that's not less work on the user, it's just different. And it also adds more API which is already a lot for such a simple task.

examples/pong/pong.rs
jojolepro: I get it now. Personally I would not use a spritesheet for different / unrelated sprite. If you want to keep it like that its fine, but add a comment explaining how the spritesheet is layouted.

amethyst_renderer/src/sprite.rs
jojolepro: https://github.com/amethyst/amethyst/pull/777

Moxinilian: Ah it's been merged. Maybe I can rebase and use it then.

examples/pong/pong.rs
Moxinilian: The thing is it then requires to introduce it in the tutorial before systems, which I feel is inappropriate...

Moxinilian:
Seems like crates.io is having issues and is breaking CI...

Moxinilian:
This is my worst experience with git so far.
I managed to make the PR actually of the correct size, but I can't squash the merge commit with the main commit.
Why are we not using the built-in GitHub feature that squashes all the commits of a PR before merging?

Xaeroxe:
@Moxinilian I'm going to take a stab at rebasing and squashing this

Xaeroxe:
My bad, working on fixing this now

Xaeroxe:
Welp I had permission to push to this branch until I accidentally closed the PR by pushing up 0 commits.  I figured out my mistake, but I can't fix it because I no longer have access to @Moxinilian 's repo, so I'll re-open this PR from my branch.

PR #791, opened 2018-06-14
feat: Add basic scene utility prefab

Rhuagh:
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/791)

### DIFF
--- a/examples/prefab/resources/display_config.ron
+++ b/examples/prefab/resources/display_config.ron
@@ -1,10 +1,4 @@
 (
-  dimensions: None,
-  max_dimensions: None,
-  min_dimensions: None,
-  fullscreen: false,
   multisampling: 0,
   title: "Prefab example",
-  visibility: true,
-  vsync: true,
 )
--- a/amethyst_controls/src/lib.rs
+++ b/amethyst_controls/src/lib.rs
@@ -16,7 +16,7 @@ mod resources;
 mod systems;
 
 pub use self::bundles::{ArcBallControlBundle, FlyControlBundle};
-pub use self::components::{ArcBallControlTag, FlyControlTag};
+pub use self::components::{ArcBallControlTag, ControlTagPrefab, FlyControlTag};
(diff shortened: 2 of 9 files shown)


examples/prefab/resources/display_config.ron
```diff
@@ -1,10 +1,4 @@
 (
-  dimensions: None,
-  max_dimensions: None,
-  min_dimensions: None,
-  fullscreen: false,
   multisampling: 0,
   title: "Prefab example",
-  visibility: true,
-  vsync: true,
```
Xaeroxe: Why are we modifying this display config?

Xaeroxe: To demonstrate how to use it so that when people want to know where the features are they can find them.

Rhuagh: There are 15 other examples with full config, I don't see the point of having it in full mode in all of them.

Xaeroxe: Ok, that's fine with me then

torkleyy:
Wish that actually worked :P

Moxinilian:
I'm sure bors is accepting PRs :p

Xaeroxe:
Most active reviewers as of this moment have approved this.. so I'm gonna bend the rules a bit.

bors r+

Moxinilian:
Yeah, the contributing guidelines should have an unanimity close to bypass the waiting time.

Xaeroxe:
Wording such a clause would get difficult, because we don't have proper unanimous agreement here we just have unanimity among everyone who I believe is going to review this.  I could be wrong so I'm hesitant to codify that.  I'm just knowingly and willingly breaking my own rules here because I think this PR is unlikely to cause controversy, and determining likelihood to cause controversy is also very holistic.

Xaeroxe:
And if I ever break the rules when I shouldn't I'm confident my fellow contributors will be eager to inform me 😛

PR #794, opened 2018-06-15
Make point lights use GlobalTransform instead of separate `center`

### LINKED ISSUE #782: PointLights can only be expressed using global coordinates
When rendered using the shaded pass, the positions of PointLights are expressed in global coordinates and thus are independent from the entity's GlobalTransform. Either removing the PointLight's position and only depending on GlobalTransforms or having it be relative to the GlobalTransform would make a lot more sense.

termhn:
Fixes #782. ~~Seems to work with the separate_sphere example which I updated, the rest of the examples need to be fixed though, likely.~~ I believe I've updated all the examples to work properly.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/794)

### DIFF
--- a/amethyst_renderer/src/pass/shaded_util.rs
+++ b/amethyst_renderer/src/pass/shaded_util.rs
@@ -2,8 +2,6 @@ use std::mem;
 
 use amethyst_core::specs::prelude::{Join, ReadStorage};
 use amethyst_core::GlobalTransform;
-use amethyst_core::cgmath;
-type Vector4 = cgmath::Vector4<f32>;
 
 use glsl_layout::*;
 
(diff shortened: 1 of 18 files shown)


amethyst_renderer/src/pass/shaded_util.rs
```diff
@@ -35,16 +37,17 @@ pub(crate) fn set_light_args(
...
-    let point_lights: Vec<_> = light
+    let point_lights: Vec<_> = (light, global)
         .join()
-        .filter_map(|light| {
+        .filter_map(|(light, transform)| {
             if let Light::Point(ref light) = *light {
                 Some(
                     PointLightPod {
-                        position: light.center.into(),
+                        position: (transform.0 * Vector4::new(0.0, 0.0, 0.0, 1.0)).truncate().into(),
```
Xaeroxe: I'm pretty sure we should remove this position field as well

termhn: I don't think so because this is where the data is actually fed into the shaders. The shaders rely on having an array of data about each light and there's not really a point in having them take an actual matrix for a point light in which the rotation/scale doesn't actually affect anything about it. If we fed the matrix into the shader then we'd have to do this matrix -> position conversion for every fragment for every light.

Rhuagh: I agree, we should feed it like this. But can we instead use transform_point

termhn:
rebased on develop and added the changelog/doc changes :)

PR #798, opened 2018-06-22
Promote `UiButton` to a fundamental Ui component

Xaeroxe:
This PR rethinks the way we do `UiButton`s in order to add more features.

- `UiButton` is now also a component similar to `UiImage` and `UiText`.  This enables...
- Buttons can now optionally switch images and text colors when hovered or clicked
- Buttons can now optionally make a noise when hovered or clicked

This also required some changes to the prefab for buttons, and additionally a formatting pass was done in this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/798)


Xaeroxe:
@jojolepro did you even look at the changes?  `UiButton` is an interface to `UiImage` and `UiText` in this instance.  I created a system that switches the images and text colors for the `UiImage` and `UiText` in response to events.  `UiButton` only contains the things unique to a button.

jojolepro:
I didn't check actually, I will now.

amethyst_ui/src/button/builder.rs
```diff
@@ -53,6 +50,12 @@ pub struct UiButtonBuilder {
     font_size: f32,
     image: Option<TextureHandle>,
     parent: Option<Entity>,
+    hover_image: Option<TextureHandle>,
+    hover_text_color: Option<[f32; 4]>,
+    press_image: Option<TextureHandle>,
+    press_text_color: Option<[f32; 4]>,
+    hover_sound: Option<SourceHandle>,
+    press_sound: Option<SourceHandle>,
```
jojolepro: This is what I said I don't like. The separation of concern isn't there. You have audio data at the same place as visual and layouting. While I understand where you are coming from, I still think it would be better to have separated components than a big struct of Option.

When we'll be doing other elements (checkbox for example), you'll need to copy paste the data for all of those in the checkbox builder.

Tell me if I'm not being clear with what I'm trying to say.

amethyst_ui/src/button/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+    /// if possible.
+    pub fn new(
+        normal_text_color: [f32; 4],
+        normal_image: TextureHandle,
+        hover_image: Option<TextureHandle>,
+        hover_text_color: Option<[f32; 4]>,
+        press_image: Option<TextureHandle>,
+        press_text_color: Option<[f32; 4]>,
+        hover_sound: Option<SourceHandle>,
+        press_sound: Option<SourceHandle>,
```
jojolepro: Instead of 1 builder per ui element,
you'll now have 1 builder, one struct and one system.

jojolepro: prefab: If you only use composition of the basic ui elements and effect components, you can create presets (like a button) that you can reference after. When using 1 struct per element type, you are much more limited with what you can do (from the user perspective that is, because you need to either recode the button for your needs, or change it, which both require a fork)

Xaeroxe: A composition based approach here gives redundancy in the prefab files.  You now have to specify the normal image and the normal text color twice.  That's what I'm trying to avoid here.

Xaeroxe: And to be honest maybe we should have different builders for common UI elements.  They're convenience structures after all and requiring users to compose their own UI from such low level primitives introduces more cognitive overhead.

jojolepro: You could make a wrapper that starts with your current builder and decomposes it into basic elements. Then in the prefabs you have the choice of going low level/composable or high level/configurable

jojolepro: >And to be honest maybe we should have different builders for common UI elements. They're convenience structures after all and requiring users to compose their own UI from such low level primitives introduces more cognitive overhead.

Exactly. Internally (what interacts with systems) should only be low level components. What the user interfaces with should be either the low level or a higher level (previous comment)

Xaeroxe: That's more or less what this PR does is it not?

jojolepro: https://github.com/amethyst/amethyst/pull/798/files#diff-697d04ad45f512eac4d00b0353d63193R30

This is my issue. It is a internal system that uses a high level struct regrouping data that shouldn't be together.

Xaeroxe: It's not possible to implement these features without that.  Where is button specific stuff supposed to be stored?  `UiImage` is too simple to permit that.

jojolepro: >Component::OnHoverImage
System::OnHoverImageSystem
Component::OnHoverSound
System::OnHoverSound
struct UiButtonBuilder{
  hoverImage: Option<..>,
  .....
  ....
}
impl UiBuilder for UiButtonBuilder{
  fn build(self,world: &mut World) -> List<Entity>{
    world.create_entity().add(OnHoverImage).add(OnHoverSound)
  }
}

Xaeroxe: It feels like more components for the sake of more components.  What do we gain from breaking it out like that?  Where would you use these other than on a button?  Keeping relevant data together helps with cognitive load and prevents `SystemData` tuples from getting obtusely large.  

I just worry it'd be `Anchor` and `Stretch` all over again.

jojolepro: OnHoverImage
Used for: UiButton, UiCheckbox, UiDropdownList, UiTableCell, UiTab, UiPopup. So its pretty re-usable I think.

Now that I think about it, it might make more sense if they were not *that* separated.
Like anchor and stretch should have been directly into UiTransform.

Maybe OnUiActionSound(uievent, sound handle), OnUiActionImage(uievent, image)?

Xaeroxe: Ok, that seems a bit more reasonable to me, but I do wonder about `OnUiActionText`.  In order to mimic current behavior it'd have to operate on the children of the `Entity` it's attached to, rather than the `Entity` it's currently attached to.

jojolepro: Okay, so the fanciest that was actually doable was this:
```
OnUiAction<UiEventType,ComponentType>(target:Entity, |event:EventType, &mut component: ComponentType| -> ());
```
This works to edit components.

To perform side effects (playing sounds for example), you need to be more specific.
OnUiActionSound<UiEventType>(sound: AudioHandle)

Xaeroxe:
@jojolepro I've broken out a few more components, let me know what you think.

Xaeroxe:
Rebased and ready for review.

amethyst_ui/src/action_components.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+/// When this component is added to a UI element
+/// it will play sounds based on mouse interaction.
+/// Requires `MouseReactive`.
+pub struct OnUiActionSound {
+    /// Sound made when this button is hovered over
+    pub(crate) hover_sound: Option<SourceHandle>,
+    /// Sound made when this button is pressed.
+    pub(crate) press_sound: Option<SourceHandle>,
```
jojolepro: release sound too? :)

amethyst_ui/src/action_components.rs
```diff
@@ -0,0 +1,58 @@
...
+    /// Image used when the mouse hovers over this element
+    pub(crate) hover_image: Option<TextureHandle>,
+    /// Image used when element is pressed
+    pub(crate) press_image: Option<TextureHandle>,
+}
+
+impl OnUiActionImage {
+    /// A constructor for this component
+    pub fn new(
+        normal_image: TextureHandle,
```
jojolepro: what if you don't want an image except you have your mouse over? empty image vs Option?

amethyst_ui/src/button/builder.rs
```diff
@@ -1,7 +1,10 @@
-use super::{Anchor, FontAsset, FontHandle, MouseReactive, Stretch, TtfFormat, UiImage, UiText,
-            UiTransform};
-///! A clickable button.
+use super::super::{
```
jojolepro: is there a better way than super::supper?

amethyst_ui/src/button/system.rs
```diff
@@ -0,0 +1,164 @@
...
+impl UiButtonSystem {
+    /// Creates a new instance of this structure
+    pub fn new() -> Self {
+        Self::default()
+    }
+}
+
+impl<'s> System<'s> for UiButtonSystem {
+    type SystemData = (
+        ReadStorage<'s, UiButton>,
```
jojolepro: that's still the same, but if you think its better then I'll let it pass ;)

amethyst_config/src/lib.rs
```diff
@@ -2,7 +2,6 @@
 //!
 
 #![crate_name = "amethyst_config"]
-#![crate_type = "lib"]
```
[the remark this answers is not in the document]
Xaeroxe: Not necessary, and rustc complained about it at one point.

amethyst_ui/src/action_components.rs
Xaeroxe: Good point, I'll add it

amethyst_ui/src/action_components.rs
Xaeroxe: Sure, sounds good!

amethyst_ui/src/button/builder.rs
Xaeroxe: Yeah, I'll use an absolute path

amethyst_ui/src/action_components.rs
Xaeroxe: Ah yeah about that, when I come back to this I find that this one is a bit of a misnomer.  We're already using this sound as the release sound.  I'll fix the name.

jojolepro: Ya, but can't you have one for click start and one for click end (and one for click hold? :D )

Xaeroxe: I haven't seen any game UI that does those things, I'm not sure there's a strong use case for it.

jojolepro: age of empires 3 does this. It does add a lot of feel to the buttons when there's a release sound that is different from the press sound.

amethyst_ui/src/action_components.rs
Xaeroxe: A truly empty texture is pretty difficult.  The best I can do is a single pixel with alpha 0.

jojolepro: Adding and removing the sprite or texture or mesh component isn't possible from a System?

Xaeroxe: It is, it just results in kind of sloppy code because we have to account for the possibility of the component not being there when performing the updates.  The behavior flow chart gets a bit out of control.

jojolepro:
Status?

Xaeroxe:
Pretty stale but fixable.  Waiting for me to find time and motivation to work on it mostly.

Xaeroxe:
Rebased, gonna address comments now.

jojolepro:
Okay, tell me when it is ready for a review :)

azriel91:
@Xaeroxe yo! I'm looking forward to this PR's additional feature, are you close to completing it?

Xaeroxe:
Yes I am, I just need to put in some time to do it.  I'll do that today, sorry it's been so long.  I've got a lot of personal stuff going on right now.

jojolepro:
yes - reviewing

jojolepro:
I ran cargo fmt and got this
```
Warning: can't set `reorder_imports = true`, unstable features are only available in nightly channel.
error: expected one of `.`, `;`, `?`, `}`, or an operator, found `self`
   --> /home/jojolepro/prog-remote/fmtcheck/amethyst_ui/src/prefab.rs:411:9
    |
410 |         self.hover_sound.trigger_sub_loading(progress, sounds)
    |                                                               - expected one of `.`, `;`, `?`, `}`, or an operator here
411 |         self.release_sound.trigger_sub_loading(progress, sounds)
    |         ^^^^ unexpected token
```
jojolepro@JojoleproServer-1:~/prog-remote/fmtcheck$ cargo fmt --version
rustfmt 0.6.1-stable (49279d7 2018-05-08)

Xaeroxe:
Woops, my bad I went and made changes without testing them

jojolepro:
Also, there are two files that needs a fmt pass
	modified:   amethyst_ui/src/button/builder.rs
	modified:   amethyst_ui/src/prefab.rs

Xaeroxe:
There, should be good now.

Xaeroxe:
And with that

bors r+

:tada:

jojolepro:
Are you not missing a second review though?

Xaeroxe:
With activity rates being what they are right now if we wait for that we're never going to get things merged.

jojolepro:
true :(

jojolepro:
bors retry

Xaeroxe:
It was just about finished, not sure why we did that.

jojolepro:
idk, its been 40 minutes since the appveyor build was done, so I thought bors froze

Xaeroxe:
No, it was still working on the second appveyor build for the post merge.  It's finished now though so I'll just override.

bors r-

jojolepro:
ah sorry about that ^^

PR #800, opened 2018-06-22
Add MP3 support

Xaeroxe:
This PR adds MP3 support to the engine by upgrading the rodio version.

These changes required me to produce an `unsafe impl Sync for AudioEmitter`.  I've verified this implementation is correct and won't cause problems.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/800)

### DIFF
--- a/amethyst_audio/Cargo.toml
+++ b/amethyst_audio/Cargo.toml
@@ -23,7 +23,7 @@ amethyst_assets = { path = "../amethyst_assets", version = "0.3.0"}
 amethyst_core = { path = "../amethyst_core", version = "0.2.0"}
 cpal = "0.8"
 log = "0.4"
-rodio = "0.7"
+rodio = "0.8"
 serde = { version = "1.0", features = ["derive"] }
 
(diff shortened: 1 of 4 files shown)


Xaeroxe:
@torkleyy After going through several layers of dependencies you'll find this thing in the MP3 decoder: https://docs.rs/slice-deque/0.1.8/slice_deque/struct.SliceDeque.html

As far as I can tell this structure should be Send + Sync but it wasn't manually written in by the author.  I created an issue on the repository for this crate asking the author if this should be Send + Sync, as they may know something I don't.  However, I believe this is safe.

torkleyy:
Can we wait for a response there and in case we merge this before a fix was released add a TODO?

Xaeroxe:
Sure!

Xaeroxe:
https://github.com/gnzlbg/slice_deque/pull/40#issuecomment-399785830 !  :tada: the MP3 encoder should auto implement Send + Sync now!

PR #808, opened 2018-07-02
Loosen up generic type bounds for InputBundle

Frizi:
This allows input axes and actions to not implement Default, which is particularly useful for enums.
Serialization traits are also not necessary if bindings are never intended to be loaded from file.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/808)

### DIFF
--- a/amethyst_input/src/bundle.rs
+++ b/amethyst_input/src/bundle.rs
@@ -27,7 +27,8 @@ use {Bindings, InputSystem};
 ///
 /// No errors returned from this bundle.
 ///
-#[derive(Default)]
+#[derive(Derivative)]
+#[derivative(Default(bound = ""))]
 pub struct InputBundle<AX, AC>
 where
     AX: Hash + Eq,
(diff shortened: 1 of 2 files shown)


amethyst_input/src/bundle.rs
```diff
@@ -27,7 +27,8 @@ use {Bindings, InputSystem};
 ///
 /// No errors returned from this bundle.
 ///
-#[derive(Default)]
+#[derive(Derivative)]
+#[derivative(Default(bound = "AX: Hash + Eq, AC: Hash + Eq"))]
```
Xaeroxe: Please leave this bound empty, otherwise the bound shows up twice for the `Default` implementation in rustdoc.

Xaeroxe: `bound = ""` is a good way to do this.

Frizi: 👍

Rhuagh:
Hmm, bindings not supposed to be loaded from file? I thought that would be a primary requirement?

Frizi:
@Rhuagh Why? It does make sense for it to be loadable in general and that should be supported by the engine, but It doesn't necessarily have to be a requirement from the user code.

Rhuagh:
Ah, I didn't read the code, just looked at your comment, which I read as "there's no need to ever load bindings from file".

PR #811, opened 2018-07-05
Make with_bindings_from_file return a Result

### LINKED ISSUE #742: InputBundle loading does not allow for checking for success
With `InputBundle::<String, String>::new().with_bindings_from_file(&binding_path)` there is no way to check if the load succeeded or not. This ends up meaning that my program will go on without any key bindings with no indicators besides a single line in the logs.

This made for an unpleasant experience while following the tutorial as I had mistyped `Up` as `up` in my `.ron` file and couldn't didn't see the log message.

It would be great if `with_bindings_from_file` returned a result and let the program handle it instead of using the default empty bindings.

For a more reliable startup I ended up going with:
```rust
let bindings = Bindings::load_no_fallback::<&str>(binding_path)
        .map_err(|e| amethyst::Error::Config(e))?;
let input_bundle = InputBundle::<String, String>::new().with_bindings(bindings);
```

On another note it might be nice if `amethyst::Error` implemented `From<ConfigError>`, should I make an issue for that?

magnonellie:
Fixes #742

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/811)

### DIFF
--- a/amethyst_input/src/bundle.rs
+++ b/amethyst_input/src/bundle.rs
@@ -2,8 +2,9 @@
 
 use std::hash::Hash;
 use std::path::Path;
+use std::result::Result as StdResult;
 
-use amethyst_config::Config;
+use amethyst_config::{Config, ConfigError};
 use amethyst_core::bundle::{Result, SystemBundle};
 use amethyst_core::specs::prelude::DispatcherBuilder;
 use serde::de::DeserializeOwned;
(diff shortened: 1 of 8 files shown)


stenverbois:
Don't forget to add an entry to CHANGELOG.md since this is a breaking change.

jojolepro:
Good  idea! :+1:

jojolepro:
You also need to add the link corresponding to your pr number.

this thing
[#811]: https://github.com/amethyst/amethyst/pull/811

jojolepro:
Thanks!

bors r+

PR #815, opened 2018-07-08
Logging improvements

### LINKED ISSUE #813: Log messages lost before ApplicationBuilder::new is called
Fern is currently initialized in the `ApplicationBuilder::new` function and since a lot of initialization is usually done before the user starts building the application object, that may potentially cause lost log messages containing errors which isn't a good user experience.

magnonellie:
This PR contains three changes:

- Various `println!()` and `eprintln!()` statements have been converted to use either `debug!()`, `info!()`, `warn!()` or `error!()`.
- Initialization of `fern` has been moved to its separate module and function which can be called before building of the `Application` object starts (fixes #813).
- The logger now takes a configuration object where you can specify if you want colored output (new!) and which log level you're interested in. This could potentially be expanded in the future with features such as ignoring logs from specific crates.

Because initializing the logger is now optional, a logger is checked for when the `Application` object starts being built. If not found, a warning is produced.

Color output has been tested on Linux (`gnome-terminal`, IntelliJ) and Windows (`mintty`, IntelliJ). Despite not working with `cmd.exe`, it doesn't produce any noise when color output is enabled there. While not tested on macOS, I can't imagine it not working with it.


src/app.rs
```diff
@@ -377,21 +375,11 @@ impl<S> ApplicationBuilder<S> {
...
-                ))
-            })
-            .level(LevelFilter::Debug)
-            .chain(io::stdout())
-            .apply()
-            .unwrap_or_else(|_| {
-                debug!("Global logger already set, default amethyst logger will not be used")
-            });
+        if !log_enabled!(Level::Error) {
+            println!(
```
azriel91: It's probably worth `eprintln`

amethyst_config/src/lib.rs
```diff
@@ -115,7 +117,7 @@ where
 {
     fn load<P: AsRef<Path>>(path: P) -> Self {
         Self::load_no_fallback(path.as_ref()).unwrap_or_else(|e| {
-            println!("1: Failed to load config: {}", e);
+            error!("Failed to load config: {}", e);
```
stenverbois: On Linux at least, this error just says `No such file or directory`. It would be very helpful to print the path here as well.

magnonellie:
That should be everything resolved thus far. Thanks for reviewing everyone! <3

Xaeroxe:
LGTM

bors r+

Stargateur:
"Because initializing the logger is now optional, a logger is checked for when the Application object starts being built. If not found, a warning is produced."

That is counter intuitive the purpose of log is to be optionnel, something optionnel shouldn't produce a warning. Also that why log implement a default loger that do nothing, with your method you warning a user even if a user is perfectly allowed to not want any log.

magnonellie:
@Stargateur I see what you're saying. The thing is that you should always want to have some sort of logger in your application. The reasoning for making the amethyst logger optional was twofold. First, it's so that you can run whatever logger you want to, basically you manage logging instead of Amethyst in which case the warning wouldn't be produced as another logger can be found. The second reasoning is kind of stupid in a way, but it's that the way it was done before, all log messages produced before the `Application` object was created would be lost which meant we had to add a function that could be called before that so a logger could be created that catches potential error messages. Since you can't mark functions as "this has to be called", the logger essentially became optional whether we wanted it or not.

Another thing of mention here is that you are correct that libraries shouldn't produce any output if that's not the purpose of them, but Amethyst can't actually be classified as a pure library anymore. Amethyst is a framework right now and most frameworks do actually take the liberty of throwing "helpful" output at you. If you take a look at the world of JavaScript frameworks, you can see that almost all of them have some kind of "don't run this package in production" or "download the dev tools for a better experience" messages.

With all that said, if you do have a better solution to handle this, we're more than welcoming of a PR! 😄

Stargateur:
Actually there is not much to do, just remove this warning/remove all print of amethyst (as you did). The old behavior where amethyst set itself the logger was completely (sorry) stupid and you have do a very good job to put it as a helper function (Well, I don't think it necessary but that don't really hurt and can help "noob" to have a quick logger setup).

But I really really don't agree about taking javascript as an exemple, it's a completely different word with different limitation and quality. A javascript lib don't have much option for help a user / to report log message, Rust have one of the best solution, that you use it. So its very strange to see that despite you use this solution, you don't follow the typical use case (You already fixed the most problematic point so it's ok tier now).

Also, Amethyst being a "Framework" don't change anything, it's just a very big library that help you a lot to structure your project. This is not an good argument to brake (good) convention.

But, I just a random dev who have see this and was very surprise to see this behavior in a Rust library, and as amethyst is getting a lot of success and still in beta, I tried to put my little advise to make a proper library. I understand if Amethyst dev don't agree, just I think is very odd in my personal opinion.

Xaeroxe:
> The old behavior where amethyst set itself the logger was completely (no offence) stupid

I don't agree with this characterization.  I can understand why we feel a library shouldn't be logging, which is when it becomes important to acknowledge Amethyst behaves more like a standalone executable than a library.  We control the flow of execution for our users while they're running Amethyst.  Really we want to be an executable but for technological reasons we are a library.

Registering a default logger was done so that new users can easily see the logging output, which contains a lot of helpful info for the developers.  Amethyst sometimes can't recognize certain patterns as errors until runtime, so it's useful for us to be able to display those errors to the user without them having to go hunting for it.

PR #816, opened 2018-07-09
Refactor amethyst_input, replacing several wrapper types by impl trait.

whentze:
[This comment](https://github.com/amethyst/amethyst/blob/develop/amethyst_input/src/lib.rs#L39) mentions that the wrapper types defined in amethyst_input/lib.rs are to be removed when impl trait is stabilized.
As this is now the case (since rust 1.26), I've implemented the suggested change in this PR.

Sadly, this change is necessarily a breaking one, since it involves altering the signature of public fns.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/816)


jojolepro:
btw, missing changelog entry for breaking change. Please add an entry with the link to this PR in the changelog file. Thanks!

whentze:
Changelog entries are done :+1:

jojolepro:
Thanks a lot! :)

bors r+

PR #817, opened 2018-07-09
Bump specs version to 0.12

stenverbois:
Bumps specs dependency version to 0.12.

Specs now requires the `Builder` trait to be in scope to create entities via the regular builder. I added the import to the prelude since it is almost always required (and can be confusing otherwise since its an invisible trait requirement).

~~Blocked on [specs-hierarchy#5](https://github.com/rustgd/specs-hierarchy/pull/5).~~

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/817)


stenverbois:
This is ready now.

Issue #824, opened 2018-07-21
[RFC] System name constants


azriel91:
Hey, I had to write this out so that I don't forget it.

---

When a bundle adds a system to a dispatcher, it provides a name which is used in system dependency ordering. The system name is effectively API, as external systems that depend on that system specify it as a dependency.

While the number of systems is small, it isn't hard to maintain a few `&'static str`s. However when there are many systems, if I make a typo, or the name has changed, using `&str`s defers failure to application startup (runtime) instead of compile time. If we can use constants or a function that returns a derivable system name, it would decrease the maintenance cost for larger applications.

Things to consider:

* `System`s are generally not public outside of a crate, so we'd have to expose the name some other way.
* What about Systems with generic parameters, such as `AnimationControlSystem<I, T>`, where the type parameters are defined by the consumer / application code

Possible useful crates:

* [`named_type`](https://github.com/cjhowe7/named_type/), but would depend on https://github.com/cjhowe7/named_type/issues/5

Alternatives:

* Do nothing &mdash; perhaps the maintenance cost isn't big enough for this effort to be undertaken.

AnneKitsune:
Currently I'm not sure this is something worth changing.
Since you usually declare all your `System`s in the main (for small to medium games. big requires states containing other dispatchers, but all systems are still declared at the same place), it is pretty easy to see what goes where, except with bundles where you have to dig in the code to find the name of the `System`s. Maybe a doc update including the names in the bundle doc would be helpful though?

For example, this https://github.com/amethyst/amethyst/blob/b24355d36fc728b19eb4f1ccebfe805acc6ffd21/amethyst_core/src/transform/bundle.rs#L12
is missing a "parent_hierarchy_system" entry

AnneKitsune:
You actually have to implement this in the shred crate. If it works there, then on the amethyst side we can add it too.

AnneKitsune:
I thought about it, and it may make sense to change the 'static str to a T: Eq in some cases.

I'm not sure how you could make this work with the engine's bundles though.

AnneKitsune:
@torkleyy do you want to take note of this somewhere? Should it be reopened as an issue on the shred or specs repo?

azriel91:
For now I'm happy to close until we decide to engineer either or both of these:

[...]

PR #827, opened 2018-07-28
Do Not Merge: Experiment using `typename` to derive system names.

azriel91:
Seeing what using the `typename` crate to get programmatic system names does. I don't like how it means downstream crates need to depend on `typename` to get to use it (they need to import the trait), and animated types need to `#[derive(TypeName)]` as well.

Issue #824

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/827)


azriel91:
Closing for now, but put here for tracking / remembering later.

jojolepro:
@Rhuagh Which are? To me, config should be data-local and not system local.

Issue #829, opened 2018-07-31
Sprite Render Pass and Sprite Animation


azriel91:
I propose a new render pass with its own shaders specifically for rendering sprites.

## Motivation

The current way sprites are drawn requires consumers to create and attach the following components:

```rust
let material = // Material { albedo: <texture>, ..MaterialDefaults.0.clone() }
let mesh = // either use the `with_sprite` helper, or make your own

let entity = world.create_entity()
    .with(mesh_handle)
    .with(material)
    .with(global_transform)
    .build();
```

This usage interface is unergonomic for at least the following reasons:

* **Boilerplate:** The `Material` concept isn't used except to be an adaptor to the 3D rendering interface
* **Partial objects:** Need to create two objects for one purpose
* **Animation:** Sprites with different dimensions need the mesh to mutate in sync with the material

## Proposal

Provide another render pass that removes the need for users to write their own boilerplate, but still achieves the same result.

## How

Some of this is based on guidance from @omni-viral in the gitter chat.

* Write shaders to draw two triangles for the sprite and map the texture coordinates.
* Write a render pass that wires through the sprite information for each entity to the shaders.
* Provide a `Component` that allows the render pass to work out the following:
    - which texture
    - sprite layout information (sprite dimensions, the sprite's texture coordinates)

Sprite animation will be mutation of the said `Component`, which means the texture and layout information all mutate at the same time.

## Tasks

* [x] experiment (#830)
* [x] sprite offsets
* [x] sprite mirroring
* [x] sprite animation

## Bikeshedding

Thing | Current name(s) | Proposed
--- | --- | ---
Render passes | `DrawSprite`, `DrawSpriteSeparate` |
Component | `SpriteRender` |
Vertex shader | `sprite.glsl` |
Fragment shader | `sprite.glsl` |
... | ... | ...

## Alternatives

Some of these are not mutually exclusive:

* Do nothing (always an option =p).
* Provide other `Component` types to make it more ergonomic to still use the `DrawFlat` pass (but probably makes `DrawFlat` unmaintainable)
* Provide another animation type that animates both `Material` and `Mesh` together

omni-viral:
Texture shouldn't be changed during animation. Only texture coordinates.

azriel91:
Hm, I've got a use case where bigger sprites are on a separate texture, and the game configuration allows a character to choose a different sprite sheet mid sequence. How should that be handled?

omni-viral:
[...]

PR #830, opened 2018-07-31
Implement `DrawSprite` render pass and `SpriteRender` animation

azriel91:
See issue #829 for discussion

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/830)


amethyst_renderer/src/pass/shaders/fragment/sprite.glsl
```diff
@@ -0,0 +1,28 @@
...
+    return offset.x + coord * (offset.y - offset.x);
+}
+
+// coords = 0.0 to 1.0 texture coordinates
+vec2 texture_coords(vec2 coords, vec2 u, vec2 v) {
+    return vec2(texture_coord(coords.x, u), texture_coord(coords.y, v));
+}
+
+void main() {
+    color = texture(albedo, texture_coords(tex_uv, albedo_offset.u_offset, albedo_offset.v_offset));
```
omni-viral: You can adjust uv in vertex shader. This way it will be calculated `6` times instead of `w * h` times

amethyst_renderer/src/sprite.rs
```diff
@@ -19,17 +21,27 @@ pub type SpriteSheetHandle = Handle<SpriteSheet>;
 pub struct SpriteSheet {
     /// Index into `MaterialTextureSet` of the texture for this sprite sheet.
     pub texture_id: u64,
+    /// Pixel width of each sprite
+    pub sprite_w: f32,
```
jojolepro: Sprites can have variable sizes in the same spritesheet

jojolepro:
Added `status: postponed` tag because this will go into amethyst 0.9 instead of 0.8 (which is currently being prepared for release)

amethyst_renderer/src/sprite.rs
azriel91: * [ ] put it at the `Sprite` level

amethyst_renderer/src/pass/shaders/fragment/sprite.glsl
```diff
@@ -0,0 +1,28 @@
+// TODO: Needs documentation.
+
+#version 150 core
+
+uniform sampler2D albedo;
+
+layout (std140) uniform AlbedoOffset {
```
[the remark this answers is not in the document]
azriel91: hm, I was piggy backing off the existing buffer set up code in `amethyst_renderer/src/pass/util.rs`; I can split out the part I need

amethyst_renderer/src/pass/shaders/fragment/sprite.glsl
azriel91: * [x] be efficient with texture coordinates calculation

azriel91:
@Moxinilian (I can't comment inline =/ )

high :five:

> imposing an indexing on the spritesheet, which: 1) might not be very clear and could behave weirdly if one decides to extend one of the spritesheet's dimensions, 2) does not work with variably-sized sprites.

The indexing probably(?) should happen at some level, otherwise for any game configuration it would be quite tedious to have to specify the sprite coordinates for each sprite. It does work for variably sized sprites if there is a corresponding sprite metadata file for the sprite sheet indicating "sprite 0 lives at these coordinates" (see [aseprite's format](https://community.aseprite.org/t/how-to-process-the-generated-json-sprite-sheet/34) -- the indexing is defined by the order of the listed sprites, even if there's no explicit index).

> Doesn't this make SpriteRenderData useless?

For 2D games perhaps; one could argue that a game with mixed 2D and 3D graphics (e.g. sprites on a 3d background) would still want that functionality.

amethyst_renderer/src/pass/shaders/vertex/sprite.glsl
```diff
@@ -0,0 +1,38 @@
...
+    uniform mat4 proj;
+    uniform mat4 view;
+    uniform mat4 model;
+};
+
+layout (std140) uniform SpriteArgs {
+    // Sprite width and height.
+    uniform vec2 sprite_dimensions;
+    // Pixel offsets for the sprite. Used to shift the sprite left and down.
+    uniform vec2 offsets;
```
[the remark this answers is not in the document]
azriel91: * [ ] use it!

jojolepro:
@azriel91 manual offsets for each sprite makes sense because you can have a wrapper function over it that creates those index for you using fixed height and width.

azriel91:
> manual offsets for each sprite makes sense

The manual specification of offsets makes sense in configuration; the indexing this implementation imposes is the way it's referenced from the `Component` that is attached to the entity. So this implementation:

* requires the consumer to loading the `SpriteSheet` (in this context I mean the layout data, not texture), which allows them to define the texture offsets for each sprite (whether regular sized or not)
* doesn't enforce that the sprites must be the same size on the sprite sheet
* doesn't enforce that the sprites must be ordered on the sprite sheet
* does enforce that the loading code loads the sprites in order, and the order that they are stored in the sprite sheet is the index at which the sprite will be referenced on the `Component`

That is, it makes sense to me that a consumer will want to reference a sprite by some sort of ID in configuration, and the ID used here is the index of the sprite on the sprite sheet. It *could* be made generic (e.g. use a `HashMap<SpriteId, Sprite>` in the `SpriteSheet`) but at this point I don't think it's necessary.

Hopefully that clarifies things

azriel91:
Ready for review!

sprite_render_example

amethyst_animation/src/sprite.rs
```diff
@@ -0,0 +1,120 @@
...
+        panic!("Cannot magnitude2 SpriteRenderPrimitive")
+    }
+
+    fn magnitude(&self) -> f32 {
+        panic!("Cannot magnitude SpriteRenderPrimitive")
+    }
+
+    fn normalize(&self) -> Self {
+        panic!("Cannot normalize SpriteRenderPrimitive")
+    }
```
jojolepro: heh? Why implement it if you can't actually use the trait at all.

amethyst_renderer/src/pass/sprite/interleaved.rs
```diff
@@ -0,0 +1,149 @@
...
+                            sprite_render,
+                            &sprite_sheet_storage,
+                            &tex_storage,
+                            &material_texture_set,
+                            camera,
+                            global.get(*entity),
+                        );
+                    }
+                }
+            }
```
jojolepro: I see this piece of code over and over again in passes. Any way to put that in utils.rs instead?

amethyst_renderer/src/pass/sprite/separate.rs
```diff
@@ -0,0 +1,157 @@
...
+use mtl::MaterialTextureSet;
+use pass::util::{draw_sprite, get_camera, setup_textures, SpriteArgs, VertexArgs};
+use pipe::pass::{Pass, PassData};
+use pipe::{DepthMode, Effect, NewEffect};
+use sprite::{SpriteRender, SpriteSheet};
+use tex::Texture;
+use types::{Encoder, Factory};
+use visibility::Visibility;
+
+/// Draw mesh without lighting
```
jojolepro: I'm not too familiar with the difference between a normal and a "separate" pass. Would it be possible to add documentation explaining the difference between the 2?

amethyst_renderer/src/sprite.rs
```diff
@@ -24,19 +27,38 @@ pub struct SpriteSheet {
...
+    pub tex_coords: TextureCoordinates,
+}
+
+/// Texture coordinates of the sprite
 ///
-/// These should be in normalized coordinates:
+/// The coordinates should be normalized to a value between 0.0 and 1.0:
 ///
 /// * X axis: 0.0 is the left side and 1.0 is the right side.
 /// * Y axis: 0.0 is the top and 1.0 is the bottom.
```
jojolepro: I don't quite remember, but I think we settled on using a bottom to up Y axis?

amethyst_renderer/src/sprite.rs
```diff
@@ -24,19 +27,38 @@ pub struct SpriteSheet {
...
+
+/// Dimensions and texture coordinates of each sprite in a sprite sheet.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct Sprite {
+    /// Pixel width of the sprite
+    pub width: f32,
+    /// Pixel height of the sprite
+    pub height: f32,
+    /// Number of pixels to shift the sprite to the left and down relative to the entity
+    pub offsets: [f32; 2],
```
jojolepro: What will this be used for? If it is used to change the entity's origin, it might be better to have a parent entity with the sprite as a child entity + offset.

amethyst_renderer/src/sprite.rs
```diff
@@ -24,19 +27,38 @@ pub struct SpriteSheet {
...
+    }
+}
+
+/// Dimensions and texture coordinates of each sprite in a sprite sheet.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct Sprite {
+    /// Pixel width of the sprite
+    pub width: f32,
+    /// Pixel height of the sprite
+    pub height: f32,
```
jojolepro: Isn't this already defined by texture coordinates?

Also, I didn't check too much of the implementation (yet), but I personally use a [0,1,1,0] camera for my games. I expect to be able to create a (0,1,1,0) mesh and scale it down using transform after that. Is this something possible using this?

amethyst_renderer/src/sprite.rs
```diff
@@ -69,6 +108,26 @@ impl From<[f32; 4]> for Sprite {
     }
 }
 
+/// Information for rendering a sprite.
+///
+/// Instead of using a `Mesh` on a `DrawFlat` render pass, we can use a simpler set of shaders to
+/// render sprites. This struct carries the information necessary for the sprite pass.
```
jojolepro: Sounds great!

amethyst_renderer/src/sprite.rs
```diff
@@ -183,14 +248,70 @@ impl<'a> WithSpriteRender for EntityBuilder<'a> {
     }
 }
 
+/// Sprite sheets used by sprite render animations
```
jojolepro: Do you mind adding a bit more doc? Thanks!

amethyst_renderer/src/sprite.rs
jojolepro: I feel like I might be unclear. What I mean to ask is: Is this actual pixel size, or is it relative to the camera's dimensions matrix?

amethyst_renderer/src/sprite.rs
jojolepro: (even the ui uses Y up to avoid confusion)

amethyst_animation/src/sprite.rs
azriel91: because [it's a trait bound](https://github.com/amethyst/amethyst/blob/4cd5f4bdc218bd6278f71ba36b66a900c1c593e9/amethyst_animation/src/resources.rs#L29), [`MaterialPrimitive`](https://github.com/amethyst/amethyst/blob/4cd5f4bdc218bd6278f71ba36b66a900c1c593e9/amethyst_animation/src/material.rs#L15) does too; it's just that both of these animations aren't able to be interpolated, and must be stepped

amethyst_renderer/src/pass/sprite/interleaved.rs
azriel91: the repetition is in the `pbm`, `shaded`, and `flat` passes which I haven't touched; this sprite pass is notably different to those in the sense that it:

* doesn't use `Material` or `Mesh`
* doesn't use non-uniform vertex buffers
* uses sprite related components / storages

if it were to be commonized, perhaps it should be in a separate PR so that something that says "implement sprites pass" doesn't inadvertently break other rendering passes unrelated to sprites

amethyst_renderer/src/pass/sprite/separate.rs
azriel91: > I'm not too familiar with the difference between a normal and a "separate" pass.

you and me both :stuck_out_tongue:. This was me copying `DrawFlat`. I'll have a go

* [ ] docs for separate vs interleaved passes

amethyst_renderer/src/sprite.rs
azriel91: I know it should be bottom up for world coordinates, I wasn't sure about sprites, but I can make this change

* [ ] bottom up texture Y coordinates

jojolepro: thanks

amethyst_renderer/src/sprite.rs
azriel91: "change the entity's origin" is sounds like it (it's, "change the origin of the sprite relative to the entity"). I wouldn't advocate for the parent with child entity + offset because:

* ergonomics: for games like a space shooter where one entity has one sprite, the consumer shouldn't have to create 2 entities just to offset the sprite rendering
* mirroring sprites: if using a parent + child entity, and they have a left/right mirror, they inadvertently have to code some graphics logic to calculate the flipped coordinates for the child entity, when what they really want to do is say `is_flipped: bool`

amethyst_renderer/src/sprite.rs
azriel91: > Isn't this already defined by texture coordinates?

I don't think it is -- you can have a big texture with 1 sprite, or a small texture with 1 sprite, and the texture coordinates will be largely different since they are proportions of the texture. You'd still need the sprite dimensions to know how much space it occupies in the world.

> [0,1,1,0] camera

what do the numbers correspond to? (forgive my lapse in knowledge :confused:). It's the actual pixel size on the texture. I don't think I understand/remember the camera dimension's matrix without re-reading =o

amethyst_renderer/src/sprite.rs
azriel91: haha, shall!

* [ ] more doc = good

amethyst_renderer/src/pass/sprite/separate.rs
azriel91: So I figured out what it's for :stuck_out_tongue:, and documented what it is in `amethyst_renderer/src/lib.rs`. I've deleted `DrawSpriteSeparate` because it isn't necessary.

azriel91:
@jojolepro woosh, thanks for reviewing. Hopefully I've addressed everything (whether through rationale / docs / fixing code). Also, got that changelog updated (was waiting for the release to happen, woo).

@Moxinilian probably good to get your eyes on this as well :eye:

azriel91:
Clip to show this being used in a slightly more complex scene:
https://imgur.com/2aHcO6k (it's mp4, so I can't embed it).

amethyst_renderer/src/sprite.rs
jojolepro: left right top bottom

I want to know if the sprite pass takes the camera's matrix into account when rendering, or it uses the raw pixel size to pixels on the screen.

azriel91: ohh, yeap it does take the camera's matrix into account. So if the camera resolution is halved then you get double sized sprites.

Oh ya, should the `TextureCoordinates` use left right top bottom as the field order then? I swapped it when inverting the Y axis direction

Moxinilian: If you got inspiration from SpriteRenderData, then yes, the Y axis is wrong in it.

amethyst_renderer/src/bundle.rs
```diff
@@ -59,7 +61,12 @@ impl<'a, 'b, 'c, B: PipelineBuild<Pipeline = P>, P: 'b + PolyPipeline> SystemBun
             );
         };
         builder.add_thread_local(
-            RenderSystem::build(self.pipe, self.config).chain_err(|| "Renderer error!")?
+            RenderSystem::build(self.pipe, self.config).chain_err(|| "Renderer error!")?,
+        );
+        builder.add(
+            Processor::<SpriteSheet>::new(),
+            "sprite_sheet_processor",
+            &[],
```
Moxinilian: Is it wise to always add a SpriteSheet processor to the rendering bundle?
When you're making a 3D, you probably don't need this at all.

amethyst_renderer/src/pass/util.rs
```diff
@@ -300,6 +325,80 @@ pub(crate) fn draw_mesh(
...
+        );
+        return;
+    }
+
+    let texture = tex_storage.get(&texture_handle.unwrap());
+    if texture.is_none() {
+        warn!(
+            "Texture not loaded for texture id: `{}`.",
+            sprite_sheet.texture_id
+        );
```
Moxinilian: Why this warning?

jojolepro:
Amethyst 0.8 has been released, so I'm removed the postponed tag. Once all reviews are addressed we can merge.

amethyst_renderer/src/sprite.rs
azriel91: Sweet, also I'll keep the parameter order to `(left, right, bottom, top)` because that's how `glOrtho` does it<sup>[[1]](https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glOrtho.xml)[[2]](https://docs.microsoft.com/en-us/windows/desktop/opengl/glortho)</sup>

jojolepro: We should normalize it in the engine. Its often unclear which order we choose.

jojolepro:
By the way, would you mind updating the tutorial to showcase everything you added in this PR?

Its optional, but would really help new engine users to understand how everything works now.
(if the change was breaking, its not optional) :P

amethyst_renderer/src/bundle.rs
azriel91: True that, it should be in the consumer application &mdash; I used `loader.load_from_data(..)`.

* [ ] Todo: Move `Processor::<SpriteSheet>` out

amethyst_renderer/src/pass/util.rs
azriel91: At the time it was, "nothing's rendering, let me add diagnostics to everything that could be missing", and texture followed the pattern of "this is an `Option<..>`, let's log it in case it's `None`". Admittedly if `loader.load(path, PngFormat, ..)` is used (which I suspect it will be most of the time), it wouldn't be `None`.

That said, I'm usually happier to have something that tells me something's missing, even if it's super unlikely.

azriel91:
> SpriteRenderData really should go

* [ ] Todo: parting ceremony for `SpriteRenderData`

> Maybe the method that builds meshes and materials can stay

cool, shall check if it's worth keeping, and remove them if not

* [ ] Todo: check if mesh/material functions are worth keeping

> would you mind updating the tutorial to showcase everything you added in this PR?

Will do. Haha the tutorial was already inheriting code changes because `Sprite` changed, but I'll also check the accompanying text to make sure it still flows.

* [ ] Todo: ensure pong tutorial is fluent

amethyst_renderer/src/pass/util.rs
Moxinilian: Actually, loader.load will return immediately, even when the texture is not loaded yet.
That means you can have a couple frames without the loaded texture.

jojolepro:
Thanks!

amethyst_renderer/src/bundle.rs
azriel91: Just realized, this should still be in the render bundle, but not enabled until the user requests it. Rationale is we shouldn't require everyone who is using the `DrawSprites` pass to create a bundle and paste this in it:

```rust
builder.add(
    Processor::<SpriteSheet>::new(),
    "sprite_sheet_processor",
    &[],
);
```

Ideally the render bundle would be able to inspect the `Pipeline` for any pass that needs `Sprite`s support and enable this automatically.

Moxinilian: Then we should get that pipeline parser going beforehand. It shouldn't be that hard and would be very helpful.

azriel91: I can't seem to find info relating to pipeline parser in github issues / source code, do you have a link / can you give me a quick rundown on how it's meant to work?

Moxinilian: I was talking about the idea you just mentioned of inspecting the pipeline from the render bundle. It shouldn't be that hard to implement I think.

azriel91: ohh, so *this is it* :man_dancing:. haha okay I'll have a go

azriel91:
Phewsh, Pong tutorial 02 has been restructured quite a bit for that last change &mdash; it now uses `DrawSprite`. I've added some intermediate "try and run this now" steps for earlier feedback, and shifted the "oh no it failed, here's what's wrong" bits earlier.

@Moxinilian I don't mind letting you hold the reins for that Pipeline parsing thing :stuck_out_tongue:, rewording the pong tutorial took quite a bit of energy.

One more thing I'd like in this before it's merged is, there's no Z sorting yet (sadface). Any of you keen to do that?

Moxinilian:
If I have two entities with the same sprite but with different Z values, would one be bigger than the other or would the Z value be ignored?

termhn:
They should be the same size but the one with the higher z value should be on top of the one with the lower z value.

I think we should just introduce a depth test tbh, easiest solution to this and it really shouldn't be a performance burden at all I don't think

Moxinilian:
What do you mean by depth test?

termhn:
Add a depth buffer and turn on depth testing. Then you could just pass in a z value with the position of the sprite and it could use that to depth test.

Moxinilian:
Ah yes absolutely. This is what I assumed we'd use for the ludum dare this week-end. Do you think it could be a long term solution? I'm afraid of the overhead introduced by such a non-specifically 2D operation.

termhn:
I'm pretty sure depth testing is quite inexpensive, especially in this case

omni-viral:
Depth testing easily fix ordering problem if you have no semi-transparent pixels.
For general solution you should draw all sprites w/o semi-transparent pixels, then sort by Z sprites w/ semi-transparent pixels and draw from far to near.

Moxinilian:
Ah so the current system only works for 0 or 1 alpha values?

termhn:
This looks great from what I looked over, one thing I think is pretty essential though is providing a `SpriteBatch` of some sort to do instanced drawing, or to automatically batch draws that use the same sprite sheet internally. This is very important to get good performance when drawing more than a couple hundred sprites per frame.

azriel91:
**Re:alpha blending**

Without changing the render pass, when you specify `Some(DepthMode::LessEqualWrite)` for the depth mode alpha blending works with the sprites example:

alpha_blending

**Re:z axis ordering**

I don't really understand why the sprites render when they do. The camera has z values: `near: 0.1,  far: 2000.0`, meaning z axis increases "into the screen". Next, the z values of each bat on the bottom row, moving from left to right are:

```
z: -10.00
z: -9.90
z: -9.80
.. // omitted, but they just increment in 0.1 each time
z: -9.10
z: -9.00
```

Shouldn't none of the bats render at all, given their z values aren't even in the camera's range?
If I change the z values to be between `0.0` and `1.1`, the bat at `1.1` doesn't render at all, suggesting some z-culling (but why `1.1`? it's not at one of the camera's limits).

If you want to test it out you can play with the sprites example:

```bash
git remote add azriel91 git@github.com:azriel91/amethyst.git
git fetch azriel91
git checkout sprites-alpha-z-test
cargo run --example sprites
```

I'm somewhat frustrated that I couldn't figure out what it's doing after trying to for a number of hours and making no progress =/. If anyone would like to do the z axis sorting please say so.

amethyst_renderer/src/pass/sprite/interleaved.rs
Rhuagh: There's no way to do that, because types are different, and have no commonality. This was the best that could be done for unification of the passes.

amethyst_renderer/src/pass/util.rs
Rhuagh: You should be using `Progress` to track asset loads before using them on entities for this reason.

azriel91:
Implemented z axis order rendering using `sprite_visiblity`.
Also added a `sprites_ordered` example with semi transparent sprites, and explains the camera (in code comments), when things are culled/not, and lets you play with the camera values / z coordinates / and the `Transparent` component to see what effect each has.

azriel91:
Mostly complete for a first cut, just missing the automatic detection of the `DrawSprite` pass from the pipeline (go go @Moxinilian!)

amethyst_renderer/src/sprite.rs
jojolepro: ui doesn't use Y = up, but I thought it did :P
anyway, sprites should be bottom up still

azriel91:
Added a flag that users call to enable the `Processor<SpriteSheet>`
Also updated Pong tutorials to use the `DrawSprite` pass, and made the tutorial snippets match `main.rs`

r?

Moxinilian:
I think it would be nice, as a final touch, to have the pong tutorial and examples updated. It really shouldn't be much work, just swapping SpriteRenderData out. I think it is important to do it right now because as this PR removes SpriteRenderData we just can't afford to have such a discrepancy between doc and code, even in the develop branch.

Actually maybe you might want to have a separate PR as we changed the doc a lot since you opened this PR. So I may just do it myself. What do you prefer?

Moxinilian:
Oh my god you're brilliant.

jojolepro:
Order of merge: 
Docs changes by @Moxinilian into develop
Merge into master
The master docs will then be freezed for the 0.8 version.
Merge this into develop (after rebase)
merge all the other PR in the queue

jojolepro:
Hi. This can be merged as soon as the tutorial files are merged. :)

Moxinilian:
I'm doing the merge because it involves the new testing system stuff.
Don't merge until I tell you it's okay tho because I want to make sure mdbook tests runs correctly.

Moxinilian:
Let me first try mdbook test before we do anything else.

Moxinilian:
LGTM

Moxinilian:
You must be kidding me right

jojolepro:
Thanks!

Issue #828, opened 2018-07-29
Keybind to ungrab the mouse: FlyCamera & ArcBallCamera


jojolepro:
Add a way to stop grabbing the mouse when using the FlyCamera or the ArcBallCamera.

zoonage:
Moving discussion in #840 to here

@jojolepro Just so I can be sure I have the right requirements here; What we need is the following

1. A `MovementToggle` resource
2. A `MovementToggleSystem` that checks for some specific keypress events (? Is this correct and if so which one should they be?)
3. Finally (the bit that I'm confused about), how does this system interact with the `FlyMovementSystem` and the `ArcBallMovementSystem` to stop them moving? Is there an example somewhere of two systems interacting?

zoonage:
@jojolepro A bucket load of questions incoming probably stemming from my complete lack of understanding of how the components/resources/systems all interact.

1. By remove from the camera do you mean remove it from the `Entity`? I.e. do the reverse of `ControlTagPrefab::load_prefab`? I'm unsure how that would be done since I can't pass an `Entity` around as `SystemData` to remove it from the `WriteStorage<'a, FlyControlTag>` or `WriteStorage<'a, ArcBallControlTag>`.

2. How can the toggling be done in the state? Would the `GrabCursorOnFocus` be added to the world as a resource by the state? Would we then need a system that checks for when this resource is added to then;
    1. Send "some event" to `CursorHideSystem` to tell it to release the cursor via the `amethyst_renderer::release_cursor` function. Since `CursorHideSystem` only responds to winit::Events` how can we trigger anything in it unless all the logic is handled inside the `CursorHideSystem`?
    2. Trigger the event described in question 1.

jojolepro:
Partially fixed by this: https://github.com/amethyst/amethyst/pull/946/files
All that is left to do is to add this line to the amethyst rotation system
https://github.com/jojolepro/amethyst-extra/blob/cd1aa58cc5ad9e20bb4ac0bec672a9c959419e9a/src/lib.rs#L1306

PR #840, opened 2018-08-04
Add toggle to CursorHideSystem

zoonage:
#828 

Adds a `toggle()` to the `CursorHideSystem` to disable the hiding of the cursor

Opening this up for discussion regarding how to actually use the toggling. Should there be an event listener on the `CursorHideSystem` that checks for this or can it be changed outside of this system?

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/840)

### DIFF
--- a/amethyst_controls/Cargo.toml
+++ b/amethyst_controls/Cargo.toml
@@ -27,3 +27,4 @@ thread_profiler = { version = "0.1", optional = true }
 [features]
 profiler = [ "thread_profiler/thread_profiler" ]
 nightly = [ "amethyst_core/nightly" ]
+


amethyst_controls/src/systems.rs
```diff
@@ -229,7 +237,7 @@ impl<'a> System<'a> for CursorHideSystem {
             match *event {
                 Event::WindowEvent { ref event, .. } => match event {
                     &WindowEvent::Focused(focused) => {
-                        if focused {
+                        if focused && !self.released {
```
Moxinilian: What if the toggled value changes without a focus event? Does that mean the cursor will be released only on the next focus change?

amethyst_controls/src/systems.rs
```diff
@@ -211,11 +211,19 @@ impl<'a> System<'a> for MouseFocusUpdateSystem {
...
 impl CursorHideSystem {
     pub fn new() -> CursorHideSystem {
-        CursorHideSystem { event_reader: None }
+        CursorHideSystem { 
+            event_reader: None,
+            released: false,
+        }
+    }
+
+    pub fn toggle(&mut self) {
```
Moxinilian: It's pretty rare to have access to the instance of a system. What about having a GrabCursor boolean resource?

jojolepro:
By grabbing I meant enabling/disabling all the controls of a fly/arc camera. That means you need to be able to toggle FlyMovementSystem ArcBallMovementSystem FreeRotationSystem CursorHideSystem

Disabling a flycamera:
Removing the flycamera tag component, make the mouse visible and ungrab it(through WindowMessages). Then you only need a way to toggle CursorHideSystem. I would suggest a resource called GrabCursorOnFocus{bool}.

Making the mouse visible is actually a matter of changing thr cursor between normal and none I think. The should be a method called disable_fly_camera(entity, &mut WindowMessages,ungrab_mouse: bool, allthestuffyouneed) that does all of this. Also an enable_fly_camera(...)

jojolepro:
Hi, are you still working on this pull request?
Thanks!

zoonage:
Heya, yes I will be, just been a hectic week

zoonage:
This shouldn't ben closed, I did a hard reset of my branch from develop and pushed it, i think GitHub then took this to mean that the branch had been merged.

Moxinilian:
You need to commit to your branch again to re-open it.

jojolepro:
rip files owo

Issue #845, opened 2018-08-06
Better example for Component in the book


rivertam:
I'm a bit confused about `Component`s. Namely, do they describe all properties of an `Entity` or does each one correspond to one property?

In [the book](https://www.amethyst.rs/book/master/concepts/component.html), the example given is just

```rust
pub struct MyComponent {
    some_property: String,
}
```

At the risk of sounding unfairly harsh, I consider this an unacceptable example as it explains nothing about what a `Component` is to me.

A good example, I think, given the previous descriptor, would be one of the following:

```rust
enum Shape {
    Sphere { radius: f32 },
    RectangularPrism { height: f32, width: f32, depth: f32 },
}
pub struct Shape {
    shape: Shape,
}

pub struct Material {
    material_name: String,
}
```

or 

```rust
enum Shape {
    Sphere { radius: f32 },
    RectangularPrism { height: f32, width: f32, depth: f32 },
}

pub struct BottleComponent {
    shape: Shape,
    material_name: String,
}
```

jojolepro:
Makes a lot of sense!
To answer your question, the right answer is the first of the two examples you wrote.
You usually want to have a single property (or property set) per component.

rivertam:
@jojolepro Great! Thanks! Will update the book. Working on a couple PRs with general changes.

Temeez:
I too am confused about components and I want to confirm something. For a in-game item component I would do something like this:
```rust
pub struct Name {
    name: String
}

pub struct Description {
    description: String
}

pub struct Value {
    value: i32
}

pub struct Quantity {
    quantity: i32
}
```

instead of this:
```rust
pub struct Item {
    name: String,
    description: String,
    value: i32,
    quantity: i32,
}
```
?

If the answer is "yes" then why? Is it more memory efficient or something along those lines?

jojolepro:
No you should be using the second piece of code.

magnonellie:
@Temeez In your case, I would probably say that you should go for the combined struct. The reason why someone might want to go with your first approach is that it's more modular. You can mix and match between having and not having different attributes. Something maybe doesn't have a name and something might not have a value just as simple examples.

rivertam:
Well now I'm confused. I thought each `Component` represented one trait an object might have?

I could see how `Item` might represent some generic "pile". Like in RuneScape there are "piles" of things that would have those attributes (though maybe not `value`, I forget). But barring the "this is a very generic bundle of one resource", why would you want to combine all of those?

[...]

PR #848, opened 2018-08-06
Put in a better example for Component in the book

rivertam:
Addresses #845

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/848)

### DIFF
--- a/book/src/concepts/component.md
+++ b/book/src/concepts/component.md
@@ -16,17 +16,13 @@ In this example, the bottle is the entity, and the properties are components.
 
 Creating a component is easy.
 
-You declare the relevant structs:
+You declare the relevant underlying data:
 
 ```rust,ignore
+/// This `Component` describes the shape of an `Entity`
 enum Shape {
     Sphere { radius: f32 },
     RectangularPrism { height: f32, width: f32, depth: f32 },
-}
-
-/// This `Component` describes the shape of an `Entity`
-pub struct Shape {
-    shape: Shape,
 }
 
 /// This `Component` describes the contents of an `Entity`
--- a/book/src/concepts/component.md
+++ b/book/src/concepts/component.md
@@ -24,10 +24,12 @@ enum Shape {
     RectangularPrism { height: f32, width: f32, depth: f32 },
 }
 
+/// This `Component` describes the shape of an `Entity`
 pub struct Shape {
     shape: Shape,
 }
 
+/// This `Component` describes the contents of an `Entity`
 pub struct Content {
     content_name: String,
 }
--- a/book/src/concepts/component.md
+++ b/book/src/concepts/component.md
@@ -16,19 +16,31 @@ In this example, the bottle is the entity, and the properties are components.
 
 Creating a component is easy.
 
-You declare a struct:
+You declare the relevant structs:
 
 ```rust,ignore
-pub struct MyComponent {
-    some_property: String,
+enum Shape {
+    Sphere { radius: f32 },
+    RectangularPrism { height: f32, width: f32, depth: f32 },
+}
+pub struct Shape {
+    shape: Shape,
+}
+
+pub struct Material {
+    material_name: String,
 }
 ```
 
-and then you implement the `Component` trait for it.
+and then you implement the `Component` trait for them:
 
 ```rust,ignore
 use amethyst::ecs::{Component, VecStorage};
-impl Component for MyComponent {
+impl Component for Shape {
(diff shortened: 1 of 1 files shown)


book/src/concepts/component.md
```diff
@@ -16,19 +16,31 @@ In this example, the bottle is the entity, and the properties are components.
...
+You declare the relevant structs:
 
 ```rust,ignore
-pub struct MyComponent {
-    some_property: String,
+enum Shape {
+    Sphere { radius: f32 },
+    RectangularPrism { height: f32, width: f32, depth: f32 },
+}
+pub struct Shape {
```
jojolepro: Missing new line before Shape struct

book/src/concepts/component.md
```diff
@@ -16,19 +16,31 @@ In this example, the bottle is the entity, and the properties are components.
...
 }
 ```
 
-and then you implement the `Component` trait for it.
+and then you implement the `Component` trait for them:
 
 ```rust,ignore
 use amethyst::ecs::{Component, VecStorage};
-impl Component for MyComponent {
+impl Component for Shape {
```
jojolepro: Missing new line

book/src/concepts/component.md
```diff
@@ -16,19 +16,31 @@ In this example, the bottle is the entity, and the properties are components.
...
-    some_property: String,
+enum Shape {
+    Sphere { radius: f32 },
+    RectangularPrism { height: f32, width: f32, depth: f32 },
+}
+pub struct Shape {
+    shape: Shape,
+}
+
+pub struct Material {
```
Moxinilian: I'm not sure the `Material` name is a good idea considering Amethyst's renderer also uses a `Material` component for object textures. Maybe you could prefer something like `Content` which indicated what an entity could contain.

jojolepro: true

rivertam: What about `MaterialComponent`?

jojolepro: Not really, because 
1) Material is already a thing
2) We shouldn't append Component to components. (you could, but we don't internally)

Moxinilian: Having `Component` at the end of a component type is not good practice.

rivertam: Gotcha. Best part about working on the docs is you get to learn as you go. ;)

I changed it to `Content` as suggested. Truth be told, we probably just don't need multiple.

jojolepro:
Also, you should add comments. For example:

/// This is the `Component` holding the content of the bottle.
pub struct Content {
    /// The content of the bottle (usually a liquid).
    pub content: String,
}

rivertam:
@jojolepro I'm on board with comments, but doesn't "the content of the bottle" limit the `Component` to just describe bottles? That is to say, couldn't a `Chest` entity also have a `Content` `Component`?

jojolepro:
Well we try to encourage re-usability at a maximum, so yes it makes sense to not specify bottle and only "content".

rivertam:
I think the comments are quite generic at this point and not particularly useful, but there they are.

book/src/concepts/component.md
```diff
@@ -16,19 +16,35 @@ In this example, the bottle is the entity, and the properties are components.
 
 Creating a component is easy.
 
-You declare a struct:
+You declare the relevant structs:
 
 ```rust,ignore
-pub struct MyComponent {
-    some_property: String,
+enum Shape {
```
Aceeri: Why not just impl Component for enum Shape?

rivertam: When you're right, you're right. Good call.

rivertam: Actually wow, I can't believe I didn't notice this won't compile. Sloppy of me.

book/src/concepts/component.md
```diff
@@ -16,19 +16,31 @@ In this example, the bottle is the entity, and the properties are components.
...
 
-and then you implement the `Component` trait for it.
+and then you implement the `Component` trait for them:
 
 ```rust,ignore
 use amethyst::ecs::{Component, VecStorage};
-impl Component for MyComponent {
+
+impl Component for Shape {
+    type Storage = VecStorage<Self>;
```
torkleyy: should probably use `DenseVecStorage`

jojolepro:
Bors fails because travis gets a permission denied on the trying branch.
Will investigate soon.

PR #849, opened 2018-08-06
Add docs badges to the concepts pages

rivertam:
Again, Prettier did stuff automatically. I think it's for the best, but
I can still take it out if you'd prefer.

I checked every link I added, but a double check would not be
unwelcome.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/849)

### DIFF
--- a/book/src/concepts/dispatcher.md
+++ b/book/src/concepts/dispatcher.md
@@ -1,4 +1,4 @@
-# Dispatcher
+# Dispatcher [![develop docs](https://img.shields.io/badge/docs-develop-blue.svg)](https://www.amethyst.rs/doc/develop/doc/amethyst/ecs/prelude/struct.Dispatcher.html) [![master docs](https://img.shields.io/badge/docs-master-blue.svg)](https://www.amethyst.rs/doc/master/doc/amethyst/ecs/prelude/struct.Dispatcher.html)
 
 ## What is a `Dispatcher`?
 
@@ -14,5 +14,4 @@ The dispatcher looks at all the `SystemData` in the systems and builds execution
 
 If you want to have the best performance possible, you should prefer immutable over mutable whenever it is possible. (`Read` instead of `Write`, `ReadStorage` instead of `WriteStorage`).
 
-_Note: Please however keep in mind that `Write` is still preferable to interior mutability, such as `Mutex` or `RwLock` for example.
-
+\_Note: Please however keep in mind that `Write` is still preferable to interior mutability, such as `Mutex` or `RwLock` for example.
--- a/book/src/concepts/component.md
+++ b/book/src/concepts/component.md
@@ -1,4 +1,4 @@
-# Component
+# Component [![develop docs](https://img.shields.io/badge/docs-develop-blue.svg)](https://www.amethyst.rs/doc/develop/doc/amethyst/ecs/prelude/trait.Component.html) [![master docs](https://img.shields.io/badge/docs-master-blue.svg)](https://www.amethyst.rs/doc/master/doc/amethyst/ecs/prelude/trait.Component.html)
 
 ## What is a `Component`?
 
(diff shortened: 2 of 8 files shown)


Moxinilian:
Can we see a screenshot of what it does?

rivertam:
@Moxinilian Screenshot added to the PR comment

Moxinilian:
What do theses badges do exactly?

jojolepro:
Links to develop and master docs of shrev

Moxinilian:
I'm not sure it is a good idea to be able to swap between develop and master so easily, it could create confusion.

rivertam:
They just link to the respective docs of each concept. I often get frustrated reading tutorials that don't link to the API documentation for more details.

jojolepro:
You should link to the exact version used by this amethyst release.

rivertam:
I'm actually not even seeing a way to visit old versions of the docs. Do you know how I can do that? I tried going to like `https://www.amethyst.rs/doc/0.6.0/doc/amethyst/` etc. to no avail.

jojolepro:
You have to go on docs.rs for that. On the amethyst.rs website we only host the master and develop branches.

LucioFranco:
@rivertam I like this idea, but I do not like having both master and develop on the same page. I don't know if there is a way a variable can be passed down in mdbook but if that is possible, I'd like to have the book on develop point just to the develop docs.

So one solution is to either embed the docs within the book so that it can be reached by a relative link. But I don't think having both is the right idea, as @jojolepro it will be just confusing.

rivertam:
@LucioFranco I've raised the issue on the mdBook repository for configurable variables. https://github.com/rust-lang-nursery/mdBook/issues/760

I could also write a specific preprocessor for this project, but that would require amethyst to move to a custom mdBook implementation rather than the CLI.

LucioFranco:
@rivertam Great, I'm curious to see how the preprocessor might be implemented. I'm not really a big fan of having a custom implementation that we will have to maintain. If we can get mdBook to insert a variable either from the environment or via a setting in the `book.toml` I would prefer that as travis can do all the work there.

Now, something we could do now is update amethyst/website to somehow insert the correct version using something like `sed` into the generated html.

rivertam:
@LucioFranco That's a good point -- we could definitely do this in JavaScript.

LucioFranco:
@rivertam that's not a bad idea, if it can be done without compromising anything, I'm all for it!

book/src/concepts/dispatcher.md
```diff
@@ -14,5 +14,4 @@ The dispatcher looks at all the `SystemData` in the systems and builds execution
 
 If you want to have the best performance possible, you should prefer immutable over mutable whenever it is possible. (`Read` instead of `Write`, `ReadStorage` instead of `WriteStorage`).
 
-_Note: Please however keep in mind that `Write` is still preferable to interior mutability, such as `Mutex` or `RwLock` for example.
-
+\_Note: Please however keep in mind that `Write` is still preferable to interior mutability, such as `Mutex` or `RwLock` for example.
```
torkleyy: Unrelated to this PR, but this line doesn't make sense. `Write` _is_ interior mutability. Maybe this should be "preferable locks"? And then, there are cases where locks actually are preferable (we do use them for a few things).

LucioFranco: @torkleyy I agree, should we open an issue for that? Since this PR is unrelated?

jojolepro: moved to https://github.com/amethyst/amethyst/issues/870

jojolepro:
If you think support for getting the current library version is not coming soon, what you can do is hardcode the current dependencies versions in for now, and open a PR later with the dynamic versionning. That way we can merge the documentation updates before starting to merge the features PR (this way, we don't have to wait until the next version to fix the docs)

jojolepro:
Hi. Is this still a thing?

rivertam:
Hey @jojolepro It's vaguely still a thing but I'm not really that actively pursuing it right now. I'd rather not hardcode the dependencies versions just because that's a lot of work for myself and a lot of work for you guys when you update the version. So for my needs it's kind of pending https://github.com/rust-lang-nursery/mdBook/issues/760, which I'm also not that actively pursuing.

LucioFranco:
@rivertam cool, we can close this then and if you want to attempt it again you can reopen this or open a new one! Thanks.

PR #868, opened 2018-08-12
Fix flat render gamma.

FelixMcFelix:
Could probably make this a uniform in future, very quick fix for now for extremely bright 2D sprite rendering.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/868)


jojolepro:
Please add a changelog entry in docs/CHANGELOG.md

Rhuagh:
Or we could use a different output surface

FelixMcFelix:
Thinking on this, if we want to make this more general then the issue is not so much the output format but the format of the textures. The fixes as applied essentially reverse gamma correction because in many cases the input textures themselves are *already* in the sRGB space. However, normal maps etc. are already in linear space, so present a case where mixed formats are present (and blind handling either way will lead to broken results).

OpenGL provides the GL_SRGB and GL_SRGB_ALPHA internal texture formats, which I suppose we can either set as the default, or otherwise intuit by looking at the image data. I'm not sure when they first became available, but we should be able to make this a per-texture thing.

Rhuagh:
The output backbuffer also have a format that needs to be aligned

FelixMcFelix:
As it turns out, `TextureMetadata` already allows this:
```rust
use gfx::format::ChannelType;

let handle = loader.load(
	"path/to/image.png",
	PngFormat,
	TextureMetadata::default()
		.with_channel(ChannelType::Srgb),
	(),
	&tex_storage,
);
```
This is probably a papercut at the least, given that none of our examples make use of this (when they likely should). Particularly in the absence of the relevant chunks, [decoders should assume the input is gamma-encoded with a reasonable value](https://www.w3.org/TR/PNG/index-noobject.html#13Decoder-gamma-handling). Reading the specific gamma level from a PNG file doesn't seem to be handled yet by the image/png crates relied upon by us, but we could always make it an option to specify a non-2.2 gamma and change the default `ChannelType` for PNG at least.

Maybe something like (for all texture formats):
```rust
enum TextureGamma {
    EncodedDefault, // Srgb ChannelType, no correction locally/in shader
    Encoded(f32), // F32 ChannelType, manually decode
    Linear, // F32 ChannelType, no correction
}
```

omni-viral:
@FelixMcFelix non-standard gamma correction would require changing values of all pixels.
We could start with making `Srgb` channel type default for textures loaded from file and `Float` for textures created from color values.

FelixMcFelix:
@omni-viral That seems like a good place to start. Changing all the pixels isn’t *that bad* down the line: it’s a one-off cost at texture load time (at most 256 exponentiations, if you build a LUT) if you assume it won’t be changed at any point.

jojolepro:
Can someone review please?

amethyst_renderer/src/pass/shaders/fragment/flat.glsl
```diff
@@ -27,5 +27,5 @@ vec2 tex_coords(vec2 coord, vec2 u, vec2 v) {
 }
 
 void main() {
-    color = texture(albedo, tex_coords(vertex.tex_coord, albedo_offset.u_offset, albedo_offset.v_offset));
+    color.rgb = texture(albedo, tex_coords(vertex.tex_coord, albedo_offset.u_offset, albedo_offset.v_offset));
```
Rhuagh: Why did we remove setting the alpha channel here? That's gonna disable blending on both Flat passes.

FelixMcFelix: Yikes, my mistake. Fixing.

FelixMcFelix: Apologies, careless mistake.

PR #871, opened 2018-08-12
Added basic mouse button support

Telzhaak:
A simple rewrite of the existing `is_key_down` and `get_key` to allow for detection of mouse button down and released.

Non-exclusive rights granted to the source code under both the MIT License and the Apache License 2.0.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/871)

### DIFF
--- a/amethyst_input/src/lib.rs
+++ b/amethyst_input/src/lib.rs
@@ -22,7 +22,9 @@ pub use self::controller::{ControllerAxis, ControllerButton};
 pub use self::event::InputEvent;
 pub use self::input_handler::InputHandler;
 pub use self::system::InputSystem;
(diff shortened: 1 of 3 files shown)


Telzhaak:
In opposite to `is_key_down`, `is_mouse_button_down` only fires when the mouse button is pressed, not held. Therefore it should be named more appropriately `was_mouse_button_pressed`.

And the pull-request number in the Changelog.doc was wrong as well. I apologize.

jojolepro:
Why did you close the pull request?

Telzhaak:
because there already is a way to use mouse input, 

and the function names/documentation was wrong (in opposite to is_key_down the mouse ElementState::Pressed only returns once, not every frame. So it should be was_mouse_down

PR #873, opened 2018-08-13
Fix error[E0412]: cannot find type `PosNormTex` in this scope

paulirotta:
Updated the replacement header to match the requirements of "amethyst new pong"

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/873)

### DIFF
--- a/book/src/pong-tutorial/pong-tutorial-01.md
+++ b/book/src/pong-tutorial/pong-tutorial-01.md
@@ -17,11 +17,12 @@ In `src` there's a `main.rs` file. Delete everything, then add these imports:
 ```rust,ignore
 extern crate amethyst;
 
-use amethyst::prelude::*;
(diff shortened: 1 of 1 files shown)


jojolepro:
I think the tutorial asks you to delete what was generated by "amethyst new pong" before inserting new code. The correct pass to use for pong is effectively PosTex and not PosNormTex.

paulirotta:
@jojolepro you may be right. I replaced the header but not the rest of the body generated by "amethyst new pong". By deleting the rest of the generated code I may not have seen the error. Instead of this fix, the wording can be changed to reflect that if others have a similar confusion.

I must say it is most excellent documentation and impressive ECS design. Thank you!

jojolepro:
Thank you! 
Keeping note of your suggestion. It seems a lot of people are getting confused and skip the line where we tell to remove the generated code.

Issue #788, opened 2018-06-13
Status of glTF import?


jondo2010:
what's the priority and difficulty of this one?

Rhuagh:
I would wait a few days to analyze this. The GLTF importer is being completely rewritten basically. 
Technically, after the next lines of PRs are merged, we will support everything except:

* Camera
* Morph targets
* Extensions

Rhuagh:
The new example is a lot better, we got a new original art piece which use the PBR system with transparency.

torkleyy:
If we manage to load https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/Sponza that would already be a cool demo I think. Definitely suitable for showing our renderer.

jojolepro:
Status?

anderejd:
Here's the example I tried to get working in #877 <http://github.khronos.org/glTF-WebGL-PBR/> (the live demo currently loads the "DamagedHelmet" reference sample as default)

I removed that experiment from #877 since it was a bit too ambitious for me right now, and it was suggested by Rhuagh. Regarding that those sample models though, would it be a good direction for Amethyst to support gltf 2.0 fully? in other words rendering the sample models the same way or close to the same as the reference implementation? (see the live demo link above).

jojolepro:
I'm impressed by how that looks. Fully supporting a standard like gltf definitely looks like something that we want. The tip I can give you is to go by iterations, with a PR for each (or combining them if they are small). Adding the required features one by one until you get them all together.

Thanks!

Rhuagh:
There are a few issues with the importer currently: 

* Material ids are just pure bugged, they should be removed
* Node names are not kept
* Joint ids/weights should be added (empty) for models that don't have them

Then there's missing support for a few things:

* Progressive jpeg (because non of the image crates supported it when the importer was created).
* Morph targets (hard)
* Cameras
* Extensions?

anderejd:
@Rhuagh Thanks for the update! The progressive jpeg support should be ready to merge #877 as far as I know.

> I would wait a few days to analyze this. The GLTF importer is being completely rewritten basically.

Is this still in the works? Where are the related PRs? In forks? Can I assist?

jojolepro:
I'm not aware of a current gltf importer rewrite, and rhuagh is taking a break from heavy developement from what I understood. Maybe we should just continue improving the current importer to fix the issues mentionned by Rhuagh yesterday.

Rhuagh:
[...]

PR #877, opened 2018-08-14
Support progressive jpeg.

anderejd:
Good day Amethystaceans!

This is a work in progress PR. In its current form it's adding progressive jpeg support, but I was trying to solve something else, read on please ;)

I'm trying to get the reference GLTF 2.0 models to load in amethyst and ran into this error:
```
[ERROR][amethyst_assets::progress] caused by: Unsupported("progressive jpeg")
```
when trying to load: <https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/DamagedHelmet>

This PR solves that problem but the model still doesn't load. The new error is instead: 
```
[ERROR][amethyst_renderer::pass::util] Required vertex attribute buffer with format [("joint_ids", Element { format: Format(R16_G16_B16_A16, Uint), offset: 0 })] missing in mesh
[ERROR][amethyst_renderer::pass::util] Required vertex attribute buffer with format [("joint_ids", Element { format: Format(R16_G16_B16_A16, Uint), offset: 0 })] missing in mesh
[ERROR][amethyst_renderer::pass::util] Required vertex attribute buffer with format [("joint_ids", Element { format: Format(R16_G16_B16_A16, Uint), offset: 0 })] missing in mesh
...
```

The bigger picture is that I'm evaluating GLTF 2.0 support in various Rust crates to base a new project on and I've wanted to try something with Amythyst for a while. I'm willing to invest some time but it's not unlimited, I need to get somewhat ok gltf-rendering done soonish. My other candidates to base the new project on are <https://github.com/bwasty/gltf-viewer> and <https://gitlab.com/mallumo/mallumo>. It would probably make more sense for me to use gltf-viewer but it would be cool to get good GLTF 2.0 rendering in Amethyst that supports all the official sample models <https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0>

What's missing, is this all wrong, is progressive jpeg support considered harmful, let's merge?

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/877)

### DIFF
--- a/amethyst_gltf/src/format/material.rs
+++ b/amethyst_gltf/src/format/material.rs
@@ -125,11 +125,12 @@ fn deconstruct_image(data: &TextureData, offset: usize, step: usize) -> TextureD
         TextureData::Image(ref image_data, ref metadata) => {
             let metadata = metadata
                 .clone()
-                .with_size(image_data.raw.w as u16, image_data.raw.h as u16)
+                .with_size(image_data.rgba.width() as u16, image_data.rgba.height() as u16)
                 .with_format(SurfaceType::R8);
             let image_data = image_data
-                .raw
-                .buf
+                .rgba
+                .clone()
+                .into_raw()
                 .iter()
                 .dropping(offset)
                 .step(step)
--- a/amethyst_renderer/src/formats/texture.rs
+++ b/amethyst_renderer/src/formats/texture.rs
@@ -1,6 +1,6 @@
-pub use imagefmt::Error as ImageError;
+use image::DynamicImage;
+use image::RgbaImage;
 
-use std::io::Cursor;
 use std::result::Result as StdResult;
 
 use amethyst_assets::{
@@ -11,8 +11,6 @@ use amethyst_core::specs::prelude::{Entity, Read, ReadExpect};
 use gfx::format::{ChannelType, SurfaceType};
 use gfx::texture::SamplerInfo;
 use gfx::traits::Pod;
-use imagefmt;
-use imagefmt::{ColFmt, Image};
 use tex::{Texture, TextureBuilder};
 use Renderer;
 
@@ -241,7 +239,27 @@ where
 #[derive(Clone, Debug)]
 pub struct ImageData {
     /// The raw image data.
-    pub raw: Image<u8>,
+    pub rgba: RgbaImage,
+}
+
+
+fn load_into_rgba8_from_memory(data: Vec<u8>, options: TextureMetadata) -> Result<TextureData> {
+    use image::load_from_memory;
+    load_from_memory(&data)
+        .map(|image| {
+            match image {
+                DynamicImage::ImageRgba8(im) => im,
+                _ => {
+                    // TODO: Log performance warning.
+                    image.to_rgba()
+                },
+            }
+        })
+        .map(|rgba| {
+            TextureData::Image(ImageData { rgba }, options)
+        })
+        // TODO: Add more context? File path or containing gltf archive?
+        .chain_err(|| "Image decoding failed")
 }
 
 /// Allows loading of jpg or jpeg files.
@@ -251,9 +269,7 @@ pub struct JpgFormat;
 impl JpgFormat {
     /// Load Jpg from memory buffer
     pub fn from_data(&self, data: Vec<u8>, options: TextureMetadata) -> Result<TextureData> {
-        imagefmt::jpeg::read(&mut Cursor::new(data), ColFmt::RGBA)
-            .map(|raw| TextureData::Image(ImageData { raw }, options))
-            .chain_err(|| "Image decoding failed")
+        load_into_rgba8_from_memory(data, options)
     }
 }
 
@@ -274,9 +290,7 @@ pub struct PngFormat;
 impl PngFormat {
     /// Load Png from memory buffer
     pub fn from_data(&self, data: Vec<u8>, options: TextureMetadata) -> Result<TextureData> {
-        imagefmt::png::read(&mut Cursor::new(data), ColFmt::RGBA)
-            .map(|raw| TextureData::Image(ImageData { raw }, options))
-            .chain_err(|| "Image decoding failed")
+        load_into_rgba8_from_memory(data, options)
     }
 }
 
@@ -302,9 +316,7 @@ impl SimpleFormat<Texture> for BmpFormat {
     fn import(&self, bytes: Vec<u8>, options: TextureMetadata) -> Result<TextureData> {
         // TODO: consider reading directly into GPU-visible memory
         // TODO: as noted by @omni-viral.
-        imagefmt::bmp::read(&mut Cursor::new(bytes), ColFmt::RGBA)
-            .map(|raw| TextureData::Image(ImageData { raw }, options))
-            .chain_err(|| "Image decoding failed")
+        load_into_rgba8_from_memory(bytes, options)
     }
 }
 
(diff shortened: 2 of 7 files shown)


anderejd:
Related to: https://github.com/amethyst/amethyst/issues/788

Rhuagh:
Your render pass descriptor is configured for skeletal animation and the model you're loading have no joint ids/weights vertex buffers defined. This should be fixed somehow, but it's a somewhat complex issue because we can't render some entities with some passes, and some other entities with some other pass currently. An interim solution would be to add empty joint id/weight buffers if there is no such information in the gltf file.

amethyst_gltf/src/format/material.rs
```diff
@@ -125,11 +125,12 @@ fn deconstruct_image(data: &TextureData, offset: usize, step: usize) -> TextureD
...
             let metadata = metadata
                 .clone()
-                .with_size(image_data.raw.w as u16, image_data.raw.h as u16)
+                .with_size(image_data.rgba.width() as u16, image_data.rgba.height() as u16)
                 .with_format(SurfaceType::R8);
             let image_data = image_data
-                .raw
-                .buf
+                .rgba
+                .clone() // Should this be avoided? The function takes data by reference.
```
Rhuagh: It's fine, it's just during load and not during use.

amethyst_renderer/src/formats/texture.rs
```diff
@@ -241,7 +241,26 @@ where
 #[derive(Clone, Debug)]
 pub struct ImageData {
     /// The raw image data.
-    pub raw: Image<u8>,
+    pub rgba: RgbaImage,
+}
+
+
+fn load_into_rgba8_from_memory(data: Vec<u8>, options: TextureMetadata) -> Result<TextureData> {
+    ::image::load_from_memory(&data)
```
Rhuagh: We try to avoid this type of crate referencing if we can, and instead import the function.

examples/gltf_new/main.rs
```diff
@@ -0,0 +1,260 @@
...
+        )
+        .with(
+            GltfSceneLoaderSystem::default(),
+            "gltf_loader",
+            &["scene_loader"],
+        )
+        .with_basic_renderer(
+            path,
+            DrawPbmSeparate::new()
+                .with_vertex_skinning()
```
Rhuagh: This is where you request skinning, which requires vertex attributes for joint ids and weights.

anderejd: Oh right. It did not occur to me to check out the example main.rs... *Facepalm*

Should the engine or the game code try to analyze the GLTF data before loading it in order to request `.with_vertex_skinning()`? I like your suggested solution:
> An interim solution would be to add empty joint id/weight buffers if there is no such information in the gltf file.

It makes sense to me that the game code should not need to know the internals of each GLTF file that might be loaded, before they are loaded. Makes sense?

amethyst_renderer/src/formats/texture.rs
anderejd: Would the following be the preferred format:
```
    use ::image::load_from_memory;                                                                   |                                                                                                      
    load_from_memory(&data)
```
?

jojolepro: yes

amethyst_gltf/src/format/material.rs
anderejd: Removing the comment.

examples/gltf_new/main.rs
Rhuagh: Yep, the reverse works fine (models with joint ids/weights in passes without skinning), so just defaulting to empty buffers for those should solve the most immediate problem.

amethyst_renderer/src/formats/texture.rs
Rhuagh: Yes.

anderejd: Ok, fixing.

examples/gltf_new/main.rs
anderejd: Should I include the empty buffers change in this PR too? In that case could you point me in the right direction? :)

anderejd:
Should the new example be part of this PR? The `gltf_new` example is just a copy-paste of the old `gltf` example but using a different path to the gltf-file, that's the only difference. The path to the new gltf file is located outside of the amethyst git repos at the moment for local testing. Should I include the Khronos sample model in this repos, or add the samples repos as a git submodule perhaps?

Rhuagh:
I would remove gltf_new, and keep the sample models outside, and add documentation to the gltf example on how to test the other models.

examples/gltf_new/main.rs
```diff
@@ -230,7 +230,7 @@ fn main() -> Result<(), amethyst::Error> {
         .with_basic_renderer(
             path,
             DrawPbmSeparate::new()
-                .with_vertex_skinning()
+                //.with_vertex_skinning()
```
jojolepro: Either remove it, or add a comment explaining why this line is commented and how it can be used if uncommented.

jojolepro: Either remove it, or add a comment explaining why this line is commented and how it can be used if uncommented.

jojolepro:
It depends how complex and tedious making the experimental PR is.
If you are unsure about the idea, an rfc is better than to spend time making an implementation that doesn't work.

anderejd:
screenshot from 2018-08-14 20-23-25

It did load with the `.with_vertex_skinning()` commented out, the coordinate system all wanky and colors looking weird, no reflections etc. This will need some more word to look good, so I'm removing the new example, just wanted to post some visual results! :)

anderejd:
> It depends how complex and tedious making the experimental PR is.
If you are unsure about the idea, an rfc is better than to spend time making an implementation that doesn't work.

Ok I'll open an RFC issue.

anderejd:
Thank you both for the review!

Can we merge? Should I change something else?

anderejd:
Thanks for the update!

jojolepro:
@Rhuagh has everything you asked been fixed?

jojolepro:
Please add a changelog entry in docs/CHANGELOG.md

anderejd:
Added entry in changelog, rebased on master and squashed.

anderejd:
I just realized `develop` is the default branch. Rebased on `develop` instead.

jojolepro:
Just gotta get the ok from rhuagh

jojolepro:
bors r+
thanks!

jojolepro:
bors r-

you need to add a link to this pr in CONTRIBUTING.md

anderejd:
> you need to add a link to this pr in CONTRIBUTING.md

I'm sorry I don't follow. Should this PR change the contents of CONTRIBUTING.md?

minecrawler:
I think @jojolepro meant that you have to add a link in the CHANGELOG.md, because you only added the change text, but no link.

See Submission Checklist Point 5 in CONTRIBUTING.md ;)

anderejd:
Added links to this PR in the CHANGELOG.md.

docs/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 
 ## [0.8.0] - 2018-08
 ### Added
+* Support for progressive jpeg loading. ([#877])
```
jojolepro: This should be under the unreleased section of the changelog. Also, just adding the pr number is no enough. You also need to create a link. Check how it was made for others in the same file. [#888]: httpsblabla
With your pr number

anderejd: I don't see any modification to `CHANGELOG.md` in #888

anderejd:
Rebased on develop and resolved merge conflicts in CHANGELOG.md.

I formatted the news entries in the changelog in the same way as the other new entries. They did not have any http links, only github issue links.

jojolepro:
Please check my last commit https://github.com/amethyst/amethyst/pull/877/commits/c22f3c1f9fd5ca859f0d98d3e9ca0e9d6c1d3612
so see how I did it.

Thanks for your contribution!

bors r+

anderejd:
> Please check my last commit c22f3c1
> so see how I did it.

Thank you!

Why add something in the markdown that doesn't render?

jojolepro:
This makes your (#877) be a link instead of just showing in plain text.
In this case, it links to this PR.

anderejd:
Ah right 😄

Issue #790, opened 2018-06-14
[RFC] Labeling entities


torkleyy:
I think we should start to give entities a `Name` / `Label` component. This can be useful

* in the editor
* for scene files
* for debugging

Unresolved: should the name be unique? Without uniqueness, using it for scene files wouldn't make a whole lot of sense, but we'd need some way to enforce it.

minecrawler:
I am not sure that it always makes sense to label entities. Especially using an ECS, I don't really care about the individual entities most of the time... Taking a look at your use-cases, I cannot see a single relevant case for the release build of a game, and personally, I can only come up with a few; for example  in low-performance scenarios to not store a reference to every single "special" entity or when a marker-component is overhead for the dev. As such, I would not enforce labeling of entities, but make it optional. There should be a way to retrieve an entity by its name (for example "Player 1").

---

> should the name be unique?

imho, yes, definitely. Otherwise, labeling would not make a whole lot of sense and one could simply use a component and join.

> but we'd need some way to enforce it.

labeling could be done using a hashmap. If there is an entry at the location, reject the name. For programmatically called labeling methods, that might be a panic, since the dev does something wrong there. For an importer, it might be an `Err()`, since "the file is corrupt (multiple entities with the same name)."

torkleyy:
Yeah, it's meant to be optional.

Rhuagh:
Something like `Tag`  in utils? That is only type tagging though, so not really the same thing, but some like it atleast.

torkleyy:
@Rhuagh this would also be useful for the prefab system so the entities could be a map rather than just a list. That would make specifying the parent a little nicer.

jojolepro:
Name helps with:
-Editor scene organisation
-Runtime debugging

Name is bad for:
-Performance when using lots of entities (which is the point of having an ecs)

I think that having a name should be optionnal, and configured by entity. It would probably make sense to have just a Named component?

world.create_entity.with(Named("ball")) (standard)
world.create_entity.named("ball") (my preferred)
world.create_named_entity("ball") (opinions?)

OvermindDL1:
> It would probably make sense to have just a Named component?

That's what I do.  Unsure why it would be done any other way?

jojolepro:
I think we agreed on making a Named (or Label) optionnal component to name entities and to be able to access and visually recognize them.

Tasks:
* Create the component.
* Add a entity builder utility function .named("the_name")

jojolepro:
*reserving this issue*

Issue #460, opened 2017-10-26
Improve errors and error handling


torkleyy:
So we need

* more error types
* include more information
* allow better handling of them (e.g. improve `amethyst_assets::Progress` stuff)

There are many locations where such improvements can be made, so I'm gonna set this to `easy` and ask for help here. Small PRs are preferred here.

torkleyy:
If there are any questions, I'll gladly assist you!

minecrawler:
What about logging?

torkleyy:
We also need logging, should be a separate issue though.

torkleyy:
I'd suggest to use `error_chain` for this.

Rhuagh:
YES!

torkleyy:
So I'm working on better error handling in the assets crate now.

torkleyy:
I did the hard part in #464 which introduces much better errors for asset handling. What is left are much easier jobs for the other crates:

* `amethyst_gltf`
* `amethyst_renderer`
* `amethyst`

All this crates would benefit from using `error_chain!`.

Binero:
This issue is strongly tied to https://github.com/amethyst/amethyst/issues/301.

Aceeri:
Thoughts on using [`failure`](https://github.com/withoutboats/failure) for this?

Binero:
What are the advantages of said crate? It says it tries to solve a lot of problems `quick-error` and `error-chain` have, but doesn't go into more detail.

Hittherhod:
From listening to a talk about `failure` it seems like `error-chain` encourages things like a single unified error type for a library as well as having a steep learning curve for newcomers (macros given pretty bad error messages if you mess them up).

omni-viral:
@hittherhod. From my experience you easily can define multiple `error_chain`s in crate and establish links between them to use `?`. Or use `chain_err` function to stack up errors for better context of what went bad where you handle errors.

Rhuagh:
I'm in favor of moving to `failure` myself.

Binero:
One major setback in using the failure crate is that it's very hard to maintain type information about the errors you're returning, when there is a series of possible errors. This is especially painful when some error variants are from external libraries, as as far as I know there is no way to implement foreign errors like in error-chain.

Furthermore, much like in error chain, I don't think there is an easy way to chain errors without losing type information.

jojolepro:
for reference https://github.com/rust-lang-nursery/failure

[...]

PR #879, opened 2018-08-14
Added Named Component

jojolepro:
Fixes https://github.com/amethyst/amethyst/issues/790

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/879)


amethyst_core/src/named.rs
```diff
@@ -0,0 +1,37 @@
...
+/// An easy way to name an `Entity` and give it a `Named` `Component`.
+pub trait WithNamed
+where
+    Self: Sized,
+{
+    /// Adds a name to the entity being built.
+    fn named(self, name: &'static str) -> Self;
+}
+
+impl<'a> WithNamed for EntityBuilder<'a> {
```
Xaeroxe: Psst, add one for the other builder types too

jojolepro: which are?

Xaeroxe: [Lazy](https://docs.rs/specs/0.12.1/specs/world/struct.LazyBuilder.html) and [EntityRes](https://docs.rs/specs/0.12.1/specs/world/struct.EntityResBuilder.html)

jojolepro: Added LazyBuilder. I cannot add EntityResBuilder without changing the signature of the interface, so I'm not sure what to do about that.

Xaeroxe: You could add a second trait with the new interface

jojolepro:
Missing changelog

amethyst_core/src/named.rs
```diff
@@ -0,0 +1,45 @@
...
+    fn named(self, name: &'static str) -> Self {
+        // Unwrap: The only way this can fail is if the entity is invalid and this is used while creating the entity.
+        self.world
+            .system_data::<(WriteStorage<'a, Named>,)>()
+            .0
+            .insert(self.entity, Named::new(name))
+            .unwrap();
```
LucioFranco: Would expect with a message be better here than just an unwrap, (granted its been a while since Ive actually written a lot of rust so I could be wrong).

Xaeroxe: This is what the rest of the builders do.  The only reason this would fail is because the entity was dead, and there's no way it's dead while it's still being initialized.

That being said, some kind of comment explaining that could go well here.

jojolepro: hmmmmmmmmmmm. This was already there?
```
// Unwrap: The only way this can fail is if the entity is invalid and this is used while creating the entity.
```

amethyst_core/src/named.rs
jojolepro: I'm not sure if it makes more sense to have a second interface for the same thing, or to just let people use 
```rs
.with(Named::new("my name"), &mut named_storage)
```
vs
```rs
use amethyst::ecs::world::EntityResBuilder;
.named("my name", &mut named_storage)
```

amethyst_core/src/named.rs
Xaeroxe: haha thanks I wasn't paying very close attention

amethyst_core/src/named.rs
Xaeroxe: Yeah I'll leave that choice to you, just wanted to clue you in to the existence of the other builders.

amethyst_core/src/named.rs
LucioFranco: Ok, yeah was just making sure, I've always been very cautious around leaving unwraps in released code but I wasn't sure if this was right or not.

jojolepro: Talking about unwraps, anyone can check on the issue related to error crates. I would like to have a better system to handle errors than making a ton of enums for them, and nobody seems to know which crate to choose yet.

amethyst_core/src/named.rs
jojolepro: I'll go with only "with" for now. I feel like having two traits would be more confusing than anything.

jojolepro: Thanks for teaching me :)

amethyst_core/src/named.rs
LucioFranco: @jojolepro open an issue unless there is already one, so that way we can discuss there.

jojolepro: There is one already https://github.com/amethyst/amethyst/issues/460

PR #888, opened 2018-08-19
reformat: Again.

Rhuagh:

PR #889, opened 2018-08-19
 fix: Remove the need to manually add GlobalTransform when using Transform

Rhuagh:
After reformat PR (#888). Relevant code is in amethyst_core/src/transform/system.rs.


jojolepro:
8 hours before merge

PR #892, opened 2018-08-20
ProgressCounter now shows errors at runtime

### LINKED ISSUE #869: Expose the error data collected by `ProgressCounter`
https://github.com/amethyst/amethyst/blob/5acc3d0bcfc53923deb2a0583ab5ed171beb17aa/amethyst_assets/src/progress.rs#L128-L155

The user should be able to retrieve the errors and inspect them using the `ProgressCounter`. This was left out in the implementation.

jojolepro:
fixes #869

### DIFF
--- a/amethyst_assets/src/progress.rs
+++ b/amethyst_assets/src/progress.rs
@@ -144,6 +144,13 @@ impl Tracker for ProgressCounterTracker {
         asset_name: String,
         error: Error,
     ) {
+        error!(
+            "Error loading handle {}, {}, with name {}, caused by: {:?}",
+            handle_id,
+            asset_type_name,
+            asset_name,
+            error
+        );
         self.errors.lock().push(AssetErrorMeta {
             error,
             handle_id,
@@ -188,13 +195,13 @@ impl Tracker for () {
         error: Error,
     ) {
         error!(
-            "error loading handle {}, {}, with name {}, caused by: {:?}",
+            "Error loading handle {}, {}, with name {}, caused by: {:?}",
             handle_id, asset_type_name, asset_name, error
         );
         error
             .iter()
             .skip(1)
             .for_each(|e| error!("caused by: {:?}", e));
-        error!("note: to handle the error, use a `Progress` other than `()`");
+        error!("Note: to handle the error, use a `Progress` other than `()`");
     }
 }


amethyst_assets/src/progress.rs
```diff
@@ -144,6 +144,17 @@ impl Tracker for ProgressCounterTracker {
...
+            "Error loading handle {}, {}, with name {}, caused by: {:?}",
+            handle_id,
+            asset_type_name,
+            asset_name,
+            error
+        );
+        error
+            .iter()
+            .skip(1)
+            .for_each(|e| error!("caused by: {:?}", e));
```
Aceeri: Probably would want to concatenate these two `error!(...)` into one just as a caution against possible out of order printing.

jojolepro:
@Aceeri fixed

amethyst_assets/src/progress.rs
```diff
@@ -188,13 +200,13 @@ impl Tracker for () {
...
     ) {
         error!(
-            "error loading handle {}, {}, with name {}, caused by: {:?}",
+            "Error loading handle {}, {}, with name {}, caused by: {:?}",
             handle_id, asset_type_name, asset_name, error
         );
         error
             .iter()
             .skip(1)
             .for_each(|e| error!("caused by: {:?}", e));
```
LucioFranco: Do you want to do the same thing as you did above to avoid the out of order errors?

jojolepro: yes

jojolepro:
@Aceeri re-review? :)

amethyst_assets/src/progress.rs
```diff
@@ -187,14 +199,18 @@ impl Tracker for () {
         asset_name: String,
         error: Error,
     ) {
-        error!(
-            "error loading handle {}, {}, with name {}, caused by: {:?}",
-            handle_id, asset_type_name, asset_name, error
+        let mut err_out = format!(
```
magnonellie: Since this is the same piece of code repeated twice in the same file, wouldn't it be better to put it inside of a separate function?

jojolepro: very true

jojolepro: we saved 7 lines yay. 
I thought we would actually gain code from this.

jojolepro:
bors r+

yeet

PR #895, opened 2018-08-22
IsometricTransforms and where to find them

Moxinilian:
This PR introduces a new kind of transform: IsometricTransform.
This type of transform is used to make 2D isometry very easy.

### Concept
These 2D transforms convert isometric-world coordinates to local coordinates. To do so, the transform pre-computes basis change matrices, and apply them into the entity's local transform in the transform system.
That way, the user can think of an isometric 2D world just like a typical 2D world, without needing any angle math. The computations are also more efficient that way as the isometric coordinates are collapsed into local at most only once per frame.

### Ordering
Isometric objects require ordering to behave correctly.
This is achieved by setting the Z coordinate of the local transform proportionally to the y value of the local transform. That way, lower tiles will be rendered over upper tiles. If the user however needs to force some tiles to be over other tiles, this PR implements a priority value (inspired by what can be found in the Unity engine) that gives a priority to specific IsometricTransforms in the ordering (by adding the value to the Z coordinate).
This has the drawback of requiring the orthographic camera to be potentially quite far back on the Z axis. This has no side effect, but needs to be done by the user nonetheless.

### Filtering types
While trying to make the example look good, I noticed there is no way to turn on linear filtering without importing gfx into the project. Therefore, I made amethyst::renderer expose the 3 gfx types necessary to turn it on. If you feel this is inappropriate and have an alternative, please comment.

### `SpriteSheetFormat` and `XYCameraSystem`
Side effect of trying to have a good isometry example and Jojo le professionnel being annoying, this PR also implements into the engine an asset format to load sprite positions from a RON file, and a system to pilot a plane 2D camera along the X and Y axis (with zooming implemented with transform scale).

### Demo
Checkout the cute new example `2d_isometry` for an usage demo.
If anyone feels like expanding the map and adding cars in the example, that would be lovely.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/895)


examples/2d_isometry/resources/display.ron
```diff
@@ -0,0 +1,10 @@
+(
+  dimensions: Some((500, 500)),
+  max_dimensions: None,
+  min_dimensions: None,
+  fullscreen: false,
+  multisampling: 0,
+  title: "Isometry Example",
+  visibility: true,
+  vsync: true,
+)
```
magnonellie: Newline

examples/2d_isometry/map.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use amethyst::core::{GlobalTransform, IsometricTransform, Parent, Transform};
+use amethyst::ecs::prelude::*;
+use amethyst::prelude::*;
+use amethyst::renderer::{SpriteRender, SpriteSheetHandle, Transparent};
+
+pub const UNIT_DIMENSIONS: (f32, f32) = (128.0, 64.0);
+const MAP_SIZE: usize = 16;
+
+// 067: Grass
```
magnonellie: This comment is probably not needed.

examples/2d_isometry/map.rs
```diff
@@ -0,0 +1,55 @@
...
+#[cfg_attr(rustfmt, rustfmt_skip)]
+const MAP: [[u8;MAP_SIZE];MAP_SIZE] = [
```
magnonellie: I would've liked to see this use Tiled's format for loading the map, but I suppose this will do for now.

amethyst_core/src/transform/components/isometric_transform.rs
```diff
@@ -0,0 +1,142 @@
...
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(default)]
+pub struct IsometricTransform {
+    /// Position of the transform on the isometric plane
+    pub translation: Vector2<f32>,
+    /// Weight influencing the likelihood of this entity to be rendered
+    /// on top of other entities. Entities are ordered by `-local_y + priority`.
+    /// The higher the priority, the higher the likelihood of this entity
+    /// to be rendered on top. 
+    pub order_priority: f32,
```
magnonellie: I don't know how I feel about this being f32. I personally would've liked it to be either u16 or u32 since you can then use an enum for the ordering which is way more flexible. Maybe this could even be renamed to layer if we go with integers?

amethyst_core/src/transform/components/isometric_transform.rs
```diff
@@ -0,0 +1,142 @@
...
+
+impl IsometricTransform {
+    /// Builds an IsometricTransform from the local dimensions of a unit.
+    /// In other words, if you consider an isometric square unit, this function
+    /// will build the isometric grid so that this unit fits in a rectangle of
+    /// size (w;h) in the local coordinates system.
+    pub fn from_unit_dimensions(w: f32, h: f32) -> Self {
+        let angle = Deg((h / w).atan());
+        //let scale = 1.0/((h * 0.5).powi(2) + (w * 0.50).powi(2)).sqrt();
+        //let scale = 1.0/(3f32).sqrt();
```
magnonellie: Should probably remove these comments

amethyst_renderer/src/lib.rs
```diff
@@ -91,6 +91,7 @@ pub use formats::{
     GraphicsPrefab, ImageData, JpgFormat, MaterialPrefab, MeshCreator, MeshData,
     ObjFormat, PngFormat, TextureData, TextureFormat, TextureMetadata, TexturePrefab,
 };
+pub use gfx::texture::{FilterMethod, SamplerInfo, WrapMode};
```
magnonellie: I don't think it's a good idea to expose gfx types. As @jojolepro mentioned, it will be removed in the future (or at least changed to hal) so these will undoubtedly break in the future for some people.

examples/2d_isometry/tiles.rs
```diff
@@ -0,0 +1,842 @@
...
+    let sprite_sheet_handle = {
+        let loader = world.read_resource::<Loader>();
+        let sprite_sheet_store = world.read_resource::<AssetStorage<SpriteSheet>>();
+        loader.load_from_data(sprite_sheet, (), &sprite_sheet_store)
+    };
+
+    sprite_sheet_handle
+}
+
+fn get_sprite(tile_type: u8) -> SpriteDescription {
```
magnonellie: Surely this can be calculated instead of being hard-coded, right?

amethyst_renderer/src/lib.rs
Moxinilian: Do you have an alternative?

examples/2d_isometry/tiles.rs
Moxinilian: No, look at the spritesheet, it's a mess.

amethyst_core/src/transform/components/isometric_transform.rs
Moxinilian: Well as I'm adding it to an f32 value it removes the casting overhead. And nothing prevents you from making an enum associated with floating point numbers.

examples/2d_isometry/map.rs
Moxinilian: Well I guess, but this PR is about isometry more than tiling.

amethyst_renderer/src/lib.rs
magnonellie: I would say make wrapper types that are essentially the same and then convert them to gfx types when needed.

Moxinilian: Then what is the difference?

examples/2d_isometry/tiles.rs
magnonellie: Oh god, you're right. I wonder what the author was thinking xD

amethyst_renderer/src/lib.rs
magnonellie: That they won't break when we change the renderer?

Moxinilian: Wrapping them would make it break anyway.

amethyst_core/src/transform/components/isometric_transform.rs
magnonellie: That's true actually. You can't do `enum_value as f32` but I suppose doing something like `emun_value as isize as f32` upon creation would be the way to go here.

examples/2d_isometry/tiles.rs
Moxinilian: Well it doesn't waste space I guess :thonking:

amethyst_renderer/src/lib.rs
magnonellie: No it wouldn't. You just change the mapping between the types.

Moxinilian: Okay, then just re-exporting also doesn't make it break. You just have to fit to the interface.

jojolepro: Make. Wrapper. Types. To. Not. Break. User. Code. :P

Normally we don't care to break the user api, but we don't want to force people to import gfx just to break it in their face in the next version.

Moxinilian: But it breaks user code either way...

examples/2d_isometry/tiles.rs
```diff
@@ -0,0 +1,842 @@
+// TODO: Have something to import TextureAtlas as an asset.
```
jojolepro: todo

examples/2d_isometry/tiles.rs
jojolepro: prefab file please.

examples/2d_isometry/tiles.rs
```diff
@@ -0,0 +1,842 @@
...
+use amethyst::ecs::World;
+use amethyst::renderer::{
+    MaterialTextureSet, PngFormat, Sprite, SpriteSheet, SpriteSheetHandle, SpriteSheetSet, Texture,
+    TextureCoordinates, TextureMetadata, FilterMethod, SamplerInfo, WrapMode
+};
+
+const TEXTURE_ID: u64 = 0;
+const SPRITESHEET_SIZE: (f32, f32) = (1024.0, 2048.0);
+
+struct SpriteDescription {
```
jojolepro: Isn't that redefining what prefab already is?

amethyst_core/src/transform/components/isometric_transform.rs
```diff
@@ -0,0 +1,140 @@
+use cgmath::{Angle, Deg, Matrix2, SquareMatrix, Vector2};
+use specs::{Component, DenseVecStorage, FlaggedStorage};
+
+/// Component describing the position of an entity on an isometric plane.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(default)]
+pub struct IsometricTransform {
+    /// Position of the transform on the isometric plane
+    pub translation: Vector2<f32>,
+    /// Weight influencing the likelihood of this entity to be rendered
```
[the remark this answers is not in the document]
Moxinilian: Hence the following words of the sentence.

README.md
```diff
@@ -106,15 +106,23 @@ If you are compiling on Linux make sure to install the following dependencies:
 ### Ubuntu
 
 ```
-$ sudo apt install libasound2-dev libx11-xcb-dev
+$ sudo apt install libasound2-dev libx11-xcb-dev libssl-dev
```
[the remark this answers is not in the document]
Moxinilian: I don't know. Nor do I know what to do about it, in fact.

amethyst_core/src/transform/components/isometric_transform.rs
Moxinilian: cf my response above

amethyst_core/src/transform/components/isometric_transform.rs
```diff
@@ -0,0 +1,140 @@
...
+    scale: Vector2<f32>,
+    angle: Deg<f32>,
+    iso_to_local: Matrix2<f32>,
+    local_to_iso: Matrix2<f32>,
+}
+
+fn matrices_from_angle(angle: Deg<f32>, scale: Vector2<f32>) -> (Matrix2<f32>, Matrix2<f32>) {
+    // This is a change of basis matrix.
+    // It goes from the usual basis to the isometric basis.
+    let local_to_iso = Matrix2 {
```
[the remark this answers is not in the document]
Moxinilian: Why?

amethyst_core/src/transform/components/isometric_transform.rs
```diff
@@ -0,0 +1,140 @@
+use cgmath::{Angle, Deg, Matrix2, SquareMatrix, Vector2};
+use specs::{Component, DenseVecStorage, FlaggedStorage};
+
+/// Component describing the position of an entity on an isometric plane.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(default)]
+pub struct IsometricTransform {
```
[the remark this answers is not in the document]
Moxinilian: Ok. What sound be added?

amethyst_core/src/transform/components/isometric_transform.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// Weight influencing the likelihood of this entity to be rendered
+    /// on top of other entities. Entities are ordered by `-local_y + priority`.
+    /// The higher the priority, the higher the likelihood of this entity
+    /// to be rendered on top. 
+    pub order_priority: f32,
+
+    scale: Vector2<f32>,
+    angle: Deg<f32>,
+    iso_to_local: Matrix2<f32>,
+    local_to_iso: Matrix2<f32>,
```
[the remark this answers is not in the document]
Moxinilian: They're pretty straight forward. Scale is the scale of your plane, angle is the angle of the isometric projection, iso_to_local and local_to_iso are the matrices you'd expect.

amethyst_core/src/transform/components/isometric_transform.rs
```diff
@@ -0,0 +1,140 @@
...
+        y: Vector2 {
+            x: scale.y * angle.cos(),
+            y: -scale.y * angle.cos(),
+        },
+    };
+
+    // The error handling is just a panic because having the parameters
+    // of the isometry be dynamic seems unrealistic.
+    // Can quite easily be changed if need be.
+    let iso_to_local = local_to_iso.invert().unwrap_or_else(|| {
```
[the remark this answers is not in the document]
Moxinilian: No.

amethyst_core/src/transform/systems.rs
```diff
@@ -39,11 +48,15 @@ impl<'a> System<'a> for TransformSystem {
     type SystemData = (
         Entities<'a>,
         ReadExpect<'a, ParentHierarchy>,
-        ReadStorage<'a, Transform>,
+        ReadStorage<'a, IsometricTransform>,
+        WriteStorage<'a, Transform>,
```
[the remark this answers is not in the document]
Moxinilian: This approach was suggested by Xaeroxe, because it is not possible to set up dependencies in an optional system.

examples/2d_isometry/camera.rs
```diff
@@ -0,0 +1,84 @@
...
+            -CAMERA_EDGE * 0.5,
+        )))
+        .with(GlobalTransform::default())
+        .with(transform)
+        .build();
+}
+
+pub struct MoveCamera;
+
+impl<'a> System<'a> for MoveCamera {
```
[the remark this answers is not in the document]
Moxinilian: I'll see what I can do.

examples/2d_isometry/tiles.rs
Moxinilian: I'm not aware of a prefab like this.

Moxinilian: Actually, you're not supposed to use prefabs for that sort of stuff.
A SpriteSheet is just an asset, it should not be in a prefab.

amethyst_core/src/transform/components/isometric_transform.rs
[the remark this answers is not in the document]
jojolepro: To not recalculate cos 4 times and not recalculate each multiplication 2 times?

amethyst_core/src/transform/systems.rs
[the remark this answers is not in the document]
jojolepro: Yes it is. You have to do it in the bundle.

amethyst_core/src/transform/components/isometric_transform.rs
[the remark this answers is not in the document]
Moxinilian: Well this is not python, rustc is not that stupid. It produces the exact same LLVM IR.

amethyst_core/src/transform/systems.rs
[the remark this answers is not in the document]
Moxinilian: I want to reuse the scratch to automatically populate Transform tho.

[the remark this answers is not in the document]
jojolepro: More parallelism and clarity is always the reason I separate systems.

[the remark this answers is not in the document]
Moxinilian: Yeah, and I don't think it beats reusing the scratch here. Most notably because this grabs a WriteStorage on Transform. Besides, it's pretty clear what it does here.

Moxinilian:
Any other complaint?

amethyst_core/src/transform/systems.rs
```diff
@@ -56,9 +78,26 @@ impl<'a> System<'a> for TransformSystem {
...
+        );
+        isoms.populate_modified(
+            self.modified_isom_id.as_mut().unwrap(),
+            &mut self.isom_modified,
+        );
+
+        // Apply IsometricTransforms to their associated Transform.
+        for (_, isom, mut local) in (&self.isom_modified, &isoms, &mut locals).join() {
+            let v = isom.local();
+            local.translation = v.extend(-v.y + isom.order_priority);
```
msiglreith: What's the motivation behind this one?

msiglreith: Reading the example code: so extending the vec2 to a vec3, where z is encodes the drawing order (somehow).

examples/2d_isometry/cars.rs
```diff
@@ -0,0 +1,138 @@
...
+
+pub fn load_cars_sprite_sheet(world: &mut World) -> SpriteSheetHandle {
+    let texture_handle = {
+        let loader = world.read_resource::<Loader>();
+        let texture_storage = world.read_resource::<AssetStorage<Texture>>();
+        loader.load(
+            "texture/vehicles_blue.png",
+            PngFormat,
+            TextureMetadata::default()
+                .with_sampler(SamplerInfo::new(FilterMethod::Trilinear, WrapMode::Border)),
```
msiglreith: nit: There is only one mipmap level so bilinear would be fine also

examples/2d_isometry/cars.rs
```diff
@@ -0,0 +1,138 @@
...
+            &texture_storage,
+        )
+    };
+
+    let mut material_texture_set = world.write_resource::<MaterialTextureSet>();
+    material_texture_set.insert(CARS_TEXTURE_ID, texture_handle);
+
+    let mut sprites = Vec::new();
+    let tex_coords = TextureCoordinates {
+        left: 1.0 / 256.0,
```
msiglreith: these are the texture dimensions?

msiglreith: Why does it start at `1.0` and the width is different from `right - left` in unnormalized texture coordinates?

amethyst_core/src/transform/systems.rs
Moxinilian: That's basically Z ordering, that's how the sprite batch does it.

amethyst_core/src/transform/components/isometric_transform.rs
```diff
@@ -0,0 +1,143 @@
...
+            x: 1.0 / w,
+            y: -1.0 / h,
+        };
+        Self::from_angle(angle, scale)
+    }
+
+    /// Builds an IsometricTransform from the angle between the usual x axis
+    /// and the isometric axes. In other words, if you consider an isometric unit,
+    /// the angle parameter is half the angle of the lateral corners.
+    pub fn from_angle(angle: Deg<f32>, scale: Vector2<f32>) -> Self {
```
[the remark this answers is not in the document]
Moxinilian: Well there's nothing inherently wrong in using crazy angles, as long as the change of basis matrix is invertible.

amethyst_controls/src/systems.rs
```diff
@@ -79,6 +79,83 @@ where
     }
 }
 
+#[derive(Default)]
+pub struct XYCameraSystem<A, B> {
```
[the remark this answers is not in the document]
Moxinilian: It's for the InputHandler, I don't know exactly why it needs it.

examples/2d_isometry/cars.rs
Moxinilian: The spritesheet is blurry which makes it hard to be pixel perfect.
Now regarding why I chose 32px, it's because I assumed it's what would make sense for the sprite size, but again look at the spritesheet, it's weird.

examples/2d_isometry/cars.rs
Moxinilian: Do we care? lol

amethyst_controls/src/systems.rs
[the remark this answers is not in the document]
jojolepro: Missing doc for this struct btw :/

jojolepro:
bors try

Moxinilian:
Do not try to merge, I derped somewhere.

jojolepro:
Derp Derp :meta:

Moxinilian:
I'm not sure if I understand what you mean, but most of the time you don't want everything on your display to look isometric, because then it becomes a pain to use the local y axis for height offsetting and stuff like that.

Moxinilian:
So I'm having trouble with a weird bug here.
I decided to make the API take an ``impl Into<Rad<f32>>`` instead of simply ``Deg`` in order to have it more flexible, and it made me notice that at a point I used a value in radian as a value in degree, which was a math error. So I fixed it, but now everything has a weird offset I don't understand.
I've spent two days trying to figure out what's going on, but I must not be able to see it as I've been in it for too long. Could somebody help me on this one?

Rhuagh:
Take impl Into<Angle> maybe? (https://docs.rs/cgmath/0.15.0/cgmath/trait.Angle.html)

Moxinilian:
Well the thing is I need to store the angle afterward, so it may as well be Rad.

Rhuagh:
Or that might create a type parameter, so ignore me.

Moxinilian:
Update: I will use @randomPoison's prototype Amethyst debugger to find the origin of that issue.

Moxinilian:
We're still talking about sprite-only isometry, right?

omni-viral:
Even if you don't care about 3d I think that camera will do the trick and doesn't require another transformation type.

Moxinilian:
That makes sense. The IsometricTransform idea should be dropped.
Whil I have the intuitive understanding of how that camera would work, I have no idea of what kind of matrix would do the proper translating. I'm not familiar enough with 3D math to do that unfortunately.

omni-viral:
The matrix math is simple here.
You need 3x3 matrix.
First row will convert `x` coordinate, second - `y`, third - `z`.
So you just say:
`[1, 0, 0]` must become `[-0.894427190999916, -0.447213595499958, 0.1]` - WSW
`[0, 1, 0]` must become `[0.894427190999916, -0.447213595499958, 0.1]` - ESE
`[0, 0, 1]` must become `[0, 1, 0]` - N.

The resulting matrix have those vertices as rows.
```
-0.894427190999916, -0.447213595499958, 0.1
0.894427190999916, -0.447213595499958, 0.1
0, 1, 0
```

What are magic numbers?
Length of the `[0.894427190999916, 0.447213595499958]` is `1` and `x` part is two times bigger. This is good for tilemap sprites.

Moxinilian:
I think I can recycle most of the current math I believe.
My only concern was the z axis, but in fact you're right it's pretty simple.
How would that do regarding sprite ordering?

omni-viral:
It will order most sprites naturally by `z` axis in screen space.
If you need something out of that order you might need to draw such objects without depth testing in order required after all other objects with depth testing enabled.

Moxinilian:
The problem is that it is very frequent to render things on top of other isometric sprites. For example, an isometric background with a tilemap on top with actors on top.
Also, would the z axis be meaningful with that specific matrix?

omni-viral:
With background set on high `z` value and depth testing enabled sprites will be ordered correctly.
Background will be deeper than tilemap, tilemap will be deeper than actors standing on it.

Moxinilian:
Oh I misunderstood what you meant earlier. Okay it makes perfect sense now. Thank you.

jojolepro:
@moxinillian did you get this to work? :)

Moxinilian:
Unfortunately I haven't even tried yet, and I'm sorry to say I won't have enough time to for a couple months.

LucioFranco:
@Moxinilian since you won't have time, should we close this and reopen when someone does?

Moxinilian:
Yes please do.

PR #896, opened 2018-08-22
Use Cow<'static, str> in Named

randomPoison:
Switch `Named` to represent the name using `Cow<'static, str>` instead of `&'static str` so that we can use `Named` for dynamically-generated names and names loaded from data files. The motivator here is that I would like to use `Named` for loading node names from glTF files, but it would have to support storing the names as `String`. Using `Cow<'static, str>` allows us to store `String` names without requiring an allocation if the users simply wants to use a string constant for the name.

I also added more documentation for `Named`, including a bunch of examples.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/896)

### DIFF
--- a/amethyst_core/src/named.rs
+++ b/amethyst_core/src/named.rs
@@ -1,14 +1,114 @@
 use specs::world::LazyBuilder;
 use specs::{Component, DenseVecStorage, EntityBuilder, WriteStorage};
+use std::borrow::Cow;
 
-/// A component that gives a name to an `Entity`
+/// A component that gives a name to an [`Entity`].
+///
+/// There are two ways you can get a name for an entity:
+///
+/// * Hard-coding the entity name in code, in which case the name would be a [`&'static str`][str].
+/// * Dynamically generating the string or loading it from a data file, in which case the name
+///   would be a `String`.
+///
+/// To support both of these cases smoothly, `Named` stores the name as [`Cow<'static, str>`].
+/// You can pass either a [`&'static str`][str] or a [`String`] to [`Named::new`], and your code
+/// can generally treat the `name` field as a [`&str`][str] without needing to know whether the
+/// name is actually an owned or borrowed string.
+///
+/// [`Entity`]: https://docs.rs/specs/*/specs/struct.Entity.html
+/// [`Cow<'static, str>`]: https://doc.rust-lang.org/std/borrow/enum.Cow.html
+/// [`String`]: https://doc.rust-lang.org/std/string/struct.String.html
+/// [str]: https://doc.rust-lang.org/std/primitive.str.html
+/// [`Named::new`]: #method.new
+///
+/// # Examples
+///
+/// Creating a name from string constant:
+///
+/// ```
+/// # extern crate amethyst;
+/// use amethyst::core::{Named, WithNamed};
+/// use amethyst::ecs::prelude::*;
+///
+/// let mut world = World::new();
+/// world.register::<Named>();
+///
+/// world
+///     .create_entity()
+///     .named("Super Cool Entity")
+///     .build();
+/// ```
+///
+/// Creating a name from a dynamically generated string:
+///
+/// ```
+/// # extern crate amethyst;
+/// use amethyst::core::{Named, WithNamed};
+/// use amethyst::ecs::prelude::*;
+///
+/// let mut world = World::new();
+/// world.register::<Named>();
+///
+/// for entity_num in 0..10 {
+///     world
+///         .create_entity()
+///         .named(format!("Entity Number {}", entity_num))
+///         .build();
+/// }
+/// ```
+///
+/// Accessing a named entity in a system:
+///
+/// ```
+/// # extern crate amethyst;
+/// use amethyst::core::Named;
+/// use amethyst::ecs::prelude::*;
+///
+/// pub struct NameSystem;
+///
+/// impl<'s> System<'s> for NameSystem {
+///     type SystemData = (
+///         Entities<'s>,
+///         ReadStorage<'s, Named>,
+///     );
+///
+///     fn run(&mut self, (entities, names): Self::SystemData) {
+///         for (entity, name) in (&*entities, &names).join() {
+///             println!("Entity {:?} is named {}", entity, name.name);
+///         }
+///     }
+/// }
+/// ```
+#[derive(Debug, Clone)]
 pub struct Named {
-    pub name: &'static str,
+    pub name: Cow<'static, str>,
 }
 
 impl Named {
-    pub fn new(name: &'static str) -> Self {
-        Named { name }
+    /// Constructs a new `Named` from a string.
+    ///
+    /// # Examples
+    ///
+    /// From a string constant:
+    ///
+    /// ```
+    /// # extern crate amethyst;
+    /// use amethyst::core::Named;
+    ///
+    /// let name_component = Named::new("Super Cool Entity");
+    /// ```
+    ///
+    /// From a dynamic string:
+    ///
+    /// ```
+    /// # extern crate amethyst;
+    /// use amethyst::core::Named;
+    ///
+    /// let entity_num = 7;
+    /// let name_component = Named::new(format!("Entity Number {}", entity_num));
+    /// ```
+    pub fn new<S>(name: S) -> Self where S: Into<Cow<'static, str>> {
+        Named { name: name.into() }
     }
 }
 
@@ -22,11 +122,11 @@ where
     Self: Sized,
 {
     /// Adds a name to the entity being built.
-    fn named(self, name: &'static str) -> Self;
+    fn named<S>(self, name: S) -> Self where S: Into<Cow<'static, str>>;
 }
 
 impl<'a> WithNamed for EntityBuilder<'a> {
-    fn named(self, name: &'static str) -> Self {
+    fn named<S>(self, name: S) -> Self where S: Into<Cow<'static, str>> {
         // Unwrap: The only way this can fail is if the entity is invalid and this is used while creating the entity.
         self.world
             .system_data::<(WriteStorage<'a, Named>,)>()
@@ -38,7 +138,7 @@ impl<'a> WithNamed for EntityBuilder<'a> {
 }
 
 impl<'a> WithNamed for LazyBuilder<'a> {
-    fn named(self, name: &'static str) -> Self {
+    fn named<S>(self, name: S) -> Self where S: Into<Cow<'static, str>> {
         self.lazy.insert::<Named>(self.entity, Named::new(name));
         self
     }
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -11,7 +11,7 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 ### Added
 * `SpriteRender` pass to draw sprites without using `Material` and `Mesh`. ([#829], [#830])
 * Sprite animation uses the `SpriteRenderChannel`. ([#829], [#830])
-* Added Named Component. ([#879])
+* Added Named Component. ([#879])([#896])
 * Support for progressive jpeg loading. ([#877])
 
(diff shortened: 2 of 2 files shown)


OvermindDL1:
Just as a note, in my system I kept my 'named' style components as having a built in array of 16 to allow for a maximum of 16 character names (though I always kept it to 15 or less so I could encode a null and I never tested that I covered all 16-length cases so eh).  This meant I could serialize out the entire storage for it (I used a single-allocation set with gravestones) with a single memcpy call.  I always tried to keep all my storage tables as serializeable with single memcpy calls and overall I was pretty successful in 95% of cases.

Moxinilian:
@OvermindDL1 Why would you want to serialize the named storage in an extremely fast manner like that?
Also, the 16 characters limitation isn't so great.

OvermindDL1:
Not just the named storage, all storage's possible.

And I always found 16 fine, enough to encode some unique entity names, plus it keeps it short enough to keep the editor display's short and keep them easily typeable.

The purpose of fast serialization was for initial network transfers and save/load.  I had near imperceptable save/load times even with hundreds of thousands of entities with a couple hundred component types.  :-)

randomPoison:
It's worth noting that Unity's ECS doesn't allow allocations in components for effectively the same reason. I don't think such a restriction makes a ton of sense for Amethyst at this point, but having no indirection in component data is a useful property when you're trying to maximize performance.

Moxinilian:
I think having a constant size pre-allocated string might indeed be a better solution.

OvermindDL1:
Approved now?  I don't see any changes?  o.O

Moxinilian:
We decided that if it becomes a problem in the future, it will be easy to change.

randomPoison:
@OvermindDL1 Ah sorry, I should clarify that I chatted briefly about this on Discord with @Moxinilian. The concerns I brought up were:

* It's better to use `Cow<'static, str>` as a first pass for usability.
* We don't yet know how `Named` is going to be used, so trying to optimize it at this point is a bit premature.
* We don't yet know what Amethyst's networking will look like, so we don't yet know if it's even necessary to regularly serialize all names (i.e. it may well be the case that you only send names infrequently if ever).
* @jojolepro also brought up the point that `Named` is meant primarily for debugging purposes, so performance is less of a concern than usability.

He agreed to go with `Cow<'static, str>` for now, though I think it would be useful to discuss alternatives in chat or in an issue here 😄

OvermindDL1:
Heh, well let me just remain as an oppose.  I *highly* am against random allocations of tiny chunks of data like that with indirections scattered around that destroy the ability to `memcpy` (or whatever Rust's equivalent is, I still need to use it more...).  For an efficient ECS with the ability to save/load quickly being able to `memcpy` entire tables is beyond useful over almost anything else.  The only things that I don't both enforcing `memcpy` in on my C++ version are things that are @Transiant and not ever serialized like active asset handles or the scene octree or so.  Being able to serialize, quite literally, 500k (that's 500,000 or 500.000 depending on your localization) entities with over 200 distinct serialized table types in <1second is invaluable.

jojolepro:
We can't really force everyone to limit their entity names to 16 characters, and prevent them from loading those from files. Its just too much of a usability limitation for the sake of performance. If we want we can even strip out Named in release build, if they are not being used as identifiers (which they probably shouldn't, except in rare cases)

OvermindDL1:
In my engine I tended to 'name' things like `Player0`/`Player1`/etc... for quick access, in addition to any special links between things and more.  I used it quite excessively in production.

Xaeroxe:
Data indirection is a necessary evil for a flexible powerful engine imo.  We try and reduce allocations as much as we can, and this PR is improving that situation, but to say none at all is far too restrictive.

OvermindDL1:
It was never too much for me, and I had indirection in many places (especially with the scripting related systems, but in general I tried to always keep tables as PoD, no pointers, handles sure, but no pointers.

Xaeroxe:
Seems like premature optimization to be honest.  I'd have just used a `String` and called it good personally.

If we really want there to be no pointers in components we'd have to completely overhaul several systems.  We're more likely to introduce overhead than to eliminate it.  Of course we only use these things when we have no better alternative but requiring a fixed size for all data types just gets out of control in a hurry.

jojolepro:
Thanks!
bors r+

OvermindDL1:
> We're more likely to introduce overhead than to eliminate it.

I'm curious as to what overhead you are referencing?  I never experienced any inefficiency from it other than occasionally a handle indirect (into trivially cacheable memory), and even then that is when a pointer would have normally been needed anyway?

Xaeroxe:
Mostly RAM usage.  Our best example of where we really need this is in the 3D animation systems.  A rig can have a wide variety of complexity, maybe for one entity we only need 5 joints but for another we need 70 joints.

If we're not allowed to use pointers that means we lose access to `Vec`s.  So our only way around this is a fixed allocation for the rig asset.  That means all of our entities now need storage for 100 joints when really we're only using the first 5.  All in the name of preventing a pointer.  Of course there are ways to optimize this, but most of them more or less involve building your own memory allocator for memory you've already requested from the OS.

Could work, but is likely to make the application a RAM hog.  We also haven't encountered a scenario that proves this kind of optimization and expenditure of time is necessary.

OvermindDL1:
Visual things like that I would never serialize, but rather build them back up based on other data (like what determines how they should be positioned and their timestep and so forth).

When I speak of things being POD for ease of memcpy'ing, I mean things that only need to be serialized, either via save/load or over a network, and I never ever serialized an animation state.

However, just playing devils advocate, if you wanted to serialize something like that, think a set of tag's on an object, which could in most cases be 0 in length, or could be 400 or whatever in between), that is when I use either a map structure designed for single allocation (C++ has quite a variety of choose from with different trade-offs), or a Handle style mapping from the entity ID's to a large allocated array of all values that index into parts of itm skipping and leaving marked gravestones when an entry increases until either a kind of GC sweep runs over it based on how fragmented it becomes, or they get reused by something small enough to fit within (a simple Bag style list of unused ranges is nice and fast for that).

But still the point stands, for ~15 years of the life of this engine I rarely ever have required anything more fancy than POD memcpy for serialization of component tables (and the few that I couldn't was generally because of interaction with external libraries that I required just to be able to build back up the mapping state, usually physics engines and such annoyances).

karroffel:
Nobody ever said no allocations, what was mentioned though are *handles*. Your animation system example nicely shows that it actually *is* totally doable without references and pointers.

Say you have something like this

```rust
struct Skeleton {
    bone_parents: Vec<i32>;
    bone_transforms: Vec<Transform>;
}

struct ResourceServer {
    // many fields here..
    skeletons: Slab<Skeleton>;
}

struct SkeletonHandle(i32);

impl ResourceServer {
    fn skeleton_create(&mut self) -> SkeletonHandle {
        let idx = self.skeletons.insert(Skeleton::new());
        // do stuff with new skeleton
        SkeletonHandle(idx)
    }

    fn skeleton_set_bone_pose(&self, handle: SkeletonHandle, bone_idx: i32, pose: Transform) {
        if !self.skeletons.contains(handle.0) { panic!("at the disco"); }
        let skeleton = &mut self.skeletons[handle.0];
        skeleton.bone_transforms[bone_idx] = pose;
    }
}
```

Then whenever someone has a SkeletonHandle in a component, it is safe to "memcpy" the component without doing any further lifetime management or any new allocations etc.

OvermindDL1:
Yes the component itself is, but now the state data for the given skeleton is stored outside of the component table and is not memcpy'able, and if that information is required to be serialized/deserialized then that ability is lost regardless.

jojolepro:
Or you can just memcopy the data in ram and write it in a separate thread. :)

OvermindDL1:
> Or you can just memcopy the data in ram and write it in a separate thread. :)
(and not save everything at once if necessary)

That is indeed what I generally do yep, but you still need to pause the game world to serialize it out so it doesn't change underneath you, which a memcpy allows for.

And no you don't have to save everything at once, but you still need a consistent snapshot of the world at the time of saving so it still has to be frozen long enough for the serialization to happen, whether that's a quick memcpy (that could then be written out on another thread) or whether it's a slow pointer-hopping tree-walking serialization horror, that part has to complete before the game world can continue.

Xaeroxe:
I also made this version:

```rust
{
        let data_a = vec!["abcdefghijklmnopqrstuvwxyz".to_string(); SIZE];
        let mut data_b = vec![String::with_capacity(26); SIZE];
        println!("Starting.");
        let start = Instant::now();
        for i in 0..SIZE {
            data_b[i].push_str(&data_a[i]);
        }
        let duration = start.elapsed();
        println!("Duration: {:?}", duration);
    }
```

I thought this would give an advantage to the pointers version.  It didn't produce a noticeable difference.

Xaeroxe:
With this in mind I propose we use this crate for the name component: https://crates.io/crates/inlinable_string

randomPoison:
Hmmmm, I agree we should strive to avoid allocations in components we want to serialize (especially if we're going to be serializing 100,000 of them in a frame), but I don't know that we necessarily should be targeting `Named` for that use case, at least not yet. We still don't get a good sense of how `Named` is going to be used (@jojolepro has argued that it be used purely for debugging purposes), so optimizing it for a use case that we don't currently have seems premature to me. I'm also worried about the general case of how we avoid allocation in components that need to be serialized; I don't expect storing data inline is always going to be a viable solution, so I think we'd be better served exploring other approaches before we try to optimize `Named` any further.

Xaeroxe:
Oh wait a minute.  I just thought of something.  We aren't using `String`, we're using `Cow`.

```rust
    {
        let data_a = vec![Cow::from("abcdefghijklmnopqrstuvwxyz"); SIZE];
        let mut data_b = Vec::with_capacity(SIZE);
        println!("Starting.");
        let start = Instant::now();
        data_b.extend(data_a.iter().cloned());
        let duration = start.elapsed();
        println!("Duration: {:?}", duration);
    }
```

This is just as fast as the inline data version, because a Cow clone doesn't actually do a mem copy.  And, names rarely change.

randomPoison:
Well, that's true so long as all your names are created from string literals. But if you were to serialize all those string literal names and send them to a remote client, it would have to deserialize all the names as `Strings`. Whether or not that actually matters depends on if that client will then need to serialize the names again, hence why I feel like we really need some more concrete use cases to inform how we try to optimize serialization.

OvermindDL1:
Yeah my strings definitely would not always be literals, plus that's still following a pointer.  ^.^;

@Xaeroxe Thanks for those tests though!  I was curious if Rust was able to help the pointer indirection much but it seems to be about as identical as I experienced in C++ a decade and a half ago when I made my requirement for everything possible to be memcpy'able.  :-)

OvermindDL1:
As an aside, I do want to point out again that for at least this specific component if I re-made it in my C++ engine nowadays I would use my [Atom type](https://github.com/OvermindDL1/OverECS/blob/master/StringAtom.hpp), it's a 64-bit integer that can pack a string long enough to be descriptive (10 chars for case-sensitive or 12 chars for case insensitive, or another character or two if you constrain the chars further and change the encoding), entirely reverseable, and can happen at compile or run time very cheaply, think of them as a global enumeration, every string of the max length maps to a valid enumeration and vice-versa, which is *fantastic* for naming things for either debugging or level identification but they remain super fast and easy to reason about.  I've not seen an atom-like type in Rust yet but a library for it would be awesome (there are a few variations of it in C++, but mine was highly customized for my use-case as I wanted perfect reverseability and not just simple compile-time collideable hashing).

jojolepro:
@Xaeroxe wouldn't your first test actually cause a reallocation of the String internal vec a couple of times also, since its basically appending one string slice at a time? I checked the rust std source and it seems to just foreach over a vec.extend_from_slice.

Xaeroxe:
It would, and thinking of this I created a version that doesn't do that.
It helps, but there's still about 1 order of magnitude difference.

Issue #898, opened 2018-08-22
Pong Tutorial - Moving The Paddles may not build as described


jackmott:
It is possible I have missed something, but I have double checked.

Following along with the tutorial here:
https://www.amethyst.rs/book/master/pong-tutorial/pong-tutorial-03.html

As described , the code in paddle.rs cannot find `Paddle` and `Side'.  The documentation mentions that Paddle and Side need to be made public, and I have done that.   The fix here may be simple but I'm still new to the Rust module system.

magnonellie:
I think you're missing this line: https://github.com/amethyst/amethyst/blob/develop/examples/pong_tutorial_03/systems/paddle.rs#L4 :wink: 

May I suggest you join our Discord where we have a help channel you can use if you need further assistance? https://discord.gg/E2Dd244

jackmott:
You are correct, that will do it. However that is completely missing from the tutorial, so we should get that added.  I'll see if I can do it.

What branch/repo do I fork to update the docs properly?

Moxinilian:
On another note, do we want to mention every single import? That's just a ton of not very useful code.
I guess we could just tell the user to assume them, but I'm not sure.

magnonellie:
I was thinking about that and I think for the sake of being as inclusive as possible, we should probably do it, yeah. If nothing else, people may read the book right after or before finishing TRPL, so being explicit is probably not a bad thing.

jackmott:
closed with merge #899

PR #899, opened 2018-08-22
add imports to pong-tutorial-03.md and make pong.rs things pub

jackmott:
This addresses #898 , I added the necessary includes to paddle.rs in the code examples. As well, on page 2, I made the elements that need to be public, pub, which I think will avoid confusion vs making them private then having to notice the offhand comment on page 3 to change them to public.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/899)

### DIFF
--- a/book/src/pong-tutorial/pong-tutorial-03.md
+++ b/book/src/pong-tutorial/pong-tutorial-03.md
@@ -141,8 +141,6 @@ impl<'s> System<'s> for PaddleSystem {
 }
 ```
 
-Note: We had to make our Paddle, Side, PADDLE_HEIGHT, and ARENA_HEIGHT public in `pong.rs`
-
 Now lets add this system to our `GameDataBuilder` in `main.rs`:
 
 ```rust,no_run,noplaypen
--- a/book/src/pong-tutorial/pong-tutorial-03.md
+++ b/book/src/pong-tutorial/pong-tutorial-03.md
@@ -100,6 +100,7 @@ We're finally ready to implement the `PaddleSystem` in `systems/paddle.rs`:
 use amethyst::core::Transform;
 use amethyst::ecs::{Join, Read, ReadStorage, System, WriteStorage};
 use amethyst::input::InputHandler;
+use pong::{Paddle, Side, ARENA_HEIGHT, PADDLE_HEIGHT};
 # pub enum Side {
 #   Left,
 #   Right,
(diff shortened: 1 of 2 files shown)


book/src/pong-tutorial/pong-tutorial-03.md
```diff
@@ -140,7 +141,7 @@ impl<'s> System<'s> for PaddleSystem {
 }
 ```
 
-Note: We had to make our Paddle and Side public in `pong.rs`
+Note: We had to make our Paddle, Side, PADDLE_HEIGHT, and ARENA_HEIGHT public in `pong.rs`
```
magnonellie: I don't think this line is needed anymore since you already declared those variables as public. What you could do instead is add a note in tutorial 2 that says they are made public for use in other modules later.

jackmott: Good idea, done.

jackmott:
I just ran it, I see two failures but they are both in state.md which I haven't touched.

jojolepro:
do you mind fixing them? :)

jackmott:
sure I'll give it a go, might be a while.

jackmott:
Ok I figured out I have to use the -L flag with mdbook test to point it at the deps, which caused it to get further and find an error in my changes.  

`use pong::{Paddle, Side, ARENA_HEIGHT, PADDLE_HEIGHT};`

Errors because of course there is no pong. I'm not sure how to finagle it so mdbook will be happy, other than tagging the whole block with ignore. any tips?

jojolepro:
You'll need to add hidden code until it compiles. It is done by adding the `#` component at the start of each line of non-visible code. You can see how it was done in other sections of the book

jackmott:
@jojolepro there appears to be no other case in the docs where the use statement isn't part of a crate.  I've tried creating a `mod pong { // stuff in here }` to no avail.

Moxinilian:
This is the reason it is not mentioned in fact, I had to cheat to make it compile and removed the usage directive. I'd recommend having them in a separate, untested block.

jackmott:
I figured that might be the case, I'll refactor it that way in a bit.

jojolepro:
bors r+

Thanks!

jojolepro:
Reeeeeeeeeee

bors r+

Issue #900, opened 2018-08-22
The from_data method of JpgFormat should have &[u8] or &mut [u8] instead of Vec<u8> as an argument.


Debily:
[Here](https://docs.rs/amethyst_renderer/0.8.0/amethyst_renderer/struct.JpgFormat.html#method.from_data) is the documentation for the function.

Some graphics are stored in static memory so it might be better to use &[u8] or &mut [u8]. Another idea is implementing a new function for &[u8] types.

Debily:
This of course applies to other fileformats too

Moxinilian:
Thank you for your PR!

PR #901, opened 2018-08-22
Changed the required Vec<u8> to &[u8] on texture loading

Debily:
Addressing [this](https://github.com/amethyst/amethyst/issues/900) issue.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/901)


Xaeroxe:
I don't see a terribly compelling reason to take a self reference in these signatures either.  Can we remove those?

Xaeroxe:
> Eventually we might just replace Vec in SimpleFormat by &[u8]

Btw IIRC we had this discussion when designing the Format API originally, and the conclusion we came to was essentially "The asset system has no further use for the raw data, so why not allow the format to take ownership of it?"  Indeed for some formats this turned out to be a good thing as we were able to avoid unnecessary memory copies of large texture assets.

jojolepro:
So is this PR good as it is? @Xaeroxe

Xaeroxe:
This PR seems to be an improvement to the functions provided.  If we wanted to go the extra mile we should also remove the `&self` in these function signatures, as taking a `&self` where self is a unit structure is kind of silly.

jojolepro:
@Debily Do you want to do that or we merge this like it is now?

Debily:
It's weird that some checks fail. I followed the contributing guidelines without any encounter of error.

Xaeroxe:
@Moxinilian can you weigh in here?

Moxinilian:
Well I think you first need to solve the merge conflict.
Then, if it does not solve the issue, I'll see what AppVeyor is trying to do.

Moxinilian:
bors retry

Moxinilian:
Could you please rebase this pull request with the develop branch? We changed some of bors' settings so you need to update them in your fork so that it can merge it correctly.

Issue #894, opened 2018-08-21
Implement sprite batching for DrawSprite


Moxinilian:
Now that the special sprite pass is merged, to be able to use it in realistic project, sprite batching must be implemented. Indeed, currently a small tilemap of 256 sprites eats 25% of a GTX 1070, which is ridiculous.
This also applies to traditional 3D meshes, even though sprites might be easier to implement first.

This should be relatively easy to implement as the sprite pass iterates over all sprite, so instead of drawing them directly it only needs to build a batch. Eventually, to get even better performances, we could consider caching the batch itself, but naive batching would already be a huge improvement.

Moxinilian:
@azriel91 If you've got some time I would really love if you could give that a look.
@termhn implemented it for ggez (but is unfortunately unable to do it for Amethyst), here is the relevant code in ggez:
https://github.com/ggez/ggez/blob/master/src/graphics/spritebatch.rs

azriel91:
This is on my radar as well, but not for at least the next two weeks -- got a lot of tasks that have to be done before that. I'll see if I can slot it in after that.

Moxinilian:
Fixed by #902 and #923.
Closing, thank you all!

PR #902, opened 2018-08-23
Sprite batching

msiglreith:
Refactor the current sprite drawing to batch consecutive draws into one instanced draw call.

The drawing part is split into two parts: Collecting (+ Sorting) and Encoding.
* Collection: Cache all sprites which are supposed to be display in the current frame. Further sorting with respect to the ordering constraints allows to move together sprites with the same texture reference.
* Encoding: Iterate over the sprite cache to find largest possible batch without state changes.
Sprite information is stored in an instance buffer.

With the move towards the approach outlined above, also a few changes to the shader are integrated (saves ~3ms):
* Use vertex buffers (instancing) with orphaning instead of updating constant buffers 
* Moved quite a bit of logic from shader to the CPU, minimizing the required data sent.

Overall went down from 22ms to 7ms (framelimiter) for the `2d_isometry` sample.

cc https://github.com/amethyst/amethyst/issues/894

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/902)

### DIFF
--- a/amethyst_renderer/src/pass/shaders/vertex/sprite.glsl
+++ b/amethyst_renderer/src/pass/shaders/vertex/sprite.glsl
@@ -5,15 +5,15 @@ layout (std140) uniform ViewArgs {
     mat4 view;
 };
 
-//
-in vec2 size;
-// Pixel offsets for the sprite. Used to shift the sprite left and down.
-in vec2 offsets;
+// Quad transform.
+in vec2 dir_x;
+in vec2 dir_y;
+in vec2 pos;
 
+// Texture quad.
 in vec2 u_offset;
 in vec2 v_offset;
 
-out vec2 uv;
 out vec2 tex_uv;
 
 const vec2 positions[6] = vec2[](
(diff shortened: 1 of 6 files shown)


amethyst_renderer/src/pass/shaders/vertex/sprite.glsl
```diff
@@ -1,31 +1,21 @@
...
-    uniform vec2 offsets;
-    // Whether to flip the sprite horizontally
-    uniform bool flip_horizontal;
-    // Whether to flip the sprite vertically
-    uniform bool flip_vertical;
-};
+//
+in vec2 size;
+// Pixel offsets for the sprite. Used to shift the sprite left and down.
+in vec2 offset;
```
jojolepro: Should be left and up, except if the coordinates issue is still not resolved.

amethyst_renderer/src/pass/sprite/mod.rs
```diff
@@ -4,7 +4,96 @@ mod interleaved;
...
+use vertex::{Attribute, AttributeFormat, Attributes, VertexFormat, With};
+
 static VERT_SRC: &[u8] = include_bytes!("../shaders/vertex/sprite.glsl");
 static FRAG_SRC: &[u8] = include_bytes!("../shaders/fragment/sprite.glsl");
 
 static TEXTURES: [TextureType; 1] = [TextureType::Albedo];
+
+#[derive(Clone, Debug)]
+pub enum Size {}
+impl Attribute for Size {
```
jojolepro: Pub stuff needs documentation

amethyst_renderer/src/pass/shaders/vertex/sprite.glsl
```diff
@@ -1,31 +1,21 @@
...
-    uniform vec2 sprite_dimensions;
-    // Pixel offsets for the sprite. Used to shift the sprite left and down.
-    uniform vec2 offsets;
-    // Whether to flip the sprite horizontally
-    uniform bool flip_horizontal;
-    // Whether to flip the sprite vertically
-    uniform bool flip_vertical;
-};
+// Quad transform.
+in vec2 size;
```
msiglreith: I need to split that actually, one for each basis.

msiglreith:
Fix sprite rotation, should be ready now.

amethyst_renderer/src/pass/sprite/interleaved.rs
```diff
@@ -101,49 +107,225 @@ impl Pass for DrawSprite {
...
+            warn!(
+                "Sprite sheet not loaded for sprite_render: `{:?}`.",
+                sprite_render
+            );
+            return;
+        }
+        let sprite_sheet = sprite_sheet.unwrap();
+
+        let texture_handle = material_texture_set.handle(sprite_sheet.texture_id);
+        if texture_handle.is_none() {
```
jojolepro: if let some... else 
might be cleaner

msiglreith:
@jojolepro yes, flipping works

Moxinilian:
Could you also squash the commits please? It will look cleaner in the commit history.

msiglreith:
Squashed down to two commits

Moxinilian:
Thank you very much for your work.
bors r+

PR #905, opened 2018-08-24
Load node names from glTF files

randomPoison:
This PR adds support for loading node names from glTF files. Node names are applied to the corresponding entities in the instantiated hierarchy via the new `Named` component. Only node names are imported; Other items in a glTF file can have names, but this PR makes not attempt to support them.

This PR also implements `PrefabData` for `Named`, which should enable broader support for specifying entity names in prefab.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/905)


randomPoison:
Does this need to be re-run or something? The AppVeyor build failed for unrelated reasons as far as I can tell (and I think AppVeyor has been disabled since I made this PR >.> )

PR #907, opened 2018-08-24
Maintenance/906/update dependencies

### LINKED ISSUE #906: Bump dependency versions (winit, glutin, gfx-window-glutin, hetseq)
Version bumps:

winit: 0.15 -> 0.17
glutin: 0.16 -> 0.18
gfx_window_glutin: 0.24 -> 0.26
hetseq: 0.1.5 -> 0.2

Reasons:
* save the world before it collapses (e.g. linux will break on Rust 1.30 for winit versions prior to 0.17.1
* stability

**Breakages to watch out for:**
Winit:
- **Breaking:** Removed `VirtualKeyCode::LMenu` and `VirtualKeyCode::RMenu`; Windows now generates `VirtualKeyCode::LAlt` and `VirtualKeyCode::RAlt` instead.
- **Breaking:** `Window::hidpi_factor` has been renamed to `Window::get_hidpi_factor` for better consistency. `WindowEvent::HiDPIFactorChanged` has been renamed to `WindowEvent::HiDpiFactorChanged`. DPI factors are always represented as `f64` instead of `f32` now.
- **Breaking:** The entire API for sizes, positions, etc. has changed. In the majority of cases, winit produces and consumes positions and sizes as `LogicalPosition` and `LogicalSize`, respectively. The notable exception is `MonitorId` methods, which deal in `PhysicalPosition` and `PhysicalSize`. See the documentation for specifics and explanations of the types. Additionally, winit automatically conserves logical size when the DPI factor changes.
- **Breaking:** All deprecated methods have been removed. For `Window::platform_display` and `Window::platform_window`, switch to the appropriate platform-specific `WindowExt` methods. For `Window::get_inner_size_points` and `Window::get_inner_size_pixels`, use the `LogicalSize` returned by `Window::get_inner_size` and convert as needed.
- **Breaking:** `Window::set_cursor_state` and `CursorState` enum removed in favor of the more composable `Window::grab_cursor` and `Window::hide_cursor`. As a result, grabbing the cursor no longer automatically hides it; you must call both methods to retain the old behavior on Windows and macOS. `Cursor::NoneCursor` has been removed, as it's no longer useful.
- **Breaking:** `Window::set_cursor_position` now returns `Result<(), String>`, thus allowing for `Box<Error>` conversion via `?`.
- **Breaking:** On iOS, `UIView` is now the default root view. `WindowBuilderExt::with_root_view_class` can be used to set the root view objective-c class to `GLKView` (OpenGLES) or `MTKView` (Metal/MoltenVK).

Glutin:
- **Breaking:** Added `OsError` variant to `ContextError`.

azriel91:
Closes #906.

Changelog change is gonna conflict, feel free to push over it if there's no other changes to be made.

Tested Pong and fly_camera examples to ensure hide_cursor works. The fly_camera example actually panics on `MaskedStorage<Named>` not existing, so I turned that off when testing the cursor change.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/907)


amethyst_renderer/src/mouse.rs
```diff
@@ -1,34 +1,42 @@
...
+///
+/// This grabs the cursor and hides it, so callers do not have to make separate calls.
 pub fn set_mouse_cursor_none(msg: &mut WindowMessages) {
-    set_mouse_cursor(msg, MouseCursor::NoneCursor);
+    msg.send_command(move |win| {
+        if let Err(err) = win.grab_cursor(true) {
+            error!("Unable to grab the cursor. Error: {:?}", err);
+        }
+        win.hide_cursor(true);
+    });
```
jojolepro: Hiding and grabbing are not the same thing. They should stay separated

amethyst_controls/src/systems.rs
```diff
@@ -248,7 +248,6 @@ impl<'a> System<'a> for CursorHideSystem {
         self.event_reader = Some(res.fetch_mut::<EventChannel<Event>>().register_reader());
 
         let mut msg = res.fetch_mut::<WindowMessages>();
-        grab_cursor(&mut msg);
```
jojolepro: ???

amethyst_renderer/src/mouse.rs
azriel91: #rookie_move
fixing

azriel91:
Yay so fixed the grab == hide thing, but we get this error message logged whenever re-focusing on a window where the cursor was previously grabbed (even if ungrabbing when unfocusing):

```
[ERROR][amethyst_renderer::mouse] Unable to grab the cursor. Error: "Cursor could not be grabbed: already grabbed by another client"
```

I've opened https://github.com/tomaka/winit/issues/639 in case it's a bug, but otherwise behaviour of hiding the cursor is as before

jojolepro:
okay

jojolepro:
When you see a PR stalling like this feel free to spam people a bit on discord to get that second review. There isnt many devs checking open pr without being asked to do it.

Moxinilian:
Unfortunately winit's issue tracker isn't very active. I think the error message is acceptable for now.

Issue #831, opened 2018-08-01
Remove all env!("CARGO_MANIFEST_DIR")


jojolepro:
env!("CARGO_MANIFEST_DIR")
doesn't work when you actually make a release product as it insteads target the environment where it was compiled.

In my game dash-raid, I am using the following piece of code to determine the game's directory at runtime:
```
let bin_path = env::args().next().expect("Failed to get binary executable path");
    let last_slash_index = bin_path.rfind("/").expect("Failed to get last slash in binary path.");
    let mut base_path = bin_path[..last_slash_index].to_string();
    
    if base_path.contains("target/"){
        base_path = String::from(".");
}
```

It is not really safe, and was tested only on linux. Also, I didn't check if that works when using middleman callers (example: nohup, steam, nautilus, etc...).

TODO:
1) Come up with a clean solution for that.
2) Make all instances of CARGO_MANIFEST_DIR use the resource folder instead (the one we declare when creating Application)
3) Update examples and docs
4) profit :ok_hand: 

Priority of important because it prevents releasing games if you use the profiler or prefabs.

omni-viral:
> `let last_slash_index = bin_path.rfind("/")`

You really shouldn't reimplement `std::path::Path::parent`

jojolepro:
Oh, I didn't know that was a thing. Thanks!

onbjerg:
What about [`std::env::current_exe`](https://doc.rust-lang.org/std/env/fn.current_exe.html)? Looks to fit, can pop off the top portion of the [`PathBuf`](https://doc.rust-lang.org/std/path/struct.PathBuf.html).

jojolepro:
If we do make a check for symlinks yes

onbjerg:
@omni-viral Except if I open up the game using an absolute path in another directory then `std::env::current_dir()` will not reflect the directory we are actually looking for (the directory with the data files).

omni-viral:
Then [std::env::current_exe](https://doc.rust-lang.org/std/env/fn.current_exe.html) is what you need.

Rhuagh:
That won't work when running from cargo, because the executable files are created in temporary directories under `build/`, and you have no way to guarantee its location as compared to the asset files.

Rhuagh:
This is the reason why we use the cargo manifest directory as the root location in the examples. This strategy should not be used in actual full games, for those you need to separate the logic for each platform. It is for this reason we created the `platform` crate.

jojolepro:
[...]

PR #911, opened 2018-08-24
New application_root_dir() function

scorbeil:
Added in amethyst_utils.  See [#831]

### DIFF
--- a/amethyst_utils/src/app_root_dir.rs
+++ b/amethyst_utils/src/app_root_dir.rs
@@ -1,0 +1,18 @@
+use std::env;
+use std::fs;
+
+// Returns the cargo manifest directory for debug builds or the directory in which the current
+// executable resides for release builds, traversing a symlink if necessary.
+pub fn application_root_dir() -> String {
+    if env::var("PROFILE").unwrap() == "release" {
+        let path = env::current_exe().expect("Failed to find executable path.");
+        String::from(
+            match fs::read_link(path) {
+                Ok(target) => target.parent(),
+                Err(_) => path.parent(), 
+            }.expect("Failed to get parent directory of the executable.").to_str().unwrap()
+        )
+    } else {
+        env!("CARGO_MANIFEST_DIR")
+    }
+}
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 * State::handle_event can now handle multiple types of events. ([#887])
 * Added Named Component. ([#879])([#896])
 * Support for progressive jpeg loading. ([#877])
+* New `application_root_dir()` function in `amethyst_utils`. ([#831])
 
 ### Changed
 * Sprites contain their dimensions and offsets to render them with the right size and desired position. ([#829], [#830])


Moxinilian:
Are you sure this works outside of cargo context?

scorbeil:
That's what the check for the PROFILE env var does.  If you run cargo with --release it will be set to "release", otherwise "debug" (per Rust docs)

scorbeil:
Oh, I see what you're saying now.  No, I'm actually not sure about that.  I'll test it out when I get back to my desk

Moxinilian:
Okay but if you run `cargo run --release` you don't want the root to be executable's position, as the executable is in `/target/release` while your assets are in `/assets`.
What about outside of cargo, too?

amethyst_utils/src/app_root_dir.rs
```diff
@@ -0,0 +1,18 @@
+use std::env;
+use std::fs;
+
+// Returns the cargo manifest directory for debug builds or the directory in which the current
+// executable resides for release builds, traversing a symlink if necessary.
+pub fn application_root_dir() -> String {
+    if env::var("PROFILE").unwrap() == "release" {
```
jojolepro: You should definitely not use unwrap here. If you run the executable directly ./gamename it will panic
If "PROFILE" is set to either release or debug, use CARGO_MANIFEST_DIR.
Otherwise, that means the game was executed using ./gamename or by clicking on it (verify that's its the case).

amethyst_utils/src/app_root_dir.rs
```diff
@@ -0,0 +1,18 @@
...
+        String::from(
+            match fs::read_link(path) {
+                Ok(target) => target.parent(),
+                Err(_) => path.parent(), 
+            }.expect("Failed to get parent directory of the executable.").to_str().unwrap()
+        )
+    } else {
+        env!("CARGO_MANIFEST_DIR")
+    }
+}
```
jojolepro: Add a new line after this one please.

amethyst_utils/src/app_root_dir.rs
```diff
@@ -0,0 +1,18 @@
+use std::env;
+use std::fs;
+
+// Returns the cargo manifest directory for debug builds or the directory in which the current
+// executable resides for release builds, traversing a symlink if necessary.
+pub fn application_root_dir() -> String {
+    if env::var("PROFILE").unwrap() == "release" {
+        let path = env::current_exe().expect("Failed to find executable path.");
+        String::from(
+            match fs::read_link(path) {
```
jojolepro: What if a symlink points to a symlink?

amethyst_utils/src/app_root_dir.rs
scorbeil: After line 18?  As in, you just want a blank line at the end of the function?

amethyst_utils/src/app_root_dir.rs
scorbeil: I was assuming that "PROFILE" was set at compilation for the binary.  I see now that it's an env var set by cargo that won't be set when run without it.  Silly mistake :/

amethyst_utils/src/app_root_dir.rs
scorbeil: Good call.  Will fix.

scorbeil:
> After fixing the app_root_dir function, _make the profiler use it_...

Not following...?

amethyst_utils/src/app_root_dir.rs
scorbeil: At the end of the file.  I see.

jojolepro: yes

scorbeil:
Not really sure what you mean in that line I quoted above still, but this version should work.  I tested it in all of the use cases you mentioned.

An important note: the decision whether to use env!("CARGO_MANIFEST_DIR") or env::current_exe() is entirely based on whether or not the binary was run with cargo.  You can move the binary wherever you want and run it without cargo and get a useless path, although there's not much we can do about people moving stuff around manually, eh? :)

I did test with multiple levels of indirection via symlinks and it worked fine.

Moxinilian:
If you've got time however it would be really appreciated if you could replace stuff that use it in the source code and examples.

scorbeil:
@Moxinilian I planned on doing so.  Just wanted to make sure the function was acceptable first.  Are you all okay with the behavior being entirely dictated by whether the binary was run with cargo?  Also, is that module acceptable?  e.g...

```
use amethyst::utils::app_root_dir;

let foo = app_root_dir::application_root_dir();
```

Moxinilian:
Add a ``pub use self::app_root_dir::*;`` to ``lib.rs`` instead, it will be easier to use.
The interesting part of this is that we now have a centralized place for the root directory of the application. For now the typical amethyst workflow is dictated by cargo, but when we will want to change it we will just have to change this function.

scorbeil:
Can you help me out with a silly issue I'm struggling with?  Not sure how to import this in the examples.

As an - ahem - example...

In `amethyst_config/examples/main.rs`, if I add `extern crate amethyst_utils;` the compiler complains [E0463]: can't find crate for 'amethyst_utils', but that seems to be the way amethyst_config was imported.

Apologies, but I'm definitely getting that practice I was looking for.

Moxinilian:
When writing code in the examples, the main Amethyst crate is exposed. This main crate re-exports all the other amethyst subcrates.
That means you only need to do `extern crate amethyst;`. Then, when you want to access the content of a subcrate, you do it like `amethyst::core::Transform` or `amethyst::utils::application_root_dir`.
The Rust submodule system is a bit confusing at first, don't worry.

scorbeil:
That's not working.  When I `extern crate amethyst;` the compiler complains that it doesn't exist, but if I remove it I get `error[E0433]: faile to resolve.  Maybe a missing 'extern crate amethyst;'?`

This is only failing in the examples.  The crate builds fine, but the examples fail when I run `cargo test --all`.

When you have a minute, take a peek in `amethyst_config/examples/main.rs` to see what I mean.  Other subcrates are being imported with (for example) `extern crate amethyst_assets;`, but that doesn't work for `extern crate amethyst_utils;`

Moxinilian:
I don't know why there is an example in `amethyst_config`.
`amethyst_utils` is higher in the dependency chain and therefore can not be accessed by stuff in the other crates.
One solution would be to move the utility function to `core`, but I don't know if it really belongs there.

scorbeil:
That's just the first one that's failing.  Many of the examples had `env!("CARGO_MANIFEST_DIR")` in them.  It's an issue with how I'm importing it; I'm certain.

I really don't think this function belongs in core, either.

Moxinilian:
All the examples in the main example directory (which is the only important one, in fact) will be able to import it.
As I said, amethyst_utils is too high in the dependency chain to be used by the other crates.

scorbeil:
Okay, well I'm going to push my work (to my fork) to save where I'm at so I can proceed on my home computer later.  Tests are failing atm, so this next push is not meant to be merged.

scorbeil:
Issues resolved by removing usage of application_root_dir() from all sub-crates' tests and examples.  I could not figure out how to import amethyst::utils in those locations, so they are still using env!("CARGO_MANIFEST_DIR").  All other locations have been updated, though.

```
$ grep -r 'env!("CARGO_MANIFEST_DIR")' ./
./amethyst_assets/examples/hl.rs:    let assets_dir = format!("{}/examples/assets/", env!("CARGO_MANIFEST_DIR"));
./amethyst_assets/examples/ll.rs:    let path = format!("{}/examples/assets", env!("CARGO_MANIFEST_DIR"));
./amethyst_assets/src/source/dir.rs:        let test_assets_dir = Path::new(env!("CARGO_MANIFEST_DIR")).join("tests/assets");
./amethyst_assets/src/source/dir.rs:        let test_assets_dir = Path::new(env!("CARGO_MANIFEST_DIR"))
./amethyst_config/examples/main.rs:    let path = format!("{}/examples/display_config.ron", env!("CARGO_MANIFEST_DIR"));
./amethyst_utils/src/app_root_dir.rs:        Ok(_) => String::from(env!("CARGO_MANIFEST_DIR")),
```

jojolepro:
You needed amethyst_utils instead of amethyst::utils

scorbeil:
Tried that.  Didn't work.

scorbeil:
Forgive the blunt answer.  I was at lunch and responding on my phone.  Here's the top of `amethyst_assets\examples\ll.rs` with your suggestion:

```
//! Defining a custom asset and format.

extern crate amethyst_assets;
extern crate amethyst_core;
extern crate amethyst_utils;
extern crate rayon;
```

And here's the compiler's response to that:

```
Compiling amethyst v0.8.0 (file:///C:/Rust/forks/amethyst)
error[E0463]: can't find crate for `amethyst_utils`
 --> amethyst_assets\examples\ll.rs:5:1
  |
5 | extern crate amethyst_utils;
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't find crate

For more information about this error, try `rustc --explain E0463`.
error: Could not compile `amethyst_assets`.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

jojolepro:
You don't need to add it to cargo, nor do you need extern crate.

In the examples, you need:
```rs
extern crate amethyst;
use amethyst::utils::whateveryouneed;
```

Sorry for the wrong answer I gave, I didn't actually checked if it worked.

jojolepro:
Ah okay, I read too fast last time and now I understand what the issue is. Basically, there's not really an easy way around cyclic dependencies in this case. I suggest you leave those examples as-is for now, except if someone comes up with a smart solution.

scorbeil:
So it sounds like our options are:

1. Move `application_root_dir()` to a different module or sub-crate.  I'm not convinced that it belongs in `amethyst_core`, but it sounds like that would solve this issue.
2. Leave the sub-crate examples as-is with `env!("CARGO_MANIFEST_DIR")` (perhaps with a note/comment that `application_root_dir()` is available in real applications.

If you'd rather do 2, then this is ready for review/merge.  If you'd rather do 1, then I can move it wherever you prefer and update the examples.

docs/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 * State::handle_event can now handle multiple types of events. ([#887])
 * Added Named Component. ([#879])([#896])
 * Support for progressive jpeg loading. ([#877])
+* New `application_root_dir()` function in `amethyst_utils`. ([#831])
```
jojolepro: Here you need to add a link to the pull request.
You need to add a line like this
```
 [#831]: https://github.com/amethyst/amethyst/pull/831
```

scorbeil:
Like so?

jojolepro:
Bors r+

jojolepro:
Bors r- 
Someone fix the conflict please

jojolepro:
```rs
error[E0425]: cannot find function `application_root_dir` in this scope
   --> src/app.rs:319:54
    |
319 |         let path = format!("{}/thread_profile.json", application_root_dir());
    |                                                      ^^^^^^^^^^^^^^^^^^^^ not found in this scope
help: possible candidates are found in other modules, you can import them into scope
    |
3   | use utils::app_root_dir::application_root_dir;
    |
3   | use utils::application_root_dir;
    |
```
doesn't compile currently.

scorbeil:
Yea, I noticed and was on it.  I wasn't running cargo test with the same flags that you're using in Travis.  Now I am.  the `--features sdl_controller` build is still going to fail, but I don't know what the error is.  Failure to link with SDL2.lib.

scorbeil:
It's all passing now.

PR #915, opened 2018-08-25
fix: Until the asset caching situation is cleared up, remove ids from…

Rhuagh:
… materials loaded from gltf

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/915)


jojolepro:
Can be merged once the conflixt is resolved

PR #916, opened 2018-08-25
feat: Add ability to control logging config via environment

Rhuagh:
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/916)

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -36,7 +36,20 @@ impl Default for LoggerConfig {
 ///
 /// If you do not intend on using the logger builtin to Amethyst, it's highly recommended you
 /// initialise your own.
-pub fn start_logger(config: LoggerConfig) {
+///
+/// Configuration of the logger can also be controlled via environment variables:
+///
+/// * AMETHYST_LOG_DISABLE_COLORS - if set, disables colors for the log output
+/// * AMETHYST_LOG_LEVEL_FILTER - sets the log level
+/// 
+pub fn start_logger(mut config: LoggerConfig) {
+    if let Ok(_) = env::var("AMETHYST_LOG_DISABLE_COLORS") {
+        config.use_colors = false;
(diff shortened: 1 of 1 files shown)


magnonellie:
Doesn't this defeat the point of using a config object at all? I mean, what's preventing you from doing this in user code instead with the `LoggerConfig` object? I'm not against taking environment variables into account, but I don't think that logic belongs in `default()`.

Rhuagh:
Because it's quite tedious to edit code or a config file to get more debug output. Just rerunning cargo or an executable with an env variable set is far faster

magnonellie:
Yeah I understand that, but I still do not think this belongs in the `default()` function. There surely has to be a better way to accomplish this.

Moxinilian:
@magnonellie What's wrong with having it be in the Default implementation?

magnonellie:
I've never seen people put anything other than straight up default values inside of a default implementation and I suspect others have similar experiences. `Default::default()` should, in my opinion, always be predictable and return the same data. Is it possible we could put this code inside of `start_logger()` instead where it would truly be an override for whatever has been requested in code?

Moxinilian:
While I don't mind having it in start_logger, I think it isn't inappropriate to have it in Default.
I think however that it better belongs in start_logger as there we can be sure the override is enforced.

Rhuagh:
That better ?

jojolepro:
Missing one review from anyone

jojolepro:
In case you didn't see my old review

>Approving if you add docs explaining this. Thanks

So it would be nice to have either a small section in the book or in a docs/ page explaining how its used.

Rhuagh:
Is doc on the start_logger function enough?

jojolepro:
bors r+

hell yeah :ok_hand: 

Thanks!

PR #920, opened 2018-08-26
Added ArcThreadPool, camera ortho, removal and timed destroy

jojolepro:
Good stuff :ok_hand:

Also renames ThreadPool to ArcThreadPool

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/920)


amethyst_utils/src/ortho_camera.rs
```diff
@@ -0,0 +1,103 @@
+use amethyst_core::Axis2;
+use amethyst_core::cgmath::Ortho;
+use amethyst_core::specs::{Component, DenseVecStorage, Join, System, ReadExpect, ReadStorage, WriteStorage};
+use amethyst_renderer::{Camera, ScreenDimensions};
+
+/// `Component` attached to the camera's entity that allows automatically adjusting the camera's matrix according to preferences in the "mode" field.
```
Moxinilian: Could you explain it in layman's terms too? Like the way I did for the isometric stuff.

amethyst_utils/src/ortho_camera.rs
```diff
@@ -0,0 +1,103 @@
...
+
+impl Component for NormalOrthoCamera {
+    type Storage = DenseVecStorage<Self>;
+}
+
+/// Settings that decide how to scale the camera's matrix when the aspect ratio changes.
+pub enum CameraNormalizeMode {
+    /// Using an aspect ratio of 1:1, tries to ajust the matrix values of the camera so
+    /// that the direction opposite to the stretch_direction is always [0,1].
+    /// Scene space can be lost on the specified stretch_direction.
```
Moxinilian: Could you please reformulate a bit? What does it mean to have a direction be always [0,1]?

amethyst_utils/src/ortho_camera.rs
```diff
@@ -0,0 +1,103 @@
...
+    Lossy {stretch_direction: Axis2},
+    
+    /// Scales the render dynamically to ensure no space is lost in the [0,1] range on any axis.
```
Moxinilian: What's that again?

amethyst_utils/src/ortho_camera.rs
```diff
@@ -0,0 +1,103 @@
...
+    Shrink,
+}
+
+impl CameraNormalizeMode {
+    /// Get the camera matrix offsets according to the specified options.
```
Moxinilian: What does that do?

amethyst_utils/src/ortho_camera.rs
```diff
@@ -0,0 +1,103 @@
...
+
+            for (mut camera, ortho_camera) in (&mut cameras, &ortho_cameras).join() {
+                let offsets = ortho_camera.camera_offsets(aspect);
+                camera.proj = Ortho {
+                    left: offsets.0,
+                    right: offsets.1,
+                    bottom: offsets.2,
+                    top: offsets.3,
+                    near: 0.1,
+                    far: 1000.0,
```
Moxinilian: I think the current default is 2000.0, any specific reason to have it be 1000.0 here?

amethyst_utils/src/removal.rs
```diff
@@ -0,0 +1,59 @@
+use amethyst_assets::{PrefabData, PrefabError};
+use amethyst_core::specs::{Component, DenseVecStorage, Entity, Join, ReadStorage, WriteStorage};
+use amethyst_core::specs::world::EntitiesRes;
+
+use std::result::Result;
+
+/// A marker `Component` used to remove entities and clean up your scene.
+/// The generic parameter `I` is the type of id you want to use.
+/// Generally a int or an enum.
```
Moxinilian: "an int". But that's a very good description here.

amethyst_utils/src/removal.rs
```diff
@@ -0,0 +1,59 @@
...
+pub struct Removal<I> {
+    id: I,
```
Moxinilian: Why not directly use the type as a marker?

Moxinilian: I guess it's fine right now because it will be better for scripts, so I'd keep it that way. But any reason?

amethyst_utils/src/removal.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+}
+
+/// Removes all entities that have the `Removal<I>` component with the specified removal_id.
+pub fn exec_removal<I: Send + Sync + PartialEq + 'static>(
+    entities: &EntitiesRes,
+    removal_storage: &ReadStorage<Removal<I>>,
+    removal_id: I,
+) {
+    for (e, r) in (&*entities, removal_storage).join() {
```
Rhuagh: for (e, r) in (&*entities, removal_storage).join().filter(|(_, r)| r.id == removal_id) {
}

amethyst_utils/src/removal.rs
```diff
@@ -0,0 +1,59 @@
...
+    type Storage = DenseVecStorage<Self>;
+}
+
+/// The prefab allowing to easily add a `Removal` `Component` to an entity.
+#[derive(Default, Clone, Deserialize, Serialize)]
+pub struct RemovalPrefab<I> {
+    id: I,
+}
+
+impl<'a, I: PartialEq + Clone + Send + Sync + 'static> PrefabData<'a> for RemovalPrefab<I> {
```
Rhuagh: Could we put the trait bounds in a where clause?

amethyst_utils/src/ortho_camera.rs
jojolepro: I was afraid it might cause issues with float precision in the matrix because of small values. Scaling of sprites are usually 0.005x or similar.

Moxinilian: I don't think float precision would be an issue here at only a scale of 2x. You'd better follow the engine convention I think.

jojolepro: not that dividing by 2 helps :joy:

Moxinilian: Yeah right?

amethyst_utils/src/removal.rs
jojolepro: Its made to be used like this:
entity1: Removal<Int>(0)
entity2: Removal<Int>(0)
entity3: Removal<Int>(1)

exec_removal(entities, storage, 0)

Its where the whole grouping idea I talked about started from.

amethyst_utils/src/ortho_camera.rs
jojolepro: changed to private. I'm not sure anyone would use this anyway.

amethyst_utils/src/ortho_camera.rs
```diff
@@ -0,0 +1,110 @@
...
+/// Settings that decide how to scale the camera's matrix when the aspect ratio changes.
+pub enum CameraNormalizeMode {
+    /// Using an aspect ratio of 1:1, tries to ajust the matrix values of the camera so
+    /// that the direction opposite to the stretch_direction always have a world size of 1.
+    /// If you use the default `Transform` (position = 0,0,0), this means that the direction opposite to stretch_direction 
+    /// will go from 0.0 to 1.0 in world coordinates.
+    /// Scene space can be lost on the specified stretch_direction.
+    Lossy {stretch_direction: Axis2},
+    
+    /// Scales the render dynamically to ensure no space is lost in the [0,1] range on any axis.
```
Moxinilian: Could you make it so that the actual [0;1] range be modifiable? Many workflow don't use [0,1].
An affine transformation should do the trick I think.

amethyst_utils/src/removal.rs
```diff
@@ -0,0 +1,60 @@
...
+        entity: Entity,
+        system_data: &mut Self::SystemData,
+        _entities: &[Entity],
+    ) -> Result<(), PrefabError> {
+        system_data.0.insert(entity, Removal::new(self.id.clone()))?;
+        Ok(())
+    }
+}
+
+/// Removes all entities that have the `Removal<I>` component with the specified removal_id.
```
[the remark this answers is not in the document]
jojolepro: When you delete an entity the associated components should be deleted

amethyst_utils/src/ortho_camera.rs
jojolepro: You just made me realize I forgot to include all the unit test I made for this

amethyst_utils/src/removal.rs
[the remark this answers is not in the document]
Rhuagh: They will be removed from the storage during world.maintain, which happens at the end of the frame.

Moxinilian:
What's the state on this my dude?

jojolepro:
I was in vacation. Going to finish it soon :tm:

jojolepro:
I'll try to finish this pr tomorow, and either work on my game project or finish the cleaning of the network pr

jojolepro:
>Could you make it so that the actual [0;1] range be modifiable? Many workflows don't use [0;1]. An affine transformation should do the trick I think.

It could be possible, but to my eyes the added complexity is not worth it. Its just a matter of preference at this point.


This PR is ready!

jojolepro:
By the way this PR also renames amethyst::core::ThreadPool to ArcThreadPool

azriel91:
Btw, do you think it's a good idea to run rustfmt, or wait until Rustfmt 1.0-rc1 is out (end of the week), and then run it over the whole repo?

jojolepro:
If its going to get release we should do that all at once. :)

jojolepro:
@rhuagh @moxinillian r? :)

jojolepro:
@Moxinilian it was only one "l" in your name lol.

amethyst_utils/src/ortho_camera.rs
```diff
@@ -0,0 +1,206 @@
...
+
+/// Settings that decide how to scale the camera's matrix when the aspect ratio changes.
+#[derive(Debug, Copy, Clone, Serialize, Deserialize)]
+pub enum CameraNormalizeMode {
+    /// Using an aspect ratio of 1:1, tries to adjust the matrix values of the camera so
+    /// that the direction opposite to the stretch_direction always have a world size of 1.
+    /// If you use the default `Transform` (position = 0, 0, 0),
+    /// this means that the direction opposite to stretch_direction 
+    /// will go from 0.0 to 1.0 in world coordinates.
+    /// Scene space can be lost on the specified stretch_direction.
```
magnonellie: I think these docs are somewhat confusing. How about linking to an image that visually explains how it works and maybe clearing up the text a bit?

amethyst_utils/src/ortho_camera.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// In other words, this ensures that you can always at least see everything 
+    /// between the world coordinates (0, 0) up to (1, 1).
+    /// If you have a non-default `Transform` on your camera,
+    /// it will just translate those coordinates by the translation of the `Transform`.
+    Shrink,
+}
+
+impl CameraNormalizeMode {
+    /// Get the camera matrix offsets according to the specified options.
+    fn camera_offsets(&self, aspect_ratio: f32) -> (f32,f32,f32,f32) {
```
magnonellie: I don't think these tuples are formatted correctly?

amethyst_utils/src/ortho_camera.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// will go from 0.0 to 1.0 in world coordinates.
+    /// Scene space can be lost on the specified stretch_direction.
+    Lossy {stretch_direction: Axis2},
+    
+    /// Scales the render dynamically to ensure no space is lost in the [0,1] range on any axis.
+    /// In other words, this ensures that you can always at least see everything 
+    /// between the world coordinates (0, 0) up to (1, 1).
+    /// If you have a non-default `Transform` on your camera,
+    /// it will just translate those coordinates by the translation of the `Transform`.
+    Shrink,
```
magnonellie: I think this one should probably be changed to `Contain` to match CSS. I'm not sure about `Lossy` however.

amethyst_utils/src/ortho_camera.rs
```diff
@@ -0,0 +1,206 @@
...
+
+            for (mut camera, ortho_camera) in (&mut cameras, &ortho_cameras).join() {
+                let offsets = ortho_camera.camera_offsets(aspect);
+                camera.proj = Ortho {
+                    left: offsets.0,
+                    right: offsets.1,
+                    bottom: offsets.2,
+                    top: offsets.3,
+                    near: 0.1,
+                    far: 2000.0,
```
magnonellie: These values should probably reuse from the previous values in case the user has something custom going on.

amethyst_utils/src/removal.rs
```diff
@@ -0,0 +1,73 @@
...
+use amethyst_core::specs::{Component, DenseVecStorage, Entity, Join, ReadStorage, WriteStorage};
+use amethyst_core::specs::world::EntitiesRes;
+use std::fmt::Debug;
+use std::result::Result;
+
+/// A marker `Component` used to remove entities and clean up your scene.
+/// The generic parameter `I` is the type of id you want to use.
+/// Generally an int or an enum.
+#[derive(Debug, Clone)]
+pub struct Removal<I>
```
magnonellie: I don't think this should be called `Removal` as you could use it for other purposes that need to identify all entities made for a specific scene. Maybe something like `SceneTag` instead?

amethyst_utils/src/time_destroy.rs
```diff
@@ -0,0 +1,56 @@
...
+impl<'a> System<'a> for TimedDestroySystem {
+    type SystemData = (
+        Entities<'a>,
+        ReadStorage<'a, DestroyAtTime>,
+        WriteStorage<'a, DestroyInTime>,
+        Read<'a, Time>,
+    );
+    fn run(&mut self, (entities, dat, mut dit, time): Self::SystemData) {
+        for (e, d) in (&*entities, &dat).join() {
+            if time.absolute_time_seconds() > d.time {
```
magnonellie: Is this optimized automatically by the compiler or will it run the function each iteration? (same for the other loop below)

amethyst_utils/src/scene.rs
```diff
@@ -23,22 +25,25 @@ use serde::Serialize;
 /// - `M`: `Format` to use for loading `Mesh`es from file
 #[derive(Deserialize, Serialize)]
 #[serde(default)]
-pub struct BasicScenePrefab<V, M = ObjFormat>
+pub struct BasicScenePrefab<V, R, M = ObjFormat>
```
magnonellie: Can't this also have a default to avoid changing all the examples like you have done?

amethyst_utils/src/removal.rs
jojolepro: Its better to use other types to identify entities. If you re-use this, all the identifiers will be in a giant list. If you use different types for identifications, it makes smaller list because they are divided into Storage<T> instead of a single (Storage<Removal<i32>> -> runtime match what you need)

amethyst_utils/src/time_destroy.rs
jojolepro: I would assume it is optimized since its not mutable.

Moxinilian: Yes, and the operation is basically free anyway.

amethyst_utils/src/scene.rs
jojolepro: No because that's really up to the user to use it. The examples don't make use of it so that's why I changed them.

magnonellie: I don't quite understand your logic right now... What does it hurt to just make it `R = ()` so that current code doesn't break and you don't have to bother adding the `()` if you don't care about it?

jojolepro: People might use just BasicScenePrefab<V>, not realize that the Removal component is a thing and skip on that feature. Tbh I use Removal in 100% of the prefabs I have made up to date.

magnonellie: They would miss it anyway. They copy from the examples and just ignore that `()` and then they'll find out later once they do need it. It's essentially the same thing, just with more initial confusion if it always has to be there. Besides, not everyone will actually use that feature but might do it some other way.

jojolepro: okay will change

amethyst_utils/src/ortho_camera.rs
jojolepro: I'm not smart enough to resolve the formula to get near and far back from the matrix, so I just skipped this one and left a comment :P

PR #923, opened 2018-08-28
Fix depth for sprites

msiglreith:
Followup to #902 
Didn't take the depth sprite into account.

cc @azriel91


azriel91:
Yay it works! Thank you very much 🙇

PR #925, opened 2018-08-29
fix: Disable backface culling for the sprite pass

### LINKED ISSUE #909: Transform.scale does not handle negative values in DrawSprite
If I take an entity with a `Transform` component and set `transform.scale.x` to a negative value, the sprite should be flipped vertically. Instead, the sprite disappears.

gif

Rhuagh:
Might fix (probably) #909.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/925)


Rhuagh:
Don't have the reporters code unfortunately.

PR #926, opened 2018-09-01
Fix the application_root_dir function

Moxinilian:
Fixes a breaking bug in the `application_root_dir` function.


Moxinilian:
Come on dude

PR #929, opened 2018-09-03
Fixed regression in sprite positioning.

azriel91:
cc: @msiglreith

The recent sprite batching change included regressions that we didn't notice before:

* sprite render position was shifted left by sprite_w, and down by sprite_h
* offsets were added instead of subtracted

This PR reverts the behaviour to what it was previously.
No changelog entry since it's a bug that's fixed before the feature was released.

Issue #829, #894

### DIFF
--- a/amethyst_renderer/src/pass/sprite/interleaved.rs
+++ b/amethyst_renderer/src/pass/sprite/interleaved.rs
@@ -281,8 +281,12 @@ impl SpriteBatch {
 
             let dir_x = transform.x * sprite_data.width;
             let dir_y = transform.y * sprite_data.height;
-            let pos =
-                transform * Vector4::new(sprite_data.offsets[0], sprite_data.offsets[1], 0.0, 1.0);
+            let pos = transform * Vector4::new(
+                -sprite_data.offsets[0],
+                -sprite_data.offsets[1],
+                0.0,
+                1.0
+            );
 
             instance_data.extend(&[
                 dir_x.x, dir_x.y, dir_y.x, dir_y.y, pos.x, pos.y, uv_left, uv_right, uv_bottom,
--- a/amethyst_renderer/src/pass/sprite/interleaved.rs
+++ b/amethyst_renderer/src/pass/sprite/interleaved.rs
@@ -281,6 +281,12 @@ impl SpriteBatch {
 
             let dir_x = transform.x * sprite_data.width;
             let dir_y = transform.y * sprite_data.height;
+
+            // The offsets are negated to shift the sprite left and down relative to the entity, in
+            // regards to pivot points. This is the convention adopted in:
+            //
+            // * libgdx: <https://gamedev.stackexchange.com/q/22553>
+            // * godot: <https://godotengine.org/qa/9784>
             let pos = transform * Vector4::new(
                 -sprite_data.offsets[0],
                 -sprite_data.offsets[1],
--- a/amethyst_renderer/src/pass/shaders/vertex/sprite.glsl
+++ b/amethyst_renderer/src/pass/shaders/vertex/sprite.glsl
@@ -38,7 +38,7 @@ void main() {
     float tex_u = positions[gl_VertexID][0];
     float tex_v = positions[gl_VertexID][1];
 
-    vec2 uv = pos + (tex_u - 1.0f) * dir_x + (tex_v - 1.0f) * dir_y;
+    vec2 uv = pos + tex_u * dir_x + tex_v * dir_y;
     tex_uv = texture_coords(vec2(tex_u, tex_v), u_offset, v_offset);
 
     vec4 vertex = vec4(uv, depth, 1.0);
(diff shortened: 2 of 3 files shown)


amethyst_renderer/src/pass/sprite/interleaved.rs
```diff
@@ -281,8 +281,12 @@ impl SpriteBatch {
 
             let dir_x = transform.x * sprite_data.width;
             let dir_y = transform.y * sprite_data.height;
-            let pos =
-                transform * Vector4::new(sprite_data.offsets[0], sprite_data.offsets[1], 0.0, 1.0);
+            let pos = transform * Vector4::new(
+                -sprite_data.offsets[0],
```
jojolepro: Coordinates are still starting at the bottom left and going towards the top right?

msiglreith: negating doesn't really make sense to me

azriel91: Yeap they're still bottom left to top right.

The negation means shift the sprite left and down, I don't mind not negating it and leaving that to the user; I don't know if there's a common convention.

On my phone now so I'll change it when I get to my computer

azriel91: Pushed! Also updated docs, pong examples + tutorial, and sprites_ordered example

msiglreith: oohh.. that's supposed to be the pivot point heh

azriel91: Yeaps! does it change your preference? (it's easy to remove, can just pop the last commit)

msiglreith: I would prefer either rename to `pivot` and keep the old one, or go with the current change.
But probably on others to decide,

azriel91: I'll keep it current unless anyone else *really* wants it negated, it makes sense in my head now :stuck_out_tongue:

jojolepro: I think I didn't understand what this is about. What was the negation actually doing to the pivot? I'm expecting it to default in the center of the sprite and when you offset it the positive coordinates go towards the top right.

azriel91: Hoping I understand what you're saying, so I'll try clarify before and after:

* Previously (before sprite batching), `offsets` meant "how many pixels to shift the sprite left and down", so:

    - When offsets were `0, 0`, the sprite is drawn like:

        ```text
          |ABCD| // ABCD indicates the top of the sprite
          |    |
          |    |
          x----| // x marks the entity position
        ```

    - When offsets were `1, 2`, the sprite is drawn like:

        ```text
         |ABCD|
         |x   | <- the sprite is shifted 2 pixels down relative to the entity
         |    |
         |----|

          ^ and one pixel to the left
        ```

* After sprite batching, instead of specifying `1, 2`, you have to specify `-1, -2` to get the same effect. I did revert this change when I opened the PR, but then was convinced that it makes sense to not negate the offsets, so undid the revert (i.e. redid it to mean you should specify `-1, -2` when you want to shift it left and down.)

There wasn't a default for "use `sprite_w / 2.0` and `sprite_h / 2.0`" when creating a `Sprite` structure directly, but the `with_sprite` function used to do this when sprites were still drawn using a `Mesh` component (before the `DrawSprite` pass was implemented).

azriel91:
Re: offsets

Things that don't negate (user specifies negated values):
* [css convention](https://stackoverflow.com/questions/1359639/how-to-find-the-proper-offsets-for-css-sprites)

Things that negate it (user specifies positive values):
* [libgdx](https://stackoverflow.com/questions/25838226/libgdx-sprite-rotation-causing-position-offset)
* [gamedev stack exchange](https://gamedev.stackexchange.com/questions/22553/why-is-my-sprite-displayed-offset-from-its-box2d-body)
* [windows phone game dev book](https://books.google.co.nz/books?id=jjJ1tH1k4uEC&pg=PA428&lpg=PA428&dq=sprite+offset+negate&source=bl&ots=kABrbbL7aL&sig=oEMWcjSkTHlMX6XU-TuCMWuuu0o&hl=en&sa=X&ved=2ahUKEwjU1t6j6qTdAhWPzmEKHXHAAzIQ6AEwBHoECAQQAQ#v=onepage&q=sprite%20offset%20negate&f=false)

Since it's more common in game dev world for the user to not negate the values, I'll remove the last commit.

amethyst_renderer/src/pass/sprite/interleaved.rs
jojolepro: Please add a comment explaining it here. Thanks

jojolepro:
You also need one more review

jojolepro:
bors r+
Thanks!

PR #933, opened 2018-09-04
fix: Joint entities could only be part of a single skin.

### LINKED ISSUE #802: glTF loader swapping materials and transform when loading multiple objects
It seems that glTF loader is randomly reordering materials and transforms of objects that are loaded at the same time. The provided repo contains three objects in two files:

- question.gtfs
  - a blue question mark
- cylinder_stage.gltf
  - a red cylinder
  - a green "floor" plane

When loading the scene from prefab, the cylinder is pretty much always coming up as blue. Often it's loaded way too big. The floor is often misplaced or missing. The behaviour seems random - every execution can yield different results.

Repository with reproduction: https://github.com/Frizi/random_towers/tree/gltf-bug-repro

The reproduction branch contains precompiled glTFs, it should be ready to just `cargo run`. The models are exported using official glTF blender exporter. The original source `.blend` files are also provided in the repo.

Rhuagh:
Fixes #802.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/933)


amethyst_animation/src/skinning/systems.rs
```diff
@@ -39,33 +34,34 @@ impl<'a> System<'a> for VertexSkinningSystem {
...
                 skin.joints
                     .iter()
-                    .map(|joint_entity| {
-                        (
-                            joints.get(*joint_entity).unwrap(),
-                            transforms.get(*joint_entity).unwrap(),
-                        )
+                    .zip(skin.inverse_bind_matrices.iter())
+                    .map(|(joint_entity, inverse_bind_matrix)| {
+                        (transforms.get(*joint_entity).unwrap(), inverse_bind_matrix)
```
jojolepro: Optionnal: add a comment explaining how this unwrap is almost always safe.
Also to prevent denial of services where a server could crash a client, i think the almost always safe wont be enough and we will need proper handling.

Moxinilian: I think server security will require a full audit of the engine anyway, so while I think the comment is important, we don't have to handle all attack vectors yet.

jojolepro:
bors r-
wtf bors

PR #934, opened 2018-09-05
TGA image support

FaultyRAM:
This adds support in amethyst_renderer for TGA images. It also fixes an implementation detail where we were calling `image::load_from_memory`, which guesses the image format, even though we always know exactly what format we want. (In fact, `image::load_from_memory` doesn't work on TGA files, so that's two birds with one stone.)

### DIFF
--- a/amethyst_renderer/src/formats/texture.rs
+++ b/amethyst_renderer/src/formats/texture.rs
@@ -1,5 +1,4 @@
-use image::DynamicImage;
-use image::RgbaImage;
+use image::{DynamicImage, ImageFormat, RgbaImage};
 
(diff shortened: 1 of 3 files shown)


jojolepro:
You also need to create the link to this PR. Just under where you added the line there is some examples.
[#934]: https://github.com/amethyst/amethyst/pull/934

Aceeri:
bors r+
Not sure why the previous merge failed (was the old travis rebased over or something?)

FaultyRAM:
@Aceeri there was a merge conflict, I had to rebase onto develop to fix it.

PR #937, opened 2018-09-07
Improve ergonomics of the Named component

magnonellie:
Adds the `Named` component to `prelude` and as a default registered component.

I'm unsure about the "registered by default" part but it's honestly the best I can think of right now. The reasoning behind it is that a lot of the time you'll use the named component, maybe even without thinking of it, without having a system that actually consumes it. That's great for forward thinking, but then you also have to manually register `Named` when building the application and remember to remove that once you do add a system which requires a `Named` storage.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -5,6 +5,7 @@ use assets::{Loader, Source};
 use core::frame_limiter::{FrameLimiter, FrameRateLimitConfig, FrameRateLimitStrategy};
 use core::shrev::{EventChannel, ReaderId};
 use core::timing::{Stopwatch, Time};
+use core::Named;
 use ecs::common::Errors;
 use ecs::prelude::{Component, World};
 use error::{Error, Result};
@@ -440,6 +441,8 @@ impl<S, E: Send + Sync + 'static> ApplicationBuilder<S, E> {
         world.add_resource(FrameLimiter::default());
         world.add_resource(Stopwatch::default());
         world.add_resource(Time::default());
+
+        world.register::<Named>();
 
         Ok(ApplicationBuilder {
             initial_state,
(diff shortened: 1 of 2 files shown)


src/app.rs
```diff
@@ -441,6 +442,8 @@ impl<S, E: Send + Sync + 'static> ApplicationBuilder<S, E> {
         world.add_resource(Stopwatch::default());
         world.add_resource(Time::default());
 
+        world.register::<Named>();
```
Rhuagh: Why do we need this? This is only needed if you never intend to use the component in a system, and only from the states. If we're serious about doing this for Named, we should do it for a whole bunch of other components tbh.

magnonellie: I explained most of my thinking in the initial comment. Basically, it's because a user may name an entity without actually using the named component in a system which means the user has to register the named component by themselves until they do decide to add a system which does use it. What other components are you thinking should also be handled this way?

omni-viral:
How can you use `Named` component without systems? Create entities with names via `World`?

magnonellie:
@omni-viral Pretty much, yes. The `WithNamed` trait allows you to do `world.create_entity().named("A name")` which I at least would like to be able to do without having to immediately add a system that uses `Named`.

omni-viral:
That's make sense unless you can make the code work without `Named` storage existing. In which case provided name would be stored in /dev/null

Moxinilian:
I think it makes sense to register the storage by default because names might be useful in a context outside of systems. Considering they have a dedicated method on the entity builder, I'd like it to work consistently.

jojolepro:
Missing a single review (or refusal lol)

src/app.rs
Rhuagh: I'm fine with it being added.

Issue #908, opened 2018-08-24
Remove unwraps


jojolepro:
Either remove all unwraps from the engine and replace them with proper error handling, or add a comment explaining how an unwrap will never panic.

usbalbin:
Hi!

Amethyst looks like a very exciting project. I would like to give this issue a try, if that is ok.

I am fairly new to rust. Been playing around with some smaller toy projects. However, I have pretty much zero experience with contributing to open source projects.

Is there anything in particular to think about regarding this issue except follow you contribution guidelines?

/Albin

Moxinilian:
Welcome!
Error handling in the Rust ecosystem is a little bit tricky because the features of the standard library regarding it are a bit limited. This is why it is rated medium.
Have you ever dealt with advanced error handling in Rust?

anderejd:
I would like to point you in the direction of this <https://doc.rust-lang.org/std/result/enum.Result.html> to begin with. One of the most important pieces of error handling with Result is the `map_err` method, which is surprisingly not mentioned once in the error handling chapter of the Rust book. Custom error enums in combination with ´Result::map_err´ and the question mark operator will provide the tools you need to do error propagation in most cases. Hope that helps to get you started 🙂

jojolepro:
Welcome!

usbalbin:
Thank you for your answers! No I have not dealt with any sort of non-trivial error handling before. Might not be the best issue for me to start with?

So for something like StateMachine::start

```rust
pub fn start(&mut self, data: StateData<T>) {
    if !self.running {
        let state = self.state_stack.last_mut().unwrap();
        state.on_start(data);
        self.running = true;
    }
}
```
would turn in to something like this
```rust
enum StateError { NoStatesPresent }

pub fn start(&mut self, data: StateData<T>) -> Result<(), StateError> {
    if !self.running {
        let state = self
            .state_stack
            .last_mut()
            .ok_or(StateError::NoStatesPresent)?;
        state.on_start(data);
        self.running = true;
    }
    Ok(())
}
```
? Or am I supposed to use amethyst::Result? Sorry if this is obvious.

Am I on the right path or should I should I start with something easier?
Please be honest :)

Moxinilian:
If you've got specific questions on the internals of the engine, feel free to contact us on the #development channel on the Discord server.

usbalbin:
Hi again!

[...]

PR #940, opened 2018-09-11
Remove unwraps in StateMachine

usbalbin:
This PR is focused on #908 and replaces some unwraps from StateMachine with some error propogation.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -187,8 +187,8 @@ impl<'a, T, E: Send + Sync + Clone + 'static> Application<'a, T, E> {
     ///
     /// See the example supplied in the
     /// [`new`](struct.Application.html#examples) method.
-    pub fn run(&mut self) {
-        self.initialize();
+    pub fn run(&mut self) -> Result<()> {
+        self.initialize()?;
         self.world.write_resource::<Stopwatch>().start();
         while self.states.is_running() {
             self.advance_frame();
@@ -206,14 +206,16 @@ impl<'a, T, E: Send + Sync + Clone + 'static> Application<'a, T, E> {
         }
 
         self.shutdown();
+        Ok(())
     }
 
     /// Sets up the application.
-    fn initialize(&mut self) {
+    fn initialize(&mut self) -> Result<()> {
         #[cfg(feature = "profiler")]
         profile_scope!("initialize");
         self.states
-            .start(StateData::new(&mut self.world, &mut self.data));
+            .start(StateData::new(&mut self.world, &mut self.data))
+            .map_err(|e| Error::StateMachine(e))
     }
 
     /// Advances the game world by one tick.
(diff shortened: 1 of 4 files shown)


src/app.rs
```diff
@@ -213,7 +213,8 @@ impl<'a, T, E: Send + Sync + Clone + 'static> Application<'a, T, E> {
         #[cfg(feature = "profiler")]
         profile_scope!("initialize");
         self.states
-            .start(StateData::new(&mut self.world, &mut self.data));
+            .start(StateData::new(&mut self.world, &mut self.data))
+            .unwrap(); //TODO: figure out whether this needs to be handled
```
usbalbin: `StateMachine::start` errors when there are no states available. If I understand the situation this only happens when it is run after it has previously been stopped by `stop` or `pop` until empty. Do we care about those situations here? Will a game ever be run multiple times?

LucioFranco: I would wrap this with a message that notifies the user that they initialized without states, I would use `expect_err` in this case.

usbalbin: Wouldn't that be just plain `expect(“some message”)`? Or did I misunderstand?
```rust
self.states
    .start(start(StateData::new(&mut self.world, &mut self.data))
    .expect_err(“Tried to start state machine without any states present”);
```
Panics every time except when `start` returns an `Err`

LucioFranco: Yup, you are right, somehow slept read that doc 😄 . Yeah, so it should be just `expect`.

src/state.rs
```diff
@@ -2,8 +2,26 @@
 
 use amethyst_input::is_close_requested;
 use ecs::prelude::World;
+use std::fmt::Result as FmtResult;
+use std::fmt::{Display, Formatter};
 use {GameData, StateEvent};
 
+#[derive(Debug)]
+pub enum StateError {
```
jojolepro: Missing documentation explaining what it is. (its obvious, but we prefer to have docs as much as possible)

src/state.rs
```diff
@@ -272,12 +290,16 @@ impl<'a, T, E: Send + Sync + 'static> StateMachine<'a, T, E> {
     ///
     /// # Panics
```
jojolepro: You forgot to update the documentation. It doesn't panic anymore but returns the StateError

usbalbin:
I hope I put my entry under the right section in the changelog

src/state.rs
```diff
@@ -342,6 +362,8 @@ impl<'a, T, E: Send + Sync + 'static> StateMachine<'a, T, E> {
             }
 
             self.state_stack.push(state);
+
+            //State was just pushed, thus pop will always succeed
```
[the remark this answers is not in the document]
usbalbin: Oh okay, I'll try to remember to next time:)

LucioFranco:
This looks good to go! Thank you for the contribution, look forward to seeing some more 😉 😄 .

git merge

bors r+

PR #941, opened 2018-09-13
Feature: Debug lines

### LINKED ISSUE #917: Debug lines
Just would be nice to have a quick system for adding debugging lines in 2d/3d.

khskarl:
Adds support for debug lines, closes #917. 

There is a new example available demonstrating the `DrawDebugLines` pass and the `DebugLines` component and resource.
Debug lines can be used as a resource for quick dirty debugging, but it's better suited as a component for "long term" debugging in a system, mainly if performance due synchronization becomes an issue.

Also requesting any suggestions on the API and for any more intuitive names for `add_as_line`, `add_as_direction`. 

Things I left out:
- Transparency, I couldn't get it working and I don't know why 
- Configurable line width, all lines have the same constant width defined inside the geometry shader. It would be ideal to be able to set a size for each line, but that would require an extra vertex attribute and a weird-ish vertex format exclusive for lines, so I thought it's best to leave this for after I get at least a second opinion.

I decided to submit as is so people can use lines as they are and leave these things to fix later in separate issues.

Looking cautiously forward for reviews. :shipit:


amethyst_renderer/src/pass/debug_lines/interleaved.rs
```diff
@@ -0,0 +1,138 @@
...
+            None => builder.with_output("color", Some(DepthMode::LessEqualWrite)),
+        };
+        builder.build()
+    }
+
+    fn apply<'a, 'b: 'a>(
+        &'a mut self,
+        encoder: &mut Encoder,
+        effect: &mut Effect,
+        mut _factory: Factory,
```
Rhuagh: Rename to factory

amethyst_renderer/src/pass/shaders/geometry/debug_lines.glsl
```diff
@@ -0,0 +1,62 @@
...
+    vec4 color;
+    vec3 normal;
+} vertex;
+
+layout (points) in;
+layout (triangle_strip, max_vertices = 4) out;
+
+uniform vec3 camera_position;
+
+const float WIDTH = 2.0 / 800.0;
```
Rhuagh: Where's the 800 from ?

amethyst_renderer/src/debug_drawing.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /// Builder method to pre-allocate a number of line.
+    pub fn with_capacity(mut self, capacity: usize) -> Self {
+        self.lines = Vec::<DebugLine>::with_capacity(capacity);
+        self
+    }
+
+    /// Adds a line to be rendered by giving a position and a direction.
+    pub fn add_as_direction(&mut self, position: [f32; 3], direction: [f32; 3], color: Rgba) {
```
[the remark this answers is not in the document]
khskarl: I agree, it's a leftover as initially I wasn't doing any operations and was using a format ready for the pass. I will change this soon.

amethyst_renderer/src/pass/debug_lines/interleaved.rs
khskarl: Nice catch

amethyst_renderer/src/pass/shaders/geometry/debug_lines.glsl
khskarl: It was meant to be `1.0 / 400.0`, as this way is easy to set the line width relative to the unit value `1.0`. As in `the line width is one line width unit, which is one four-hundredth the unit size`. This also makes it easier to use intuitive values for the line width instead of small values like 0.0001, 0.015, 0.003... As simple using line width 1.0 will result in in way too thick lines. 

The value `2.0 / 800.0` was chosen arbitrarily as it didn't look too narrow nor too thick.
I will add a comment right above to make it clearer.

Rhuagh: Thanks

amethyst_renderer/src/pass/debug_lines/interleaved.rs
```diff
@@ -0,0 +1,138 @@
...
+impl<V> DrawDebugLines<V>
+where
+    V: Query<(Position, Color, Normal)>,
+{
+    /// Create instance of `DrawDebugLines` pass
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    /// Enable transparency { Currently not supported, but should be useful }
```
jojolepro: The method should not be there or at least should have the unimplemented! Macro in it if it doesn't work

amethyst_renderer/src/pass/shaders/geometry/debug_lines.glsl
```diff
@@ -0,0 +1,62 @@
...
+    pos = vertex_in[id].position + vertex_in[id].normal + width_vector;
+    gl_Position = proj * view * vec4(pos, 1.0);
+    EmitVertex();
+    EndPrimitive();
+}
+
+void main()
+{
+    EmitLine(0);
+}
```
jojolepro: Missing a new line at the end

amethyst_renderer/src/pass/debug_lines/interleaved.rs
khskarl: Well, I didn't think that way, fair enough! I'm adding it.

amethyst_renderer/src/debug_drawing.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /// Builder method to pre-allocate a number of line.
+    pub fn with_capacity(mut self, capacity: usize) -> Self {
+        self.lines = Vec::<DebugLine>::with_capacity(capacity);
+        self
+    }
+
+    /// Adds a line to be rendered by giving a position and a direction.
+    pub fn add_as_direction(
```
Moxinilian: As far as I understood, this needs to be called on every frame for a line to be rendered.
Wouldn't something like `render_as_direction` make it clearer?

khskarl: You understood right. 
True, from a user perspective `add` may sound like there is a `remove` equivalent and something should be removed.

What about `draw_direction`? In my opinion `draw`  is shorter and more intuitive. 

I'm hesitant about `draw_as_direction` because it sounds like `direction` and `line` are going to be rendered in different ways, when the only thing that changes is that one uses a direction and the other an explicit end position for the line.

jojolepro:
Add_as_line -> add_from_points ?

amethyst_renderer/src/debug_drawing.rs
Moxinilian: That seems ideal to me.

jojolepro:
Also I didn't realize the data was cleared each frame. I think it doesn't make much sense to have DrawLines as a Component in this case, since it only consist of temporary data.

Moxinilian:
Would it be good to have the data in the components permanent while the data in the resources would be volatile? It makes sense but at the same time creates inconsistencies.

jojolepro:
That would actually be pretty good. I would like it if it was two different structs to clearly show the difference.

khskarl:
Originally the data in the components wasn't volatile, but it became cumbersome using the component with dynamic lines, e.g. a character's direction, velocity. As there is no way simple way to update one specific line, e.g. My character entity has a DebugLines component rendering which is rendering the characters up direction and it's velocity, but I only want to update the velocity line.

It works well for long term static lines such as the grid and the axis in the example, but I was tired when I did this change and didn't realize this. Plus in my mind at the time I was focused on the disadvantage of using the debug lines resource being only syncing write access to the resource.

Moxinilian:
I actually now get your point. But then, everything you do with components can be done with the resource. I think you should either make the components not volatile, or simply remove the components interface entirely.

Moxinilian:
I personally prefer the separation of two different names.

amethyst_renderer/src/pass/debug_lines/interleaved.rs
```diff
@@ -0,0 +1,138 @@
...
+where
+    V: Query<(Position, Color, Normal)>,
+{
+    /// Create instance of `DrawDebugLines` pass
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    /// Enable transparency
+    pub fn with_transparency(
```
Aceeri: Should we even have this method if it is unimplemented? If so we might want a comment in the docs about it.

amethyst_renderer/src/pass/debug_lines/interleaved.rs
```diff
@@ -0,0 +1,138 @@
...
+){
+        trace!("Drawing debug lines pass");
+        let debug_lines = {
+            let mut lines = Vec::<DebugLine>::new();
+
+            for debug_lines_component in (&lines_components).join() {
+                lines.extend(&debug_lines_component.lines);
+            }
+
+            match lines_resource {
```
Aceeri:
```rust
if let Some(mut lines_resources) = lines_resource {
    ...
}
```
Is probably more clear here.

amethyst_renderer/src/pass/shaders/geometry/debug_lines.glsl
```diff
@@ -0,0 +1,63 @@
...
+    EmitVertex();
+
+    vertex.color = vertex_in[id].color;
+    pos = vertex_in[id].position + vertex_in[id].normal + width_vector;
+    gl_Position = proj * view * vec4(pos, 1.0);
+    EmitVertex();
+    EndPrimitive();
+}
+
+void main()
```
Aceeri: Should probably put the opening brace on the same line for consistencies sake.

LucioFranco:
@khskarl closing this PR as we will be moving to master branch only scheme, please if you want to continue this awesome work, just reopen this PR versus master and link to this one. Thanks.

amethyst_renderer/src/pass/debug_lines/interleaved.rs
khskarl: Probably it's better removed, so far I haven't been able to make it work at all and it's not a lot of code.

PR #944, opened 2018-09-14
Bugfix/943/hidpi scaling

### LINKED ISSUE #943: HiDPI factor from winit update missed scaling UI
When Winit was updated to `0.17`, the hidpi factor wasn't taken into account to scale UI components.

Symptoms:

* UI button in the middle of the screen is rendered further up and left, but clickable area is still in the middle.

This issue tracks the fix.

azriel91:
Fixes #943.

Effect of fix: https://imgur.com/P2kmlPo.mp4

Ignore the sprites bit at the end, I thought there was a scaling issue with it but there wasn't, so I removed the commit that changed it.

### DIFF
--- a/amethyst_ui/src/pass.rs
+++ b/amethyst_ui/src/pass.rs
@@ -248,10 +248,12 @@ impl Pass for DrawUi {
                     proj_vec: proj_vec.into(),
                     // Coordinates are middle centered. It makes it easier to do layouting in most cases.
                     coord: [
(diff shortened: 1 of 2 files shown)


azriel91:
oh yes something seems to be broken, shall try to fix it today.

azriel91:
To be reopened on `master` when `develop` goes away, and I actually figure out the fix. Most of the code shouldn't have to change, it's the text position that is wrong, not the background sections etcetera, so we can't shift "just everything".

Issue #924, opened 2018-08-29
Move back to only master branch


LucioFranco:
Originally we had decided to use a develop branch and a master branch due to high code churn. Now that the engine is more stable I suggest we move back to a master only situation, where `staging` and `trying` are reserved for @bors[bot]. This would mean stable and released code would be found under  the release tag in github releases which will correspond with crates.io releases.

Things that need to change to support this:
- [x] Merge develop into master, remove develop and set master as the main branch for the repo.
  - [x] Verify that travis config works without a develop branch
  - [x] Verify that appveyor config works without develop branch (even though we don't use it at the moment we still want to keep the config consistent)
  - [x] Verify bors works without a develop branch
  - [x] Make sure all new PR's are open vs master and all previous PR's get merged into develop before it gets deleted.
- [x] Remove references to the `develop` branch in the main repo
  - [x] Book
  - [x] Readme
  - [x] Contributing guideline
- [x] Remove referenes on the amethyst/website repo
  - [x] Change the docs page

## Questions

~Should we remove `release-x.x` branches and just leave releases to git tags?~

cc @amethyst/engine-devs

Xaeroxe:
> Should we remove release-x.x branches and just leave releases to git tags?

No, because these are very important for patch releases.  Maintaining the ability to diverge code for old releases is very important.

LucioFranco:
@Xaeroxe great point, I like the ability to apply *hotfixes* in a sense. 👍

OvermindDL1:
> No, because these are very important for patch releases. Maintaining the ability to diverge code for old releases is essential for fixing high priority bugs in older versions.

I entirely agree here.  Although some projects use just tags to mark releases, that means applying patches to them is more difficult, where always generating a release (and tagging it too) allows for future patches far more easily and consistently so I always vote for release branches.

Rhuagh:
@Xaeroxe I would suggest only creating those branches on need, i.e. when a patch release needs to be done. Branching from a tag isn't that hard.

LucioFranco:
@Rhuagh right, but then we have branches that exist just for minor fixes instead of branches that exist for that release and any minor changes that need to come after its release?

Xaeroxe:
@Rhuagh Yeah that's a good point.  I suppose thanks to the power of git thinking ahead that much isn't necessary and might just be creating extra work for us.

Most releases historically haven't needed patch releases, so perhaps it's better to start with the assumption they won't be necessary.

[...]

PR #948, opened 2018-09-16
Remove references to the develop branch

LucioFranco:
This change is related to RFC #924, this change is in preparation for removing the develop branch.

This should solve most of the main repo issues that deal with removing develop. A few second set of eyes would be helpful to make sure evetything is correct.


LucioFranco:
I've updated the commit to include the `develop` branch in travis so we can continue to test incoming branches as we need those to be closed or reopened before we can move to master.

jojolepro:
Wait til everything is ready to not break the CI for weeks :P

PR #949, opened 2018-09-17
Derive Serialize/Deserialize for Named

randomPoison:
This makes `Named` consistent with other components that contain simple, serializable data. I specifically need `Named` to be serializable for use in the editor. It seems like we're not bothering to make serde support optional, so seems fine to just derive the impls directly?


jojolepro:
bors r+

not 24 hours but I think we'll survive ;)

PR #950, opened 2018-09-19
add JsonFormat to formats

robinvd:
This does add serde_json to the dependencies, but it can be very useful for (external) config files in json.

### DIFF
--- a/amethyst_assets/src/formats.rs
+++ b/amethyst_assets/src/formats.rs
@@ -23,3 +23,25 @@ where
         Ok(val)
     }
 }
+
+/// Format for loading from Json files.
+#[derive(Default, Clone, Debug)]
+pub struct JsonFormat;
+
+impl<T> SimpleFormat<T> for JsonFormat
+where
+    T: Asset,
+    T::Data: for<'a> Deserialize<'a> + Send + Sync + 'static,
+{
+    const NAME: &'static str = "Json";
+    type Options = ();
+
+    fn import(&self, bytes: Vec<u8>, _: ()) -> Result<T::Data, Error> {
+        use serde_json::de::Deserializer;
+        let mut d = Deserializer::from_slice(&bytes);
+        let val = T::Data::deserialize(&mut d).chain_err(|| "Failed parsing Json file")?;
+        d.end().chain_err(|| "Failed parsing Json file")?;
+
+        Ok(val)
+    }
+}
--- a/amethyst_assets/src/formats.rs
+++ b/amethyst_assets/src/formats.rs
@@ -25,9 +25,11 @@ where
 }
 
 /// Format for loading from Json files.
+#[cfg(feature="json")]
 #[derive(Default, Clone, Debug)]
 pub struct JsonFormat;
 
+#[cfg(feature="json")]
 impl<T> SimpleFormat<T> for JsonFormat
 where
     T: Asset,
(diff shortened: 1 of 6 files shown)


docs/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 
 ## Unreleased
 ### Added
+* Added JsonFormat.
```
LucioFranco: Please link the PR in the change log entry.

amethyst_assets/src/formats.rs
```diff
@@ -23,3 +23,25 @@ where
...
+    T: Asset,
+    T::Data: for<'a> Deserialize<'a> + Send + Sync + 'static,
+{
+    const NAME: &'static str = "Json";
+    type Options = ();
+
+    fn import(&self, bytes: Vec<u8>, _: ()) -> Result<T::Data, Error> {
+        use serde_json::de::Deserializer;
+        let mut d = Deserializer::from_slice(&bytes);
+        let val = T::Data::deserialize(&mut d).chain_err(|| "Failed parsing Json file")?;
```
LucioFranco: Similar to how its done above for Ron, I would say `Failed deserializing JSON file` so that there are not two places that the same message can be produced.

magnonellie:
I really don't mean to sound rude or dismissive here, but I don't feel that this belongs in Amethyst itself since it can very easily be added through a crate or something similar. Amethyst has already chosen to go all in on Ron and as such, I think it makes sense to push that unless something else is absolutely required. If you have a more concrete use case in mind however, please tell me and I may change my mind. Again, this is nothing personal, so please don't take it that way :smile_cat:

LucioFranco:
@magnonellie Agreed, but since the implementation is very small, I don't see the issue, unless this impacts maintainability down the line?

magnonellie:
@LucioFranco Yeah, it's not gonna be hard to maintain, I agree with that. I'm thinking more in terms of keeping a consistent message to users that we encourage the use of Ron. :wink:

docs/CHANGELOG.md
robinvd: Is there a way to see the github issue number beforehand? Fixed it

amethyst_assets/src/formats.rs
robinvd: I agree with you that that would be usefull, but Ron also has the same error message in these positions: "Failed deserializing Ron file". Changed it

robinvd:
@LucioFranco it is now conditionally compiled behind a flag 'json'

docs/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 
 ## Unreleased
 ### Added
+* Added JsonFormat ([#950]).
```
jojolepro: You also need to link `#950` to the pr link
like this:
`[#950]: https://github.com/amethyst/amethyst/pull/950`

robinvd: Ah i thought it would work like github references: #950. Fixed

jojolepro: Thanks!

jojolepro:
Needs @LucioFranco approval and 20 more hours to wait :3

robinvd:
The flag can always be made default/removed later, so i think this is a good choice.

PR #962, opened 2018-09-23
Update an example from the System book chapter

### LINKED ISSUE #918: Docs: Example in the book invalid in most cases
The example regarding ActiveCamera is out of date:
https://www.amethyst.rs/book/develop/concepts/system.html#getting-a-component-of-a-specific-entity

It can panic, because no ActiveCamera Resource is found,

Moxinilian:
Fixes #918 and a simple warning in renderer.

### DIFF
--- a/book/src/concepts/state.md
+++ b/book/src/concepts/state.md
@@ -189,7 +189,7 @@ struct MyEvent {
     data: i32,
 }
 
-struct MyState;
+struct GameplayState;
 
 impl State<(), MyEvent> for GameplayState {
     fn handle_event(&mut self, _data: StateData<()>, event: StateEvent<MyEvent>) -> Trans<(), MyEvent> {
--- a/book/src/concepts/system.md
+++ b/book/src/concepts/system.md
@@ -4,6 +4,8 @@
 
 A system is where the logic of the game is executed. In practice, it consists of a struct implementing a function executed on every iteration of the game loop, and taking as an argument data about the game.
 
+Systems can be seen as a small unit of logic. All systems are run by the engine together (even in parallel when possible), and do a specialized operation on one or a group of entities.
+
 ## Structure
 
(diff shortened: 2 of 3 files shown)


book/src/concepts/state.md
```diff
@@ -183,6 +183,7 @@ If you were to replace `()` by `MyEventType`, then you could react to them from
 # use amethyst::prelude::*;
 # use amethyst::renderer::VirtualKeyCode;
 # use amethyst::input::is_key_down;
+# struct GameplayState;
```
jojolepro: https://github.com/amethyst/amethyst/pull/962/files#diff-9c67de6309fc1e25b360604449cc8301R193

what is MyState used for here?

book/src/concepts/system.md
```diff
@@ -66,24 +68,24 @@ Sometimes, it can be useful to get a component in the storage for a specific ent
 
 ```rust,no_run,noplaypen
 # extern crate amethyst;
-# use amethyst::ecs::{System, ReadExpect, WriteStorage};
+# use amethyst::ecs::{Entity, System, WriteStorage};
 # use amethyst::core::Transform;
-# use amethyst::renderer::ActiveCamera;
-struct CameraGoesUp;
+struct WalkPlayerUp {
+    player: Entity,
```
jojolepro: This works, but it is a bad practice to store runtime data in systems, as it reduces re-usability.
Would it be possible to have a "MovePlayerUp" component and the system join over (Transform, MoveUp) instead.
MoveUp can have a float inside called move_speed and the system multiplying move_speed by time.delta_seconds() when moving the player.

book/src/concepts/state.md
Moxinilian: I don't know, git blame you.

book/src/concepts/system.md
Moxinilian: The point is that the reader does not know join at that point.
And no.

PR #964, opened 2018-09-23
Default font loading

### LINKED ISSUE #761: Error because of missing default font used by UiButtonBuilder::new()
This is a minor issue, but UiButtonBuilder::new() is using DEFAULT_FONT_NAME: &'static str = "font/square.ttf" if it doesn't exist, you get an error in stdin. But it doesn't matter because you can set a new font later .with_font. So, there should be some way to suppress the error in stdin or set the font before Builder tries to use DEFAULT_FONT_NAME.

jojolepro:
fixes #761

### DIFF
--- a/amethyst_ui/src/font/default.rs
+++ b/amethyst_ui/src/font/default.rs
@@ -10,7 +10,8 @@ pub fn get_default_font(loader: &Loader, storage: &AssetStorage<FontAsset>) -> F
 
 	match system_font {
 		Ok(handle) => match handle {
-			FontKitHandle::Path{..} => unimplemented!(),
+			FontKitHandle::Path{..} => unimplemented!("Default system font was provided as a path, this is not yet supported.
+				If you see this message, open an issue so that we know we need to implement it."),
 			FontKitHandle::Memory{bytes, ..} => {
 				let font_data = TtfFormat.import(bytes.to_vec(), ());
 				match font_data {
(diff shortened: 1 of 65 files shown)


jojolepro:
changelog incoming in 10 minutes

jojolepro:
Ready for review!

amethyst_ui/src/font/systemfont.rs
```diff
@@ -0,0 +1,32 @@
...
+	let family_handles = families.into_iter().flat_map(|fam| source.select_family_by_name(&fam));
+	let font_handles = family_handles.map(|fam| fam.fonts().to_owned()).flatten().collect::<Vec<_>>();
+    Ok(font_handles)
+}
+
+/// Returns the default system font.
+pub fn default_system_font() -> Result<Handle, SelectionError> {
+	let source = SystemSource::new();
+	// TODO: remove once font_kit accepts titles with &str
+	let default_fonts = &[FamilyName::Title("arial".to_string()), FamilyName::SansSerif, FamilyName::Monospace, FamilyName::Fantasy];
```
randomPoison: `FamilyName::Fantasy` as the final fallback? Is that a common thing to do?

amethyst_ui/src/font/default.rs
```diff
@@ -0,0 +1,26 @@
...
+use font_kit::handle::Handle as FontKitHandle;
+
+/// Get the system default fonts.
+/// If unable to, gets the local square.ttf font.
+pub fn get_default_font(loader: &Loader, storage: &AssetStorage<FontAsset>) -> FontHandle {
+	let system_font = default_system_font();
+
+	match system_font {
+		Ok(handle) => match handle {
+			FontKitHandle::Path{..} => unimplemented!(),
```
randomPoison: It would probably be good to include a message when using `unimplemented` so that anyone who hits this case get's a more useful message, e.g. `unimplemented!("Default system font was provided as a path, this is not yet supported")`.

Issue #928, opened 2018-09-02
UiText doesn't show multiple lines


Temeez:
If I put `\n` in the String for UiText.text then it'll only render the text before the `\n` when it should render multiple lines of text.

```rust
        let text_debug_transform = UiTransform::new(
            "debug_text:".to_string(),
            Anchor::Middle,
            0.0, 0.0, 1.0,
            width, height, // ScreenDimensions
            0
        );
```

```rust
        let text_debug = world
            .create_entity()
            .with(text_debug_transform)
            .with(UiText::new(
                font.clone(),
                "Input Device: Mouse\nPlayer Position: 0 0".to_string(),
                [0.0, 0.0, 0.0, 1.0],
                20.0
            ))
            .build();
```
Text appears on the top left corner of the screen and the result of above code is:
```
Input Device: Mouse

```

When it should be:
```
Input Device: Mouse
Player Position: 0 0
```

I am on the most recent develop branch.
```
[dependencies]
amethyst = { git = "https://github.com/amethyst/amethyst", rev = "f6cbdfe6e5c38838190e59a35b6e6d35115af0cb" }
```

jojolepro:
That's because internally we use the single line mode of gfx-glyph. I have a fork where I was working on that (and text alignment) but it is not ready yet as the text doesn't actually align properly.

PR #965, opened 2018-09-23
Text alignment and multiline support

### LINKED ISSUE #594: UiText alignment
Feature request

Align left default----
--align center--
----align right

Also, if the text doesn't fit into the mesh holding it, it might be a good idea to have an "overflow" option.
if we reverse the problem, changing the text size so it fits onto the mesh could be another option.

Also top,center and bottom vertical alignements.
When I use my layouting PR to set the text to be fill a rectangle, having the text centered in the middle would be nice ;)
Less testing required to align stuff this way.

Since this requires modifying the pass, I'd consider this average difficulty.

jojolepro:
Fixes #594
Fixes https://github.com/amethyst/amethyst/issues/928

image

### DIFF
--- a/amethyst_ui/src/font/systemfont.rs
+++ b/amethyst_ui/src/font/systemfont.rs
@@ -4,10 +4,6 @@ use font_kit::source::SystemSource;
 use font_kit::family_name::FamilyName;
 use font_kit::properties::{Properties, Style};
 
-// Temporarily disabled because const can't contain heap allocations.
-// Fantasy contains comic sans btw
-//const DEFAULT_FONTS: &[FamilyName] = &[FamilyName::Title("arial".to_string()), FamilyName::SansSerif, FamilyName::Monospace, FamilyName::Fantasy];
-
 /// Lists all installed font families on the system.
 pub fn list_system_font_families() -> Result<Vec<String>, SelectionError> {
 	let source = SystemSource::new();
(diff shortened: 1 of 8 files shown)


amethyst_ui/src/font/systemfont.rs
```diff
@@ -0,0 +1,32 @@
+use font_kit::handle::Handle;
+use font_kit::error::SelectionError;
+use font_kit::source::SystemSource;
+use font_kit::family_name::FamilyName;
+use font_kit::properties::{Properties, Style};
+
+// Temporarily disabled because const can't contain heap allocations.
+// Fantasy contains comic sans btw
+//const DEFAULT_FONTS: &[FamilyName] = &[FamilyName::Title("arial".to_string()), FamilyName::SansSerif, FamilyName::Monospace, FamilyName::Fantasy];
```
Moxinilian: Not interfering with the merge, but I'd recommend PRing Arial as a default family name to that crate you use.

jojolepro:
image

Moxinilian:
Beautiful

amethyst_ui/src/font/systemfont.rs
jojolepro: I would actually PR a way to declare FamilyName without an allocation if possible

Xaeroxe: Should remove this, const will never have the ability to contain heap allocations.

jojolepro:
Ready to go.
There's a bug where if you do the following input sequence on an input text, a char can become invisible. I don't know where the bug comes from and didn't see anything that looked suspicious on the amethyst side (I checked text.rs line 524).

```
select edit text
erase all text
press a
press backspace
press a (invisible char)
press a (both "a" are visible)
Works with any char key, not just "a"
```

PR #532, opened 2018-01-07
[Review] Networking

jojolepro:
Early PR for discussions and suggestions.

Currently trying to find a way to annotate the serialized data with the type of the struct(s) inside and to deserialize using that data. (Types are known at compile time using world::register, but you will only know what type you receive using that annotation).

The types (read components + resources + events) that are registered are going to be different on the client and on the server. A server won't have the `Sprite` component registered, for example. This means that incrementing an id is not an option. Also, sending the full type name isn't either, because of how long and costly that string would be to send.

Progress tracker:
- [X] Serialization/Deserialization
- [X] Basic socket implementation
- [x] Connection management
- [X] Message Passing (NetEvent) + extensibility
- [x] Mio integration (per request)
- [x] Clean

What isn't in the scope of this PR:
-Everything not listed. If you want the list of planned features, see #335 at the "10 Jan" mark.

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/532)

Issue #335, opened 2017-09-06
RFC: Networking


LucioFranco:
## Motivation

In the day and age of high-speed internet, nobody wants to play alone. For this, we need an easy way for users to add multiplayer networking to their games without being networking masters.

## Implementation

The main goal of the initial implementation is to create three main features:

- Syncable components/entities, with either remote or local syncing
- RPC's that can either be sent from client -> server or server -> client via macros
- Server/Client that runs in its own thread via `Tokio`

### Syncing Components

The goal with this feature is to have the ability to really easily sync the state of a component to the server and clients. There are two ways that this syncing can occur, one is remote and the other is local. Remote Syncing means that the data that needs to be synced comes from the server then is applied to the component locally. For example, this would be like an enemy player that gets its data from the server. Local Syncing is where the local data is then replicated across the clients. For example, this would be the current player's controllable entity. In Pong, this would be the current local player's plank.

The syncing of the components can be accomplished in two ways. One way is to sync the Transform components of an entity, this can be done by the `NetTransformSystem` which will collect all entities with the `Transform` and `RemoteSync\LocalSync`, with these components, it will then communicate with the `Server/Client` to replicate the data. The second way is to build a custom system to sync certain components. In the future, I think code generation might help this but this system will use built-in functions to serialize custom components.

### RPC's

The implementation of the RPC's is done with macros. This would be for example to trigger game events that are not directly related to a component. It could be to play a certain sound across the different clients.

### Other implementation details

There will also be the need to add a Resource so any `System` can access the current connection. This will be done via a `ConnectionInfo` resource that can be fetched from any `System`. This will maintain details related to the connection. For example, store if the connection is active or not.

As for behind the scenes, we will have another thread running a `UdpSocket` from `tokio-core`. This will then communicate to the game thread via `mspc` channels. Then `bincode` with `serde` will encode the components and messages that will be sent.

The `Server` will be implemented by a trait that will support the ability to react to events on the server side. Then the user can use the NetworkManager to communicate with the network IO thread.

To communicate with the server there will be a concept of a `LocalClient` and `RemoteClient`. The `LocalClient` is designed to communicate to the `Server` running locally without going over the wire. The `RemoteClient` will communicate to the `Server` running remotely over `UDP`.

### Future

In the future, some other planned features is a light weight `TCP` like protocol, this way you can set what type of replication you want. If you want the messages to *always* arrive or just attempt to arrive and don't attempt to resend.

[...]

PR #969, opened 2018-09-23
Base Networking implementation

LucioFranco:
This PR is in reference to #335. This PR originates from #665 and #532, this also stems from @jojolepro work in both those PRs.

This provides an initial implementation for networking just presenting the most basic way of communicating to a server and client that can be integrated into an amethyst application. In its current state, the only working example is `net_server` and `net_client`. When run at the same time, they send messages back and forth. There is also some commented code, mainly in the components folder of the `amethyst_network` crate due to not fitting the scope of the base implementation, but will be worked on after this PR is merged.

Opening this for inital review and comments.

Once, we agree this is good to go, I will add the changelog entry and squash.


Moxinilian:
With the networking thing, I'm mostly concerned about security.
There's no way we can encourage using Amethyst on a server, I mean it's Rust, but we are not even close to the necessary security practices for servers.
I'm sure there's a billion denial of service flaws, and potentially worse in some hidden unsafe blocks.
What's the deal with that?

jojolepro:
We don't have unsafe blocks, thanks jojo :3 
As for denial of services, its the same as the rest of the engine: Remove the unwraps and expects before 1.0 release.

amethyst_network/src/bundle.rs
```diff
@@ -0,0 +1,69 @@
...
+pub struct NetworkBundle<'a, T> {
+    /// The local ip to bind to.
+    ip: &'a str,
+    /// The local port to bind to.
+    port: Option<u16>,
+    /// The filters applied on received network events.
+    filters: Vec<Box<NetFilter<T>>>,
+    /// The server to automatically connect to.
+    /// You would usually want this if you set is_server = false.
+    connect_to: Option<SocketAddr>,
```
jojolepro: This isn't used anymore.

amethyst_network/src/components/mod.rs
```diff
@@ -0,0 +1,5 @@
+//! ECS Components
+
+mod netsync;
+
+pub use self::netsync::*;
```
jojolepro: new line

amethyst_network/src/filter.rs
```diff
@@ -0,0 +1,39 @@
...
+        FilterConnected { _pd: PhantomData }
+    }
+}
+
+impl<T> NetFilter<T> for FilterConnected<T>
+where
+    T: PartialEq + Send + Sync,
+{
+    /// Checks if the event is from a connected client.
+    fn allow(&mut self, _source: &SocketAddr, event: &NetEvent<T>) -> bool {
```
jojolepro: Indicate in the doc that this is not usable currently.

amethyst_network/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+extern crate log;
+#[macro_use]
+extern crate serde;
+extern crate fern;
+extern crate shred;
+extern crate shrev;
+extern crate uuid;
+
+mod bundle;
+// TODO: Readd components and entity sync
```
jojolepro: readd :D

amethyst_network/src/network_socket.rs
```diff
@@ -0,0 +1,193 @@
...
+/// The System managing the network state and connections.
+/// The T generic parameter corresponds to the network event type.
+/// Receives events and filters them.
+/// Received events will be inserted into the NetReceiveBuffer resource.
+/// To send an event, add it to the NetSendBuffer resource.
+///
+/// If both a connection (Connect or Connected) event is received at the same time as another event from the same connection,
+/// only the connection event will be considered and rest will be filtered out.
+// TODO: add Unchecked Event type list. Those events will be let pass the client connected filter (Example: NetEvent::Connect).
+// Current behaviour: hardcoded passthrough of Connect and Connected events.
```
jojolepro: The comments and docs content are fine, but the comment is split in half by the doc in the middle xD

jojolepro: line 33

amethyst_network/src/network_socket.rs
```diff
@@ -0,0 +1,193 @@
...
+                            receive_queue
+                                .send(RawEvent {
+                                    byte_count: amt,
+                                    data: buf[..amt].iter().cloned().collect(),
+                                    source: src,
+                                }).unwrap();
+                        }
+                        Err(e) => {
+                            if e.kind() == ErrorKind::WouldBlock {
+                                //error!("WouldBlock: {}", e);
```
jojolepro: Forgotten comment

amethyst_network/src/utils.rs
```diff
@@ -0,0 +1,37 @@
+extern crate bincode;
```
jojolepro: That's from all the way when I started rust. The extern crate should go into lib.rs

examples/net_server/main.rs
```diff
@@ -0,0 +1,73 @@
...
+    fn run(&mut self, (mut connections,): Self::SystemData) {
+        let mut count = 0;
+        for (mut conn,) in (&mut connections,).join() {
+            if self.reader.is_none() {
+                self.reader = Some(conn.receive_buffer.register_reader());
+            }
+            for ev in conn.receive_buffer.read(self.reader.as_mut().unwrap()) {
+                count += 1;
+                match ev {
+                    &NetEvent::TextMessage { ref msg } => println!("{}", msg),
```
jojolepro: Could use info instead of println

TimonPost:
Why one system/bundle for both client and server?  Is it not better to separate the concerns here?

amethyst_network/src/bundle.rs
LucioFranco: Good catch

amethyst_network/src/filter.rs
LucioFranco: Okay, could you lay out for me how you initially saw this trait being used for? You can ping me on discord.

LucioFranco:
@jojolepro thanks for the comments, gonna take a crack at those later today.

@TimonPost Currently, there is no connect of a "Server" or "Client". Really, the only concept currently implemented is a simple form of two-way communication over UDP. Once, this is merged my next goal is to make the ping tutorial multiplayer and probably during that phase I will uncover stuff that will lead to a better implementation and possibly having these concepts encoded as systems.

TimonPost:
@LucioFranco You might want to look at this: https://github.com/TimonPost/amethyst/tree/net/amethyst_network. 

I have been also working on the networking part on @jojolepro his version. Today I copied my code to your branch since you had done a cleanup. 

I had contact with the maker of the crate [game_udp](https://github.com/acmcarther/gaffer_udp) implementation. In this issue you can see or [conversation history](https://github.com/acmcarther/gaffer_udp/issues/10).

So basically what that crate is, is an implementation of a game network environment described by Gaffer. He describes in several blogs the best practices for game networking. So I went an first implemented it by my self but then I found an already implemented create exactly as he described in his blog posts.

As you can see int the chat history this creates foundations where solid so I went implementing it for Amethyst network part. And that is what all my changes are about?

One question do you want me to make a pull request now or when this pull request is merged?

I think these changes made by me are helpful since now there is reliable UDP witch you definitely want for game networking. As you can see [here](https://github.com/TimonPost/amethyst/blob/net/amethyst_network/README.md) these are the points we need to keep in mind when implementing networking.

jojolepro:
I have plans to have different channels to separate rudp and normal udp

LucioFranco:
@TimonPost @jojolepro I want to round out this convo, so just moving forward, we do want all of this but we need this base here first. 

@TimonPost I will be taking a look at the PR you made against my branch and the docs. But I do not want to merge your PR into my branch but would rather have that exist as a PR against the master branch of this repo. So, once this is through, we can work on that part.

amethyst_network/src/bundle.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+impl<'a, T> NetworkBundle<'a, T> {
+    /// Creates a new NetworkBundle in client mode.
+    pub fn new_client(ip: &'a str, port: Option<u16>, filters: Vec<Box<NetFilter<T>>>) -> Self {
```
fhaynes: You may want to forbid ports < 1024 here, as those require root. Or do a permissions check to tell the user they need to sudo.

jojolepro: nice catch

LucioFranco: I added a warning here just in case someone is crazy enough to use it.

amethyst_network/src/network_socket.rs
```diff
@@ -57,6 +57,11 @@ where
         addr: SocketAddr,
         filters: Vec<Box<NetFilter<E>>>,
     ) -> Result<NetSocketSystem<E>, Error> {
+        if addr.port() < 1024 {
+            // Just warning the user here, just in case they want to use the root port.
+            warn!("Using a port below 1024, this will require root permission and should not be done.");
```
jojolepro: Can you also specify that running as root is probably the worst idea ever too? :)
People might think you are telling them to run as root.

amethyst_network/src/test.rs
```diff
@@ -60,17 +61,17 @@ mod test {
         {
             let mut sto = WriteStorage::<NetConnection<()>>::fetch(&world_cl.res);
             for mut cmp in (&mut sto).join() {
-                for _i in 0..10000 {
+                for _i in 0..100 {
```
jojolepro: Hmmm? The method was called 100k packets so fill up the system send buffer of the client.

jojolepro: Though we might not want to do that with travis lol

LucioFranco: Exactly why I reduced it, I could reproduce the test passing locally but it seems Travis couldn't keep up.

amethyst_network/src/network_socket.rs
```diff
@@ -0,0 +1,188 @@
...
+                if net_connection.target == raw_event.source {
+                    matched = true;
+                    // Get the event
+                    let net_event = deserialize_event::<E>(raw_event.data.as_slice());
+                    match net_event {
+                        Ok(ev) => {
+                            // Filter events
+                            let mut filtered = false;
+
+                            if !filtered {
```
Xaeroxe: There's no opportunity for `filtered` to be true here.  Why are we doing this?

LucioFranco: Thanks for catching this, I've removed it. I want to look back into filtering once this is merged.

Xaeroxe:
Let's get this show on the road

bors r+

PR #976, opened 2018-09-30
Logger improvements

Ristarg:
Implements following changes:
+ Add optional logging to file (with new env variable)
+ Add toggle for logging to stdout (with new env variable)
+ General refactoring

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -1,28 +1,31 @@
 pub use log::LevelFilter;
 
-use fern::{self, FormatCallback};
-use log;
-use std::{env, fmt, io, str::FromStr};
+use fern;
+use std::{env, io, path::PathBuf, str::FromStr};
 
 /// Logger configuration object.
-#[derive(Clone, Copy)]
+#[derive(Clone)]
 pub struct LoggerConfig {
+    /// Determines whether to log to the terminal or not.
+    pub use_stdout: bool,
     /// Whether to use color output when logging to the terminal or not.
     pub use_colors: bool,
     /// Sets the overarching level filter for the logger.
     pub level_filter: LevelFilter,
+    /// If set, enables logging to file at the given path.
+    pub log_file: Option<PathBuf>,
 }
 
 impl Default for LoggerConfig {
     fn default() -> LoggerConfig {
-        let use_colors = env::var_os("AMETHYST_LOG_DISABLE_COLORS").is_none();
-        let level_filter = env::var("AMETHYST_LOG_LEVEL_FILTER").ok()
-            .and_then(|lf| LevelFilter::from_str(&lf).ok())
-            .unwrap_or(LevelFilter::Debug);
-
         LoggerConfig {
-            use_colors,
-            level_filter,
+            use_stdout: env::var_os("AMETHYST_LOG_DISABLE_STDOUT").is_none(),
+            use_colors: env::var_os("AMETHYST_LOG_DISABLE_COLORS").is_none(),
+            level_filter: env::var("AMETHYST_LOG_LEVEL_FILTER").ok()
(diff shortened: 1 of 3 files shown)


src/logger.rs
```diff
@@ -38,41 +36,74 @@ impl Default for LoggerConfig {
...
+        }
+    }
+
+    dispatch.apply().unwrap_or_else(|_| {
+        debug!("Global logger already set, default Amethyst logger will not be used")
+    });
+}
+
+fn env_var_override(config: &mut LoggerConfig) {
+    if env::var("AMETHYST_LOG_DISABLE_STDOUT").is_ok() {
```
torkleyy: I think this should better use the positive form and check if the value is `0` / `1` and override respectively.

src/logger.rs
```diff
@@ -1,33 +1,31 @@
...
-use std::env;
-use std::io;
+
+use fern;
+use std::{env, io, path::PathBuf, str::FromStr};
 
 /// Logger configuration object.
-#[derive(Clone, Copy)]
+#[derive(Clone)]
 pub struct LoggerConfig {
```
torkleyy: Could add a `allow_env_override` or something like that so @Khionu is happy with this, too ;)

src/logger.rs
```diff
@@ -38,41 +36,85 @@ impl Default for LoggerConfig {
 /// initialise your own.
 ///
 /// Configuration of the logger can also be controlled via environment variables:
-///
-/// * AMETHYST_LOG_DISABLE_COLORS - if set, disables colors for the log output
+/// * AMETHYST_LOG_ENABLE_STDOUT - toggles the logging to the terminal
```
torkleyy: `AMETHYST_LOG_STDOUT` would be enogh IMO, same for colors

src/logger.rs
```diff
@@ -1,33 +1,31 @@
...
 pub use log::LevelFilter;
-use std::env;
-use std::io;
+
+use fern;
+use std::{env, io, path::PathBuf, str::FromStr};
 
 /// Logger configuration object.
-#[derive(Clone, Copy)]
+#[derive(Clone)]
```
torkleyy: Ah, another idea, this should derive `Serialize` so we change logging without recompiling.

Ristarg: Do you mean both `Serialize` and `Deserialize`? Also, `serde` is currently listed as a dev dep, should I promote it then?

magnonellie: Be careful with this as the current ways of deserializing in Amethyst rely on the log macros for error handling, meaning that you wouldn't get an error message when initializing the logger if your config is wrong.

src/logger.rs
```diff
@@ -1,33 +1,31 @@
...
 pub struct LoggerConfig {
+    /// Determines whether to log to the terminal or not.
+    pub use_stdout: bool,
     /// Whether to use color output when logging to the terminal or not.
     pub use_colors: bool,
```
magnonellie: Perhaps the two variables above should be combined into an enum since `use_colors` depends on `use_stdout`.

src/logger.rs
Ristarg: Actually, `LevelFilter` doesn't implement `Serialize`, so I'm not sure if this is possible (I should probably note that I haven't worked with `serde` before)

torkleyy:
It does, if the serde feature of log is enabled.

Ristarg:
Ah, gotcha gotcha

src/logger.rs
torkleyy: @magnonellie That's only in `Config`, which shouldn't be used here.

torkleyy:
Thanks, merging this now.

bors r+

Ristarg:
Oh wow. *facepalm*
I'll fix this later today.

PR #981, opened 2018-09-30
Refactor `TextureMetadata`

### LINKED ISSUE #966: Make the texture filtering API public
Currently, there is no way to change the filtering for textures without manually importing gfx. This forces the textures to be displayed with nearest filtering, which usually makes games and demos look unimpressive.

We should find a way to expose that to the user, possibly in a backend agnostic way. At least exporting the 3 gfx types needed could also do the trick.

**This is especially needed for 2D stuff hence why it's important for 0.9**

torkleyy:
Fixes #966 

## TODO

* [x] add example that shows custom filtering
* [x] ~~decide on a default filter method~~ (postponed)
* [x] changelog

### DIFF
--- a/amethyst_gltf/src/format/material.rs
+++ b/amethyst_gltf/src/format/material.rs
@@ -25,6 +25,7 @@ pub fn load_material(
             buffers,
             source.clone(),
             name,
+            true,
         ).map(|(texture, _)| TexturePrefab::Data(texture))?,
     );
 
@@ -41,6 +42,7 @@ pub fn load_material(
         buffers,
         source.clone(),
         name,
+        false,
     ).map(|(texture, factors)| deconstruct_metallic_roughness(texture, factors[0], factors[1]))
     .map(|(metallic, roughness)| {
         (
@@ -59,13 +61,14 @@ pub fn load_material(
             buffers,
             source.clone(),
             name,
+            false,
         ).map(|(texture, _)| TexturePrefab::Data(texture))?,
     );
 
     // Can't use map/and_then because of Result returning from the load_texture function
     prefab.normal = match material.normal_texture() {
         Some(normal_texture) => Some(
-            load_texture(&normal_texture.texture(), buffers, source.clone(), name)
+            load_texture(&normal_texture.texture(), buffers, source.clone(), name, false)
                 .map(|data| TexturePrefab::Data(data))?,
         ),
 
@@ -75,7 +78,7 @@ pub fn load_material(
     // Can't use map/and_then because of Result returning from the load_texture function
     prefab.ambient_occlusion = match material.occlusion_texture() {
         Some(occlusion_texture) => Some(
-            load_texture(&occlusion_texture.texture(), buffers, source.clone(), name)
+            load_texture(&occlusion_texture.texture(), buffers, source.clone(), name, false)
                 .map(|data| TexturePrefab::Data(data))?,
         ),
 
@@ -142,10 +145,11 @@ fn load_texture_with_factor(
     buffers: &Buffers,
     source: Arc<Source>,
     name: &str,
+    srgb: bool,
 ) -> Result<(TextureData, [f32; 4]), GltfError> {
     match texture {
         Some(info) => Ok((
-            load_texture(&info.texture(), buffers, source, name)?,
+            load_texture(&info.texture(), buffers, source, name, srgb)?,
             factor,
         )),
         None => Ok((TextureData::color(factor), [1.0, 1.0, 1.0, 1.0])),
@@ -157,11 +161,15 @@ fn load_texture(
     buffers: &Buffers,
     source: Arc<Source>,
     name: &str,
+    srgb: bool,
 ) -> Result<TextureData, GltfError> {
     let (data, format) = get_image_data(&texture.source(), buffers, source, name.as_ref())?;
 
-    // TODO: use Unorm where appropriate
-    let metadata = TextureMetadata::srgb().with_sampler(load_sampler_info(&texture.sampler()));
+    let metadata = match srgb {
+        true => TextureMetadata::srgb(),
+        false => TextureMetadata::unorm(),
+    };
+    let metadata = metadata.with_sampler(load_sampler_info(&texture.sampler()));
     Ok(match format {
         ImageFormat::Png => PngFormat::from_data(&data, metadata),
         ImageFormat::Jpeg => JpgFormat::from_data(&data, metadata),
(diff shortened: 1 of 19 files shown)


amethyst_gltf/src/format/material.rs
```diff
@@ -161,7 +156,9 @@ fn load_texture(
     name: &str,
 ) -> Result<TextureData, GltfError> {
     let (data, format) = get_image_data(&texture.source(), buffers, source, name.as_ref())?;
-    let metadata = TextureMetadata::default().with_sampler(load_sampler_info(&texture.sampler()));
+
+    // TODO: use Unorm where appropriate
```
torkleyy: Oh and that is TODO, too.

amethyst_renderer/src/formats/texture.rs
```diff
@@ -7,71 +7,104 @@ use amethyst_assets::{
...
+    pub format: SurfaceType,
+    /// The channel type which describes the data format of the channels (e.g. how the red value
+    /// is stored).
+    ///
+    /// This is usually `Srgb` for color textures, normalmaps & similar mostly use `Unorm`
+    /// (a floating-point number between `0.0` and `1.0`.
+    pub channel: ChannelType,
 }
 
+/*
```
torkleyy: should be removed before merging; we realized there is no proper default thus I removed it.

amethyst_renderer/src/formats/texture.rs
```diff
@@ -7,71 +7,104 @@ use amethyst_assets::{
...
+    ///
+    /// The default is `R8_G8_B8_A8`.
+    #[serde(default = "SurfaceFormat::get_surface_type")]
+    pub format: SurfaceType,
+    /// The channel type which describes the data format of the channels (e.g. how the red value
+    /// is stored).
+    ///
+    /// This is usually `Srgb` for color textures, normalmaps & similar mostly use `Unorm`
+    /// (a floating-point number between `0.0` and `1.0`.
+    pub channel: ChannelType,
```
torkleyy: Note: this is the only field here that has no default.

examples/assets/prefab/sphere.ron
```diff
@@ -6,7 +6,7 @@ Prefab (
                 graphics: (
                     mesh: Shape((shape: Sphere(32, 32),)),
                     material: (
-                        albedo: Data(Rgba((0.0, 0.0, 1.0, 1.0,), (),)),
+                        albedo: Data(Rgba((0.0, 0.0, 1.0, 1.0,), (channel: Srgb),)),
```
torkleyy: This is unfortunate, but required with our current architecture. We might be infer that from the `Material`.

cc @Rhuagh

amethyst_renderer/src/formats/texture.rs
```diff
@@ -499,6 +513,64 @@ impl SimpleFormat<Texture> for TextureFormat {
...
 }
 
+mod serde_helper {
+    use super::SamplerInfo as SamplerInfoComplex;
+    use serde::{Serialize, Deserialize, Serializer, Deserializer};
+    use tex::{FilterMethod, WrapMode};
+
+    #[derive(Serialize, Deserialize)]
+    #[serde(untagged)]
+    enum SamplerInfo {
```
torkleyy: It's not a very clean solution, but I didn't see another way to allow specifying the filter without making it very verbose.

amethyst_renderer/src/formats/texture.rs
```diff
@@ -7,71 +7,104 @@ use amethyst_assets::{
...
+    /// In simpler words, this defines the color format, e.g. RGBA 32-bit.
+    ///
+    /// The default is `R8_G8_B8_A8`.
+    #[serde(default = "SurfaceFormat::get_surface_type")]
+    pub format: SurfaceType,
+    /// The channel type which describes the data format of the channels (e.g. how the red value
+    /// is stored).
+    ///
+    /// This is usually `Srgb` for color textures, normalmaps & similar mostly use `Unorm`
+    /// (a floating-point number between `0.0` and `1.0`.
```
omni-viral: Not sure when does `Unorm` became `floating-point` :)

amethyst_renderer/src/formats/texture.rs
omni-viral: Maybe `Unorm` could be default. If it is possible to calculate later then `Option<ChannelType>` could be used. with `None` meaning is "get from asset or derive from usage or use `Unorm` as default"

amethyst_renderer/src/formats/texture.rs
omni-viral: This *is* clever solution.

amethyst_renderer/src/formats/texture.rs
torkleyy: I un-implemented `Default` on purpose. I want to force us to immediately specify the channel type because later the information is not available anymore.

amethyst_renderer/src/formats/texture.rs
torkleyy: Actually, the easiest and best solution would be adding defaults to gfx.

amethyst_gltf/src/format/material.rs
```diff
@@ -114,12 +114,7 @@ fn deconstruct_image(data: &TextureData, offset: usize, step: usize) -> TextureD
     use gfx::format::SurfaceType;
     match *data {
         TextureData::Image(ref image_data, ref metadata) => {
-            let metadata = metadata
-                .clone()
-                .with_size(
```
Rhuagh: Why remove size? You won't know size from the data itself from gltf.

torkleyy: Yes, it's in the image data. Let me check one sec.

torkleyy: Fixed.

amethyst_renderer/src/formats/texture.rs
```diff
@@ -7,71 +7,106 @@ use amethyst_assets::{
...
 use gfx::traits::Pod;
-use tex::{Texture, TextureBuilder};
+use tex::{FilterMethod, Texture, TextureBuilder};
+use types::SurfaceFormat;
 use Renderer;
 
-/// Texture metadata, used while loading
+/// Additional texture metadata that can be passed to the asset loader or added to the prefab.
+///
+///
```
jojolepro: 2 empty lines?

torkleyy: Don't know how they got there. I'm myself allergic to those ;P

torkleyy:
@Moxinilian I'm not sure if it should be `Anisatropic`. Also, I'm a bit unsure if it's good that there's a per-texture setting for the anisatropic filtering. That's something you want to set globally.

omni-viral:
@torkleyy you definitely want to set filtering per-texture. Nothing stops you from setting it from value stored globally.

torkleyy:
@omni-viral I think I want to set filtering per sampler and sampler per texture, but yeah.

torkleyy:
I'm gonna open an issue and let you guys merge my PR if you're happy with it.

Issue #890, opened 2018-08-19
Transparent image parts ignore the window background


ab0v3g4me:
Here's the code used for drawing my `.png` file. I'm using Wayland as my display server.
```rust
pub fn initialise_player(world: &mut World, spritesheet: TextureHandle) {
    /* . . . . .  */
    let mut transform = Transform::default();
    transform.translation = Vector3::new(200.0,200.0,0.5);
    world.create_entity()
        .with_sprite(&sprite, spritesheet.clone(), (32.0,32.0))
        .expect("Failed to add sprite render on player")
        .with(Player::new())
        .with(GlobalTransform::default())
        .with(transform)
        .build();
}
```

```rust 
    fn on_start(&mut self, data: StateData<GameData>) {
        let world = data.world;
        let spritesheet = {
            let loader = world.read_resource::<Loader>();
            let texture_storage = world.read_resource::<AssetStorage<Texture>>();
            loader.load(
                "player.png",
                PngFormat,
                Default::default(),
                (),
                &texture_storage,
            )
        };
        player::initialise_player(world, spritesheet);
    }
```

```rust
    let pipe = Pipeline::build().with_stage(
        Stage::with_backbuffer()
        .clear_target([0.0, 0.0, 0.0, 1.0], 1.0)
        .with_pass(DrawFlat::<PosTex>::new()),
    );
```
Here's an image on how it looks like. 

![image](https://user-images.githubusercontent.com/19503519/44313219-76843380-a404-11e8-919c-f28768591326.png)

Telzhaak:
@Moxinilian 
Transparency wasn't fixed with using a different png encoding, it just so happened to fill the alpha with the same colour as the background (black):
![unbenannt2](https://user-images.githubusercontent.com/29286474/44574012-77142580-a788-11e8-987e-f4bc4b147437.png)

For other .png encodings (do not know which this is):
![heart](https://user-images.githubusercontent.com/29286474/44574164-e8ec6f00-a788-11e8-8144-e1f5407837ad.png)
gets displayed as:
![unbenannt](https://user-images.githubusercontent.com/29286474/44574145-e12cca80-a788-11e8-85bf-09fcb07f28cd.PNG)

minecrawler:
untested, but I think transparency is missing completely from the [flat shader](https://github.com/amethyst/amethyst/blob/develop/amethyst_renderer/src/pass/shaders/fragment/flat.glsl). It might be done with the following main function (just replace it and test again):

```glsl
[...]

PR #984, opened 2018-10-01
Add workaround for Wayland issues

torkleyy:
Addresses #890 

This just asks winit to use X11 instead of Wayland.

Known side effect: if no dimensions are specified, the default window size on Wayland will differ from the one on X11

### DIFF
--- a/amethyst_renderer/src/renderer.rs
+++ b/amethyst_renderer/src/renderer.rs
@@ -160,7 +160,9 @@ impl RendererBuilder {
         RendererBuilder {
             config: DisplayConfig::default(),
             events: el,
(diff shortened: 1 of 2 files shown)


torkleyy:
should we hardcode a default size to make the side effect disappear?

jojolepro:
I just realized. That workaround can't work on mobile. :/

torkleyy:
On Android this will be ignored anyway?

torkleyy:
What should we do about the dimensions?

jojolepro:
Use dev defined size. If not present, default to either screensize or 800x600?

torkleyy:
Done.

jojolepro:
@Moxinilian r & bors ?

PR #988, opened 2018-10-02
Ungrab camera

Ristarg:
Implements #828.
Not sure if this is idiomatic code, but it works.


Ristarg:
Misunderstood what the issue was about, apologies.

Issue #939, opened 2018-09-09
Mapping MouseWheel to an Axis


yeliknewo:
I am making a strategy game. I need to be able to zoom in and out with the mouse wheel. The current input system doesn't directly handle MouseWheel. My work around is to handle_event the MouseWheel and store the movement in a Resource.

jojolepro:
this https://docs.rs/winit/0.17.2/winit/enum.MouseScrollDelta.html
needs to be added here https://github.com/amethyst/amethyst/blob/develop/amethyst_input/src/button.rs#L8
as well as the handling

yeliknewo:
How hard would this be for me to implement?

jojolepro:
Pretty easy. You add a variant in the second link of my comment for the event found in the first link, then you add a handling in amethyst_renderer/src/system (or renderer?)
that's all of it

OvermindDL1:
Isn't a mousewheel effectively just more buttons in the OS?  I'm unsure how well that would map, especially as it 'clicks', like does it reset back to 0 when read once or...?

magnonellie:
When using touchpads and similar the scroll is analog which I think most input systems have adapted at this point even for regular scroll wheels.

OvermindDL1:
Internally the touchpads don't send an analog signal, since the OS doesn't work that way, rather it just translates it in to `keydown`/`keyup` messages, you can even listen for those and see.

magnonellie:
Wayland would like to have a word with you: https://wayland.freedesktop.org/libinput/doc/latest/api/group__event__pointer.html#ga81ad7d8a95c456731a874e584c4c8dda 😝

OvermindDL1:
Oooo wayland changed that?  I'm liking that.  Yep then I'm +1 on relative (not absolute) axis then.  :-)

Xaeroxe:
Fixed in #991

PR #991, opened 2018-10-03
Added MouseWheelMoved event and input_handler therefore

mh84:
This adds basic support for `winit::DeviceEvent::MouseWheel` to amethyst and handles #939


mh84:
We actually shall not care about both events happening at the same time, because their invoking [implementation](https://github.com/tomaka/winit/blob/089816d9bab498c0ec002d7e560ad940b78d476e/src/platform/linux/wayland/pointer.rs#L146) will trigger only one of them at any time.

I will change the input bindings later this day.

mh84:
So from what i've seen, and please correct me if i'm misunderstanding some parts, we will need another InputEvent like `ActionWheelMoved`, glued to something like `Button::MouseWheel`, and triggered that new Action event when a MouseWheel event is detected?

jojolepro:
https://github.com/amethyst/amethyst/blob/master/amethyst_input/src/button.rs#L8

amethyst_input/src/input_handler.rs
```diff
@@ -213,6 +213,18 @@ where
                 } => {
                     event_handler.single_write(MouseMoved { delta_x, delta_y });
                 }
+                DeviceEvent::MouseWheel {
```
torkleyy: Won't there be both events emitted?

mh84: No, there wont't be both of them emitted at the same time, as mentioned here:

> 
> 
> We actually shall not care about both events happening at the same time, because their invoking [implementation](https://github.com/tomaka/winit/blob/089816d9bab498c0ec002d7e560ad940b78d476e/src/platform/linux/wayland/pointer.rs#L146) will trigger only one of them at any time.

torkleyy: Oh, sorry for asking the same question twice.

amethyst_input/src/button.rs
```diff
@@ -17,6 +17,9 @@ pub enum Button {
     /// Mouse buttons
     Mouse(#[serde(with = "LocalMouseButton")] MouseButton),
 
+    /// Mouse wheel
+    MouseWheel,
```
torkleyy: I think we need to add this as axis to the `InputHandler`. Also, an example would be nice ;)

mh84: I've thought about handling that as Axis too, but we have to consider that we have to split this into 2 Axis that are handled by this Event, with delta_x = horizontal scrolling, delta_y = vertical scrolling, and i'm currently not having a good idea how to integrate the emulated axis with that.

I'm thinking about an example implementation too, may its possible to extend the arc_ball_camera example with distance-control referring to the mousewheel, but this was just my first idea.

amethyst_input/src/scroll_direction.rs
```diff
@@ -0,0 +1,7 @@
+#[derive(Eq, PartialEq, Debug, Copy, Clone, Serialize, Deserialize)]
+pub enum ScrollDirection {
```
jojolepro: amethyst::utils::Axis2

jojolepro: Sorry I linked the wrong spot. I was saying that you basically reimplemented this thing here: https://github.com/amethyst/amethyst/blob/master/amethyst_core/src/axis.rs

I'd prefer if you would use that instead of having a copy of it under a different name.

amethyst_input/src/scroll_direction.rs
```diff
@@ -0,0 +1,7 @@
...
+	ScrollUp,
+	ScrollDown,
+	ScrollLeft,
+	ScrollRight,
+}
```
jojolepro: Can you add a new line after this one?

PR #994, opened 2018-10-04
Add missing reexport in logger

Ristarg:
Followup to #976. I tested the new functionality on the internal examples, so I didn't catch the missing visibility; sorry about that. Took this chance to also update the docs.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -28,6 +28,7 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 * New vertices() method on MeshCreator trait. ([#946])
 * Support for text alignment (align left, center, right). ([#965])
 * Support for multiline text. ([#965])
+* Support for logging to file, toggle for logging to stdout. ([#976], [#994])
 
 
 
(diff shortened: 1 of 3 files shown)


torkleyy:
@jojolepro I don't think that should be a concern here. We've got modules for that, so you don't need to prefix stuff. You can always just import the module or rename the item when importing.

Ristarg:
I think Amethyst could definitely utilize namespacing better than it does right now, instead of polluting the crate's namespace with reexports from private modules. But that should probably come in the context of a bigger PR.

jojolepro:
Its a matter of preference I guess. I'll need to play with crossterm's Stdout and amethyst's logger Stdout in the same time when I integrate the terminal duplex

torkleyy:
@jojolepro Yes, so using modules or renaming would be much better than hoping that random word additions avoid collisions. It's understandable in a single crate that clearly differentiable names are better (for the understanding, not the convenience of not renaming). But deciding to rename an item here (even making it longer) because it collides with other crates doesn't make that much sense. I mean, we even have items that collide with very common things (think of `Result`).

In short: Rename for the purpose of clarity, not to avoid collisions.

jojolepro:
I'll change the logger in a few days so I'll see there what I can do for the name. Thanks!
bors r+

PR #995, opened 2018-10-04
Separation of state and state event from the base crate

jojolepro:
Pre-requisite to more event-driven features.

Mainly:
* Terminal
* Better ui event handling
* Futures


randomPoison:
Could you elaborate on what this enables? The code changes themselves look good, it's just not clear to me why this is necessary (sorry if this was already explained in discord or somewhere else, though).

jojolepro:
The amethyst_terminal crate which I will be adding will need to reference the Trans struct. This cannot be done while this struct is in the base crate, as the base crate depends on all other sub crates (cyclic dependency).
Then I need to move everything else for the same reason.
StateMachine -> (Trans -> State) -> StateEvent
and its not possible to go back up the crate hierarchy

Rhuagh:
Why exactly does terminal need to depend on Trans? Can't you make a command system that is generic enough for that ?

torkleyy:
@jojolepro I agree with Simon, we would like to keep the state machine an optional component.

amethyst_state/src/game_data.rs
```diff
@@ -213,7 +226,7 @@ impl<'a, 'b> GameDataBuilder<'a, 'b> {
     {
         bundle
             .build(&mut self.disp_builder)
-            .map_err(|err| Error::Core(err))?;
+            .map_err(|err| Error(err.0,err.1))?;
```
azriel91: does this compile? `error-chain` [docs](https://docs.rs/error-chain/0.12.0/error_chain/example_generated/struct.Error.html) say the second field is private.

The returned error type changes from `amethyst::error::Error(core::Error(..))` to `core::Error(..)`.  `amethyst::error::Error` does implement `From<core::Error>`, so most apps that simply use `.with_bundle(..)?` can live with a straight recompile, but if any of them actually match over the error type will break. So at minimum we need a changelog entry for this.

also, missing a space after the comma.

amethyst_state/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+// TODO: FeatureGate
+extern crate amethyst_input;
+// TODO: FeatureGate
+extern crate amethyst_renderer;
+// TODO: FeatureGate
+extern crate amethyst_ui;
+#[macro_use]
+extern crate derivative;
+
+pub use self::state::{EmptyState, EmptyTrans, SimpleState, SimpleTrans, State, StateData, StateError, StateMachine, Trans,};
```
azriel91: rustfmt probably will wrap the line

src/prelude.rs
```diff
@@ -4,6 +4,6 @@ pub use app::{Application, ApplicationBuilder};
 pub use config::Config;
 pub use core::WithNamed;
 pub use ecs::prelude::{Builder, World};
-pub use game_data::{DataInit, GameData, GameDataBuilder};
+pub use {DataInit, GameData, GameDataBuilder};
```
azriel91: these should be in the `use state::{..}` line

src/app.rs
```diff
@@ -9,12 +9,12 @@ use core::Named;
 use ecs::common::Errors;
 use ecs::prelude::{Component, World};
 use error::{Error, Result};
-use game_data::DataInit;
+use DataInit;
```
azriel91: pedantic: rustfmt should reorder this to the bottom iirc.

super pedantic (you can ignore this one -- it's not done in this repo): I normally split import groups into std, all other crates, and current crate, separated by an empty line.

```rust
use std::marker::PhantomData;

use not_this_crate::TypeA;
use another_crate::TypeB;

use local_module::TypeC;
```

jojolepro:
re: amethyst_terminal Trans dependency
The amethyst_terminal crate is the parser that converts the command string into the concrete types that need to be sent into the EventChannels. For example, the command 
```
$ statechange "MainState"
```

will try to parse the string as ron data, and if it succeeds, it will push a Trans instance through the global EventChannel<Trans>.

I'm not sure how I could make that generic enough, considering all commands have a different way of being parsed, some have side effects, etc.

Most commands will be in the amethyst_command crate, because that crate will be access by most other crates like core::TransformSystem to check for "Teleportation" events for example.
For the Trans command however, it cannot be inside of amethyst_commands, because that will create a cyclic dependency (almost everything will depend on amethyst_commands, so it cannot depend on anything in return).

The StateMachine will be behind a feature gate. I forgot to include it in the list I made yesterday on discord.

azriel91:
okay, so this post is code heavy, but it's how I've done commands in my game ([youtube](https://youtu.be/jpk2MTeWz3w)). It shouldn't force people to use the state machine, since event types are "plug in if you need it".

## Crate: `stdio_view` (probably analogous to `amethyst_commands`)

* Reads `stdin` strings, uses `shell_words` to parse into separate tokens.
* Parses the first token into an `AppEventVariant` to determine which `AppEvent` the tokens correspond to. On success, it sends a tuple: `(AppEventVariant, Vec<String>)` (the tokens) to an `EventChannel<(AppEventVariant, Vec<String>)>`.

Changes if put into Amethyst:

* `StdinSystem` would be generic over top level types `E` and `EVariant`, which would take in `AppEvent` and `AppEventVariant`.

## Crate: `application_event`

* Contains  `AppEvent` and `AppEventVariant`.
* `AppEvent` is an enum over all custom event types, `AppEventVariant` is derived from `AppEvent`, without the fields.

    Example:

    ```rust
    use character_selection_model::CharacterSelectionEvent;
    use map_selection_model::MapSelectionEvent;

    #[derive(Clone, Debug, Display, EnumDiscriminants, From, PartialEq)]
    #[strum_discriminants(
        name(AppEventVariant),
        derive(Display, EnumIter, EnumString),
        strum(serialize_all = "snake_case")
    )]
    pub enum AppEvent {
        /// `character_selection` events.
        CharacterSelection(CharacterSelectionEvent),
        /// `map_selection` events.
        MapSelection(MapSelectionEvent),
    }
    ```

This would be an application specific crate, so it wouldn't go into Amethyst. If I want to have `State` event control, this will include an additional variant `State(StateEvent)` from `use amethyst_state::StateEvent;`, where `StateEvent` carries the information of what to do (e.g. `Pop` or `Switch`).

## Crate: `stdio_spi`

* `StdinMapper` is a trait with the following associated types:

    ```rust
    use structopt::StructOpt;

    use Result;

    /// Maps tokens from stdin to a state specific event.
    pub trait StdinMapper {
        /// Resource needed by the mapper to construct the state specific event.
        ///
        /// Ideally we can have this be the `SystemData` of an ECS system. However, we cannot add
        /// a `Resources: for<'res> SystemData<'res>` trait bound as generic associated types (GATs)
        /// are not yet implemented. See:
        ///
        /// * <https://users.rust-lang.org/t/17444>
        /// * <https://github.com/rust-lang/rust/issues/44265>
        type Resource;
        /// State specific event type that this maps tokens to.
        type Event: Send + Sync + 'static;
        /// Data structure representing the arguments.
        type Args: StructOpt;
        /// Returns the state specific event constructed from stdin tokens.
        ///
        /// # Parameters
        ///
        /// * `tokens`: Tokens received from stdin.
        fn map(resource: &Self::Resource, args: Self::Args) -> Result<Self::Event>;
    }
    ```

    `Args` is a `T: StructOpt` which we can convert the `String` tokens from before we pass it to the `map` function. `Resource` is there because the constructed `AppEvent` can contain fields that are constructed based on an ECS resource.

* This crate also provides a generic `MapperSystem` that reads from `EventChannel<(AppEventVariant, Vec<String>)>` from the `stdio_view` crate. If the variant matches the `AppEventVariant` this system is responsible for, it passes all of the tokens to a `T: StdinMapper` that understands how to turn them into an `AppEvent`, given the `Resource`.

    ```rust
    /// Type to fetch the application event channel.
    type MapperSystemData<'s, SysData> = (
        Read<'s, EventChannel<VariantAndTokens>>,
        Write<'s, EventChannel<AppEvent>>,
        SysData,
    );

    impl<'s, M> System<'s> for MapperSystem<M>
    where
        M: StdinMapper + TypeName,
        M::Resource: Default + Send + Sync + 'static,
        AppEvent: From<M::Event>,
    {
        type SystemData = MapperSystemData<'s, Read<'s, M::Resource>>;

        fn run(&mut self, (variant_channel, mut app_event_channel, resources): Self::SystemData) {
        // ...
        let args = M::Args::from_iter_safe(tokens.iter())?;
        M::map(&resources, args)
        // ... collect each event

        app_event_channel.drain_vec_write(&mut events);
    }
    }
    ```

## Crate: `character_selection_stdio` (or any other crate that supports stdin -> AppEvent)

* Implements the `stdio_spi`.
* The `Args` type:

    ```rust
    #[derive(Clone, Debug, PartialEq, StructOpt)]
    pub enum MapSelectionEventArgs {
        /// Select event.
        #[structopt(name = "select")]
        Select {
            /// Slug of the map or random, e.g. "default/eruption", "random".
            #[structopt(short = "s", long = "selection")]
            selection: String,
        },
    }
    ```

* The `StdinMapper` type:

    ```rust
    impl StdinMapper for MapSelectionEventStdinMapper {
        type Resource = MapAssets;         // Read resource from the `World`, I take a `MapHandle` from it
        type Event = MapSelectionEvent;    // Event to map to
        type Args = MapSelectionEventArgs; // Strong typed arguments, rather than the String tokens

        fn map(map_assets: &MapAssets, args: Self::Args) -> Result<Self::Event> {
            match args {
                MapSelectionEventArgs::Select { selection } => {
                    Self::map_select_event(map_assets, &selection)
                }
            }
        }
    }
    ```

* The bundle, which adds a `MapperSystem<MapSelectionEventStdinMapper>`:

    ```rust
    builder.add(
        MapperSystem::<MapSelectionEventStdinMapper>::new(AppEventVariant::MapSelection),
        &MapperSystem::<MapSelectionEventStdinMapper>::type_name(),
        &[],
    );
    ```

I'm happy to push my code up for the reusable parts (`stdio_spi` should be usable as is, `stdio_view` probably needs a re-write).

Probably should stick this in an RFC issue somewhere so it doesn't get lost.

torkleyy:
RFC, yes, please :+1:

jojolepro:
> StateEvent carries the information of what to do (e.g. Pop or Switch).

That proves why this PR is necessary.

The design you have is effectively what I have on paper. Following rhuagh comments on discord however, the enum containing the different commands will be replaced by something more dynamic. See the opened PR talking about state handle_event made by rhuagh

azriel91:
RFC Issue: #999

jojolepro:
closing following discussions

PR #1003, opened 2018-10-07
Derive Serialize (and other traits) for more types

randomPoison:
I've derived `Serialize` impls for a number of types in order to support displaying them in the editor via [amethyst-editor-sync](https://github.com/randomPoison/amethyst-editor-sync). I've also derived various other commonly-used traits where appropriate. Please let me know if any of these traits shouldn't be added, or if there are any other traits you think I should go ahead and add as part of this PR 🙂

### DIFF
--- a/amethyst_utils/src/removal.rs
+++ b/amethyst_utils/src/removal.rs
@@ -7,7 +7,7 @@ use std::result::Result;
 /// A marker `Component` used to remove entities and clean up your scene.
 /// The generic parameter `I` is the type of id you want to use.
 /// Generally an int or an enum.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Removal<I>
 where
     I: Debug + Clone,
--- a/amethyst_controls/src/components.rs
+++ b/amethyst_controls/src/components.rs
@@ -3,7 +3,7 @@ use amethyst_core::specs::prelude::{Component, Entity, HashMapStorage, NullStora
 
 /// Add this to a camera if you want it to be a fly camera.
 /// You need to add the FlyControlBundle or the required systems for it to work.
-#[derive(Default)]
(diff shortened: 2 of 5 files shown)


amethyst_utils/src/removal.rs
```diff
@@ -7,7 +7,7 @@ use std::result::Result;
 /// A marker `Component` used to remove entities and clean up your scene.
 /// The generic parameter `I` is the type of id you want to use.
 /// Generally an int or an enum.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Removal<I>
```
Aceeri: Unsure of whether this should be put in or not, I guess it doesn't hurt.

randomPoison: Yeah, seems like an odd one to me, but also doesn't seem like it hurts so ¯\\\_(ツ)_/¯

jojolepro: Removal are often in prefabs, so deserialize can make sense.
I would be curious to know if this forces `I` to be Serialize though.

randomPoison:
@jojolepro I'm not sure if implementing `Serialize` for `UiText` makes sense at this point. Reasonably I can use `#[serde(skip)]` on the private members to avoid serializing them, however the `font` field is a `FontHandle` and won't serialize, and I don't think it makes sense to skip that field. I have a more robust solution to this drafted in randomPoison/amethyst-editor-sync#7, but I think it'll take me a little while before I can implement it.

amethyst_utils/src/removal.rs
randomPoison: It doesn't force `I` to be `Serialize`, but it does mean that `Removal<I>` will only be `Serialize` if `I` is.

jojolepro:
Thanks!
bors r+

PR #1007, opened 2018-10-08
Clean up text editing code

Xaeroxe:
A few things were bothering me here so I fixed them.

This should be functionally identical to the prior code, just a bit easier on the eyes.  I also cut out an unnecessary Vec allocation 😉

### DIFF
--- a/amethyst_ui/src/text.rs
+++ b/amethyst_ui/src/text.rs
@@ -7,7 +7,7 @@ use amethyst_core::specs::prelude::{
 use amethyst_core::timing::Time;
 use amethyst_renderer::ScreenDimensions;
 use clipboard::{ClipboardContext, ClipboardProvider};
-use gfx_glyph::PositionedGlyph;
+use gfx_glyph::{Point, PositionedGlyph};
 use hibitset::BitSet;
 use std::cmp::Ordering;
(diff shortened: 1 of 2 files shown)


amethyst_ui/src/text.rs
```diff
@@ -374,98 +340,40 @@ impl<'a> System<'a> for UiKeyboardSystem {
...
                                 .join()
                                 .filter(|&(_, t)| {
                                     t.pixel_x - t.width / 2.0 <= self.mouse_position.0
                                         && t.pixel_x + t.width / 2.0 >= self.mouse_position.0
                                         && t.pixel_y - t.height / 2.0 <= self.mouse_position.1
                                         && t.pixel_y + t.height / 2.0 >= self.mouse_position.1
-                                }).collect::<Vec<_>>();
-                            // In instances of ambiguity we want to select the element with the
-                            // highest Z order, so we need to find the highest Z order value among
-                            // eligible elements.
```
jojolepro: Can we keep the doc?

Xaeroxe: Sure

jojolepro: thanks :)

PR #1008, opened 2018-10-09
fix: Rename functions on PrefabData to be easier to understand

Rhuagh:
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/1008)


Rhuagh:
Well, the function should only add components to a single entity (and/or to resources). It will be called individually for each entity in the prefab.

jojolepro:
Ah right!

PR #1009, opened 2018-10-10
Kill some unwraps

Xaeroxe:
A little progress towards #908.  I might come back and kill some more later but I figured I should PR what I had so far.

### DIFF
--- a/amethyst_animation/src/material.rs
+++ b/amethyst_animation/src/material.rs
@@ -170,27 +170,29 @@ impl AnimationSampling for Material {
         channel: &Self::Channel,
         extra: &Read<MaterialTextureSet>,
     ) -> Self::Primitive {
+        const ERR_MSG: &'static str = "Unable to get requested channel from MaterialTextureSet.";
+
         match *channel {
             MaterialChannel::AlbedoTexture => {
-                MaterialPrimitive::Texture(extra.id(&self.albedo).unwrap())
+                MaterialPrimitive::Texture(extra.id(&self.albedo).expect(ERR_MSG))
             }
             MaterialChannel::EmissionTexture => {
-                MaterialPrimitive::Texture(extra.id(&self.emission).unwrap())
+                MaterialPrimitive::Texture(extra.id(&self.emission).expect(ERR_MSG))
             }
             MaterialChannel::NormalTexture => {
-                MaterialPrimitive::Texture(extra.id(&self.normal).unwrap())
+                MaterialPrimitive::Texture(extra.id(&self.normal).expect(ERR_MSG))
             }
             MaterialChannel::MetallicTexture => {
-                MaterialPrimitive::Texture(extra.id(&self.metallic).unwrap())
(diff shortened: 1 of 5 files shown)


amethyst_animation/src/resources.rs
```diff
@@ -499,7 +499,7 @@ where
             .iter_mut()
             .filter(|t| t.control_id == control_id)
             .filter(|t| t.state != ControlState::Done)
-            .map(|c| (samplers.get(&c.sampler).unwrap(), c))
+            .filter_map(|c| samplers.get(&c.sampler).map(|s| (s, c)))
```
magnonellie: I'm not super familiar with the animation crate, but is it really better to silently fail here?

Rhuagh: No, this should probably fail hard, because if you're referring to a missing sampler you're in deep shit

src/state.rs
```diff
@@ -446,6 +446,7 @@ mod tests {
         let mut world = World::new();
 
         let mut sm = StateMachine::new(State1(7));
+        // Unwrap here is fine because start can only fail when there are no states in the machine.
```
torkleyy: This is a test :sweat_smile: no need to justify unwraps here ;)

amethyst_animation/src/material.rs
```diff
@@ -170,27 +170,29 @@ impl AnimationSampling for Material {
         channel: &Self::Channel,
         extra: &Read<MaterialTextureSet>,
     ) -> Self::Primitive {
+        const ERR_MSG: &'static str = "Unable to get requested channel from MaterialTextureSet.";
+
         match *channel {
             MaterialChannel::AlbedoTexture => {
-                MaterialPrimitive::Texture(extra.id(&self.albedo).unwrap())
+                MaterialPrimitive::Texture(extra.id(&self.albedo).expect(ERR_MSG))
```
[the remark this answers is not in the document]
Xaeroxe: Would require altering the trait interface.  Not sure if that's desirable or not.

src/state.rs
Xaeroxe: Eh I already wrote it, not much harm in leaving it right?

amethyst_animation/src/resources.rs
Xaeroxe: @Rhuagh should I also unwrap [here](https://github.com/amethyst/amethyst/blob/master/amethyst_animation/src/systems/sampling.rs#L65)?

amethyst_animation/src/material.rs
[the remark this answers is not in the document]
Rhuagh: So this would only be triggered in the following scenario:

* You start at texture animation for a specific texture, say albedo
* The animation have a sampler for the albedo for a specific entity
* Said entity does not already have a albedo texture defined in its material

If we want to change the return value of this function, I think it should be changed to an Option, not Result.

src/state.rs
torkleyy: No :)

PR #1010, opened 2018-10-10
Nix gitter refs

Xaeroxe:
A few of these were still straggling.


Xaeroxe:
Trivial and noncontroversial so

bors r+

PR #1014, opened 2018-10-11
Tiny Pong tutorial improvements

Adrijaned:
There has been some text remaining from before a block of code in chapter 4.1 was changed, plus some missing whitespace from the same commit.

### DIFF
--- a/book/src/pong-tutorial/pong-tutorial-01.md
+++ b/book/src/pong-tutorial/pong-tutorial-01.md
@@ -49,10 +49,12 @@ impl<'a, 'b> SimpleState<'a, 'b> for Pong {
 
 The `SimpleState` already implements a bunch of stuff for us, like the `update` 
 and `handle_event` methods that you would have to implement yourself were you 
-using just a simple `State`. In particular, the default implementation for
+using just a regular `State`. In particular, the default implementation for
 `handle_event` returns `Trans::Quit` when a close signal is received
 from your operating system, like when you press the close button in your graphical
-environment. This allows the application to quit as needed.
+environment. This allows the application to quit as needed. The default 
+implementation for `update` then just returns `Trans::None`, signifying that
+nothing is supposed to happen.
 
 Now that we know we can quit, let's add some code to actually get things
 started! We'll start with our `main` function, and we'll have it return a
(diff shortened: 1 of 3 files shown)


book/src/pong-tutorial/pong-tutorial-01.md
```diff
@@ -44,19 +43,16 @@ just implement two methods:
...
-transitions. In this case, we're watching for the Escape keycode, and the
-`CloseRequested` event from the Window. If we receive it, we'll return
-`Trans::Quit` which will be used to clean up the `State` and close the application.
-All other keyboard input is ignored for now.
-
-The `update` method is executed after events have happened.  In this instance
-we're just using it to execute our dispatcher.  More on that later.
+The `SimpleState` already implements a bunch of stuff for us, like the `update` 
+and `handle_event` methods that you would have to implement yourself were you 
+using just a simple `State`. In particular, the default implementation for
```
Rhuagh: change `simple State` to `normal State` or something. I feel it might be misunderstood as `simple State` is fairly close to `SimpleState`

book/src/pong-tutorial/pong-tutorial-01.md
```diff
@@ -44,19 +43,16 @@ just implement two methods:
...
 ```
 
-The `handle_event` method is executed for every event before updating, and it's
-used to react to events. It returns a `Trans`, which is an enum of state machine
-transitions. In this case, we're watching for the Escape keycode, and the
-`CloseRequested` event from the Window. If we receive it, we'll return
-`Trans::Quit` which will be used to clean up the `State` and close the application.
-All other keyboard input is ignored for now.
-
-The `update` method is executed after events have happened.  In this instance
```
torkleyy: Why did you remove this?

Moxinilian: I think it is because it is not actually used and so feels out of place.
I'd suggest instead introducing it by mentioning that the following method can be implemented, and what they do.

Adrijaned: This text was written for the `latest` version of the book, and as you can see on https://www.amethyst.rs/book/latest/pong-tutorial/pong-tutorial-01.html, the impl is filled with the mentioned functions there. But in the `master` version, the text just gives the impression that something is missing from the example. Also, I previously understood that the comment from  @torkleyy aims just at the `update` method info having no alternative in my suggested version, so I changed that in the latest commit. If you feel like that the info is unnecessary/confusing there, just let me know and I'll remove it again.

jojolepro:
bors r+

timedout

PR #1016, opened 2018-10-11
Implement configurable line width for debug lines

NCrashed:
Problem: debug lines are too thin for orthographic projection with large top/bottom/left/right ranges (usually used for 2D setups).

Solution:
* Use uniform for line width.
* Pass line width via resource.

Also added the usage of the new feature to example.

How to test:
* Change `DebugLinesParams` in `debug_lines` example and run

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -101,6 +101,7 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 [#994]: https://github.com/amethyst/amethyst/pull/994
 [#1006]: https://github.com/amethyst/amethyst/pull/1006
 [#1008]: https://github.com/amethyst/amethyst/pull/1008
+[#1016]: https://github.com/amethyst/amethyst/pull/1016
 [winit_017]: https://github.com/tomaka/winit/blob/master/CHANGELOG.md#version-0172-2018-08-19
 [glutin_018]: https://github.com/tomaka/glutin/blob/master/CHANGELOG.md#version-0180-2018-08-03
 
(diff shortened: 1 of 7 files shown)


examples/debug_lines/main.rs
```diff
@@ -2,19 +2,16 @@
...
-        cgmath::{Deg, Point3, Vector3},
-        transform::{GlobalTransform, Transform, TransformBundle},
-    },
-    ecs::{Read, System, Write},
-    input::InputBundle,
-    prelude::*,
-    renderer::*,
-    utils::application_root_dir,
-};
+use amethyst::controls::{FlyControlBundle, FlyControlTag};
```
magnonellie: Did rustfmt do this? It was literally changed a few days ago to the format you removed :stuck_out_tongue:

docs/CHANGELOG.md
```diff
@@ -30,7 +30,7 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 * Support for multiline text. ([#965])
 * Support for logging to file, toggle for logging to stdout. ([#976], [#994])
 * Added `shadow_update()` and `shadow_fixed_update()` to the `State` trait. ([#1006])
-
+* Added configurable width for debug lines ([#1016])
```
magnonellie: You need to also add a link definition below

examples/debug_lines/main.rs
NCrashed: Yep, it is rustmft `rustfmt 0.99.1-stable (da17b68 2018-08-04)`

magnonellie: Well then. I guess @jojolepro was right after all. :sweat_smile:

NCrashed: So, I can revert the particular changes.

torkleyy: Why are they doing this?

docs/CHANGELOG.md
NCrashed: Done

NCrashed:
There is a problem that all lines from components and resource are merged in single mesh. So, all lines are forced to have the same width now. Possible solution is construct several meshes, one per width value.

jojolepro:
Ah that makes sense, you can keep this as-is for now until the new renderer comes and we can have batching (to not kill fps while using debug lines in the mean time)

Rhuagh:
Bors r+

PR #1017, opened 2018-10-12
Remove unwrap() in a few places

distransient:
Let me know if I should put these in smaller PR's or larger ones.

Part of #908

### DIFF
--- a/amethyst_animation/src/skinning/systems.rs
+++ b/amethyst_animation/src/skinning/systems.rs
@@ -40,11 +40,17 @@ impl<'a> System<'a> for VertexSkinningSystem {
 
     fn run(&mut self, (joints, global_transforms, mut skins, mut matrices): Self::SystemData) {
         self.updated.clear();
-        global_transforms
-            .populate_modified(&mut self.updated_id.as_mut().unwrap(), &mut self.updated);
-        global_transforms
-            .populate_inserted(&mut self.inserted_id.as_mut().unwrap(), &mut self.updated);
+
+        global_transforms.populate_modified(
+            &mut self.updated_id.as_mut().expect("VertexSkinningSystem missing updated_id."), 
+            &mut self.updated);
+
+        global_transforms.populate_inserted(
+            &mut self.inserted_id.as_mut().expect("VertexSkinningSystem missing inserted_id."), 
+            &mut self.updated);
+
         self.updated_skins.clear();
+
         for (_, joint) in (&self.updated, &joints).join() {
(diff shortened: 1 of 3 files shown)


amethyst_animation/src/skinning/systems.rs
```diff
@@ -40,11 +40,19 @@ impl<'a> System<'a> for VertexSkinningSystem {
...
     fn run(&mut self, (joints, global_transforms, mut skins, mut matrices): Self::SystemData) {
         self.updated.clear();
-        global_transforms
-            .populate_modified(&mut self.updated_id.as_mut().unwrap(), &mut self.updated);
-        global_transforms
-            .populate_inserted(&mut self.inserted_id.as_mut().unwrap(), &mut self.updated);
+
+        match self.updated_id.as_mut() {
+            Some(x) => { global_transforms.populate_modified(x, &mut self.updated) },
+            None => { warn!("VertexSkinningSystem missing updated_id.") },
```
Moxinilian: Is there a realistic scenario where this would happen and not be a bug?

amethyst_utils/src/app_root_dir.rs
```diff
@@ -14,7 +14,7 @@ pub fn application_root_dir() -> String {
             path.parent()
                 .expect("Failed to get parent directory of the executable.")
                 .to_str()
-                .unwrap(),
+                .expect("Failed to get valid string from path."),
```
Moxinilian: "[...] from path to the executable."

amethyst_animation/src/skinning/systems.rs
distransient: Only if Setup() is not ran or if the value is modified to None but I don't think it hurts, I guess.

distransient: I would assume of course that Setup() is ran for every system before it's Run()

amethyst_utils/src/app_root_dir.rs
distransient: Pushed.

amethyst_animation/src/skinning/systems.rs
distransient: Found the diff, I guess this is expected. https://github.com/amethyst/amethyst/commit/d949f6c603bb8c3efbc3e12ee93d26fd2b1ff734#diff-91175d8ffab11be5a6cef3d04164cf82R48

Fixed.

distransient:
Murderously squashed.

PR #1022, opened 2018-10-12
Make release_cursor call release_cursor, not grab_cursor

karyon:
i guess? (spotted randomly while browsing the code, haven't tested anything)

### DIFF
--- a/amethyst_renderer/src/mouse.rs
+++ b/amethyst_renderer/src/mouse.rs
@@ -11,7 +11,7 @@ pub fn hide_cursor(msg: &mut WindowMessages) {
 /// Set the cursor back to normal/visible.
 pub fn release_cursor(msg: &mut WindowMessages) {
     msg.send_command(move |win| {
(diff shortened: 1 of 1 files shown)


karyon:
yeah, there is no release_cursor in winit, and I missed the bool parameter.

now that we've both been appropriately embarrassed, I think i'll close this PR, call it a day and go to bed.

amethyst_renderer/src/mouse.rs
```diff
@@ -11,7 +11,7 @@ pub fn hide_cursor(msg: &mut WindowMessages) {
 /// Set the cursor back to normal/visible.
 pub fn release_cursor(msg: &mut WindowMessages) {
     msg.send_command(move |win| {
-        if let Err(err) = win.grab_cursor(false) {
```
Xaeroxe: You'll notice this passes in `false` and the grab_cursor function passes in `true.

PR #1028, opened 2018-10-14
Reduce Travis workload by only building with bors

torkleyy:


torkleyy:
Well, with all the activity we're having on the repo and the slow builds I think it makes sense if you're building PRs yourself. `bors try` can still be used to do that before the approval.

torkleyy:
I'm not 100% sure either, but with the changed situation I thought it makes sense because it speeds up the merging with bors (no other PR will block the build).

Moxinilian:
I understand the rationale, but I don't feel it's worth it.
Besides faster bors, why do we need this?
I like fast CI because I can get feedback fast, but bors being a bit slow is okay to me. To me CI speed is the most useful when you are still (potentially) working on the PR, not when you fired (and forgot) the merging process.

torkleyy:
Convinced.

PR #1037, opened 2018-10-15
[wip] 3d object picker

plredmond:
Goal:
* Enable picking 3d objects in the scene with the mouse.

Status:
* Working. WIP. Asking for help. See questions below.
* Try with `cargo run --example picker`
* hover

Approach:
* Add a new component to specify an entity's bounding box.
* Add three systems to assist in event emitting when hovering or clicking entities.

Summary of changes:
* Add new `amethyst_physics` crate per @jojolepro because of the new `collision` dependency.
* Add `Pickable<T>` component to specify collision hull.
* Add `MouseRaySys` to generate a world-coordinates ray from the camera point which captures the location of the mouse. Output is written to the `MouseRay` resource.
* Add `PickSys` to intersect the `MouseRay` against every entity with `Pickable` and write the nearest one to the `Picked` resource.
* Add `PickEventSys` adapted from `amethyst_ui::UiMouseSystem` to observe the `Picked` resource and emit hover and click events on `EventChannel<UiEvent>`

Questions:
  1. Should I revert 570bf65? That commit makes the `Pickable` component generic over implementations of `ContinousTransformed`, so that it can be used for [Primitive3](https://docs.rs/collision/0.18.0/collision/primitive/enum.Primitive3.html) and things like [Plane](https://docs.rs/collision/0.18.0/collision/struct.Plane.html). However, the commit introduced a `'static` on the lifetime of `T` in `Pickable<T>`'s implementations of other traits.
  1. Should I try optimizing `MouseRaySys` by caching the pre-multiplied inverted matrix?
  1. Should I add a new `PickBundle` to make using these new systems easier? This was made harder by the addition of `T` in 570bf65.
  1. Any concern with the `TODO`, `FIXME` and `XXX` in the code?
  1. Overall, any other suggested improvements?


amethyst_physics/src/picking/mouseray.rs
```diff
@@ -0,0 +1,109 @@
...
+use amethyst_core::specs::prelude::*;
+use amethyst_core::transform::Transform;
+use amethyst_input::InputHandler;
+use amethyst_renderer::ActiveCamera;
+use amethyst_renderer::{Camera, ScreenDimensions};
+
+use collision::Ray3;
+
+/// Resource which contains a ray from the `ActiveCamera` to the mouse in world space coordinates.
+pub struct MouseRay {
```
Rhuagh: Use `collision::Ray3` directly instead?

amethyst_physics/src/picking/mouseray.rs
```diff
@@ -0,0 +1,109 @@
...
+    //   unproject
+    //   overwrite z
+    v = proj.invert().expect("invert projection matrix") * v;
+    v.z = -1.; // XXX: in Camera forward is -z
+
+    // TODO: try doing Z before proj.invert(), and then cache the combination of proj*view
+
+    // World
+    //   transform & normalize
+    v = view * v; // XXX: why is this just the normal view matrix??
```
Rhuagh: It's not, because `view` here is actually the world `Transform` for the `Camera`, the actual view matrix is the inverse of this matrix.

Moxinilian:
Are you aware of the planned move from cgmaths to nalgebra?
The purpose of this initiative is to open amethyst_physics, a wrapper around nphysics.

torkleyy:
Could we call it amethyst_nphysics instead?

Rhuagh:
It's not using nphysics yet though. And don't spend too much time on this now, since nphysics will soon be a thing<tm>

plredmond:
Hey all, thanks for checking this out. Did anyone have any response to the
questions I posed in the PR description?

@moxinilian yes, @jojolepro mentioned there was a planned move to
`nalgebra`. I guess that also means there's a move away from `collision`,
so I'm trying not to explose much of that in this.

Given than there's already a
https://github.com/amethyst/amethyst-rhusics/pull/36, what should I do with
this?

Telzhaak:
@plredmond 
Rhusics is likely going to be dropped once the merge to nalgebra is done, since rhusics is written with cgmath.

I don't know if a ray-picker is part of nphysics, so I'll just say, that it would be worthwhile to finish this PR.

Rhuagh:
I will provide pick rays when nphysics is integrated

plredmond:
Ok, it sounds then like this pr is redundant and can be closed?

Let me know where else I might be able to help. :)

torkleyy:
Thank you @plredmond for your work! Unfortunately the interaction with physics is very unclear here.

I would like to propose we close this PR for now. I think this is a useful feature, but there's no way forward for this currently. I think it makes sense to reopen the discussion when the physics situation is clear @plredmond.

plredmond:
No problem! :) Keep up the good work.

PR #1038, opened 2018-10-15
Added "saveload" feature to amethyst

Telzhaak:
The feature "saveload" enables users to use `specs/saveload` module to serialize and deserialize entities and components.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,9 @@ sdl_controller = [
 json = [
     "amethyst_assets/json"
 ]
+saveload = [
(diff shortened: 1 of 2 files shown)


Telzhaak:
Would it make sense to also export `serde` and `serde_derive` from amethyst, just like `specs`?

torkleyy:
I'm personally not a big fan of the reexport structure, so I'd probanly say no to exporting serde.

LucioFranco:
bors r=Rhuagh,jojolepro

PR #1042, opened 2018-10-15
Fix UI caching

### LINKED ISSUE #1041: UI entities get drawn multiple times
I noticed that the FPS in the UI examples on master is constantly decreasing and tracked it down to the `cached_draw_order.cache` in the UI render pass constantly growing with new entities (or maybe the same added multiple times?), resulting in the UI elements being drawn multiple times in an ever increasing number.

Xaeroxe:
Fixes #1041


Xaeroxe:
@jojolepro Other passes won't have this, the bug arose because of how I chose to manage draw order in this pass.  The draw order cache is unique to the UI crate.

Issue #1021, opened 2018-10-12
Provide And/Or logic for input bindings


Xaeroxe:
Currently all multi-bindings use OR logic.  We should extend this provide AND logic and possibly combinations of AND/OR logic.

mh84:
Some solution could be providing the current parts of each multi-binding to be a combination of multiple bindings AND-combined. May i can work on a prove of concept for this.

PR #1043, opened 2018-10-16
[RFC] combinations of buttons triggering actions

mh84:
This will take care of  #1021 

Current state of progress is that single button action bindings are replaced with a `vec!` of buttons, each of them needed to be pressed to trigger that action. Using this will actually break current action bindings in code or input.ron, because these need to be convert to arrays.


mh84:
As this is planned for 0.10 i guess an API change on input bindings is okay, the needed changes on hard coded input bindings / input.ron-files only require some minimal / minor changes, so i guess we will be fine the way it is actually.

But i'm open for suggestions how/what to change to make this more handy.

jojolepro:
This looks pretty good, I'm not sure what you want me to comment me on ;)

Moxinilian:
I think @Xaeroxe might be interested in reviewing that.
I have conflicted feelings: this is a useful feature, but I'm afraid it could create an overhead in usability.
What do you think?

amethyst_input/src/bindings.rs
```diff
@@ -15,7 +15,7 @@ where
     AC: Hash + Eq,
 {
     pub(super) axes: HashMap<AX, Axis>,
-    pub(super) actions: HashMap<AC, SmallVec<[Button; 4]>>,
+    pub(super) actions: HashMap<AC, SmallVec<[Vec<Button>; 4]>>,
```
jjedelsky:
```suggestion
    pub(super) actions: HashMap<AC, SmallVec<[SmallVec<[Button; 2]>; 4]>>,
```
I believe that most actions will still only need a few buttons (mostly just 1 button)

amethyst_input/src/input_handler.rs
```diff
@@ -121,13 +118,13 @@ where
                                 .iter()
                                 .cloned(),
                         );
-                        for (k, v) in self.bindings.actions.iter() {
-                            for &button in v {
-                                if Button::Key(key_code) == button {
-                                    event_handler.single_write(ActionReleased(k.clone()));
+                        for (action, combinations) in self.bindings.actions.iter() {
+                            for combination in combinations {
+                                if combination.contains(&Button::Key(key_code)) {
```
Xaeroxe: I'm not super certain this logic works.  I think if you have for example buttons A and B, each with their own action, and then they can also combo with A + B you'll get a release event for the combo if A is pressed and released even if B was never touched.  If I've misread what's happening here please let me know.

mh84: You missed the next line, the action would be released in that place, so for an action bind to a+b, this will be released when either a or b will be released.

mh84: Okay now i misread your comment, i will have a look at this later.

Xaeroxe: I just thought of a pretty simple way to resolve this.  Only send this release event if every other button in the combo is currently pressed.

mh84: Thats pretty much what cames to my mind too while resolving the merge issues and integrating jjedelskys improvements :)

Xaeroxe:
@Moxinilian brings up a good point that this will require a pretty clear explanation in documentation as to how it works.

amethyst_input/src/event.rs
```diff
@@ -86,9 +86,15 @@ pub enum InputEvent<T> {
...
-    /// The associated action had one of its keys pressed.
+    /// The associated action had any related button or combination pressed.
+    ///
+    /// If a combination is bound to an action, it will be pressed
+    /// if all buttons within are pressed.
     ActionPressed(T),
-    /// The associated action had one of its keys released.
+    /// The associated action had any related button or combination released.
+    ///
+    /// If a combiantion is bound to an action, it will be released
```
Xaeroxe: Spelling error

s/combiantion/combination/g

amethyst_input/src/bindings.rs
```diff
@@ -63,19 +65,19 @@ where
         self.axes.keys().cloned().collect::<Vec<AX>>()
     }
 
-    /// Add a button to an action.
+    /// Add a button or button combination to an action.
     ///
-    /// This will insert a new binding between this action and the button.
-    pub fn insert_action_binding<A>(&mut self, id: A, binding: Button)
+    /// This will insert a new binding between this action and the button(s).
+    pub fn insert_action_binding<A>(&mut self, id: A, binding: SmallVec<[Button; 2]>)
```
Xaeroxe: I'm not super sure I want to expose SmallVec here as part of our API, it should just be an implementation detail.  I noticed at one point this was implemented with iterators, why did we change that?

torkleyy: I don't think `SmallVec` should be used here - http://troubles.md/posts/improving-smallvec/

mh84: I'm not sure when or where it got lost, but i restored that too :)

Moxinilian:
What ended up being the consensus regarding SmallVec?

Moxinilian:
Oops sorry

mh84:
We don't expose SmallVec through `insert_action_binding<A, B: IntoIterator<Item = Button>>(&mut self, id: A, binding: B)` as mentioned earlier

torkleyy:
Yeah, I commented the same on the original implementation (where it is still used), but it was ignored back then.

torkleyy:
Can be another PR, can be done here, I'll let you decide. This PR makes sense as it is.

torkleyy:
Thanks @mh84!
bors r=Xaeroxe,torkleyy

PR #1045, opened 2018-10-18
Clippy suggestion cleanup.

0ndorio:
I spent some time to fix most of the clippy suggestions that do not interfere with the function signatures or internal code structuring. As this is neither an API change nor a new feature I skipped the update of `docs/CHANGELOG.md`.

Feel free to nitpick If this includes any unwanted changes or if I missed something important. I will try to react as fast as possible.

I would also suggest to review the two commits independently as the second one only contains automatically applied changes made by rustfmt (`rustc 1.29.2 (17a9dc751 2018-10-05)
`, `rustfmt 0.99.1-stable (da17b68 2018-08-04)`).

### DIFF
--- a/amethyst_animation/src/resources.rs
+++ b/amethyst_animation/src/resources.rs
@@ -134,7 +134,8 @@ where
 /// Defines the hierarchy of nodes that a single animation can control.
 /// Attached to the root entity that an animation can be defined for.
 /// Only required for animations which target more than a single node or entity.
-#[derive(Debug, Clone)]
+#[derive(Derivative, Debug, Clone)]
+#[derivative(Default(bound = ""))]
 pub struct AnimationHierarchy<T> {
     /// A mapping between indices and entities
     pub nodes: FnvHashMap<usize, Entity>,
@@ -155,10 +156,7 @@ where
 {
     /// Create a new hierarchy
     pub fn new() -> Self {
-        AnimationHierarchy {
-            nodes: FnvHashMap::default(),
-            m: marker::PhantomData,
-        }
+        Self::default()
     }
 
     /// Create a new hierarchy containing a single given entity
@@ -807,7 +805,7 @@ where
         rate_multiplier: f32,
         command: AnimationCommand<T>,
     ) {
-        if let Some(_) = self.animations.iter().find(|a| a.0 == id) {
+        if self.animations.iter().any(|a| a.0 == id) {
             return;
         }
         self.animations.push((
@@ -833,7 +831,7 @@ where
         wait_for: I,
         wait_deferred_for: DeferStartRelation,
     ) {
-        if let Some(_) = self.animations.iter().find(|a| a.0 == id) {
+        if self.animations.iter().any(|a| a.0 == id) {
             return;
         }
         self.deferred_animations.push(DeferredStart {
@@ -851,7 +849,7 @@ where
 
     /// Insert an animation directly
     pub fn insert(&mut self, id: I, control: AnimationControl<T>) {
-        if let Some(_) = self.animations.iter().find(|a| a.0 == id) {
+        if self.animations.iter().any(|a| a.0 == id) {
             return;
         }
         self.animations.push((id, control));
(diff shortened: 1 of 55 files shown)


src/app.rs
```diff
@@ -422,6 +429,7 @@ impl<S, E, X> ApplicationBuilder<S, E, X> {
     /// game.run();
     /// ~~~
 
+    // TODO: Unused lifetime. Is it required?
```
Xaeroxe: I'd venture a "no"

Rhuagh: No, it's a remnant of old code that's been refactored atleast three times :)

amethyst_animation/src/systems/control.rs
```diff
@@ -439,13 +439,15 @@ where
...
+    let still_loading = animation
         .nodes
         .iter()
         .any(|&(ref node_index, _, ref sampler_handle)| {
             !hierarchy.nodes.contains_key(node_index)
                 || sampler_storage.get(sampler_handle).is_none()
-        }) {
+        });
+
+    if still_loading {
```
Rhuagh: What's the motivation for this change?

0ndorio: The idea was to take the "complex" if statement and introduce an intermediate variable name that represents the intention of the statement. I assume that the comment above the statement could be seen as doing the same as the its only the check for a precondition.

amethyst_renderer/src/pipe/pass.rs
```diff
@@ -16,8 +16,10 @@ pub trait Pass: for<'a> PassData<'a> {
     /// The pass is given an opportunity to compile shaders and store them in an `Effect`
     /// which is then passed to the pass in `apply`.
     fn compile(&mut self, effect: NewEffect) -> Result<Effect>;
+
     /// Called whenever the renderer is ready to apply the pass.  Feed commands into the
     /// encoder here.
+    // TODO: Not one of the implementations uses lifetime 'b. Is it required?
```
0ndorio: What about this one?

amethyst_animation/src/resources.rs
```diff
@@ -134,7 +134,7 @@ where
 /// Defines the hierarchy of nodes that a single animation can control.
 /// Attached to the root entity that an animation can be defined for.
 /// Only required for animations which target more than a single node or entity.
-#[derive(Debug, Clone)]
+#[derive(Debug, Default, Clone)]
```
torkleyy: * should use the default in `new`
* should use `derivative` as `derive` only implements it for `T: Default`

0ndorio: thx! totally missed that.

amethyst_animation/src/systems/control.rs
0ndorio: Restored the original version.

amethyst_renderer/src/pipe/pass.rs
Rhuagh: This one is needed, because the pass data does not have the same lifetime as the pass.

0ndorio: Okay. Was just curious because all implementations of `Pass` I could discover either utilized `'a` a second time (e.g. `Pass for DrawSprite`) or the anonymous lifetime  (e.g. `Pass for DrawUi`) but never `'b` for `PassData`.

0ndorio:
Should I go an rebase on current master as the branch was updated in the mean time?

amethyst_renderer/src/pipe/effect/mod.rs
```diff
@@ -94,8 +90,11 @@ impl Effect {
         match self.const_bufs.get(name.as_ref()) {
             Some(i) => {
                 let raw = &self.data.const_bufs[*i];
-                enc.update_buffer::<T>(unsafe { mem::transmute(raw) }, &data[..], 0)
-                    .expect("Failed to update buffer (TODO: replace expect)");
+                enc.update_buffer::<T>(
+                    unsafe { &*(raw as *const RawBuffer<_> as *const Buffer<_, _>) },
```
jojolepro: I can't read this, so I'd like to know if this was reviewed by the other reviewers.

torkleyy: Yes, it was ;)

magnonellie:
@0ndorio Sorry for pinging you like this, but I couldn't find you on Discord :sweat_smile: 

How did you get clippy to run on Amethyst? For me it's just throwing errors on dependencies using the rustup install method. I would like to integrate clippy into our CI at least in some capacity so being able to get it running locally is a pretty good first step I think. :stuck_out_tongue:

PR #1051, opened 2018-10-19
[Feature] Create HiddenUi via prefab field

Telzhaak:
Aims to make it possible to hide parts or the complete Ui of a prefab,
both via the prefab, and the `UiCreator::create()` function.

Current usage:
```ron
transform:(
// snip
hidden: true,
)
```
This adds a `HiddenPropagate` component to this entity. When using the `HideHierarchySystem` this allows hiding and unhiding of every entity below it at run-time (which would not be possible with only a `Hidden` component. It is still advised to set this variable for EVERY child entity of the entity you hide, otherwise the child entities might be drawn once before the `HideHierarchySystem` kicks in (depends on the order of dependencies)

The `create` function is work in progress.


Telzhaak:
I admit defeat.
No idea how to add a component from the `Loader.load()` function...

It works from inside a prefab at least.

jojolepro:
> No idea how to add a component from the Loader.load() function...
You are not supposed to. This returns a handle to the stored asset in AssetStorage<T>

Telzhaak:
I may not be supposed to, but i'd still want to be able to load Ui in a hidden state and only make it visible once it is needed.

For now the prefab-method will suffice

amethyst_ui/src/prefab.rs
```diff
@@ -83,6 +85,12 @@ impl UiTransformBuilder {
         self
     }
 
+    /// Set to event transparent
```
Moxinilian: What does that mean?

Moxinilian:
That's what prefabs are for by the way, the loader only loads assets with no context, just plain data, when a prefab loads assets and context in the world.

amethyst_ui/src/prefab.rs
Telzhaak: I have no idea

Telzhaak:
No idea why rustfmt is complaining though.

I'm on latest stable (just updated) and nothing is changed by running `cargo fmt`

Issue #1055, opened 2018-10-20
Fetch used in the book


Rhuagh:
I the appendix, Fetch is still being used, should use Read/ReadExpect instead. The example have correct code.

LucioFranco:
@Rhuagh should we put this under `0.9` seems easy to fix and should go out with that release.

PR #1056, opened 2018-10-20
Fetch removed that was used in book

highway900:
I noticed this the other day and then someone bumped into it on discord. Easy win I suppose :)

fixes (I hope) #1055


jojolepro:
I'll wait a bit more just in case. Someone else can start the merge after approving

PR #1058, opened 2018-10-21
Lifetime documentation

jojolepro:
Opened to get feedback.
Not planned for the 0.9 release.

1) I might be very wrong about my explanations. If so, correct me.
2) Also, if this isn't something we want in the book, I don't mind closing the issue.
3) If you have ideas of topics you want covered (or where covered on discord), let me know.

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -29,6 +29,7 @@
     * [`SpriteRender` Component](./sprites/sprite_render_component.md)
     * [Orthographic Camera](./sprites/orthographic_camera.md)
 * [Prefabs](./prefab.md)
+* [Advanced Concepts](./advanced.md)
 * [Glossary](./glossary.md)
 * [Appendix A: Config Files](./appendices/a_config_files.md)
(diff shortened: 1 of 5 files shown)


Temeez:
Tried to leave a "single comment" on the files changed but it doesn't work so I'll write here.

> Why are the lifetimes present two times?

I have no idea what I'm reading past this. This would be easier to understand if you first answer the question and then explain the answer.

Rhuagh:
I feel that this doesn't belong in the Amethyst book. I don't think we should have to explain everything, a better solution would be to point users in the direction of good documentation for things, such as the rust and specs books.

Rhuagh:
I feel that this doesn't belong in the Amethyst book. I don't think we should have to explain everything, a better solution would be to point users in the direction of good documentation for things, such as the rust and specs books.

Rhuagh:
I feel that this doesn't belong in the Amethyst book. I don't think we should have to explain everything, a better solution would be to point users in the direction of good documentation for things, such as the rust and specs books.

PR #1059, opened 2018-10-21
Add README for amethyst_network

LucioFranco:


jojolepro:
Manual merge because bors is way too slow at the moment and the pull requests count is going up.

PR #1064, opened 2018-10-22
Fix the pong example's scoreboard

Moxinilian:
This brings back the pong scoreboard.
I also fixed the rendering problems with larger score, and capped them at 999 to avoid overlapping.


Moxinilian:
bors r=Jojolepro,Moxinilian

PR #1069, opened 2018-10-25
Trans Queue

jojolepro:
You can add a Trans to the queue and it will be executed before the next State::update.
Useful for event-driven programming (futures, ui, network, error handling).


src/app.rs
```diff
@@ -289,6 +291,19 @@ where
             states.stop(StateData::new(world, &mut self.data));
         }
 
+        // Read the Trans queue and apply changes.
+        {
+            let trans = {
```
azriel91: why drain and collect and then `into_iter()` later though? why not just grab the `Write<VecDeque<_>>` and `into_iter()` directly?

src/app.rs
```diff
@@ -1,5 +1,6 @@
 //! The core engine framework.
 
+use state::Trans;
```
azriel91: can merge into line 17

src/app.rs
```diff
@@ -123,7 +125,7 @@ pub struct CoreApplication<'a, T, E = StateEvent, R = StateEventReader> {
 /// [log]: https://crates.io/crates/log
 pub type Application<'a, T> = CoreApplication<'a, T, StateEvent, StateEventReader>;
 
-impl<'a, T, E, R> CoreApplication<'a, T, E, R>
+impl<'a, T: 'static, E, R> CoreApplication<'a, T, E, R>
```
azriel91: prefer adding predicates to the `where` clause instead of in the declaration

src/state.rs
```diff
@@ -54,13 +54,14 @@ pub enum Trans<T, E> {
     /// if there are none.
     Pop,
     /// Pause the active state and push a new state onto the stack.
-    Push(Box<State<T, E>>),
+    Push(Box<State<T, E> + Send + Sync>),
```
torkleyy: That would disallow dispatchers in states.

src/state.rs
```diff
@@ -413,7 +414,7 @@ impl<'a, T, E: Send + Sync + 'static> StateMachine<'a, T, E> {
 
     /// Performs a state transition, if requested by either update() or
```
torkleyy: Docs are outdated now.

src/app.rs
jojolepro: Because I need to empty the queue and read the collected elements. If it wasn't from the borrow issues, I would have write_storage().drain(..).into_iter()

src/app.rs
```diff
@@ -289,6 +291,19 @@ where
...
         }
 
+        // Read the Trans queue and apply changes.
+        {
+            let trans = {
+                let mut v = self.world.write_resource::<TransQueue<T, E>>();
+                let x = v.drain(..).collect::<VecDeque<_>>();
+                x
+            };
+            let states = &mut self.states;
```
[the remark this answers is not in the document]
jojolepro: Its just the way it was written before.
We have some people who like to separate the borrow and the usage.
states.transition at line 303 requires a mutable reference, so here it is create before the loop instead of inside (doesn't really matter, the compiler would have optimized it to be outside of the loop anyway)

examples/state_dispatcher/main.rs
```diff
@@ -0,0 +1,49 @@
...
+use amethyst::prelude::*;
+use amethyst::Error;
+
+struct StateA;
+
+impl SimpleState<'static, 'static> for StateA {
+    fn update(&mut self, data: &mut StateData<GameData>) -> SimpleTrans<'static, 'static> {
+        println!("StateA::update()");
+        // Shows how to push a `Trans` through the event queue.
+        // If you do use TransQueue, you will be forced to use the 'static lifetime on your states.
```
azriel91: I'm curious, maybe if there is a `T: 'a` bound in `CoreApplication`, it might relax this constraint; not sure about the second lifetime though,

src/app.rs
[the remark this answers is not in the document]
Rhuagh: No, the reason it was done that way is to tell the compiler it's safe to have mutable references to multiple parts of self. If you don't do this you will get a borrow error between self.states and self.world.

src/state.rs
Rhuagh: Indeed.

src/app.rs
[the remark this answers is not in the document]
jojolepro: ah okay

jojolepro:
Will update this either tonight or tomorow

examples/state_dispatcher/main.rs
jojolepro:
```rs
error[E0310]: the parameter type `T` may not live long enough                                                                                              
   --> src/app.rs:307:40                                                                                                                                   
    |                                                                                                                                                      
137 | impl<'a, T, E, R> CoreApplication<'a, T, E, R>                                                                                                       
    |          - help: consider adding an explicit lifetime bound `T: 'static`...                                                                          
...                                                                                                                                                        
307 |                 let mut v = self.world.write_resource::<TransQueue<T, E>>();                                                                         
    |                                        ^^^^^^^^^^^^^^                                                                                                
    |                                                                                                                                                      
note: ...so that the type `std::collections::VecDeque<std::boxed::Box<(dyn std::ops::Fn() -> state::Trans<T, E> + std::marker::Send + std::marker::Sync + 'static)>>` will meet its required lifetime bounds
   --> src/app.rs:307:40                                                                                                                                   
    |                                                                                                                                                      
307 |                 let mut v = self.world.write_resource::<TransQueue<T, E>>();                                                                         
    |                                        ^^^^^^^^^^^^^^                                                                                                
                                                                                                                                                           
error[E0310]: the parameter type `T` may not live long enough                                                                                              
   --> src/app.rs:496:15                                                                                                                                   
    |                                                                                                                                                      
395 | impl<'a, S, T, E, X> ApplicationBuilder<S, T, E, X>                                                                                                  
    |             - help: consider adding an explicit lifetime bound `T: 'static`...                                                                       
...                                                                                                                                                        
496 |         world.add_resource(TransQueue::<T, StateEvent>::with_capacity(2));                                                                           
    |               ^^^^^^^^^^^^                                                                                                                           
    |                                                                                                                                                      
note: ...so that the type `std::collections::VecDeque<std::boxed::Box<dyn std::ops::Fn() -> state::Trans<T, state_event::StateEvent> + std::marker::Send + std::marker::Sync>>` will meet its required lifetime bounds
   --> src/app.rs:496:15                                                                                                                                   
    |                                                                                                                                                      
496 |         world.add_resource(TransQueue::<T, StateEvent>::with_capacity(2));                                                                           
    |               ^^^^^^^^^^^^                                                                                                                           
                                                                                                                                                           
error: aborting due to 2 previous errors                                                                                                                   
                                                                                                                                                           
For more information about this error, try `rustc --explain E0310`.                                                                                        
error: Could not compile `amethyst`.
```

In my code I have T: 'a

src/app.rs
```diff
@@ -292,6 +297,19 @@ where
             states.stop(StateData::new(world, &mut self.data));
         }
 
+        // Read the Trans queue and apply changes.
+        {
```
DoumanAsh: Consider checking len of `TransQueue` before entering this part, and maybe it would be better to move the part of reading `TransQueue` into separate function

Rhuagh:
I'm not sure about this, because of the Send/Sync requirement on States. This will disallow using state specific dispatchers as an example.

jojolepro:
@Rhuagh you already said that and no, States don't need to be Send+Sync as shown in the state_dispatcher example.

Rhuagh:
Try doing a normal Push on a state with a lifetime.

Rhuagh:
(Outside the trans queue).

jojolepro:
Done

Rhuagh:
I stand corrected.

jojolepro:
The impl State is 'static because in src/app.rs it seems that Resources wants to own values that live for the 'static lifetime, thus preventing the use of an 'a lifetime (the one from Application)

src/app.rs
```diff
@@ -292,6 +300,31 @@ where
...
+            /*let trans = {
+                let mut v = self.world.write_resource::<TransQueue<T, E>>();
+                let x = v.drain(..).collect::<VecDeque<_>>();
+                x
+            };*/
+            let mut world = &mut self.world;
+            let states = &mut self.states;
+            let reader = &mut self.trans_reader_id;
+
+            /*let trans = world.exec(|ev: Read<EventChannel<TransEvent<T, E>>>| {
```
DoumanAsh: Clean comment?
Here and everywhere else

src/app.rs
```diff
@@ -50,6 +55,8 @@ pub struct CoreApplication<'a, T, E = StateEvent, R = StateEventReader> {
     #[derivative(Debug = "ignore")]
     events: Vec<E>,
     event_reader_id: ReaderId<Event>,
+    #[derivative(Debug = "ignore")]
```
torkleyy: Why is this ignored?

src/state.rs
```diff
@@ -411,9 +418,10 @@ impl<'a, T, E: Send + Sync + 'static> StateMachine<'a, T, E> {
         }
     }
 
-    /// Performs a state transition, if requested by either update() or
-    /// fixed_update().
-    fn transition(&mut self, request: Trans<T, E>, data: StateData<T>) {
+    /// Performs a state transition.
+    /// Usually called by update or fixed_update.
+    /// Can also be called when TransQueue is used.
```
torkleyy: Can you add a bit more extensive documentation on the behaviour etc?

src/app.rs
CBenoit: Yeah, commented code is probably not updated, better remove it.

docs/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 * `PrefabData` can now be derived for many situations, see the book for more information ([#1035])
 * Support for DirectionalLight and SpotLight in PBM pass. ([#1074], [#1081])
 * `amethyst_ui::get_default_font` supports loading system font from Path. ([#1108])
+* Added Trans event queue. Used to trigger state transitions from systems. Also used to trigger multiple state transitions at once. (For example, to `Trans::Pop` two states. ([#1069])
```
azriel91: unmatched `(`

examples/state_dispatcher/main.rs
```diff
@@ -0,0 +1,58 @@
+//! An example showing how to create a dispatcher inside of a State.
+
+extern crate amethyst;
+
+use amethyst::ecs::{Dispatcher, DispatcherBuilder};
+use amethyst::prelude::*;
+use amethyst::shrev::EventChannel;
+use amethyst::Error;
+use std::marker::PhantomData;
```
azriel91: can you split `std` into its own import group, and group the `amethyst::` ones?

src/app.rs
```diff
@@ -292,6 +300,23 @@ where
...
+            let states = &mut self.states;
+            let reader = &mut self.trans_reader_id;
+
+            let trans = world
+                .read_resource::<EventChannel<TransEvent<T, E>>>()
+                .read(reader)
+                .map(|e| e())
+                .collect::<Vec<_>>();
+            for tr in trans {
+                //let t = tr();
```
azriel91: `//remove me`

amethyst_utils/src/ortho_camera.rs
```diff
@@ -255,7 +255,7 @@ mod test {
         let y = mat4.w.z; // c3r2
         let near = (y + 1.0) / x;
         let far = (x - 1.0) / y;
-
+
```
azriel91: undo

azriel91:
> Adding this to the concepts page would be super confusing, because it is mostly most usages of it involve complex data flows. When we make an "advanced usages" section, this would be a good candidate to have its (very small) chapter.
I agree -- it's not a common thing to do, and not introducing it early could overwhelm.

amethyst_utils/src/ortho_camera.rs
jojolepro: fixed

src/app.rs
jojolepro: fixed

jojolepro:
Will merge in a few hours if there is no more complaints :3

PR #1071, opened 2018-10-26
[Fix] Reformat and fix Concept/Resource example code

yukicode:
Reformatted example code to meet documentation guideline. 
Fixed a wrong return type in comment.
Made the example code compile.

### DIFF
--- a/book/src/concepts/resource.md
+++ b/book/src/concepts/resource.md
@@ -36,12 +36,20 @@ Fetching a resource can be done like this:
 ```rust,no_run,noplaypen
 # extern crate amethyst;
 # use amethyst::ecs::{Resources};
-# struct MyResource;
+# #[derive(Debug, PartialEq)]
+# struct MyResource {
+#   pub game_score: i32,
+# }
 # fn main() {
 #   let mut resources = Resources::new();
+#   let my = MyResource{
+#     game_score: 0,
+#   };
+#   resources.insert(my);
   // try_fetch returns a Option<<Fetch<MyResource>>
   let fetched = resources.try_fetch::<MyResource>()
                        .expect("No MyResource present in Resources");
+  assert_eq!(*fetched, MyResource{ game_score: 0, });
 # }
 ```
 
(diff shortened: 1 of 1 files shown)


book/src/concepts/resource.md
```diff
@@ -61,10 +71,20 @@ If you want to change a resource that is already inside of `Resources`:
...
-let mut fetched = resources.try_fetch_mut::<MyResource>().expect("No MyResource present in Resources");
+#   let my = MyResource{
+#     game_score: 0,
+#   };
+#   resources.insert(my);
+  // try_fetch_mut returns a Option<FetchMut<MyResource>>
+  let mut fetched = resources.try_fetch_mut::<MyResource>()
+                             .expect("No MyResource present in Resources");
+  assert_eq!(fetched.game_score, 0);
+  fetched.game_score = 10;
```
jojolepro: Could you add another assert_eq after this line showing that the score is now 10?
Thanks!

yukicode: Sure! I was debating between adding it or not :)

book/src/concepts/resource.md
```diff
@@ -36,11 +36,20 @@ Fetching a resource can be done like this:
...
 #   let mut resources = Resources::new();
-// Returns a Option<MyResource>
-let fetched = resources.try_fetch::<MyResource>().expect("No MyResource present in Resources");
+#   let my = MyResource{
+#     game_score: 0,
+#   };
+#   resources.insert(my);
+  // try_fetch returns a Option<Fetch<MyResource>>
+  let fetched = resources.try_fetch::<MyResource>()
+                       .expect("No MyResource present in Resources");
```
torkleyy: Would be nice to not panic here, because that's the point of `try_fetch` ;)

yukicode: Good point. I could use 
`If let Some(mut fetched_resource)... else...` 
to replace the expect().
What do you think?

torkleyy: Sounds good!

yukicode: Changed :)

PR #1073, opened 2018-10-27
Update CI

magnonellie:
Hopefully this will fix some of the problems we've been having with the CI. Unfortunately, I'm still not sure the docker image will automatically update, but my hope is that it will since it's tracking the `latest` tag now.


Moxinilian:
As agreed on Discord, let's do an early merge.
bors r+

Issue #855, opened 2018-08-07
Lighting/Pass compatibility


ghost:
# Tasks
- Add missing lights to renderer

I've noticed some holes in the supported [lights](https://docs.rs/amethyst_renderer/0.8.0/amethyst_renderer/enum.Light.html) with different rendering passes. I'll catalog them here, so that maybe I or somebody can work on fixing them eventually. I'll post my test program when I get a chance.

Pass|Area|Directional|Point|Spot|Sun
-|-|-|-|-|-
`DrawShaded<PosNormTex>`||Lit|Lit||**Dark**
`DrawShadedSeparate<PosNormTex>`
`DrawPbm<PosNormTangTex>`||**Dark**|Lit||**Dark**
`DrawPbmSeparate<PosNormTangTex>`|||Bugged||**Dark**
`DrawFlat<PosTex>`|*n/a*|*n/a*|*n/a*|*n/a*|*n/a*
`DrawFlatSeparate<PosTex>`|*n/a*|*n/a*|*n/a*|*n/a*|*n/a*

Pinging @omni-viral per @jojolepro for guidance.

omni-viral:
In `Pbm` shaders passes `Directional` light wasn't implemented. Should be.
What's `Sun` light type? How it is different from `Directional`

minecrawler:
As per the comment above the struct:

> /// A realistic disk-shaped sun light source.

So I guess that a directional light has parallel light rays, but a sun's rays are spreading, which creates a different effect...

What would be interesting is knowing what technique is used to calculate the lights and shadows and if it provides modern effects, like light bleeding, ambient occlusion, multi-reflection and scattering (getting rid of ambient illumination), hard->soft shadows, etc.

Imho, especially the PBS shaders should be able to handle such realistic features. There should also be algorithms for weaker machines!

Aceeri:
https://seblagarde.files.wordpress.com/2015/07/course_notes_moving_frostbite_to_pbr_v32.pdf#page=24&zoom=auto,-122,784
Good diagram of punctual vs area and such.

omni-viral:
With deferred shading it would be easier to add new lightsource types to lighting pass.

magnonellie:
We would still need to support forward rendering for stuff like VR that pretty much require MSAA.

termhn:
Clustered forward: some of the benefits of deferred and most of the benefits of forward http://www.cse.chalmers.se/~uffe/clustered_shading_preprint.pdf
[...]

PR #1074, opened 2018-10-27
added directional light support to PBM shader

karroffel:
This PR implements directional light processing in the PBM shader. See #855 

image

This just takes the existing lighting code and refactors it to use it for directional lights as well, so any previous issues will most likely not be fixed by this. But that could be a challenge for next time :wink:


Moxinilian:
There is an issue with Concourse testing here. It does not take into account glsl changes.
Maybe an edge case here would be nice. @magnonellie

jojolepro:
You can also add a changelog entry if you want. ^^

PR #1075, opened 2018-10-28
Refactor unwraps for several amethyst crates

barskern:
**This PR contains a small breaking change in the public API of `Transform`**.

Crates where unwraps have been replaced:

- [x] amethyst_animation
- [x] amethyst_assets
- [x] amethyst_audio (already no unwraps)
- [x] amethyst_config (already no unwraps)
- [x] amethyst_controls
- [x] amethyst_core
- [x] amethyst_derive (already no unwraps)
- [x] amethyst_gltf
- [x] amethyst_input
- [x] amethyst_locale (already no unwraps)
- [x] amethyst_network
- [x] amethyst_renderer
- [x] amethyst_ui
- [x] amethyst_utils (already no unwraps)

Each crate will receive its own commit. 

Closes: #908

### DIFF
--- a/amethyst_assets/src/prefab/mod.rs
+++ b/amethyst_assets/src/prefab/mod.rs
@@ -247,9 +247,11 @@ impl<T> Prefab<T> {
     ///
     /// ### Panics
     ///
-    /// If sub asset loading have not been triggered
+    /// If sub asset loading has not been triggered.
     pub fn progress(&self) -> &ProgressCounter {
-        self.counter.as_ref().unwrap()
+        self.counter
+            .as_ref()
+            .expect("Sub asset loading has not been triggered")
     }
 
     /// Trigger sub asset loading for the asset
--- a/amethyst_assets/src/prefab/system.rs
+++ b/amethyst_assets/src/prefab/system.rs
@@ -93,8 +93,12 @@ where
             &**pool,
             strategy,
         );
-        prefab_handles
-            .populate_inserted(self.insert_reader.as_mut().unwrap(), &mut self.to_process);
+        prefab_handles.populate_inserted(
+            self.insert_reader.as_mut().expect(
+                "`PrefabLoaderSystem::setup` was not called before `PrefabLoaderSystem::run`",
+            ),
+            &mut self.to_process,
+        );
         self.finished.clear();
         for (root_entity, handle, _) in (&*entities, &prefab_handles, &self.to_process).join() {
             if let Some(prefab) = prefab_storage.get(handle) {
@@ -112,10 +116,16 @@ where
                                 Parent {
                                     entity: self.entities[parent],
                                 },
-                            ).unwrap();
+                            ).expect("Unable to insert `Parent` for prefab");
                     }
-                    tags.insert(new_entity, PrefabTag::new(prefab.tag.unwrap()))
-                        .unwrap();
+                    tags.insert(
+                        new_entity,
+                        PrefabTag::new(
+                            prefab.tag.expect(
+                                "Unreachable: Every loaded prefab should have a `PrefabTag`",
+                            ),
+                        ),
+                    ).expect("Unable to insert `PrefabTag` for prefab entity");
                 }
                 // create components
                 for (index, entity_data) in prefab.entities.iter().enumerate() {
@@ -125,7 +135,7 @@ where
                                 self.entities[index],
                                 &mut prefab_system_data,
                                 &self.entities,
-                            ).unwrap();
+                            ).expect("Unable to add prefab system data to entity");
                     }
                 }
             }
--- a/amethyst_assets/src/storage.rs
+++ b/amethyst_assets/src/storage.rs
@@ -154,7 +154,10 @@ impl<A: Asset> AssetStorage<A> {
         F: FnMut(A::Data) -> Result<ProcessingState<A>>,
     {
         {
-            let requeue = self.requeue.get_mut().unwrap();
+            let requeue = self
+                .requeue
+                .get_mut()
+                .expect("The mutex of `requeue` in `AssetStorage` was poisoned");
             while let Some(processed) = self.processed.try_pop() {
                 let assets = &mut self.assets;
                 let bitset = &mut self.bitset;
--- a/amethyst_controls/src/systems.rs
+++ b/amethyst_controls/src/systems.rs
@@ -150,7 +150,10 @@ where
 
     fn run(&mut self, (events, mut transform, tag, focus, hide): Self::SystemData) {
         let focused = focus.is_focused;
-        for event in events.read(&mut self.event_reader.as_mut().unwrap()) {
+        for event in
+            events.read(&mut self.event_reader.as_mut().expect(
+                "`FreeRotationSystem::setup` was not called before `FreeRotationSystem::run`",
+            )) {
             if focused && hide.hide {
                 if let Event::DeviceEvent { ref event, .. } = *event {
                     if let DeviceEvent::MouseMotion { delta: (x, y) } = *event {
@@ -188,7 +191,9 @@ impl<'a> System<'a> for MouseFocusUpdateSystem {
     type SystemData = (Read<'a, EventChannel<Event>>, Write<'a, WindowFocus>);
 
     fn run(&mut self, (events, mut focus): Self::SystemData) {
-        for event in events.read(&mut self.event_reader.as_mut().unwrap()) {
+        for event in events.read(&mut self.event_reader.as_mut().expect(
+            "`MouseFocusUpdateSystem::setup` was not called before `MouseFocusUpdateSystem::run`",
+        )) {
             if let Event::WindowEvent { ref event, .. } = *event {
                 if let WindowEvent::Focused(focused) = *event {
                     focus.is_focused = focused;
--- a/amethyst_core/src/named.rs
+++ b/amethyst_core/src/named.rs
@@ -136,12 +136,11 @@ impl<'a> WithNamed for EntityBuilder<'a> {
     where
         S: Into<Cow<'static, str>>,
     {
-        // Unwrap: The only way this can fail is if the entity is invalid and this is used while creating the entity.
         self.world
             .system_data::<(WriteStorage<'a, Named>,)>()
             .0
             .insert(self.entity, Named::new(name))
-            .unwrap();
+            .expect("Unreachable: Entities should always be valid when just created");
         self
     }
 }
--- a/amethyst_core/src/transform/systems.rs
+++ b/amethyst_core/src/transform/systems.rs
@@ -61,18 +61,23 @@ impl<'a> System<'a> for TransformSystem {
         self.global_modified.clear();
 
         locals.populate_inserted(
-            self.inserted_local_id.as_mut().unwrap(),
+            self.inserted_local_id
+                .as_mut()
+                .expect("`TransformSystem::setup` was not called before `TransformSystem::run`"),
             &mut self.local_modified,
         );
         locals.populate_modified(
-            self.modified_local_id.as_mut().unwrap(),
+            self.modified_local_id
+                .as_mut()
+                .expect("`TransformSystem::setup` was not called before `TransformSystem::run`"),
             &mut self.local_modified,
         );
 
-        for event in hierarchy
-            .changed()
-            .read(self.parent_events_id.as_mut().unwrap())
-        {
+        for event in hierarchy.changed().read(
+            self.parent_events_id
+                .as_mut()
+                .expect("`TransformSystem::setup` was not called before `TransformSystem::run`"),
+        ) {
             match *event {
                 HierarchyEvent::Removed(entity) => {
                     // Sometimes the user may have already deleted the entity.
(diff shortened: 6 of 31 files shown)


amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -74,6 +74,12 @@ impl Transform {
     pub fn matrix(&self) -> Matrix4<f32> {
         // This is a hot function, so manually implement the matrix-multiply to avoid a load of
         // unnecessary +0s.
+
+        // FIXME This operation assumes that `self.rotation` is a unit quaternion,
+        // which might not always be true. Perhaps this should be
+        // "foolproofed"? The probability that a user has altered a single
+        // component of the quaternion is small, but its possible as the
+        // rotation is a public field.
```
Moxinilian: This will be fixed thanks to the UnitQuaternion type in nalgebra.

barskern: Nice. I just saw it so I figured I would comment on it :smile:

barskern:
Of course, I just wanted to highlight the breaking change so it would not be overlooked hence two commits.

amethyst_core/src/named.rs
```diff
@@ -136,12 +136,11 @@ impl<'a> WithNamed for EntityBuilder<'a> {
     where
         S: Into<Cow<'static, str>>,
     {
-        // Unwrap: The only way this can fail is if the entity is invalid and this is used while creating the entity.
         self.world
             .system_data::<(WriteStorage<'a, Named>,)>()
             .0
             .insert(self.entity, Named::new(name))
-            .unwrap();
+            .expect("entity should always be valid when just created");
```
jojolepro: The unwrap was actually fine and had an explanation. There wasn't much reason for the change.

Moxinilian:
Ah, yes, about that, you will have to update it where it is used in ametyst_renderer.
Tests will fail accordingly so you'll know where to do it.

Moxinilian:
lol how did they succeed

Moxinilian:
My bad, I was confusing it with GlobalTransform, everything is fine.

barskern:
Yeah I grepped for `view_matrix` but I could not find any places it was used

amethyst_core/src/named.rs
barskern: True I figured for concistency and better reports if it will ever fail because of a bug. Should I not do this in other places + revert this one?

jojolepro: You could change the message to
"unreachable: Entities should always be valid when just created"
?

:)

Moxinilian: I think replacing all unwraps with expects makes it much better.
Last time, when we had that camera crash, it took too long to discover where the bug was from without an expect telling us.

barskern: Sure! Should I then do the same if I encounter other unwraps like this?

Moxinilian: Yep, seems good to me.

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -254,10 +260,17 @@ impl Transform {
     /// Calculates the inverse of this transform, which we need to render.
     ///
     /// We can exploit the extra information we have to perform this inverse faster than `O(n^3)`.
-    pub fn view_matrix(&self) -> Matrix4<f32> {
-        // todo
+    ///
+    /// # Return
+    ///
+    /// This will return in `None` in the following cases:
```
torkleyy:
```suggestion
    /// This will return in `None` the following cases:
```

amethyst_core/src/transform/systems.rs
```diff
@@ -60,18 +60,23 @@ impl<'a> System<'a> for TransformSystem {
         self.global_modified.clear();
 
         locals.populate_inserted(
-            self.inserted_local_id.as_mut().unwrap(),
+            self.inserted_local_id
+                .as_mut()
+                .expect("'TransformSystem::setup' was not called before 'TransformSystem::run'"),
```
torkleyy: This should use a backtick (\`) instead of a single quote (').

amethyst_core/src/transform/components/local_transform.rs
torkleyy:
```suggestion
    /// This will return `None` in the following cases:
```

amethyst_core/src/transform/systems.rs
barskern: In an error message?

amethyst_core/src/transform/components/local_transform.rs
barskern: Nice catch!

amethyst_core/src/transform/systems.rs
torkleyy: Why do you feel backticks are not appropriate in error messages? It's just the standard style we use for code there.

Rhuagh:
@Moxinilian the view matrix function is never used because we use the `GlobalTransform` in the renderer and perform the inversion there

amethyst_core/src/transform/systems.rs
barskern: Hmm good question. Change is on the way

amethyst_assets/src/prefab/mod.rs
```diff
@@ -246,7 +246,9 @@ impl<T> Prefab<T> {
     ///
     /// If sub asset loading have not been triggered
     pub fn progress(&self) -> &ProgressCounter {
-        self.counter.as_ref().unwrap()
+        self.counter
+            .as_ref()
+            .expect("Sub asset loading have not been triggered")
```
Moxinilian: has*
The mistake was there before in the doc in fact, could you fix it in the process?

amethyst_assets/src/prefab/system.rs
```diff
@@ -89,8 +89,12 @@ where
             &**pool,
             strategy,
         );
-        prefab_handles
-            .populate_inserted(self.insert_reader.as_mut().unwrap(), &mut self.to_process);
+        prefab_handles.populate_inserted(
+            self.insert_reader
+                .as_mut()
+                .expect("`PrefabLoaderSystem::setup` was called before `PrefabLoaderSystem::run`"),
```
Moxinilian: I think you swapped the two methods here

amethyst_assets/src/prefab/system.rs
```diff
@@ -102,26 +106,32 @@ where
...
+                            Parent {
+                                entity: self.entities[parent],
+                            },
+                        ) {
+                            error!("Unable to insert new entity with a `Parent`: {}", e);
+                        }
+                    }
+                    if let Err(e) = tags.insert(
+                        new_entity,
+                        PrefabTag::new(prefab.tag.expect("Prefab tag is not present")),
```
Moxinilian: Would this happen and not be a bug? If it is always a bug, could you mention it in the expect?

amethyst_assets/src/prefab/system.rs
```diff
@@ -102,26 +106,32 @@ where
...
-                                self.entities[index],
-                                &mut prefab_system_data,
-                                &self.entities,
-                            ).unwrap();
+                        if let Err(e) = prefab_data.add_to_entity(
+                            self.entities[index],
+                            &mut prefab_system_data,
+                            &self.entities,
+                        ) {
+                            error!("Unable to insert prefab system data for entity: {}", e);
```
Moxinilian: Again, could this happen and not be a bug?
If so, it might be beneficial to have the engine crash with a meaningful message.

amethyst_assets/src/storage.rs
```diff
@@ -141,7 +141,7 @@ impl<A: Asset> AssetStorage<A> {
         F: FnMut(A::Data) -> Result<ProcessingState<A>>,
     {
         {
-            let requeue = self.requeue.get_mut().unwrap();
+            let requeue = self.requeue.get_mut().expect("Mutex was poisoned");
```
Moxinilian: Could you add a word for context?

amethyst_controls/src/systems.rs
```diff
@@ -146,7 +146,12 @@ where
 
     fn run(&mut self, (events, mut transform, tag, focus, hide): Self::SystemData) {
         let focused = focus.is_focused;
-        for event in events.read(&mut self.event_reader.as_mut().unwrap()) {
+        for event in events.read(
+            &mut self
+                .event_reader
+                .as_mut()
+                .expect("`FreeRotationSystem::setup` was called before `FreeRotationSystem::run`"),
```
Moxinilian: I think you swapped the two here as well.

amethyst_controls/src/systems.rs
```diff
@@ -184,7 +189,9 @@ impl<'a> System<'a> for MouseFocusUpdateSystem {
     type SystemData = (Read<'a, EventChannel<Event>>, Write<'a, WindowFocus>);
 
     fn run(&mut self, (events, mut focus): Self::SystemData) {
-        for event in events.read(&mut self.event_reader.as_mut().unwrap()) {
+        for event in events.read(&mut self.event_reader.as_mut().expect(
+            "`MouseFocusUpdateSystem::setup` was called before `MouseFocusUpdateSystem::run`",
```
Moxinilian: Same

amethyst_assets/src/prefab/mod.rs
barskern: Thank you for fast feedback @Moxinilian! I sorry about all the mishaps, corrections incoming 👍

amethyst_assets/src/prefab/system.rs
barskern: Hmm I looked at it a lot before I decided to just log because from what it seemed like in the in definitions of `Prefab`, it seems to be a lot of `None`s all over the place before there is inserted an actual value. Hence I was actually going to ask for some help on this one.

amethyst_assets/src/storage.rs
barskern: Will do

amethyst_assets/src/prefab/system.rs
barskern: More specifically missing a `not` if I would write them similar as `amethyst_core` and earlier similar changes.

amethyst_assets/src/prefab/system.rs
barskern: This one aswell. I haven't been able to figure out if its a bug or a possible occurence. Could I have some guidance perhaps?

amethyst_assets/src/prefab/system.rs
Moxinilian: You should get the input of @Rhuagh, he wrote the prefab system.

torkleyy: This should not be able to fail, since every loaded prefab has such a tag. If the prefab is not loaded, it cannot be retrieved at this point.

amethyst_assets/src/prefab/system.rs
barskern: @Rhuagh I saw Moxinilian said that it was you who created the asset system. Could you help me understand if a `None` in this case is possible or would be a bug?

torkleyy:
Yeah, more progress :tada: 

One idea I've had earlier, maybe there are some places where we can add `#[no_panic]` to assert that it can indeed not panic. What do you think? I'm sure this won't work everywhere we use `.expect`, but it sounds useful for when it does.

barskern:
The way I understood it `no_panic` can verify at compile time if a test contains input which would cause a panic? So in essence it would catch panics when tests are complied instead of after they are ran?

torkleyy:
Hmm, yes. I'm not sure if that makes too much sense, just an idea I wanted to comment somewhere. Not important to this PR specifically.

barskern:
It's a good suggestion, but I am not sure how much value it would bring for the added complexity.

barskern:
This PR is ready for review @torkleyy :smile:

barskern:
Cool! That's probably outside the scope of this PR hopefully because I have to admit that even though it's been interesting, removing unwraps haven't been the most rewarding form of programming. But thank you so much for nice follow up and motivation 😁❤️ should I squash as planned?

torkleyy:
Could you please rebase the PR? I'll merge as soon as possible :)

magnonellie:
Haha that was probably my fault I would assume. I'll look into if we can't integrate that as a check the same way we do with rustfmt. My only concern with using expect is that it bloats the final binary, so I guess what I would kinda want is a `.cant_fail(explaination: &'static str)` function that just gets taken out in production builds.

barskern:
You have a good point @magnonellie, that's the disadvantage about using `expect`. However as @Moxinilian mentioned, there is another side to the story aswell:

> I think replacing all unwraps with expects makes it much better.
Last time, when we had that camera crash, it took too long to discover where the bug was from without an expect telling us.

torkleyy:
Thanks!
bors r+

magnonellie:
bors r-

Please fix rustfmt 😅

barskern:
Fixed 😅

torkleyy:
Thanks, now, let's try again ^^

bors r+

PR #1096, opened 2018-10-31
Add tm symbol everywhere

Xaeroxe:
cc @magnonellie


magnonellie:
I don't think we need to trademark the current logo as we have a new one right around the corner.

Moxinilian:
I'm a bit split on this one.
In one hand we indeed need to protect the branding if we want to move on. But on the other hand sticking TMs everywhere just feels weird in an OSS context.

Xaeroxe:
Since no one is saying no on merging this I'm gonna go ahead and 

bors r+

torkleyy:
Well, quite frankly I still don't really like it.

Issue #919, opened 2018-08-26
Follow Camera using Parent -> PlayerEntity can lead to difficult to track error


Telzhaak:
Scenario:
- Renderer pipeline with SpritePass
- Camera with Player as Parent, so that it should follow the parent's transform

Result:
-Panic at runtime on https://github.com/amethyst/amethyst/blob/690acabf6c578896280673e6273472cd6117d94c/amethyst_renderer/src/pass/util.rs#L252 according to @Moxinilian 

Reason: 
- Player Z gets copied, too, therefore Camera is on the same Plane as the PlayerSprite/Sprites,  
leading to the View matrix of the Global Transform becoming singular/non-invertable

Amethyst 0.8.0 + SpriteRender:
```
[dependencies]
shred = { version = "0.7.0", features =["nightly"]}
amethyst = { git = "https://github.com/amethyst/amethyst", rev = "44c87a512cacbbfd774433bd6f6d935ed97b6c0d", features = ["nightly"]}
```

Example:
```rust
let pipe = Pipeline::build()
        .with_stage(
            Stage::with_backbuffer()
                .clear_target([0.0, 0.0, 0.2, 1.0], 1.0)
                .with_pass(DrawSprite::new()),
        );

let game_data = GameDataBuilder::default()
        .with_bundle(InputBundle::<String, String>::new()
            .with_bindings_from_file(
                &keybindings_path
            )?,
        )?
        .with_bundle(
            TransformBundle::new()
                .with_dep(&["position_mapping_system"])
        )?
        .with_bundle(
            RenderBundle::new(
                pipe, 
                Some(config)
            )
            .with_sprite_sheet_processor()
        )?;
```

```rust
let player_entity = world
            .create_entity()
                .with(
                    sprite_render
                )
                .with(
                    GlobalTransform::default()
                )
                .with(
                    Transform::default()
                )
            .build());

 world
        .create_entity()
            .with(
                Camera::from(
                    Projection::orthographic(
                        0.0,
                        100,
                        100,
                        0.0,
                    )
                )
            )
            .with(
                GlobalTransform(
                    Matrix4::from_translation(
                        Vector3::new(
                            0.0, 0.0, 1.0
                        )
                    ).into()
                )
            )
            .with(
                Parent{
                    entity: player_entity,
                }
            )
            .with(
                Transform::default()
            )
        .build(); 
```

Rhuagh:
[...]

PR #1099, opened 2018-11-01
Create sprite_camera_follow example

halfbro:
This PR creates an example which shows creation of a Camera that has a sprite parent, which follows the sprite around. This is related to issue #919.


examples/sprite_camera_follow/main.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+
+struct Example;
+
+impl<'a, 'b> SimpleState<'a, 'b> for Example {
+    fn on_start(&mut self, data: StateData<GameData>) {
+        let world = data.world;
+        let sprite_sheet_handle = load_sprite_sheet(world);
+
+        let _anchor = init_sprite(world, &sprite_sheet_handle);
```
Telzhaak: I assume this sprite is meant to allow seeing the player's movement?
Maybe add a little comment here:
`// A sprite with fixed position, to see how the player moves`

examples/sprite_camera_follow/main.rs
```diff
@@ -0,0 +1,182 @@
...
+        .create_entity()
+        .with(transform)
+        .with(Player)
+        .with(sprite)
+        .build()
+}
+
+fn init_camera(world: &mut World, parent: Entity) {
+    let mut transform = Transform::default();
+    transform.translation.z = 1.0;
```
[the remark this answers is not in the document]
halfbro: I'll have to do some testing, but I thought that's how the `Parent` component worked? e.g. the `Camera` will add it's transform onto it's Parent's transform, since it isn't a `GlobalTransform`. I'm new to the project though so let me know if that isn't how it works. I'll test it later today otherwise.

examples/sprite_camera_follow/main.rs
halfbro: Yes it is, I'll add a comment

examples/sprite_camera_follow/main.rs
```diff
@@ -0,0 +1,182 @@
...
+    fn on_start(&mut self, data: StateData<GameData>) {
+        let world = data.world;
+        let sprite_sheet_handle = load_sprite_sheet(world);
+
+        let _anchor = init_sprite(world, &sprite_sheet_handle);
+        let parent_sprite = init_player(world, &sprite_sheet_handle);
+        init_camera(world, parent_sprite);
+    }
+
+    fn handle_event(&mut self, _: StateData<GameData>, event: StateEvent) -> SimpleTrans<'a, 'b> {
```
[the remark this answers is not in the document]
halfbro: Didn't know that it was default, I'll remove it then.

examples/sprite_camera_follow/main.rs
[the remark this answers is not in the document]
halfbro: I'll confirm it later before I resolve

halfbro:
Hmm, how do I squash the commits? Do I have to do it on my branch before doing the PR?

examples/sprite_camera_follow/main.rs
```diff
@@ -0,0 +1,160 @@
...
+    let mut transform = Transform::default();
+    transform.translation.x = 100.0;
+    transform.translation.y = 0.0;
+    let sprite = SpriteRender {
+        sprite_sheet: sprite_sheet.clone(),
+        sprite_number: 1,
+        flip_horizontal: false,
+        flip_vertical: false,
+    };
+    world.create_entity().with(transform).with(sprite).build()
```
Moxinilian: Little nitpick here: could you add a second reference point so that is clear to the user that the player is moving, and not the reference point? Also maybe a background could help on that, do you think it would be interesting?

halfbro: I think a background could be good, though it might take a bit to create. Is a background just a very large sprite in this case, or is there a better way to do it?

Moxinilian: I would have went for a very large sprite so that the code does not become too complex.

halfbro: Aight, I'll add it now.

halfbro: I went ahead and added a simple background sprite, and added a little more documentation for some of the "gotchas" I found, like sprite ordering and transparency.

halfbro:
Alright I had some issues with git for a moment there, but I'm pretty sure it's all figured out, updated, etc. It's ready to get merged in whenever.

examples/sprite_camera_follow/main.rs
```diff
@@ -0,0 +1,192 @@
...
+        .with_bundle(TransformBundle::new())?
+        .with_bundle(InputBundle::<String, String>::new().with_bindings_from_file(&input_path)?)?
+        .with(MovementSystem, "movement", &[])
+        .with_bundle(
+            RenderBundle::new(pipe, Some(config))
+                .with_sprite_sheet_processor()
+                .with_sprite_visibility_sorting(&[]), // Let's us use the `Transparent` component
+        )?;
+
+    let mut game = Application::build("./", Example)?.build(game_data)?;
```
Rhuagh: This would be where you set the resources path to `format!("{}/examples/sprite_camera_follow/resources", application_root_dir())` instead of the current `"./"`.

examples/sprite_camera_follow/main.rs
```diff
@@ -0,0 +1,192 @@
...
+impl<'a, 'b> SimpleState<'a, 'b> for Example {
+    fn on_start(&mut self, data: StateData<GameData>) {
+        let world = data.world;
+        let resources_path = format!(
+            "{}/examples/sprite_camera_follow/resources",
+            application_root_dir()
+        );
+        let circle_sprite_sheet_handle = load_sprite_sheet(
+            world,
+            &format!("{}/Circle_Spritesheet.png", resources_path),
```
Rhuagh: You can then use `"Circle_Spritesheet.png"` here.

examples/sprite_camera_follow/main.rs
```diff
@@ -0,0 +1,192 @@
...
+        .with(transform)
+        .build();
+}
+
+struct Example;
+
+impl<'a, 'b> SimpleState<'a, 'b> for Example {
+    fn on_start(&mut self, data: StateData<GameData>) {
+        let world = data.world;
+        let resources_path = format!(
```
Rhuagh: This should really just set the asset directory when creating Application, and then using relative names instead.

halfbro: Didn't realize that field was setting the Asset directory, that would definitely be the right way to do it.
I can change it tonight.

halfbro: Changed

examples/sprite_camera_follow/main.rs
torkleyy: I'm not sure anymore if we wanted to have examples use the relative path instead, since it's not good practice to hardcode paths to the compilation dir.

halfbro:
This is ready to merge, unless there are any more comments.

docs/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 * `PrefabData` can now be derived for many situations, see the book for more information ([#1035])
 * Support for DirectionalLight and SpotLight in PBM pass. ([#1074], [#1081])
 * `amethyst_ui::get_default_font` supports loading system font from Path. ([#1108])
+* `sprite_camera_follow` example showing how to use a Camera that has a sprite Parent ([#919])
```
jojolepro: You forgot to add the link for this changelog entry.

```rs
[#1099]: https://github.com/amethyst/amethyst/pull/1099
```

examples/sprite_camera_follow/main.rs
```diff
@@ -0,0 +1,184 @@
...
+        let y_move = input.axis_value("entity_y").unwrap();
+
+        for (_, transform) in (&players, &mut transforms).join() {
+            transform.translation.x += x_move as f32 * 5.0;
+            transform.translation.y += y_move as f32 * 5.0;
+        }
+    }
+}
+
+fn load_sprite_sheet(world: &mut World, png_path: &str, ron_path: &str) -> SpriteSheetHandle {
```
jojolepro: This thing really need to get merged into the engine. I keep seeing it copy pasted everywhere... 
Feel free to not do it for this pr, but it would be nice if someone could do it at some point.

examples/sprite_camera_follow/main.rs
```diff
@@ -0,0 +1,184 @@
...
+        .build()
+}
+
+fn init_camera(world: &mut World, parent: Entity) {
+    let mut transform = Transform::default();
+    transform.translation.z = 1.0;
+    world
+        .create_entity()
+        .with(Camera::from(Projection::orthographic(
+            -250.0, 250.0, 250.0, -250.0,
```
jojolepro: If this gets merged after the nalgebra pr (depending on how fast you are to fix my review, you might be before), you'll need to change the order of some of those values.

docs/CHANGELOG.md
halfbro: Ah, I misunderstood how those links worked. Fixed.

examples/sprite_camera_follow/main.rs
Moxinilian: It can't be done in a very nice way as it requires the user provides a texture ID. But it is going away thanks to Rhuagh's work, we will slightly modify it when we will redesign it.

halfbro:
Updated to use the new nalgebra translations

examples/sprite_camera_follow/main.rs
Rhuagh: Just add a `SpriteSheetLoader` helper that abstract this shit away tbh.

Issue #1093, opened 2018-10-30
Document PrefabData required use statements


ISibboI:
When using the PrefabData derive macro, I still need to add use directives for the types used by the macro myself. Specifically there are
```rust
use amethyst::ecs::{Entity};
use amethyst::assets::{PrefabData, PrefabError, ProgressCounter};
```
I propose to make the macro add these on its own, if they are not present.

jojolepro:
That isn't something we can fix. This is a limitation of how macro are built in rust currently. The required imports should be listed in the documentation.

I went over this quickly and I think there isn't any mention of the required `use` https://www.amethyst.rs/book/master/prefab.html

There isn't any docs here either, which is the doc-visible part of the macro https://www.amethyst.rs/doc/master/doc/amethyst_derive/fn.prefab_data_derive.html

halfbro:
The documentation for this is merged, should this issue be closed?

PR #1100, opened 2018-11-01
Add documentation to prefab_data_derive

halfbro:
Adds document describing which imports are needed in order to successfully derive `PrefabData`. Related to #1093.


Telzhaak:
PR so good, JoJo approves twice

jojolepro:
wow that was pretty fast

PR #1107, opened 2018-11-03
make default logger more configurable

qqwa:
`fern::dispatch` gets build and can be returned to the user instead
off starting it directly, this allows the user to add extra options to
the logger like `level_for()` before starting the logger.

Example:
```
let logger = amethyst::Logger::from_config(amethyst::LoggerConfig::default());
logger.get_dispatch()
    .level_for("gfx_device_gl", log::LevelFilter::Warn)
    .level_for("gfx_glyph", log::LevelFilter::Error)
    .apply();
```

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -39,12 +39,22 @@ impl Default for LoggerConfig {
     }
 }
 
+/// Allows the creation of a logger with a set of custom configurations. If no custom configuration
+/// is required [`start_logger`] can be used instead.
+///
+/// # Examples
+/// ```
+/// amethyst::Logger::from_config(amethyst::LoggerConfig::default())
+///     .level_for("gfx_device_gl", log::LevelFilter::Warn)
+///     .level_for("gfx_glyph", log::LevelFilter::Error)
+///     .start();
+/// ```
 pub struct Logger {
     dispatch: fern::Dispatch,
 }
 
 impl Logger {
-    pub fn new() -> Self {
+    fn new() -> Self {
         let dispatch = fern::Dispatch::new().format(|out, message, record| {
             out.finish(format_args!(
                 "[{level}][{target}] {message}",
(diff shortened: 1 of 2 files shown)


src/logger.rs
```diff
@@ -39,6 +39,76 @@ impl Default for LoggerConfig {
     }
 }
 
+pub struct Logger {
```
LucioFranco: I think ci is upset because there are no docs, but an example here would be really helpful.

magnonellie:
I'm not sure I like this because it exposes something to the user that we may eventually want to change. Fern may not always be the logging library of choice so being able to swap it out without having to bother the user seems like the best option to me. Is it possible you could implement the features you want from fern into the current setup instead of outright exposing fern?

src/logger.rs
qqwa: Yeah I will write the documentation when everything else is ok

magnonellie:
What I meant is that if you don't expose the implementation details of the logger then we could potentially add new features, change the underlying library and more all without touching user facing APIs. Because this change exposes fern to the user, that means that if we ever remove fern in favor of something else, all users will have to change their code to match the new library which isn't ideal. That's why I think it's better to implement the features you need from ferns dispatcher into the current implementation and then forward that to fern from there.

qqwa:
`fern::Dispatch` is no longer exposed. I made the `Logger::new` private, because there is currently no other way then using `Logger::from_config(..)` to attach stdio or a file.

src/logger.rs
```diff
@@ -92,22 +145,7 @@ fn env_var_override(config: &mut LoggerConfig) {
...
-                target = record.target(),
-                message = message,
-            ))
-        })
-}
-
-fn colored_stdout() -> fern::Dispatch {
-    let color_config = fern::colors::ColoredLevelConfig::new();
-
+fn colored_stdout(color_config: fern::colors::ColoredLevelConfig) -> fern::Dispatch {
```
magnonellie: This is also exposing fern types which I suppose isn't a huge deal but it would be awesome if you felt like fixing it ;)

qqwa: it's private

magnonellie: Duh I'm sorry xD

Issue #1072, opened 2018-10-27
[RFC] Amethyst UI


jojolepro:
# Ui RFC

Here we go...

*Warning: Things might be out of order or otherwise hard to understand. Don't be afraid to jump between sections to get a better view.*

So here, I will be describing the requirements, the concepts, the choices offered and the tradeoffs as well as the list of tasks.

Surprisingly enough, most games actually have more work in their ui than in the actual gameplay. Even for really small projects (anything that isn't a prototype), games will have a ui for the menu, in game information display, settings screen, etc...

Let's start by finding use cases, as they will give us an objective to reach.
I will be using huge games, to be sure to not miss out on anything.

Since the use cases are taken from actual games and I will be listing only the re-usable/common components in them, nobody can complain that the scope is too big. :)

## Use cases: Images

![Endless Space](https://wallscover.com/images/endless-space-wallpaper-19.jpg)
![Endless Space](https://static1.squarespace.com/static/5438bd1be4b0446f6692b906/596457d2e3df28be4294d6c1/59656f7672af65c2088a286a/1499819904593/Endless+Space+2+-+Technology+Tree.jpg)
![Path of exiles](https://i.ytimg.com/vi/8k6fSqIsAuY/maxresdefault.jpg)
![Wow](https://i.ytimg.com/vi/K4KpsllRvmY/maxresdefault.jpg)
![COD_BO4](https://i.ytimg.com/vi/i4kXqRn87Rc/maxresdefault.jpg)

## Use cases: Text

So now, let's extract the use cases from those pictures.

* 2d text
* 3d ui (render to texture)
* 3d positionned flat text
* 3d positionned text with depth 
* 3d text can be visible through 3d elements or occluded by them (including partial occlusion)
* 2d world text (billboard)
* images
* color box
* color patterns (gradient)
* color filter (change saturation, grayscale, alpha, etc) ![color filter](https://i.imgur.com/Ly9rL2H.png)
* multiple color in same text vs multiple text segments aligned
* locale support
* display data
* clicky buttons
* clicky checkboxes
* draggable elements (free positioning)
* draggable elements (constrained) (can be used to scroll through scroll views, or to move sliders heads)
* drag and drop (with constraints on what can be dropped where)
* Scroll views
* Sliders
* tab view
* menu bar
* editable text
* focusable elements
* keyboard, mouse, controller inputs
* on screen keyboard (for use with mouse or controller)
* overlays (a.k.a help bubbles/popups)
* automatic layouting
[...]

PR #1112, opened 2018-11-05
Custom UI widget

jjedelsky:
Adds custom UI widgets.

### DIFF
--- a/amethyst_ui/src/prefab.rs
+++ b/amethyst_ui/src/prefab.rs
@@ -605,6 +605,7 @@ where
         self,
         _: Self::PrefabData,
     ) -> (UiWidget<A, I, F, NoCustomUi>, Self::PrefabData) {
+        // self can not exist
         unreachable!()
     }
 }
--- a/amethyst_ui/src/prefab.rs
+++ b/amethyst_ui/src/prefab.rs
@@ -496,6 +496,64 @@ where
     Custom(Box<C>),
 }
 
+impl<A, I, F, C> UiWidget<A, I, F, C>
+where
+    A: Format<Audio, Options = ()>,
+    I: Format<Texture, Options = TextureMetadata>,
+    F: Format<FontAsset, Options = ()>,
+    C: ToNativeWidget<A, I, F>,
+{
+    /// Convenience function to access widgets `UiTransformBuilder`
+    pub fn transform(&self) -> Option<&UiTransformBuilder> {
+        match self {
+            UiWidget::Container { ref transform, .. } => Some(transform),
+            UiWidget::Image { ref transform, .. } => Some(transform),
+            UiWidget::Text { ref transform, .. } => Some(transform),
+            UiWidget::Button { ref transform, .. } => Some(transform),
+            UiWidget::Custom(_) => None,
+        }
+    }
+
+    /// Convenience function to access widgets `UiTransformBuilder`
+    pub fn transform_mut(&mut self) -> Option<&mut UiTransformBuilder> {
+        match self {
+            UiWidget::Container {
+                ref mut transform, ..
+            } => Some(transform),
+            UiWidget::Image {
+                ref mut transform, ..
+            } => Some(transform),
+            UiWidget::Text {
+                ref mut transform, ..
+            } => Some(transform),
+            UiWidget::Button {
+                ref mut transform, ..
+            } => Some(transform),
+            UiWidget::Custom(_) => None,
+        }
+    }
+
+    /// Convenience function to access widgets `UiImageBuilder`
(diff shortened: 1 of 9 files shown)


amethyst_ui/src/prefab.rs
```diff
@@ -513,56 +637,70 @@ type UiPrefabData<A = AudioFormat, I = TextureFormat, F = FontFormat> = (
...
         let mut d =
             Deserializer::from_bytes(&bytes).chain_err(|| "Failed deserializing Ron file")?;
-        let root: UiWidget<A, I, F> =
+        let root: UiWidget<A, I, F, C> =
             UiWidget::deserialize(&mut d).chain_err(|| "Failed parsing Ron file")?;
         d.end().chain_err(|| "Failed parsing Ron file")?;
 
         let mut prefab = Prefab::new();
-        walk_ui_tree(root, 0, &mut prefab);
+        walk_ui_tree(root, 0, &mut prefab, Default::default());
```
jojolepro: If you have to put default everywhere, doesn't that indicate that it should be taking Option?

amethyst_ui/src/prefab.rs
```diff
@@ -499,11 +617,17 @@ where
     None
 }
 
-type UiPrefabData<A = AudioFormat, I = TextureFormat, F = FontFormat> = (
+type UiPrefabData<
+    A = AudioFormat,
+    I = TextureFormat,
+    F = FontFormat,
+    CD = <NoCustomUi as ToNativeWidget>::PrefabData,
```
Rhuagh: Idiomatic rust suggests we use single letter type parameters.

amethyst_ui/src/prefab.rs
```diff
@@ -490,6 +511,103 @@ where
...
+pub trait ToNativeWidget<A = AudioFormat, I = TextureFormat, F = FontFormat>
+where
+    A: Format<Audio, Options = ()>,
+    I: Format<Texture, Options = TextureMetadata>,
+    F: Format<FontAsset, Options = ()>,
+    Self: Sized,
+{
+    /// Additional data used when loading UI prefab
+    type PrefabData: for<'a> PrefabData<'a> + Default + Send + Sync + 'static;
+    /// Create native `UiWidget` and custom prefab data from custom UI
```
Rhuagh: MIght be worth documenting that this could turn into infinite recursion unless care is taken when implementing the trait.

amethyst_ui/src/prefab.rs
jjedelsky: It could use `Option<ToNativeWidget::PrefabData>`, but that would mean that:
- in order to say that not custom data is used, custom UI implementation would have to use `ToNativeWidget::PrefabData = !` (never type is unstable, custom `enum Never {}` would need to implement `for<'a> PrefabData<'a>` and others which increases boilerplate) where in current form it may use `ToNativeWidget::PrefabData = ()`
- there is no way how to flag native UI defined in prefab (now you can use `default` to do it)

Implementation can use  `ToNativeWidget::PrefabData = Option<...>`.

jjedelsky:
> Have you seen this rfc? #1072
Yes and I believe that it would improve usability of UI a lot. However I think it take a while until it is implemented. Unless there is new progress?

amethyst_ui/src/prefab.rs
jjedelsky: Updated

amethyst_ui/src/prefab.rs
jjedelsky: Updated

examples/assets/ui/custom.ron
```diff
@@ -0,0 +1,25 @@
+#![enable(implicit_some)]
+Custom(
+    Repeat(
```
Xaeroxe: We're on the verge of being able to create a neat little "text shadow" effect here if we add color and z deltas for each repeat.

jojolepro: oh right! and if you draw the text with the black color and a bit bigger under the original, you also get borders

minecrawler: @jojolepro don't forget about letter spacing and font kerning, though, which might cause issues with this approach. Are advanced font rendering techniques like that even supported, yet?

jojolepro: not currently

jjedelsky:
> Can you add a change log entry in the docs/CHANGELOG.md file?

Done

jojolepro:
> Don't forget to add the associated link; see previous links in the same file for an example.

jjedelsky:
Oh, right, updated.

PR #1113, opened 2018-11-05
Introducing a very simple gradient skybox pass

photex:
Howdy!

This is the start of a simple skybox and I'm creating this PR before I'd call it "done" so that I can have some early feedback on what needs to happen to bring it into shape.

I'd say that for starters there are some places where values are hardcoded instead of uniform parameters that can be adjusted by game code. Additionally I haven't even tried any texturing.

The geometry is internally managed by the pass (just an inverted cube) but maybe it should be possible to also specify other basic geometry.

I've modified the arc_ball_camera example to use this pass.

### DIFF
--- a/amethyst_renderer/src/pass/shaders/vertex/skybox.glsl
+++ b/amethyst_renderer/src/pass/shaders/vertex/skybox.glsl
@@ -9,12 +9,10 @@ layout (std140) uniform VertexArgs {
 };
 
 in vec3 position;
-in vec3 normal;
 in vec2 tex_coord;
 
 out VertexData {
     vec3 position;
-    vec3 normal;
     vec2 tex_coord;
 } vertex;
 
@@ -24,7 +22,6 @@ void main() {
     vec4 vertex_position = model * vec4(position, 1.0);
 
     vertex.position = vertex_position.xyz;
-    vertex.normal = mat3(model) * normal;
     vertex.tex_coord = tex_coord;
 
     gl_Position = (proj * view_without_translation * vertex_position).xyww;
--- a/amethyst_renderer/src/pass/shaders/vertex/skybox.glsl
+++ b/amethyst_renderer/src/pass/shaders/vertex/skybox.glsl
@@ -1,0 +1,31 @@
+// TODO: Needs documentation.
+
+#version 150 core
+
+layout (std140) uniform VertexArgs {
+    uniform mat4 proj;
+    uniform mat4 view;
+    uniform mat4 model;
+};
+
+in vec3 position;
+in vec3 normal;
+in vec2 tex_coord;
+
+out VertexData {
+    vec3 position;
+    vec3 normal;
+    vec2 tex_coord;
+} vertex;
+
+void main() {
+    mat4 view_without_translation = view;
+    view_without_translation[3].xyz = vec3(0.0f, 0.0f, 0.0f);
+    vec4 vertex_position = model * vec4(position, 1.0);
+
+    vertex.position = vertex_position.xyz;
+    vertex.normal = mat3(model) * normal;
+    vertex.tex_coord = tex_coord;
+
+    gl_Position = (proj * view_without_translation * vertex_position).xyww;
+}
--- a/amethyst_renderer/src/pass/skybox/interleaved.rs
+++ b/amethyst_renderer/src/pass/skybox/interleaved.rs
@@ -1,0 +1,116 @@
+//! Skybox pass
+
+use amethyst_core::{
+    cgmath::{Matrix4, One},
+    specs::{Read, ReadStorage},
+    transform::GlobalTransform,
+};
+use {
+    error::Result,
+    get_camera,
+    pipe::{
+        pass::{Pass, PassData},
+        DepthMode, Effect, NewEffect,
+    },
+    pass::util::set_attribute_buffers,
+    set_vertex_args, ActiveCamera, Camera, Encoder, Factory, Mesh, Normal,
+    PosNormTex, Position, Query, Shape, TexCoord, VertexFormat,
+};
+use gfx::pso::buffer::ElemStride;
+use glsl_layout::{mat4, Uniform};
+use std::marker::PhantomData;
+
+use super::{VERT_SRC, FRAG_SRC};
+
+#[repr(C, align(16))]
+#[derive(Clone, Copy, Debug, Uniform)]
+pub(crate) struct VertexArgs {
+    proj: mat4,
+    view: mat4,
+    model: mat4,
+}
+
+/// Draw a simple gradient skybox
+///
+/// # Type Parameters:
+///
+/// * `V`: `VertexFormat`
+#[derive(Clone, Debug)]
+pub struct DrawSkybox<V> {
+    _marker: PhantomData<V>,
+    mesh: Option<Mesh>,
+}
+
+impl<V> DrawSkybox<V>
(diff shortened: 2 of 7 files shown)


amethyst_renderer/src/pass/skybox/interleaved.rs
```diff
@@ -0,0 +1,116 @@
...
+    model: mat4,
+}
+
+/// Draw a simple gradient skybox
+///
+/// # Type Parameters:
+///
+/// * `V`: `VertexFormat`
+#[derive(Derivative, Clone, Debug)]
+#[derivative(Default(bound = "V: Query<(Position, Normal, TexCoord)>"))]
```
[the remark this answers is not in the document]
photex: I wasn't able to derive Default without this... but then again.. is a 'default' impl really needed here?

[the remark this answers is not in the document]
photex: I removed the need for Default.

amethyst_renderer/src/pass/shaders/vertex/skybox.glsl
```diff
@@ -0,0 +1,31 @@
...
+
+out VertexData {
+    vec3 position;
+    vec3 normal;
+    vec2 tex_coord;
+} vertex;
+
+void main() {
+    mat4 view_without_translation = view;
+    view_without_translation[3][0] = 0.0f;
```
Rhuagh: `view_without_translation[3].xyz = vec3(0.0f, 0.0f, 0.0f);` maybe ?

amethyst_renderer/src/pass/skybox/interleaved.rs
```diff
@@ -0,0 +1,116 @@
...
+        effect: &mut Effect,
+        mut _factory: Factory,
+        (active, camera, global): <Self as PassData<'a>>::Data,
+    ) {
+        let camera = get_camera(active, &camera, &global);
+
+        let mesh = self
+            .mesh
+            .as_ref()
+            .expect("Failed to get reference to skybox mesh.");
```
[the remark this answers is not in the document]
photex: Updated.

photex:
The cube is no longer being inverted and I've set the effect up without back face culling.

In regards to the draw order I'm not setup correctly for that to work. Currently if it's not drawn first it'll completely overwrite anything previously rendered.

Rhuagh:
Just add depth testing for the pass and it shouldn't overdraw ?

amethyst_renderer/src/pass/shaders/vertex/skybox.glsl
photex: :+1: I continue to impress myself with my lack of swizzling

photex:
With `DepthMode::LessEqualTest` active in the effect, and the skybox pass draw first, things look correct.
If I then place it last it'll overwrite everything. 
If I use `DepthMode::LessEqualWrite` it'll replace the other passes regardless of the order of the passes.

photex:
Ok, issues with depth buffer resolved with assistance from Viral and sillesta in Discord!

photex:
Whoops, I see I'm missing some docs and rustfmt. I'll get that resolved in a bit (off to do some parent stuff now).

amethyst_renderer/src/pass/skybox/interleaved.rs
```diff
@@ -0,0 +1,118 @@
...
+where
+    V: Query<(Position, Normal, TexCoord)>,
+{
+    fn compile(&mut self, mut effect: NewEffect) -> Result<Effect> {
+        let verts = Shape::Cube.generate_vertices::<Vec<PosNormTex>>(None);
+        self.mesh = Some(Mesh::build(verts).build(&mut effect.factory)?);
+
+        effect
+            .simple(VERT_SRC, FRAG_SRC)
+            .without_back_face_culling()
```
magnonellie: You can probably enable culling, no?

amethyst_renderer/src/pass/skybox/interleaved.rs
```diff
@@ -0,0 +1,118 @@
...
+            .with_raw_constant_buffer(
+                "VertexArgs",
+                std::mem::size_of::<<VertexArgs as Uniform>::Std140>(),
+                1,
+            ).with_raw_vertex_buffer(PosNormTex::ATTRIBUTES, PosNormTex::size() as ElemStride, 0)
```
magnonellie: Preferably we wouldn't actually handle the vertices this way but instead have them directly within the shader since they're always gonna be a box. The synchronization is just necessary although I don't remember if you can do shader defined vertices in OpenGL... What you definitely can do is removing the vertex format since it doesn't matter what it actually is in the end.

amethyst_renderer/src/pass/skybox/interleaved.rs
photex: This was the only function I could see that enabled back faces. Is there another way to configure culling modes?

amethyst_renderer/src/pass/skybox/interleaved.rs
photex: Yeah, I admit that this is a bit of cargo-cult dev by starting with what other passes have done. The box only needs positions honestly.

photex:
gradient_skybox_01

Quick screencap of the results. (banding in gradient is due to gif compression)

amethyst_renderer/src/pass/skybox/interleaved.rs
torkleyy: But a skybox isn't always a box?

photex: I just tested an update that allows you to specify any `amethyst::renderer::Shape` you like and the way I've currently written the shader there is no real visual distinction between the results. 

That would change if it were possible to supply a texture or material I suppose. I'm happy to experiment with that a bit but perhaps its useful to start with the most simple thing and expand into the territory of being more flexible once the work on rendy is underway?

amethyst_renderer/src/pass/skybox/interleaved.rs
magnonellie: Ah okay so the mesh should really be generated differently. I'll probably clean this up later so this is fine for now :smile:

amethyst_renderer/src/pass/skybox/interleaved.rs
magnonellie: I still believe a skybox should be enough to cover most any cases, but I'll just trust you guys that people actually want to use different geometry :sweat_smile: 

That said, the other part of this comment is still valid and needs fixing. There's no point specifying which vertex type you can use and it just opens up for user error where they can waste memory unnecessarily.

photex: I'd very much like your advice regarding that. I do want to make it as simple as possible.

photex: By advice I mean help to understand more of how to implement this the way you suggest. I've only looked at other passes as examples so far.

photex: Thanks for the explanation on discord @magnonellie. I've updated the pass and have removed the type parameter for the vertex format.

magnonellie: Great! Now you just need to remove it from the shaders too and we're good to merge! :+1:

PR #1120, opened 2018-11-07
Minor fixes of pong tutorial

Canop:
No code change. I don't think this warrants a change in the CHANGELOG but I'm new to amethyst.


jojolepro:
bors r+

Thanks!

PR #1124, opened 2018-11-10
Fix offsets field in define_the_sprite_sheet.md

Techno-coder:
Add `Some(...)` around offsets array so the field can be properly deserialized. Without it, the asset loading errors with `Failed to parse Ron file for SpriteSheet`.


azriel91:
bors r=torkleyy,jojolepro,azriel91

PR #1125, opened 2018-11-10
Callback Queue: Future support

jojolepro:
This adds support to use futures (and any other asynchronous context) as simple callbacks.

If you want a more in-depth explanation, you can read the rfc here https://github.com/amethyst/rfcs/pull/4
Be aware that you need to know how async computing (futures & others) work to understand it.


jojolepro:
The rfc is just here to explain what I'm doing for those who can't see what problem this pr fixes. Feel free to ask questions there if you think this implementation could be better.

torkleyy:
Err..that's not an RFC then.

jojolepro:
"Fixed" :+1:

jojolepro:
Crossbeam avoids having a Mutex + EventChannel.

I could use mpsc, but apparently crossbeam's channel is more performant.

LucioFranco:
@jojolepro two things, could you possibly create a small example and I would like to see what actual benefits we get from using crossbeam-channel vs just std mpsc. I'd like to not bring in another dep.

jojolepro:
1) Small example.
https://github.com/amethyst/amethyst/pull/1125/files#diff-870ccdada9e859df05ecd0d7496f31a7R21

There isn't a full fledged example because there is no use case to do this in amethyst without adding dependencies.
The best example I could do would be to spawn a thread that sleeps for a second, and then triggers a Trans::Quit. Do you want me to do that?

2) Crossbeam
I didn't really check the differences between crossbeam's channel and mpsc. Someone suggested I use it. They got an explanation on why its better than mpsc here https://docs.rs/crossbeam-channel/0.3.1/crossbeam_channel/

azriel91:
Summary of discord discussion:

* crossbeam channels are `Send + Sync`, we'd need to use `std::mpsc::SyncSender` if we swap back to `std`
* crossbeam (in the past) was profiled to be faster than std channels, not sure of current state
* crossbeam is already a dependency of one of our dependencies, so we're not adding much compilation time

Leaving this open so @torkleyy can finish review

PR #1126, opened 2018-11-11
Added render utilities

jojolepro:


amethyst_utils/src/render.rs
```diff
@@ -0,0 +1,57 @@
+//! Module containing utility methods to easily create `Material` and `Texture` handles using a minimal amount of code.
```
torkleyy: The whole module seems to be about `albedo` only materials, this should be indicated somehow. Possibly in

* module path or
* documentation (or both)

jojolepro: Currently its like that, but it will have more eventually. I just ported code that I already had on hand.

Rhuagh:
Oh, and please add Progress parameters on the functions that create textures.

amethyst_utils/src/render.rs
```diff
@@ -0,0 +1,45 @@
...
+#[derive(SystemData)]
+pub struct MaterialCreator<'a> {
+    loader: ReadExpect<'a, Loader>,
+    storage: Read<'a, AssetStorage<Texture>>,
+    defaults: ReadExpect<'a, MaterialDefaults>,
+}
+
+impl<'a> MaterialCreator<'a> {
+    /// Generate a `Material` from a color.
+    pub fn material_from_color<T: Progress>(
```
torkleyy: Let's call this `material_from_albedo`

amethyst_utils/src/render.rs
```diff
@@ -0,0 +1,45 @@
...
+        progress_counter: T,
+    ) -> Material {
+        let albedo = self
+            .loader
+            .load_from_data(color.into(), progress_counter, &self.storage);
+        self.material_from_texture(albedo)
+    }
+
+    /// Generate a `Material` from a path pointing to a png image.
+    pub fn material_from_png<T: Progress>(&self, path: String, progress_counter: T) -> Material {
```
torkleyy: Could use `S: Into<String>`

amethyst_utils/src/render.rs
```diff
@@ -0,0 +1,45 @@
...
+        color: [f32; 4],
+        progress_counter: T,
+    ) -> Material {
+        let albedo = self
+            .loader
+            .load_from_data(color.into(), progress_counter, &self.storage);
+        self.material_from_texture(albedo)
+    }
+
+    /// Generate a `Material` from a path pointing to a png image.
```
torkleyy: Should specify the source used.

jojolepro: file system path?

amethyst_utils/src/render.rs
jojolepro: Technically its material_from_albedo_from_color. Are you sure you want it called albedo?

Rhuagh:
Looks good, just one nit: Some examples in the docs on usage would be stellar

amethyst_utils/src/render.rs
torkleyy: Dismissing...

jojolepro:
ready

jojolepro:
zzz this is going to miss the release because of the missing review.

amethyst_utils/src/render.rs
```diff
@@ -0,0 +1,58 @@
...
+            path,
+            PngFormat,
+            TextureMetadata::srgb(),
+            progress_counter,
+            &self.storage,
+        ))
+    }
+
+    /// Generate a `Material` from a texture handle.
+    pub fn material_from_texture(&self, texture: Handle<Texture>) -> Material {
```
termhn: I'm not sure how useful this actually is at the moment, it seems like just doing the contained code whenever you want a texture with an albedo is pretty simple... perhaps we could make this https://github.com/amethyst/amethyst/blob/master/amethyst_renderer/src/pass/util.rs#L24 a public part of the `mtl` module and then make this method `material_from_textures` something like:

```rs
pub fn material_from_textures(&self, textures: &[(TextureType, Handle<Texture>)] -> Material {
    ...
}
```
Which inserts the given textures into the correct slots and then uses default for the rest?

jojolepro:
Waiting for rhuagh's review to squash.

Rhuagh:
@jojolepro sorry for the holdup, lgtm

jojolepro:
We'll let it build then we can merge :3

jojolepro:
bors r+

you can do it!

jojolepro:
yaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa yeet

PR #1127, opened 2018-11-11
Auto Save System

jojolepro:


amethyst_utils/src/autosave.rs
```diff
@@ -0,0 +1,95 @@
...
+};
+
+/// System used to automatically save a Resource T to a file.
+/// On load, it will attempt to load it from the file and if it fails, it will use T::default().
+/// The resource in question will be wrapped into a `Dirty<T>` value inside of specs to keep track of changes made to the resource.
+/// This `System` will save the resource each time there is a modification.
+/// It is best used with resources that are modified less than once every second.
+pub struct AutoSaveSystem<T> {
+    /// Absolute path.
+    save_path: String,
```
torkleyy: Please use the path API.

amethyst_utils/src/autosave.rs
```diff
@@ -0,0 +1,95 @@
...
+            } else {
+                error!("Failed to read content of save file: {}", self.save_path);
+            }
+        } else {
+            warn!(
+                "Failed to load save file: {}. It will be created during the next save.",
+                self.save_path
+            );
+        }
+        Self::SystemData::setup(res);
```
torkleyy: Why do this at the end?

amethyst_utils/src/autosave.rs
```diff
@@ -0,0 +1,95 @@
...
+    }
+    fn run(&mut self, (mut data,): Self::SystemData) {
+        if data.dirty() {
+            data.clear();
+            let value = data.read();
+            let string_data = ron::ser::to_string(&value).expect(&format!(
```
torkleyy: We should allow arbitrary seriliazation logic. So, there should be two type parameters, one for serialization and one for deserialization, and two closures accepting `Write` / `BufRead`.

amethyst_utils/src/autosave.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+impl<'a, T> System<'a> for AutoSaveSystem<T>
+where
+    T: Serialize + DeserializeOwned + Default + Send + Sync + 'static,
+{
+    type SystemData = (Write<'a, Dirty<T>>,);
+    fn setup(&mut self, res: &mut Resources) {
+        // attempt loading
+        if let Ok(mut f) = File::open(&self.save_path) {
```
torkleyy: Can you please use `map` logic instead of nesting if / else?

amethyst_utils/src/autosave.rs
```diff
@@ -0,0 +1,95 @@
...
+use std::{
+    fs::File,
+    io::{Read, Write as WriteIo},
+    marker::PhantomData,
+};
+
+/// System used to automatically save a Resource T to a file.
+/// On load, it will attempt to load it from the file and if it fails, it will use T::default().
+/// The resource in question will be wrapped into a `Dirty<T>` value inside of specs to keep track of changes made to the resource.
+/// This `System` will save the resource each time there is a modification.
```
torkleyy: That's hardly a good strategy. I'm not sure if we should put that into Amethyst. The proper way would be a simple event / flag that indicates the resource should be saved. That would also remove the need to wrap it in `Dirty`.

jojolepro: The other way is to have a trait on the resource, but I figured it wouldn't be usable for things already implemented like DisplayConfig

DoumanAsh: I think having file per resource is not really good idea as it would either force user to have big resources or accept many files

amethyst_utils/src/autosave.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// Absolute path.
+    save_path: String,
+    _phantom_data: PhantomData<T>,
+}
+
+impl<T> AutoSaveSystem<T> {
+    /// Create a new `AutoSaveSystem`.
+    /// Save path is an absolute path.
+    pub fn new(save_path: String) -> Self {
+        AutoSaveSystem {
```
DoumanAsh: Add `debug_assert!` to assert that path is aboslute

jojolepro:
Thanks for the reviews, but I'm afraid I'll have to close this PR.

>I think having file per resource is not really good idea as it would either force user to have big resources or accept many files

With the way I designed this system, this cannot really be avoided, so I'll scrap that idea, or come up with a better design.

Issue #1110, opened 2018-11-04
Split TimedDestroySystem


jojolepro:
Small nitpick. TimedDestroySystem can be split into two systems to improve parallelism instead of checking DestroyAtTime then DestroyInTime sequentially.

So the logic can be separated into two systems, one for each of the Destroy*Time components.

If someone from specs thinks it will make the performance worse by doing that, let me know. (keep in mind its possible to have only one of the 2 system in the dispatcher, removing a join call)

torkleyy:
If the resources do not intersect (so not even immutably), it is an improvement for sufficiently large input.

Xaeroxe:
> so not even immutably

Huh?  I thought that was the whole point of having Read separate from Write.

Xaeroxe:
That's not what I understood from his comment.

If I have SystemA with SystemData

```
Read<'a, ResourceA>,
Write<'a, ResourceB>,
```

and SystemB with

```
Read<'a, ResourceA>,
Write<'a, ResourceC>,
```

then these systems should be able to run at the same time, even though they intersect immutably.

torkleyy:
Yes, it is. There might also be an improvement if they only intersect mutably, but that depends.
There are two extra costs:

* virtual dispatch
* atomic operations

The latter one will be duplicated if dependencies intersect, even `Read`. But all this is super micro, so it **doesn't matter at all**.

So generally, and I'm repeating myself, just do what makes sense logically and don't try to count all the (negligible) micro costs.

TL;DR: DO IT :)

jojolepro:
;) 
Cleaner code yay!

Rhuagh:
DOOOEEEEEETTT!

torkleyy:
Sounds good 👍

PR #1129, opened 2018-11-11
Split `TimedDestroySystem`.

re-l124c41plus:
Split into `DestroyAtTimeSystem` and `DestroyInTimeSystem`.
See #1110.


torkleyy:
Ah, this is a breaking change, so a note in the changelog would make sense.

PR #1133, opened 2018-11-12
use reexported types instead of log::LevelFilter

qqwa:


LucioFranco:
Thank you! 🎉 

bors r=Rhuagh,azriel91,jojolepro,TimonPost

Issue #1136, opened 2018-11-13
Implement laminar


TimonPost:
Change the raw UDP socket to use laminar.

TimonPost:
See #1137

PR #1137, opened 2018-11-13
Replace std sockets with laminar sockets

TimonPost:
I've implemented laminar for `amethyst_network`. 

Solves: #1136

### DIFF
--- a/amethyst_network/Cargo.toml
+++ b/amethyst_network/Cargo.toml
@@ -29,3 +29,4 @@ log = "0.4"
 uuid = { version = "0.7.1", features = ["v4","serde"] }
 thread_profiler = { version = "0.3" , optional = true }
 fern = "0.5"
+laminar = "0.1"
--- a/amethyst_network/src/lib.rs
+++ b/amethyst_network/src/lib.rs
@@ -9,6 +9,7 @@ extern crate log;
 extern crate serde;
 extern crate bincode;
 extern crate fern;
+extern crate laminar;
 extern crate shred;
 extern crate shrev;
 extern crate uuid;
@@ -28,14 +29,16 @@ pub use {
     network_socket::NetSocketSystem,
 };
 
-use std::net::{SocketAddr, UdpSocket};
+use std::net::SocketAddr;
 
 use bincode::{deserialize, serialize, ErrorKind};
+use laminar::net::UdpSocket;
+use laminar::Packet;
 use serde::{de::DeserializeOwned, Serialize};
 
 /// Sends an event to the target NetConnection using the provided network Socket.
 /// The socket has to be bound.
-pub fn send_event<T>(event: &NetEvent<T>, target: &SocketAddr, socket: &UdpSocket)
+pub fn send_event<T>(event: &NetEvent<T>, addr: &SocketAddr, socket: &mut UdpSocket)
 where
     T: Serialize,
 {
@@ -43,7 +46,8 @@ where
     match ser {
         Ok(s) => {
             let slice = s.as_slice();
-            match socket.send_to(slice, target) {
+            // send an unreliable `Packet` from laminar which is basically just a bare UDP packet.
+            match socket.send(&Packet::unreliable(*addr, slice.to_vec())) {
                 Ok(_qty) => {}
                 Err(e) => error!("Failed to send data to network socket: {}", e),
             }
--- a/amethyst_network/src/network_socket.rs
+++ b/amethyst_network/src/network_socket.rs
@@ -3,14 +3,16 @@
 use std::{
     clone::Clone,
     io::{Error, ErrorKind},
-    net::{SocketAddr, UdpSocket},
+    net::SocketAddr,
     sync::mpsc::{channel, Receiver, Sender},
     thread,
 };
 
-use serde::{de::DeserializeOwned, Serialize};
-
 use amethyst_core::specs::{Join, Resources, System, SystemData, WriteStorage};
+use laminar::error;
+use laminar::net::UdpSocket;
+use laminar::{DeliveryMethod, NetworkConfig, Packet};
+use serde::{de::DeserializeOwned, Serialize};
 
 use super::{deserialize_event, send_event, ConnectionState, NetConnection, NetEvent, NetFilter};
 
@@ -26,6 +28,28 @@ struct RawEvent {
     pub byte_count: usize,
     pub data: Vec<u8>,
     pub source: SocketAddr,
+    pub delivery_method: DeliveryMethod,
+}
+
+impl From<Packet> for RawEvent {
+    fn from(packet: Packet) -> Self {
+        RawEvent {
+            byte_count: packet.payload().len(),
+            data: packet.payload().to_vec(),
+            source: packet.addr(),
+            delivery_method: packet.delivery_method(),
+        }
+    }
+}
+
+impl From<RawEvent> for Packet {
+    fn from(raw_event: RawEvent) -> Self {
+        Packet::new(
+            raw_event.source,
+            raw_event.data.into_boxed_slice(),
+            raw_event.delivery_method,
+        )
+    }
 }
 
 // If a client sends both a connect event and other events,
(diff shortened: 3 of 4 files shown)


amethyst_network/src/network_socket.rs
```diff
@@ -77,41 +80,49 @@ where
...
                         // Data received
-                        Ok((amt, src)) => {
-                            if let Err(_) = receive_queue.send(RawEvent {
-                                byte_count: amt,
-                                data: buf[..amt].to_vec(),
-                                source: src,
-                            }) {
+                        Ok(Some(packet)) => {
+                            let payload = packet.payload();
+                            let raw_event = RawEvent {
```
LucioFranco: We should implement From<Packet> for raw packet here

amethyst_network/src/lib.rs
```diff
@@ -28,22 +29,24 @@ pub use {
...
+pub fn send_event<T>(event: &NetEvent<T>, addr: &SocketAddr, socket: &mut UdpSocket)
 where
     T: Serialize,
 {
     let ser = serialize(event);
     match ser {
         Ok(s) => {
             let slice = s.as_slice();
-            match socket.send_to(slice, target) {
+            match socket.send(&Packet::unreliable(*addr, slice.to_vec())) {
```
LucioFranco: I thought we moved to using slices?

amethyst_network/src/network_socket.rs
```diff
@@ -77,41 +80,49 @@ where
...
+                                source: packet.addr(),
+                            };
+
+                            if let Err(_) = receive_queue.send(raw_event) {
                                 error!("`NetworkSocketSystem` was dropped");
                                 break 'outer;
                             }
-                        }
+                        },
+                        Ok(None) => { info!("Received data, but awaiting still other data.") }
```
LucioFranco: Do we really need to log this?

amethyst_network/Cargo.toml
```diff
@@ -29,3 +29,4 @@ log = "0.4"
 uuid = { version = "0.7.1", features = ["v4","serde"] }
 thread_profiler = { version = "0.3" , optional = true }
 fern = "0.5"
+laminar = "0.1.0"
```
LucioFranco: This should be “0.1”

amethyst_network/src/lib.rs
TimonPost: Our public API  not. I thought too. We can make an issue in laminar for this one.

amethyst_network/src/network_socket.rs
TimonPost: It's up to you, we could just ignore it.

TimonPost: I'll remove it.

amethyst_network/src/lib.rs
LucioFranco: Yes, please open an issue in laminar and add a `TODO` with the issue number in laminar. so that we can keep track of this.

amethyst_network/src/network_socket.rs
```diff
@@ -28,6 +30,16 @@ struct RawEvent {
     pub source: SocketAddr,
 }
 
+impl From<Packet> for RawEvent {
+    fn from(packet: Packet) -> Self {
+        RawEvent {
+            byte_count: packet.payload().len(),
+            data: packet.payload().to_owned(),
```
LucioFranco: I think you can drop this `to_owned` now? you should be able to move the payload out.

amethyst_network/src/network_socket.rs
```diff
@@ -28,6 +30,16 @@ struct RawEvent {
     pub source: SocketAddr,
 }
 
+impl From<Packet> for RawEvent {
```
LucioFranco: I would also maybe implement `From<RawEvent>` for `Packet`.

amethyst_network/src/network_socket.rs
```diff
@@ -77,43 +90,44 @@ where
...
+                                        break;
+                                    } else {
+                                        error!("Could not receive datagram: {}", e);
+                                    }
+                                },
+                                _ => { error!("Could not receive datagram: {:?}", e); }
                             }
-                        }
+                        },
+                        _ => {},
```
LucioFranco: I would prefer to be explicit here, and use `Ok(None) => {}`.

amethyst_network/src/network_socket.rs
```diff
@@ -3,13 +3,15 @@
...
     io::{Error, ErrorKind},
-    net::{SocketAddr, UdpSocket},
+    net::SocketAddr,
     sync::mpsc::{channel, Receiver, Sender},
     thread,
 };
 
 use serde::{de::DeserializeOwned, Serialize};
-
+use laminar::net::{UdpSocket};
```
LucioFranco: CI might yell at you for this, make sure you are rustfmt your code.

amethyst_network/src/lib.rs
TimonPost: It's going to be difficult since `Packet` would need a lifetime of function `send()` I think.

amethyst_network/src/network_socket.rs
TimonPost: Not sure if we can implement something on a type we don't own and is from another crate.

amethyst_network/src/network_socket.rs
TimonPost: I am not sure but you move ownership of `Packet` to this function I think you won't be able to use it afterward at all.

amethyst_network/src/network_socket.rs
TimonPost: Fixed it thanks!

amethyst_network/src/lib.rs
torkleyy: Should the reliability be documented?

TimonPost: It is in laminar-tester book. And later on we should make this configurable for the user of amethyst network.

amethyst_network/src/network_socket.rs
```diff
@@ -62,7 +83,8 @@ where
             warn!("Using a port below 1024, this will require root permission and should not be done.");
         }
 
-        let socket = UdpSocket::bind(addr)?;
+        let mut socket = UdpSocket::bind(addr, NetworkConfig::default())
```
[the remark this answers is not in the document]
TimonPost: True

amethyst_network/src/lib.rs
torkleyy: Yes, but I meant for the docs on this function. Even if it's going to change, I think documentation would be nice.

amethyst_network/src/lib.rs
```diff
@@ -58,4 +61,4 @@ where
     T: DeserializeOwned,
 {
     deserialize::<NetEvent<T>>(data)
-}
+}
```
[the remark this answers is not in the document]
TimonPost: Not sure what you mean here.... Like `deserialize` stops at 63 and the brace is at 64. That's normal right?

amethyst_network/src/lib.rs
TimonPost: Oke, let me do that.

LucioFranco: Yeah, just add a doc that at least links to laminar docs explaining what type of message is sent. thanks.

LucioFranco:
@TimonPost I think we're overall happy with this, can you rustfmt your code and make CI happy?

amethyst_network/src/lib.rs
```diff
@@ -28,22 +29,25 @@ pub use {
...
 
 use bincode::{deserialize, serialize, ErrorKind};
+use laminar::net::UdpSocket;
+use laminar::Packet;
 use serde::{de::DeserializeOwned, Serialize};
 
 /// Sends an event to the target NetConnection using the provided network Socket.
 /// The socket has to be bound.
-pub fn send_event<T>(event: &NetEvent<T>, target: &SocketAddr, socket: &UdpSocket)
+pub fn send_event<T>(event: &NetEvent<T>, addr: &SocketAddr, socket: &mut UdpSocket)
```
LucioFranco: Just a thought but this would be a good thing to turn into a trait.

```rust
trait Transport {
  type Event;
  type Error;

  fn send_event(&mut self, addr: &SocketAddr, event: &Event);

  fn receive_event(&self) -> Result<Event, Error>;
}
```

LucioFranco:
@TimonPost can you squash your commits btw?

TimonPost:
bors r=jojolepro,LucioFranco

TimonPost:
bors r=jojolepro,LucioFranco

PR #1144, opened 2018-11-14
Bugfix: Dynamically generated texture buffers have the wrong size

Stoeoef:
Hi there!
I've stumbled upon an issue when trying to create a dynamic texture (that is, not loaded from a file). The size of the opengl texture buffer for non-image textures is always 1x1 (that's why defining a texture with only one color works), but custom sizes defined via `TextureMetadata` are 'forgotten'.

I've never done any serious error handling in my other projects, so please have a look at the error that I use to indicate that a texture of invalidate size is created. Also, the code does not `Err` if "too many" pixels are defined... that probably also should result in an error.

### DIFF
--- a/amethyst_renderer/src/tex.rs
+++ b/amethyst_renderer/src/tex.rs
@@ -176,11 +176,19 @@ where
             let (w, h, _, _) = self.info.kind.get_dimensions();
             let w = w as usize;
             let h = h as usize;
+            if w * h * pixel_width != data.len() {
+                let error = format!(
+                    "Texture size mismatch: Expected pixel data vector of length {:?} (actual: {:?})",
+                    w * h * pixel_width,
+                    data.len()
+                );
+                bail!(::error::Error::PixelDataMismatch(error))
+            }
             for y in 0..h {
                 for x in 0..(w * pixel_width) {
                     v_flip_buffer.push(data[x + (h - y - 1) * w * pixel_width]);
                     // Uncomment this if you need to debug this.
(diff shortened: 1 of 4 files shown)


amethyst_renderer/src/tex.rs
```diff
@@ -176,11 +176,14 @@ where
             let (w, h, _, _) = self.info.kind.get_dimensions();
             let w = w as usize;
             let h = h as usize;
+            if w * h * pixel_width > data.len() {
```
torkleyy: Shouldn't this check for equality?

amethyst_renderer/src/tex.rs
```diff
@@ -176,11 +176,14 @@ where
...
+                return Err(CreationError::Data(data.len()).into());
```
torkleyy: This seems like it wouldn't result in a very helpful error message.

Stoeoef:
Thanks for the feedback!
I've updated the PR accordingly.

jojolepro:
Name suggestion: `PixelSizeMismatch`

torkleyy:
Can somebody confirm the examples all work?

Stoeoef:
I've tested most examples, all look fine. The gltf example did not work properly, however, a rebase onto the current master solved this - so I guess that is related to other changes.
All examples seem to either load an image (which is unaffected by this change) or used a single color as texture, which "accidentally" worked before, as they all had dimensions 1x1.

PR #1146, opened 2018-11-15
Introduce SystemExt to support pausable systems

udoprog:
This is a work in progress for better support to pause systems, see #1147 for more context.

### DIFF
--- a/amethyst_core/src/system_ext.rs
+++ b/amethyst_core/src/system_ext.rs
@@ -1,4 +1,4 @@
-//! Extra system utilities.
+//! Extension system utilities.
 //!
 //! This modules contains an extension trait for the System trait which adds useful transformation
 //! functions.
@@ -6,8 +6,8 @@
 use shred::{RunningTime, SystemData};
 use specs::prelude::{Read, System};
 
-/// Extra functionality associated systems.
-pub trait SystemExtra {
+/// Extension functionality associated systems.
+pub trait SystemExt {
     /// Make a system pausable by tying it to a specific value of a resource.
     ///
     /// # Examples
@@ -69,7 +69,7 @@ pub trait SystemExtra {
         V: Send + Sync + Default + Eq;
 }
 
-impl<'s, S> SystemExtra for S
+impl<'s, S> SystemExt for S
 where
     S: System<'s>,
 {
--- a/amethyst_core/src/system_ext.rs
+++ b/amethyst_core/src/system_ext.rs
@@ -20,7 +20,7 @@ pub trait SystemExt {
     ///     prelude::*,
     /// };
     ///
-    /// #[derive(PartialEq, Eq)]
+    /// #[derive(PartialEq)]
     /// enum CurrentState {
     ///     Disabled,
     ///     Enabled,
@@ -66,7 +66,7 @@ pub trait SystemExt {
     fn pausable<V: 'static>(self, value: V) -> Pausable<Self, V>
     where
         Self: Sized,
-        V: Send + Sync + Default + Eq;
+        V: Send + Sync + Default + PartialEq;
 }
 
 impl<'s, S> SystemExt for S
@@ -76,7 +76,7 @@ where
     fn pausable<V: 'static>(self, value: V) -> Pausable<Self, V>
     where
         Self: Sized,
-        V: Send + Sync + Default + Eq,
+        V: Send + Sync + Default + PartialEq,
     {
         Pausable {
             system: self,
@@ -95,7 +95,7 @@ impl<'s, S, V: 'static> System<'s> for Pausable<S, V>
 where
     S::SystemData: SystemData<'s>,
     S: System<'s>,
-    V: Send + Sync + Default + Eq,
+    V: Send + Sync + Default + PartialEq,
 {
     type SystemData = (Read<'s, V>, S::SystemData);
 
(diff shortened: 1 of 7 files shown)


src/system_extra.rs
```diff
@@ -0,0 +1,105 @@
...
+    ///     fn run(&mut self, mut number: Self::SystemData) {
+    ///         *number = self.0;
+    ///     }
+    /// }
+    ///
+    /// let mut world = World::new();
+    ///
+    /// let mut dispatcher = DispatcherBuilder::default()
+    ///     .with(Example(42), "set_number", &[])
+    ///     .with(Example(84).enabled_on(State::Example), "set_number_2", &[])
```
torkleyy: Thinking about this, this only allows intersection, but no unions. I'm wondering if we can allow both.

udoprog: Interesting!

I messed around with two different approaches:
* Store a collection to check if the enum is a member of that collection (e.g. `HashSet<State>`).
* Store the value as-is right now, but implement more complex testing associated with the system (e.g. `OneOf<HashSet<State>>`).

I landed on the second approach, because it ended up being significantly cleaner. More generic parameters could be elided instead of explicitly specified.

Here it is:
https://github.com/udoprog/amethyst/blob/enabled-system-predicates/src/system_extra.rs#L182

What do you think?

torkleyy: Part of me likes the pattern, but at the same time I'm unsure if this is actually helpful. I think if we show in the docs how to do more complex things (with concrete types), it helps them more than a generic API. So my suggestion to go forward with this would be making this very simple, just a `Pausable` system without any values and to show in the docs and book how to do more complex checks.

udoprog: Oh, I think I understand now, you were expecting this to be used only for pausing a _single_ set of systems?

I envisioned this to be used with having a set of systems _per state_ that should be running, not only for the pause functionality.

As an example, you have a set of systems running for:
* The game menu (UI + Menu Inputs).
* The main game (Game UI + Game Logic + Game Inputs).
* The game is paused (Pause Inputs).

src/system_extra.rs
```diff
@@ -0,0 +1,112 @@
...
+    ///
+    /// ```rust
+    /// use amethyst::{
+    ///     ecs::{System, Write},
+    ///     shred::DispatcherBuilder,
+    ///     prelude::*,
+    /// };
+    ///
+    /// #[derive(PartialEq, Eq)]
+    /// enum State {
```
jojolepro: the `State` name is confusing for pretty obvious reasons. Mind changing it? :)

udoprog: Not at all!

amethyst_core/src/system_extra.rs
```diff
@@ -0,0 +1,113 @@
+//! Extra system utilities.
+//!
+//! This modules contains an extension trait for the System trait which adds useful transformation
+//! functions.
+
+use shred::{RunningTime, SystemData};
+use specs::prelude::{Read, System};
+
+/// Extra functionality associated systems.
+pub trait SystemExtra {
```
torkleyy: Okay, this is the only thing that's debatable IMO:

* should this be in core? What's the reasoning for it?
* should we call it `SystemExtra` (which implies more functionality will go into this)?

jojolepro: @torkleyy see my review for the move to amethyst_core.

I'm pretty sure that could even be moved to an external `specs_pausable` crate if they really wanted too. Its generic enough to be that reusable.

torkleyy: I'm just don't think this is a core thing. I'll let a third party decide.

LucioFranco:
@udoprog What about `SystemExt`? System extension

LucioFranco:
@udoprog I think all that is left is to fix the merge conflict and we can merge!

amethyst_core/src/system_ext.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+impl<'s, S> SystemExt for S
+where
+    S: System<'s>,
+{
+    fn pausable<V: 'static>(self, value: V) -> Pausable<Self, V>
+    where
+        Self: Sized,
+        V: Send + Sync + Default + Eq,
```
azriel91: Can we relax the `Eq` to `PartialEq`? In case the resource used to determine pausing is `PartialEq` only, e.g. an enum with a variant that contains `f32`

amethyst_core/src/system_ext.rs
```diff
@@ -0,0 +1,113 @@
...
+    ///     prelude::*,
+    /// };
+    ///
+    /// #[derive(PartialEq, Eq)]
+    /// enum CurrentState {
+    ///     Disabled,
+    ///     Enabled,
+    /// }
+    ///
+    /// impl Default for CurrentState {
```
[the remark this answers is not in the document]
udoprog: Non-default resources require `ReadExpect`, and have the "Tried to fetch a resource" error when they can't be found which is rather cryptic:

```
Tried to fetch a resource, but the resource does not exist.
Try adding the resource by inserting it manually or using the `setup` method.
You can get the type name of the missing resource by enabling `shred`'s `nightly` feature
```

This has generally tipped me more in favor of using resources which `Default`, since it guards against this kind of error in the type system.

PR #1150, opened 2018-11-15
Fix regression in GLTF loading after #1075 (fixes #1145)

### LINKED ISSUE #1145: GLTF example does not render after refactor
Puffy doesn't render (though it seems the animation logic is working). Just get a black screen.

image

```
...
[DEBUG][amethyst_assets::storage] "animation::Sampler": Asset "<Data>" (handle id: Handle { id: 232 }) has been loaded successfully[DEBUG][amethyst_assets::storage] "PREFAB": Asset "mesh/puffy.gltf" (handle id: Handle { id: 0 }) has been loaded successfully
[DEBUG][amethyst_assets::storage] "PREFAB": Asset "prefab/puffy_scene.ron" (handle id: Handle { id: 0 }) has been loaded successfully
Running animation 0
Running animation 1
[INFO][amethyst::app] Engine is shutting down
```

The example works on ac851627 (after #1116), but not c50e2709 (after #1075). I've spent a while re-reviewing the code but can't find it. Someone else might have better eyes for this.

Reproducible on Ubuntu and Windows.

udoprog:
We should probably have a test case for this.

### DIFF
--- a/amethyst_gltf/src/format/mod.rs
+++ b/amethyst_gltf/src/format/mod.rs
@@ -357,7 +357,7 @@ fn load_node(
             // we need to add each primitive as a child entity to the node
             for (mesh, material_index, bounds) in graphics {
                 let mesh_entity = prefab.add(Some(entity_index), None);
-                let prefab_data = prefab.data_or_default(entity_index);
+                let prefab_data = prefab.data_or_default(mesh_entity);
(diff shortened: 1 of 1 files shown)


azriel91:
Whoa thank you! Nice eyes 👀

Issue #1148, opened 2018-11-15
Add InputEvent<T> to StateEvent


jojolepro:
By default, users and amethyst examples are using the winit VirtualKeyCode to handle keyboard events. This should be changed to show goos pratices by using the events coming out of the EventChannel<InputEvent<AC>> in State::handle_event. To be able to do this, there will need to be a variant Input(InputEvent<T>) added to StateData. The generic type on StateData should default to String.

udoprog:
I'll rephrase what I raised in another issue here.

If this is implemented we would have two places where we handle input events. For documentation purposes, and to reduce confusion, I'd like clarification on what is expected to be handled where. Or that we have a clear signal what the preferred way of handling input is.

I prefer handling _all_ input in systems. There is a clearly defined data dependency, and the handler is managed by the ECS like almost everything else. I'm curious what others think about this.

jojolepro:
I only handle inputs in states **if** they are directly triggering a state `Trans`.
Otherwise, I put everything else into systems too. (There's no harm in handling input in multiple places)

torkleyy:
Request by @jojolepro, let's implement this for now and discuss alternatives or improvements later given this is a big issue that would block progress on this.

fhaynes:
Closing this due to age and lack of activity....and it was resolved, maybe?

jojolepro:
I'll try implementing it tomorrow. Reopening so I don't forget.

PR #1154, opened 2018-11-15
book: how to interact between states and systems using resources

udoprog:
This adds concrete documentation for how to actually implement the state transitions hinted at during the `State` section. This was one of the pain points I encountered while [writing my game](https://github.com/udoprog/asteroids-amethyst).

I'm also actively working on trying to improve the ergonomics around this approach, so far the only relevant issue is: #1146 - but in general I'm trying to create a stronger association between a group of states and systems. If this becomes relevant, I'll update the documentation to reflect it :).

The other approach I'm aware of to handle this would be to set up event channels. But that section doesn't really talk about how to set up readers in states (it probably should!). Event channels is however a bit more complicates since the reader needs to be setup.

## Future TODOs?

Add concrete documentation for input handling - I mention adding an action called `open_menu`, but there's no great page to link to that discusses input handling generically, apart from [`pong-tutorial-03.md`](https://github.com/amethyst/amethyst/blob/master/book/src/pong-tutorial/pong-tutorial-03.md).
It would be great if we had a canonical section that dealt with only input handling in particular!


book/src/concepts/system.md
```diff
@@ -290,6 +290,153 @@ write_storage.remove(entity);
...
+    ecs::{System, prelude::*},
+    input::InputHandler,
+};
+
+struct MyFirstSystem;
+
+impl<'s> System<'s> for MyFirstSystem {
+    type SystemData = (
+        Read<'s, InputHandler<String, String>>,
+        Write<'s, Gameplay>,
```
jojolepro: Also, what you can do is react to `InputEvent<String>` events in the state's handle method. Those are the events emitted by the InputHandler. The strings you get is the key used in your input configuration (actions).

Note: We need to add `InputEvent<T = String>` to the default `StateEvent` of amethyst for events to actually be forwarded to the State::handle_event method. I opened an issue for this yesterday as a reminder.

book/src/concepts/system.md
```diff
@@ -290,6 +290,153 @@ write_storage.remove(entity);
 
 Keep in mind that inserting a component on an entity that already has a component of the same type **will overwrite the previous one**.
 
+## Changing states through resources
+
+In a previous section we talked about [`States`][s], and how they are used to organize your game
+into different logical sections.
+Since systems are responsible for processing user input, they must somehow also be responsible for
+causing a switch in states.
```
jojolepro: pssst, that isn't exactly true. The best way to do it is to have systems only to data processing (change components, modify resources, render things). When you have systems that manage states, they get too specialized and less re-usable. 

The best practice for those case would be to emit an event specific to your game from the System. For example, a "TogglePause" event. Then, inside of your state's handle_event method, you can react to this event (in this case, by returning a Trans when you see the "TogglePause" event).

udoprog: > pssst, that isn't exactly true. The best way to do it is to have systems only to data processing (change components, modify resources, render things). When you have systems that manage states, they get too specialized and less re-usable.
> 
> The best practice for those case would be to emit an event specific to your game from the System. For example, a "TogglePause" event. Then, inside of your state's handle_event method, you can react to this event (in this case, by returning a Trans when you see the "TogglePause" event).

To me it seems like a system still seems to be responsible for the state change? How would you like the phrasing to change?

book/src/concepts/system.md
udoprog: This does cover state transitions caused immediately by keyboard inputs, but not by other things, like the player dying because of a collision.

Honestly though, it seemed a bit off to me to handle input events both in states and in systems. I'd much rather have a clear signal to a new user to do it one way or another. I'd love more input on this though.

udoprog:
@jojolepro thanks for the review! <3

book/src/concepts/system.md
```diff
@@ -290,6 +290,153 @@ write_storage.remove(entity);
...
+            }
+            // do nothing for other states.
+            _ => {}
+        }
+    }
+}
+```
+
+Now whenever you are playing the game and you press the button associated with the `open_menu`
+action, the `MenuState` will resume and the the `GameplayState` will pause.
```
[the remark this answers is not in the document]
udoprog: Thanks, I ended up renaming this, but this line also had a typo (`the the`) which I missed.

book/src/concepts/system.md
```diff
@@ -290,6 +290,153 @@ write_storage.remove(entity);
...
+> A [`Resource`][r] is any type that stores data that you might need for your game AND that is not
+> specific to an entity.
+
+The data in a resource is available both to systems and resources.
+We can use this to our advantage!
+
+Let's say you have the following two states:
+
+ * `MainMenuState` - which starts the game by transitioning into the GameplayState.
+ * `GameplayState` - which we enter after the main menu, while the game is running.
```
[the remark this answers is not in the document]
udoprog: Done!

book/src/concepts/system.md
```diff
@@ -290,6 +290,153 @@ write_storage.remove(entity);
...
 Keep in mind that inserting a component on an entity that already has a component of the same type **will overwrite the previous one**.
 
+## Changing states through resources
+
+In a previous section we talked about [`States`][s], and how they are used to organize your game
+into different logical sections.
+Since systems are responsible for processing user input, they must somehow also be responsible for
+causing a switch in states.
+
+So how can we effect states from systems?
```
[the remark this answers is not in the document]
udoprog: Thanks, fixed!

udoprog:
@azriel91 

> It sounds as though MainMenuState is below the GameplayState in the stack, but the code has the GameplayState as the bottom state.

I renamed the `MainMenuState` to `GameMenuState`, as it was intended.

> Design wise, I'm still undecided between two options: /*  */

Right now I want to make sure we capture _some_ status quo in the book that allows people to use the engine more independently. As it evolves, we either find or build better ways to do this, the documentation should be updated to reflect that.

> Must end this with a great work!

Thank you for the review! <3

jojolepro:
image


I made a "pretty" drawing of how I mentally think about the ECS.

Of course, you can have Systems that are aware of the `State`, but in all cases you'll end up with game-specific systems that cannot be re-used. If you have a System that listens for a "PauseKey" input event, and emit a "PauseGame" event, you can reuse that system in all your games that know about the "PauseGame" event.

Also, you wouldn't want a system doing a State `Trans` to PauseMenu while you are inside of the MainMenu. To prevent that, you'd need pausable systems. So now, in addition to your `State`'s current State, you end up also managing which Systems are running and which aren't, which is more maintenance.

I hope I was able to explain that correctly.

udoprog:
I'll reiterate some things that were discussed over Discord. Sorry if I miss things.

It's important to emphasize that this PR is not the do-all, end-all way to trigger state transitions.
If another suggestion or new pattern emerges tomorrow which is superior at this stage of an Amethyst developer's journey, then I'd personally want to make sure that gets into the book.
For now I'd encourage us to at least include something and not letting perfect getting in the way of good. 

I do believe that triggering state transitions through resources is actually not _that bad_ all things considered.

#### These is probably a better way to trigger state transitions

* We can capture input events directly in SimpleState after #1148 is implemented. This does however not handle non-input based state transitions (e.g. player dying because their health ran out).
* The correct way to communicate events is to implement `handle_state`. This does however require the reader to deviate from `SimpleState`, which is a _big_ task at this stage of the documentation.

#### There was concerns raised about coupling systems to states

That would be the `Game` resource mentioned above, and the `CurrentState` enum.
The `Game` resource acts as a bridge between states and systems, but it's important to note that it is a declared data dependency (`Write<'s, Game>`), just like any other resource in the ECS. That means it doesn't care where the resource originates from.

We do maintain the value of the `CurrentState` enum. This is intended to implement conditional logic in systems, and is something I eventually want to solve better with #1146 (pausable systems).

udoprog:
@azriel91 @jojolepro poke?

book/src/concepts/system.md
torkleyy: This is an interesting discussion. In any case, we can currently process input in both location, so I think this statement can be confusing.

udoprog: I added some clarification with an example. Please check it out!

jojolepro:
Moxi wants to review

book/src/concepts/system.md
```diff
@@ -290,6 +290,179 @@ write_storage.remove(entity);
...
+
+So how can we affect states from systems?
+There are a couple of ways, but this section will detail the easiest one: using a [`Resource`][r].
+
+Before that, let's just quickly remind ourselves what a resource is:
+
+> A [`Resource`][r] is any type that stores data that you might need for your game AND that is not
+> specific to an entity.
+
+The data in a resource is available both to systems and resources.
```
Moxinilian: What does that mean?

book/src/concepts/system.md
```diff
@@ -290,6 +290,179 @@ write_storage.remove(entity);
...
+We can use this to our advantage!
+
+Let's say you have the following two states:
+
+* `GameplayState`: State in which the game is running.
+* `GameMenuState`: State where the game is paused and we interact with a game menu.
+
+The following example shows how to keep track of which state we are currently in.
+This allows us to do a bit of conditional logic in our systems to determine what to do depending on
+which state is currently active. And manipulating the states by setting flags:
```
Moxinilian: ", and"

book/src/concepts/system.md
udoprog: typo, s/resources/states

book/src/concepts/system.md
udoprog: Sentence would become a bit long. Do you think this is an issue?

Moxinilian: Nah I think it will be fine

udoprog: Ok, done!

torkleyy:
This should go into 0.10, please approve or request further changes.

azriel91:
I've changed the example code a bit: https://github.com/azriel91/amethyst/commit/b407a9cee65680b122febe7c27ab296ab762b428, but github isn't letting me PR into the PR.

Essentially I just changed the boolean flags with an enum, I guess manually upholding the state invariant felt like a footgun

udoprog:
@azriel91 applied your change, thanks!

Moxinilian:
Seems like bors was tired.

bors=Moxinilian,azriel91,torkleyy

Moxinilian:
Oh we're stupid.

bors r=Moxinilian,azriel91,torkleyy

PR #1159, opened 2018-11-18
implement common traits for TgaFormat

gheoan:
`Clone` is required for the `Format` trait. I did not find any issues regarding this.


gheoan:
Added Debug as suggested.

gheoan:
Will take a look and make sure it compiles on beta rust.

gheoan:
Should now compile on beta rust too. I hope these macro_use declarations were useless.

Xaeroxe:
It's just the travis windows setup being dumb..  So I'm gonna override here.

PR #1161, opened 2018-11-20
Move windows to allowed failures.

Xaeroxe:
I'm tired of this causing problems, so I'm opening this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ rust:
 matrix:
   fast_finish: true
   allow_failures:
+  - os: windows
   - rust: nightly
 
 branches:


jojolepro:
lol rust is just generally buggy on windows and mac

Xaeroxe:
Hard disagree, the travis servers are just bad.  I do pretty much 100% of my Rust dev on windows

However that's neither here nor there.

jojolepro:
Well my rust compiler freezes 50% of the time on both my windows installs.

Xaeroxe:
You're probably engaging select mode for cmd or powershell on accident.  If you just hit Escape it leaves it.  select mode is quite possibly the single worst shell feature I've ever seen, but that is most definitely not rust's fault :P

torkleyy:
But this effectively disables all checks for Windows?

Xaeroxe:
Yeah it does, and that's ok for now in my opinion.  We don't write enough platform specific code to be losing as much time to these failures as we are.  As soon as the travis windows implementation is stable we'll move it out of allowed failures.

Xaeroxe:
bors try

Xaeroxe:
bors retry

Xaeroxe:
Huh.  I guess you can only do that once per PR?

Xaeroxe:
Ok, had to wait for the last one to fail.  Let's try again with the macro_use fix applied.

bors try

jojolepro:
With rendy and laminar getting more integrated, youll have plenty of platform specific bugs popping up. There are already runtime bugs on windows currently...

Xaeroxe:
Travis wouldn't be able to catch that anyways, it would only catch compile errors.

Perhaps during the rendy and laminar integration we could open the travis builds ourselves and look to see if windows is passing.  If it fails due to typical travis operational problems then we'll have to test it on windows manually.

Xaeroxe:
Our builds have errored out 13 times since we added Windows to travis, not including the trying branch.

Xaeroxe:
On another note, if we get through the entire rendy integration without once manually testing on Windows and making sure rendering output looks the way we want it to then we are failing at our jobs.

jojolepro:
Are you counting that time when I restarted the mac builds 6 times in a row because rustc broke? Windows was enabled at that time.

Xaeroxe:
@jojolepro if you only restarted the mac build and didn't send in a whole new commit then no I'm not counting that.

jojolepro:
I would start it, then it would fail to get the rustup update. It usually did push a commit to trying.

Xaeroxe:
Can you link the PR?  I'll double check my numbers.

Xaeroxe:
Oh might not be necessary as I was only counting the staging branch

Xaeroxe:
Ok so after a recount where I was only counting builds that Windows had a timeout on, we got 10 failures.

Xaeroxe:
@jojolepro @torkleyy if you two feel like I haven't addressed your concerns can you please re-iterate which concerns remain?  If you do feel like I've addressed them then can I please get an approval?

torkleyy:
I don't have a problem with commenting "bors r+" multiple times, but with having to fix master because a PR broke it. That's why I'm unsure.

magnonellie:
For the record, doing this would significantly speed up merge times because we aren't gonna be hitting the 5 runner limit for required jobs.

torkleyy:
I see your point, but we don't need merges to happen faster now. If it becomes a problem in the future (because we are building up a backlog due to Windows bugs, which should be solved after the beta), we can still look into this again.

I see the approvals so far are in favour of this, and if there are approvals from the other members I'll simply abstain.

Xaeroxe:
It's not ideal, but it's temporary.  So I'll go ahead and

bors r+

PR #1162, opened 2018-11-21
enable event == UiEventType::Click

PSeitz:


amethyst_ui/src/event.rs
```diff
@@ -14,7 +14,7 @@ use transform::UiTransform;
 
 /// The type of ui event.
 /// Click happens if you start and stop clicking on the same ui element.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
```
Xaeroxe: Why not `Eq` too?

PSeitz: added Eq

PR #1164, opened 2018-11-21
Deny unknown fields for BasicScenePrefab

nikolads:
This adds `#[serde(deny_unknown_fields)]` to `BasicScenePrefab` so that an invalid field causes an error, instead of being silently ignored. I had trouble figuring out why my custom components aren't being added, so this should help. The message looks like:

```
Some(Message("unknown field `custom_component`, expected one of `graphics`, `transform`, `light`, `camera`, `control_tag`, `removal`"))
```

### DIFF
--- a/amethyst_utils/src/scene.rs
+++ b/amethyst_utils/src/scene.rs
@@ -19,15 +19,16 @@ use removal::Removal;
 ///
 /// ### Type parameters:
 ///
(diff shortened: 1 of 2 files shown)


LucioFranco:
bors r=torkleyy, Xaeroxe, jojolepro, magnonellie

magnonellie:
Thank you! :smile:

PR #1168, opened 2018-11-22
Remove note about Esc

### LINKED ISSUE #1158: Pressing Esc does not exit the default start project
At the [beginning of the Pong tutorial](https://www.amethyst.rs/book/latest/pong-tutorial/pong-tutorial-01.html#opening-and-closing-a-window) is mentioned:
> Let's start a new project:
>
>`amethyst new pong`
>
>If you run this project with `cargo run`, you'll end up with a window titled "pong" that renders a really delightful shade of green. Press `Esc` to quit. [...] 

However, pressing `Esc` does not exit / close the window for me, whether on a newly created project or throughout the Pong tutorial. I don't know if it's a problem local to me, a bug in the engine, or the documentation needing an update.

I'm running it on Arch Linux, Rust stable 1.30.1, and Amethyst 0.9.0. 
Let me know if you need more info.

torkleyy:
Fixes #1158


clemarescx:
Rather than removing the note, would it be really hard to implement close window on Esc instead? I've been trying to do it myself to learn how to deal with `actions` input and events, but I'm struggling to find the relevant info.

jojolepro:
It was closing on esc by default. That was removed because it was preventing users to make a pause menu :)

As for quitting on pressing escape, we could add it to an example. Here for example: https://github.com/amethyst/amethyst/blob/master/examples/window/main.rs#L13

All you have to do is: 
1) Add the handle_event method
2) Use the method https://github.com/amethyst/amethyst/blob/master/amethyst_input/src/util.rs#L29
3) If it returns true, return Trans::Quit from the handle_event method.

clemarescx:
Alright! I'll try to go from that example. If I manage to do it I'll send a PR, but don't hold your breath :sweat_smile:

clemarescx:
@jojolepro @torkleyy FYI just made a PR for the `window` example https://github.com/amethyst/amethyst/pull/1173

jojolepro:
It would be nice to add to the book that the window example shows how to close the window by pressing escape. Do you want to add it to this pr or do we merge it has is @torkleyy ?

udoprog:
Poke @torkleyy, minor thing. But we also need some conflict resolution unfortunately :/

torkleyy:
Yeah, was just looking at this :)

torkleyy:
This was fixed elsewhere.

PR #1170, opened 2018-11-22
Ui Refactor: text.rs cleaning

jojolepro:
Not to be merged.

Opened early for reviews (especially by the ui team). 

#### Incremental
After 0.10 is released, the easiest course of action would be to merge this as is (with the unimplemented stuff), and to then make the required prerequisite changes in other pull request.

vs

#### Branched
If this isn't wanted, we can blob all the changes into a separated branch, and then create a massive pull request at the end.

Opinions?

## Feature diff
* Added multi selection. See `Selectable` component.

## Breaking changes
* Selectable ui elements should now have an added "selectable" field in prefabs as shown here: <https://github.com/jojolepro/amethyst/blob/uirefactor/examples/assets/ui/example.ron#L83>
* Selectable ui elements should have an added `Selectable` component, if created manually.
* `MouseReactive` was renamed to `Interactable`
* UiTransform::tab_order was removed

## Unimplemented
* Blink cursor management system. Requires drawing colors without TextureHandle (other feature).
* Lots of TODO everywhere, which mostly depend on other features.


amethyst_ui/src/prefab.rs
```diff
@@ -8,72 +8,61 @@ use amethyst_assets::{
...
     /// Hides an entity by adding a [`HiddenPropagate`](../amethyst_renderer/struct.HiddenPropagate.html) component
     pub hidden: bool,
+    /// Makes the UiTransform selectable through keyboard inputs, mouse inputs and other means.
+    /// # Ordering
+    /// The UI element tab order.  When the player presses tab the UI focus will shift to the
+    /// UI element with the next highest tab order, or if another element with the same tab_order
+    /// as this one exists they are ordered according to Entity creation order.  Shift-tab walks
+    /// this ordering backwards.
+    // TODO: Make full prefab for Selectable.
+    pub selectable: Option<u32>,
```
DoumanAsh: Is 0 valid order? Can we consider 0 as None?

amethyst_ui/src/selection.rs
```diff
@@ -0,0 +1,268 @@
...
+                            }
+                        } else {
+                            // Down
+                            cached.cache.get(highest + 1).unwrap_or(cached.cache.first()
+	                			.expect("unreachable: A highest ui element was selected, but none exist in the cache."))
+                        };
+                        selecteds
+                            .insert(target.1, Selected)
+                            .expect("unreachable: We are inserting");
+                    } else {
```
DoumanAsh: else if?

amethyst_ui/src/selection.rs
```diff
@@ -0,0 +1,268 @@
...
+    #[new(default)]
+    /// A multi selection group. When multiple entities are in the same selection group, they can be selected at
+    /// the same time by holding shift or control and clicking them.
+    /// You can also select the first element, then hold shift and press the keyboard arrow keys.
+    // TODO: Holding shift + arrow keys to select more.
+    // TODO: Pressing the arrow keys could optionally be binded to change the selected ui element.
+    pub multi_select_group: Option<G>,
+    #[new(default)]
+    /// Indicates if you can select multiple entities at once without having to press the shift or control key.
+    pub auto_multi_select: bool,
```
DoumanAsh: I wonder if using bitflags would be more storage-effective in future instead of bools

amethyst_renderer/src/blink.rs
```diff
@@ -0,0 +1,78 @@
...
+};
+
+use Hidden;
+
+/// # Blink Component
+/// Periodically adds and removes a `Hidden` Component on the entity this is attached to.
+///
+/// ## Visibility Period
+/// During the first half period, the entity is visible.
+/// [0, delay/2[
```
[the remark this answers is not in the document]
jojolepro: ?
That's an interval opened at the end

amethyst_renderer/src/blink.rs
```diff
@@ -0,0 +1,78 @@
...
+            }
+
+            // Reset timer because we ended the last cycle.
+            // Keeps the overflow time.
+            if blink.timer > blink.delay {
+                blink.timer = blink.timer - blink.delay;
+            }
+
+            // We could cache the division, but that would require a stricter api on Blink.
+            let on = blink.timer < blink.delay / 2.0;
```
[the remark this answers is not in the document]
jojolepro: I'll probably cache the division result tbh.

amethyst_ui/src/bundle.rs
```diff
@@ -74,6 +90,10 @@ where
             "ui_button_system",
             &["ui_mouse_system"],
         );
+
+        // Required for text editing.
```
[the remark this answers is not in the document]
jojolepro: You don't need to have it depend on the system except if you want the blink to be frame perfect. Also its not used yet

amethyst_ui/src/cursor.rs
```diff
@@ -0,0 +1,58 @@
...
+	fn run(&mut self, (entities, mut transforms, editings, parents, selecteds, mut cursors, mut blinks, mut textures, colors, config): Self::SystemData){
+		// Go through all text editing entities.
+		for (entity, _) in (&*entities, &editings).join() {
+
+			// Finds child cursor of current text editing entity.
+			let cursor = (&*entities, &parents, &cursors).join().filter(|t| t.1.parent == entity).map(|t| t.0).next();
+			let selected = selecteds.contains(entity);
+
+			if let Some(c) = cursor {
+				if !selected {
```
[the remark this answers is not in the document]
jojolepro: Cargo fmt should have removed my tabs normally...

amethyst_ui/src/pass.rs
```diff
@@ -476,7 +476,7 @@ impl Pass for DrawUi {
                                 &tex_storage,
                             )).map(|tex| (tex, ed))
                     }) {
-                        let blink_on = editing.cursor_blink_timer < 0.5 / CURSOR_BLINK_RATE;
+                        let blink_on = editing.cursor_blink_timer < 0.25;
```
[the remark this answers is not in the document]
jojolepro: Its in the pass, and the pass is going to to the trash soon. BlinkSystem replaces that line

amethyst_ui/src/prefab.rs
jojolepro: 0 is valid currently

amethyst_ui/src/selection.rs
jojolepro: Maybe, but I would consider moving multiselect to its own component before that, and I'm not even sure its worth the extra complexity.

amethyst_renderer/src/blink.rs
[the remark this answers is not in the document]
DoumanAsh: ehhh? What do you mean?
To me it looks like `[0, delay/2[` unclosed `[`

amethyst_ui/src/selection.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// Indicates if this requires the inputs (except Tab) be ignored when the component is focused.
+    #[new(default)]
+    pub require_input: bool,
```
happenslol: Am I understanding correctly that with this option will basically "swallow" the inputs when the component is focused? If so I find the name a little confusing, shouldn't it rather be `swallows_inputs` or `consumes_inputs` or `stops_propagation` or something?

amethyst_ui/src/selection_order_cache.rs
```diff
@@ -0,0 +1,128 @@
...
+    //
+    //
+    //
+    //
+    /// Returns the index of the highest cached element (index in the cache!) that is currently selected.
+    pub fn highest_order_selected_index(
+        &self,
+        selected_storage: &WriteStorage<Selected>,
+    ) -> Option<usize> {
+        self.cache
```
happenslol: Depending on how many times you would theoretically read from this and how many elements will be in here, it might make more sense to calculate this value once on insert and keep it around on a field on the struct. You could even keep track of the next highest tab order in here and then skip the calculation if you push a new element that is below it.

amethyst_ui/src/selection_order_cache.rs
```diff
@@ -0,0 +1,128 @@
...
+where
+    G: PartialEq + Send + Sync + 'static,
+{
+    type SystemData = (
+        Entities<'a>,
+        Write<'a, CachedSelectionOrder>,
+        ReadStorage<'a, Selectable<G>>,
+    );
+    fn run(&mut self, (entities, mut cache, selectables): Self::SystemData) {
+        {
```
happenslol: Is there any way you could cut down on the work being done here by implementing some sort of dirty checking? i.e. have new selectables been added or removed, etc. I think 99% of the time this won't do any actual work

amethyst_ui/src/text_editing.rs
```diff
@@ -0,0 +1,360 @@
...
+            {
+                match *event {
+                    Event::WindowEvent {
+                        event: WindowEvent::ReceivedCharacter(input),
+                        ..
+                    } => {
+                        // Ignore obsolete control characters, and tab characters we can't render
+                        // properly anyways.  Also ignore newline characters since we don't
+                        // support multi-line text at the moment.
+                        if input < '\u{20}' {
```
happenslol: Maybe put all of these conditions into a `should_skip_char` or something, will probably make this easier to read

amethyst_ui/src/selection_order_cache.rs
jojolepro: I'll add this to the todo list. It is a very good idea. Thanks!

amethyst_ui/src/selection_order_cache.rs
jojolepro: We could eventually move to FlaggedStorage, yes.

amethyst_renderer/src/blink.rs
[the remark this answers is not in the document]
skreborn: @DoumanAsh It's a mathematical notation meaning that the range is closed (inclusive) on one end and open (exclusive) on the other. In this case, anything _greater than or equal to `0`_ **and** _strictly less than `delay/2`_ will fall into this range.

amethyst_renderer/src/blink.rs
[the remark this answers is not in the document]
jojolepro: Yeah I don't think its worth the trouble caching a single division here.

jojolepro:
Should be ready for review. When the next release is out, I want to merge this one to keep the changes incremental. The only breaking change introduced is the removal of the tab_order parameter in UiTransform::new.

Future PR will be breaking a lot, merging this pull request will help keep the change size small for the other ones.

jojolepro:
Squashing and adding back the UiImage -> TextureHandle change that I accidentally removed.

jojolepro:
Only the squash is left, I'll do that after the reviews.

LucioFranco:
Bump r? @amethyst/ui-dev

LucioFranco:
Bump r? @amethyst/ui-devs

amethyst_renderer/src/blink.rs
```diff
@@ -0,0 +1,78 @@
...
+            // Reset timer because we ended the last cycle.
+            // Keeps the overflow time.
+            if blink.timer > blink.delay {
+                blink.timer -= blink.delay;
+            }
+
+            // We could cache the division, but that would require a stricter api on Blink.
+            let on = blink.timer < blink.delay / 2.0;
+
+            if on {
```
happenslol: I'd combine these for better readability, something like
```rust
if on && !hidden.contains(entity) {}
else if !on && hidden.contains(entity) {}
```

amethyst_ui/src/cursor.rs
```diff
@@ -0,0 +1,59 @@
+//! Module containing the system managing the text editing cursor create, deletion and position.
+
+// TODO: Complete this and remove the logic from the UI Pass. Scheduled for completion after Transform2D.
+// File currently ignored because it is not added to lib.rs
+
+/// Tag component placed on the cursor of a text field being edited.
+pub struct TextEditingCursor;
```
happenslol: I feel like the naming for this and as well for the file could be better? `cursor` in `ui` sounds like the mouse cursor to me. Maybe just `TextCursor` is enough? Then the file could be `text_cursor.rs` as well. I think it make sense since you'd only ever have cursors in editable text anyways.

amethyst_ui/src/cursor.rs
```diff
@@ -0,0 +1,59 @@
...
+            let selected = selecteds.contains(entity);
+
+            if let Some(cursor_entity) = cursor {
+                if !selected {
+                    // Shouldn't have a cursor.
+                    entities.delete(cursor_entity);
+                    continue;
+                }
+            } else {
+                if selected {
```
happenslol: `else if`?

amethyst_ui/src/cursor.rs
```diff
@@ -0,0 +1,59 @@
...
+                    // TODO: Should have a cursor.
```
happenslol: This is already done, or is there something missing?

amethyst_ui/src/cursor.rs
```diff
@@ -0,0 +1,59 @@
...
+                    let cursor_entity = entities.create_entity();
```
happenslol: The function for creating a cursor should probably be moved into its own function, I think. You'll probably need it in more places and it makes it easier to read

amethyst_ui/src/event.rs
```diff
@@ -48,13 +62,13 @@ impl UiEvent {
 /// Will only work if the entity has a UiTransform component attached to it.
 /// Without this, the ui element will not generate events.
 #[derive(Default, Serialize, Deserialize, Clone)]
-pub struct MouseReactive;
+pub struct Interactable;
```
happenslol: So, `Interactable` now means that the entity can react to both mouse and keyboard events? Or at least, it will in the future? Maybe it make sense to keep mouse in the name somehow, since it might be preferable if we can split keyboard and mouse events

amethyst_ui/src/prefab.rs
```diff
@@ -170,13 +158,17 @@ impl<'a> PrefabData<'a> for UiTransformBuilder {
         }
         system_data.0.insert(entity, transform)?;
         if self.mouse_reactive {
```
happenslol: should rename `mouse_reactive` here

amethyst_ui/src/selection.rs
```diff
@@ -0,0 +1,268 @@
...
+					add single
+				else
+				    add replace
+			else
+				add replace
+		*/
+
+        // Checks if tab was pressed.
+        // TODO: Controller support/Use InputEvent in addition to keys.
+        for event in window_events.read(self.window_reader_id.as_mut().unwrap()) {
```
happenslol: I think there would be a very nice usecase for the event retrigger mechanism here, listening to and dispatching custom select/focus events. Can be a refactor later, but maybe it would be nice to add a todo so we don't forget about it.

amethyst_ui/src/selection.rs
```diff
@@ -0,0 +1,268 @@
...
+                        // Inside of the loop because its possible that the user clicks two times in a frame while pressing shift.
+                        let highest = cached.highest_order_selected_index(&selecteds);
+
+                        if let Some(highest) = highest {
+                            // Safe unwraps, we just got those values from the cache.
+
+                            let (highest_is_select, auto_multi_select) = {
+                                let highest_multi_select_group = &selectables
+                                    .get(cached.cache.get(highest).unwrap().1)
+                                    .unwrap()
```
happenslol: Shouldn't this still be `expect`s with a note that this is an invalid codepath

amethyst_ui/src/selection.rs
```diff
@@ -0,0 +1,268 @@
...
+                                        target_selectable.auto_multi_select,
+                                    )
+                                };
+                                (
+                                    highest_multi_select_group == target_multi_select_group,
+                                    auto_multi_select,
+                                )
+                            };
+
+                            if highest_is_select {
```
happenslol: I'd turn around if and else here, makes it way easier to read since you don't have to scroll to get to the else

amethyst_ui/src/selection_order_cache.rs
```diff
@@ -0,0 +1,118 @@
...
+                        .iter()
+                        .position(|&(cached_t, _)| selectable.order < cached_t);
+                    match pos {
+                        Some(pos) => inserts.push((pos, (selectable.order, entity))),
+                        None => pushes.push((selectable.order, entity)),
+                    }
+                }
+            }
+            inserts.iter().for_each(|e| cache.cache.insert(e.0, e.1));
+            pushes.iter().for_each(|e| cache.cache.push(*e));
```
happenslol: Can't you do the pushes directly, without the intermediate vec?

amethyst_ui/src/cursor.rs
jojolepro: This isn't done yet

amethyst_ui/src/cursor.rs
jojolepro: its the only place

amethyst_ui/src/event.rs
jojolepro: Interactable for both mouse and keyboard. If you have `Selected` a checkbox and press the space key, it will count as either a `Click` or as a `PressAction`, depending on the path we decide to take with eventing.

amethyst_ui/src/prefab.rs
jojolepro: That would be a breaking change. I plan on removing the field from UiTransform in the next PR.

amethyst_ui/src/selection.rs
jojolepro: I don't see how the retrigger fits here. There's code logic changing components on various entities. Event retrigger is mostly to re-emit events.

amethyst_ui/src/selection.rs
jojolepro: Its that way to reflect the algorithm written in comment at the top.

amethyst_ui/src/selection_order_cache.rs
jojolepro: image

amethyst_ui/src/prefab.rs
happenslol: Understandable

amethyst_ui/src/selection.rs
happenslol: I thought it would make sense to separate the system that handles the focusing (highlighting, checking focus order, and so on) from the system that dispatches the actual event. So in the end, an `Interactable` would be a retrigger that dispatches focus events, which can be handled in isolation.

amethyst_ui/src/selection.rs
happenslol: I see, that's fair

amethyst_ui/src/selection.rs
jojolepro: Do you mind adding more details? I'm not sure I see how that would work.
This system checks for raw input events and emits the initial "selected" events(not there yet) and manages the `Selected` components. You can't have a retrigger manage the components like that I think. Also you have to keep in mind that there can be concurrent events, like a mouse click and the `tab` key in the same frame. I want to keep the ordering deterministic for those.

happenslol: Hm yeah you're right, it's probably better to keep this specialized. My idea would've just been to have 1 more level of indirection. Instead of this system directly handling the `selected` events, there would be a retrigger that dispatches them. But it's fine like this, was just a thought

amethyst_renderer/src/blink.rs
```diff
@@ -0,0 +1,74 @@
...
+            // Reset timer because we ended the last cycle.
+            // Keeps the overflow time.
+            if blink.timer > blink.delay {
+                blink.timer -= blink.delay;
+            }
+
+            // We could cache the division, but that would require a stricter api on Blink.
+            let on = blink.timer < blink.delay / 2.0;
+
+            if on && !hiddens.contains(entity) {
```
Xaeroxe: I'd be happier if we structured this as

```rust
match (on, hiddens.contains(entity)) {
    (true, false) => {
        // insert
    },
    (false, true) => {
        // remove
    },
    _ => {},
}
```

It's slightly more efficient and makes the control flow more clear imo.

amethyst_ui/src/bundle.rs
```diff
@@ -59,21 +53,47 @@ where
...
+        builder.add(
+            SelectionMouseSystem::<G, A, B>::new(),
+            "ui_mouse_selection",
+            &[],
+        );
+        builder.add(
+            SelectionKeyboardSystem::<G>::new(),
+            "ui_keyboard_selection",
+            // Because when you press tab, you want to override the previously selected elements.
+            &["ui_mouse_selection"],
```
Xaeroxe: Since these can't execute in parallel and they handle roughly the same thing I have to wonder why they're separate systems.

amethyst_ui/src/button/builder.rs
```diff
@@ -9,43 +9,47 @@ use amethyst_core::{
...
+#[derive(Debug, Clone)]
+pub struct UiButtonBuilder<G> {
     name: String,
     x: f32,
     y: f32,
     z: f32,
     width: f32,
     height: f32,
-    tab_order: i32,
+    tab_order: u32,
```
Xaeroxe: Allowing these to be negative can make quick UI edits a bit easier if say I want to add a control to the top of a mostly completed form.  If I can't use negatives then I'd have to change the tab order in everything else too.

amethyst_ui/src/selection_order_cache.rs
```diff
@@ -0,0 +1,118 @@
...
+use crate::{Selectable, Selected};
+
+// TODO: Optimize by using a tree. Should we enforce tab order = unique? Sort on insert.
+/// A cache sorted by tab order and then by Entity.
+/// Used to quickly find the next or previous selectable entities.
+#[derive(Debug, Clone, Default)]
+pub struct CachedSelectionOrder {
+    /// The cached bitset.
+    pub cached: BitSet,
+    /// The cache holding the z position and the corresponding entity.
```
Xaeroxe: I get the feeling this is a copy paste error 🙂

amethyst_ui/src/text.rs
```diff
@@ -139,503 +134,132 @@ impl Component for TextEditing {
...
-                let ret = t1.cmp(&t2);
-                if ret == Ordering::Equal {
-                    return e1.cmp(e2);
-                }
-                ret
-            });
-
-        for text in (&mut text).join() {
+        // Normalize text to ensure we can properly count the characters.
+        // TODO: Possible improvement to be made if this can be moved only when inserting characters into ui text.
```
Xaeroxe: It's also possible the UI text might be programmatically modified by the game developer, so this isn't possible.  I'd already had that thought the first time around 😆

amethyst_ui/src/text.rs
```diff
@@ -139,503 +134,132 @@ impl Component for TextEditing {
...
+        // TODO: Finish TextEditingCursorSystem and remove this
         {
-            let mut focused_text_edit = focused
-                .entity
-                .and_then(|entity| zip_options(text.get_mut(entity), editable.get_mut(entity)));
-            if let Some((ref mut _focused_text, ref mut focused_edit)) = focused_text_edit {
-                focused_edit.cursor_blink_timer += time.delta_real_seconds();
-                if focused_edit.cursor_blink_timer >= 1.0 / CURSOR_BLINK_RATE {
-                    focused_edit.cursor_blink_timer = 0.0;
+            if let Some((text_editing, _)) = (&mut text_editings, &selecteds).join().next() {
```
Xaeroxe: Ah I was wondering how we were going to handle this.  Maybe somewhere in the docs we should make it clear that you can't edit multiple text fields simultaneously.

amethyst_ui/src/text_editing.rs
```diff
@@ -0,0 +1,368 @@
...
+            // Process events for the focused text element
+            if let Some((ref mut focused_text, ref mut focused_edit, _)) =
+                (&mut texts, &mut editables, &selecteds).join().next()
+            {
+                match *event {
+                    Event::WindowEvent {
+                        event: WindowEvent::ReceivedCharacter(input),
+                        ..
+                    } => {
+                        if should_skip_char(input) {
```
Xaeroxe: 👍 Much better than my approach

amethyst_ui/src/bundle.rs
jojolepro: Code separation, which is the whole point of this PR ;)

amethyst_ui/src/text.rs
jojolepro: I meant having a set_text method ensuring the normalization.

amethyst_ui/src/bundle.rs
Xaeroxe: Both systems have an input of `winit` events and an output of `Selected` components.  So it doesn't seem like we're separating concerns.  I see two reasons you might want to separate these

1. It's considered important that keyboard selection always overrides mouse selection.  In which case, why?

2. We don't like large match statements, and want to separate them into a mouse section and a keyboard section.  This could be achieved with functions like `handle_keyboard_event(&mut self, event: &Event)` and `handle_mouse_event(&mut self, event: &Event)`.  Personally I don't see the issue with large match statements, as it allows you to observe all of the possible control flows for selection in a single place.

amethyst_ui/src/button/builder.rs
jojolepro: From a logical perspective, it doesn't make sense ^^

I agree it does make prototyping easier, but you can fix that with:
1. Automatic tab_order generation
2. Partitioning your tab_order properly (leave gaps between elements, don't start at 0)
3. Planning your UI in advance

amethyst_ui/src/text.rs
jojolepro: If you are crazy I assume you could. Its not impossible either, sublime text does have the concept of multiple cursors.

Xaeroxe: Yeah but that's not a common need for videogames so I'd say shelve that enhancement and just point out that it's not there.

Xaeroxe:
@jojolepro okay, as soon as we come to a conclusion on the remaining discussions I'll shift to approval.

amethyst_ui/src/text.rs
Xaeroxe: We can shelve this for now though, the enhancement is desirable it just needs to be executed carefully.

amethyst_ui/src/bundle.rs
jojolepro: Systems don't cost anything. Also you can decide to enable just one of the two (if you want keyboard only and no mouse.) In my head, the keyboard input takes priority over the mouse for selecting stuff. Maybe that's not true, but enforcing the order prevents non-deterministic events. If you run two times the game with input simulation, you need one to be before the other to be deterministic.

amethyst_ui/src/button/builder.rs
jojolepro: That's true. I just changed it to u32 because its hard to reason about what a tab_order of -10 means in reality. Its not intuitive.

Also a lot of people seemed to think a tab_order of -1 meant disabled, when in reality it wasn't the case.

amethyst_ui/src/text.rs
jojolepro: I don't understand why you don't want this to be allowed? To select multiple things at once, the developer has to opt-in the multi select group for the text fields, so its not like the users will be able to select both at once if they are not supposed to.

amethyst_ui/src/bundle.rs
Xaeroxe: being able to enable only one of the two actually makes a lot of sense as I can definitely imagine contexts where that might be valuable, so we can keep it.  As an aside

- It's not about run time cost as much as maintenance cost
- There's no reason for keyboard to take priority, only one form of selection is used at a time typically.
- A single system is just as deterministic as two systems if the order of events is the same.  If the order of events is different then the behavior should be different and can no longer be said to be deterministic.

That being said you made a good point about selective handling for device input, so I'll resolve this.

amethyst_ui/src/text.rs
Xaeroxe: Sorry I must not be making myself very clear, I just want some comments to say "we don't support editing multiple text fields right now."

jojolepro: But in theory we do :thinking:

Xaeroxe: This will only retrieve and process events for the first selected text field it finds.  That's what `.join().next()` does.

amethyst_ui/src/bundle.rs
jojolepro: It's not about run time cost as much as maintenance cost

To me having small, separated system, lowers the maintenance cost by a lot. (At least from my experience)

amethyst_ui/src/button/builder.rs
jojolepro: Up to you. I don't think it will matter when we have a prefab editor, because we can totally automate this. We can have a "increment all" button, or a "resolve conflicts" button which works a bit like when you resolve merge conflicts, asking you which one you want first.

amethyst_ui/src/text.rs
jojolepro: oops will fix asap :3

amethyst_ui/src/button/builder.rs
Xaeroxe: Huh those are good points.  Let's leave this as is then.

jojolepro:
@Xaeroxe r? :3

jojolepro:
Fixed

jojolepro:
YAAAAAAAAAAAAAAAAAAAY

Issue #1141, opened 2018-11-13
Remove all expects() from amethyst_network


TimonPost:
Remove all `expects()` from amethyst network.

Here are a few: 
1. https://github.com/amethyst/amethyst/blob/master/amethyst_network/src/network_socket.rs#L69
2. https://github.com/amethyst/amethyst/blob/master/amethyst_network/src/network_socket.rs#L155
3. https://github.com/amethyst/amethyst/blob/master/amethyst_network/src/network_socket.rs#L151
4. https://github.com/amethyst/amethyst/blob/master/amethyst_network/src/bundle.rs#L36

This kind of relates to error handling from this [issue](#1140).

Plasticcaz:
I think I'd like to have a go of this if that's okay.

Plasticcaz:
I should note I've opened [pull request  #1176 ](https://github.com/amethyst/amethyst/pull/1176) for this issue.

jojolepro:
Fixed?

TimonPost:
Yes, it could be closed for now. The left expects could be looked at later.

PR #1176, opened 2018-11-24
[WIP]: Issue #1141 Remove expects from amethyst_network

Plasticcaz:


Plasticcaz:
I'll run rustfmt later to fix the above failure...

image
I am currently unsure how we should go about removing the above- expects found in network_socket.rs. The method doesn't return a result, so we can't offload responsibility for handling this error further up the chain. 

The first expect pictured there seems like an assertion that it should never fail, so I think we should leave it there.

I'm not very familiar with the nitty-gritty of networking code, but I was wondering if maybe I can downgrade the second expect to just log a warning if we reach it? It doesn't seem to be too serious, unless it would continually keep sending stop requests to no avail. Any suggestions?

amethyst_network/src/network_socket.rs
```diff
@@ -89,9 +89,12 @@ where
         let mut socket = UdpSocket::bind(addr, NetworkConfig::default())
             .map_err(|x| Error::new(ErrorKind::Other, x.to_string()))?;
 
-        socket
-            .set_nonblocking(true)
-            .expect("Unable to set `UdpSocket` to non-blocking mode");
+        socket.set_nonblocking(true).map_err(|_| {
+            Error::new(
+                ErrorKind::Other,
+                "Unable to set `UdpSocket` to non-blocking mode".to_owned(),
```
udoprog: `.to_owned()` is not necessary, all though it is worth noting that it still [does convert it into a String internally](https://doc.rust-lang.org/src/std/error.rs.html#271).

Plasticcaz: Hmmm... I could have sworn it wasn't letting me pass in a string slice. Must have been from when I was trying something else. I'll remove it.

I don't think it matters so much about allocating the String, as it only happens on the non-happy path.

TimonPost:
I come back at your question. I encountered the same indeed.

azriel91:
Yo @Plasticcaz! do you expect to remove more expects in this PR, or are we good to merge?

torkleyy:
I'd personally recommend not fixing too many of those at once, because that can become quite boring over time.

TimonPost:
Let's merge this I'm fine with the expects there are currently.

TimonPost:
bors r=TimonPost,LucioFranco,Xaeroxe,udoprog

TimonPost:
bors r=TimonPost,LucioFranco,Xaeroxe,udoprog

Plasticcaz:
> 
> 
> Yo @Plasticcaz! do you expect to remove more expects in this PR, or are we good to merge?

This issue was just to remove expects from networking code. As far as I could see there were 4 instances of expect. I removed two of them by returning error results instead. The other two didn't look so easy to get rid of, and I was waiting on feedback.

Looks like it's been decided to leave them for now. Thanks everyone!

torkleyy:
Thank you @Plasticcaz!

PR #1179, opened 2018-11-25
Allow searching for named entities

torkleyy:
* [x] Required utility (caching storage)
* [x] Searching functionality
* [x] Tests
* [ ] Replace `UiFinder` to use the new facilities


amethyst_core/src/util.rs
```diff
@@ -0,0 +1,53 @@
+use hibitset::BitSetLike;
+use specs::{storage::UnprotectedStorage, Component};
+
+pub trait Cache<T> {
```
jojolepro: Shouldn't that be called `EntityCache`? If it was a generic cache, you would probably want a generic for the key.

torkleyy:
Hold on, I didn't push everything yet. Still some work to do before this is
ready.

jojolepro:
Ah you are right. I didn't see the "working" tag.

amethyst_core/src/util.rs
torkleyy: It's just utilty that's not exposed, plus it doesn't cache entities (but arbitrary data).

jojolepro: **pub** trait Cache?

torkleyy: ~~pub~~ mod util; :wink:

amethyst_core/src/named.rs
```diff
@@ -154,3 +306,135 @@ impl<'a> WithNamed for LazyBuilder<'a> {
...
+#[derive(Debug, Default)]
+pub struct NameCache {
+    map: HashMap<Cow<'static, str>, u32>,
+}
+
+impl Cache<Named> for NameCache {
+    fn on_get(&self, _: u32, _: &Named) {}
+
+    fn on_update(&mut self, id: u32, val: &Named) {
+        self.map.insert(val.name.clone(), id);
```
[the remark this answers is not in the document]
torkleyy: I'm not sure whether we should allow duplications or if it should be possible (without warning). I think that's what we should discuss first.

amethyst_core/src/named.rs
```diff
@@ -1,6 +1,160 @@
...
+
+use util::{Cache, CachedStorage};
+
+/// Extension trait to find an entity based on its `Named` component.
+/// Can only be used on the `Named` component storage.
+pub trait FindNamed {
+    /// Finds an entity by its name by querying it from a map.
+    ///
+    /// For performance reasons, this should only be used once, and the returned `Entity` shall
+    /// be stored to avoid future lookups.
```
Xaeroxe: Might reword to "this is a mildly expensive operation, so please avoid it calling for large groups of entities every frame.  To avert this, you can store the returned `Entity` value in a resource or component."

amethyst_core/src/named.rs
```diff
@@ -1,6 +1,160 @@
...
+    ///
+    /// # Conflict behavior
+    ///
+    /// Only the last `Named` component with name `s` will be returned; if
```
Xaeroxe: "last" is very arbitrary and poorly defined from a user perspective, so we should probably make this an error case.  If a named entity that is expected to be unique is not unique that is extremely likely to be a coding error, and should be caught.

amethyst_core/src/named.rs
```diff
@@ -114,18 +251,33 @@ impl Named {
...
+    /// the `EntityBuilder` type and will add a `Named` component to it.
+    /// See the `Named` type for usage examples.
+    ///
+    /// # Conflict behavior
+    ///
+    /// There can be two entities with the same `Named` component, however,
+    /// only the last inserted will be returned by `FindNamed::find_named`.
+    /// If you want to retrieve all entities with a given name, you need to join
+    /// over all entities and filter for the name (or use the `find_with_name`
+    /// convenience method).
```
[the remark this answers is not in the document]
Xaeroxe: There seem to be two "modes" here.  In one mode we're using a name as a tag, in which case it's most definitely not unique, but in another we're using a name as an identifier, which is most definitely unique.  We can't know which is intended until the functions are used though.  So, see my comment above regarding enforcing the unique constraint.

Xaeroxe:
Thanks for your work!

amethyst_core/src/named.rs
Xaeroxe: I should clarify: I want this to return `Result`.

torkleyy:
> Joining-then-filtering seems like it's less performant as joining over hierarchical bit sets, would that be considered an anti-pattern?

Yes, it's generally preferable to have a bitset. The main use-case this PR is trying to address is loading a scene and fetching very specific entities from it.

Given that, and the existence of tags, it made me think if multiple entities with the same name actually make sense.

Xaeroxe:
Just re-read some of the prior review comments from @torkleyy , and I think that names should be required to be unique.  If we want a tag system there are better ways to implement that.

torkleyy:
> Joining-then-filtering seems like it's less performant as joining over hierarchical bit sets, would that be considered an anti-pattern?

Yes, it's generally preferable to have a bitset. The main use-case this PR is trying to address is loading a scene and fetching very specific entities from it.

Given that, and the existence of tags, it made me think if multiple entities with the same name actually make sense.

torkleyy:
Just noticed I never sent this comment...

Xaeroxe:
[But ya did...](https://github.com/amethyst/amethyst/pull/1179#issuecomment-441796142)

torkleyy:
Err, okay, not sure why it ended up in this comment form then... :sweat_smile:

jojolepro:
I was reading my trello and saw this:
EntityId<I: PartialEq>
EntityId could have uniqueness enforced, and named not. Also having a generic has obvious advantages for this.

torkleyy:
How do you imagine this to work? @jojolepro 

I don't intend to merge this before 0.10.

jojolepro:
You would have Named like now, and a generic EntityIdentifier component. The generic parameter I is the unique key identifying that entity. You can search in O(n) by joining then filtering. The better solution is to use a hashmap like here. The important part is only to have the id generic so you can use enums. The id type would default to String.

jojolepro:
I personally would use it at runtime too. ie on button click, find entity with id Id::MyText and set text to "hello". Using it like this prevents caching that text entity for the whole game duration.

jojolepro:
@torkleyy anything going on here?

amethyst_core/src/named.rs
Rhuagh: I would actually prefer this to return all matched entities so the caller can decide what to do with them. Consider a scenario where you have a "Goblin" name in a prefab, and then spawn 300 goblins using that prefab, you'll have 300 entities with Named("Goblin"). If I wanted to apply some logic to all 300, retrieving all of them in this way could be beneficial.

Rhuagh: If you want you could add separate functions for each scenario, that use a base one that gives `Iterator<Item=Entity>` or some such.

PR #1189, opened 2018-11-27
Implement EventRetriggers and refactor UiButton

happenslol:
I'm opening this up for preliminary review, since this introduces a lot of changes and I'd like some feedback on it before I actually implement the new patterns for all use cases. Here's what this PR basically does:

* Introduce a new generic `System` called `EventRetriggerSystem`. It works very similarly to the one proposed in #1072, allowing you to basically trigger follow-up events for any events.
* Refactor `UiButton` and its builder to use the new system to trigger its click and hover actions (This is only implemented for `SetTextColor` so far)
* Refactor the UI prefab builder to also use these (Haven't started on this, shouldn't be too much work though).

Here's things I'm still unsure about/not satisfied with:

* How to keep track of the state the buttons are in/what has changed due to which event. At the moment, I'm basically keeping a stack of changes, where the first element is the original state. This feels a little hacky, and might be nicer to read with an additional struct that contains the original state as a field, plus a `Vec` of changes. Let me know if you see a better way to do this.
* How to structure the `UiButtonRetrigger` component. Currently it's a struct containing `Vec`s for all event types (`on_hover`, `on_click`, etc.) and I've been told it's not a good idea to keep `Vec`s in components. It might be a good idea here to use `SmallVec` to keep the memory contiguous, but I'm not sure.
We definitely need some dynamically sized array here, since with a builder or at any point at runtime, the user can add/remove events from this. Since there's no variant typing and multiple events of the same type might occur for one action, it's also not feasible to add a separate component for every one.

If you want to see what's working so far in action, please take a look at the `simple_ui` example. I'm still not sure if I want to flesh out this example a bit while I develop this, but I think it might be a good idea to have an example around that programmatically constructs a UI with some simple effects.


amethyst_ui/src/action_components.rs
```diff
@@ -38,11 +62,11 @@ impl Component for OnUiActionImage {
 /// Requires `MouseReactive`.
 pub struct OnUiActionSound {
     /// Sound made when this button is hovered over
-    pub(crate) hover_sound: Option<SourceHandle>,
+    pub(crate) _hover_sound: Option<SourceHandle>,
```
udoprog: `#[allow(unused)]`?

amethyst_ui/src/button/builder.rs
```diff
@@ -54,13 +54,11 @@ pub struct UiButtonBuilder {
...
     image: Option<TextureHandle>,
     parent: Option<Entity>,
-    hover_image: Option<TextureHandle>,
-    hover_text_color: Option<[f32; 4]>,
-    press_image: Option<TextureHandle>,
-    press_text_color: Option<[f32; 4]>,
-    hover_sound: Option<SourceHandle>,
-    press_sound: Option<SourceHandle>,
-    release_sound: Option<SourceHandle>,
+    on_click: Vec<UiActionType>,
```
udoprog: Use [SmallVec](https://docs.rs/smallvec)?

amethyst_ui/src/event_retrigger.rs
```diff
@@ -0,0 +1,64 @@
...
+    },
+};
+
+use event::TargetedEvent;
+
+pub trait EventRetrigger: Component {
+    type In: Clone + Send + Sync + TargetedEvent;
+    type Out: Send + Sync;
+
+    fn apply(&self, event: &Self::In) -> Vec<Self::Out>;
```
udoprog: This signature requires the implementation to allocate a vector. An alternative could be to use a different trait to describe how data can be read out. That way the receiver can decide:

```rust
pub trait Receiver<T> {
    fn receive(&mut self, values: &[T]);
}

impl<T> Receiver<T> for Vec<T>
where
    T: Clone,
{
    fn receive(&mut self, values: &[T]) {
        self.extend(values.iter().cloned());
    }
}

pub trait Trigger {
    type In;
    type Out: Send + Sync;

    fn apply<R>(&self, event: &Self::In, out: &mut R)
    where
        R: Receiver<Self::Out>;
}

pub struct Example;

impl Trigger for Example {
    type In = ();
    type Out = u32;

    fn apply<R>(&self, event: &Self::In, out: &mut R)
    where
        R: Receiver<Self::Out>,
    {
        out.receive(&[42]);
    }
}

fn main() {
    let mut out = Vec::new();
    Example.apply(&(), &mut out);
    println!("out: {:?}", out);
}
```

amethyst_ui/src/action_components.rs
happenslol: This will go once I'm done with this, the `OnUi...` structs will be replaced completely by the new `UiAction`s.

amethyst_ui/src/button/builder.rs
happenslol: Sounds like a good idea

amethyst_ui/src/event_retrigger.rs
happenslol: That's actually a very good idea! I was struggling a lot with how to return these events, since returning a `Vec` always feels a little bad, especially in this context. I'll try this out, but maybe it might also be a good idea to somehow pass the `EventChannel` to it, so that it can write it's items there directly? That would reduce the overhead of having to construct some sort of container on the receiving end

happenslol: Done!

amethyst_ui/src/button/builder.rs
happenslol: Also done!

amethyst_ui/src/button/builder.rs
```diff
@@ -307,9 +340,13 @@ impl UiButtonBuilder {
                 image_entity,
                 UiButton {
                     normal_text_color: self.text_color,
-                    hover_text_color: self.hover_text_color,
-                    press_text_color: self.press_text_color,
+                    hover_text_color: None,
+                    press_text_color: None,
                 },
+<<<<<<< HEAD
```
jojolepro: :thinking:

happenslol:
In general this is working as I hoped it would. The sound system is already usable by everything else, so we could make an issue for other things that emit sounds to use it.

One thing I'm still unsure about is the `ActionChangeStack` (horrible name, I'm sorry). What it basically does is it keeps track of what actions have changed the state of the button - i.e. if you hover on it and then click on it, then let go, it will go back to the hovered texture/color instead of the normal one. However if you stop hovering before you release the click, it will go back to normal directly.
I made the struct to make the system easier to read, but I'm not sure if this is something that will be needed elsewhere. 

The last thing is that I'm still gonna move the `UiButtonActions` over to `SmallVec`s, for data locality.

amethyst_ui/src/event.rs
```diff
@@ -44,6 +48,12 @@ impl UiEvent {
     }
 }
 
+impl TargetedEvent for UiEvent {
+    fn get_target(&self) -> Entity {
+        self.target.clone()
```
jojolepro: Entity is `Copy`, no need to clone.

happenslol:
I kinda agree with you - it doesn't feel idiomatic, but to be honest, I think it comes down to the implementation. I think we definitely want some kind of "side effects" concept to model reactions to events, the hard part is to figure out how to do this without too much friction so it doesn't feel like we're making compromises while still keeping the pattern intact.

I think this PR can be a good first step in that direction though, as it would already allow for lots of things to be simplified and implemented with the right pattern, which will make it easier to swap out in the future if we do land on another implementation eventually.

khionu:
@jojolepro can you re-review? The newer commits invalidated your approval

happenslol:
Squashed

amethyst_ui/src/bundle.rs
```diff
@@ -10,7 +10,7 @@ use amethyst_core::{
 };
 use amethyst_renderer::TextureFormat;
 
-use super::*;
+use crate::*;
```
khionu: We're trying to remove wildcard imports

amethyst_ui/src/prefab.rs
```diff
@@ -12,7 +12,7 @@ use amethyst_core::specs::{
 };
 use amethyst_renderer::{HiddenPropagate, Texture, TextureFormat, TextureMetadata, TexturePrefab};
 
-use super::*;
+use crate::*;
```
khionu: Again, wildcard

khionu:
Also, we can squash on merge, there's no need to squash before then

jojolepro:
If we use bors to merge you need to squash yourself.

jojolepro:
bors r+
shoooooooooooooooooooooooooooot

happenslol:
bors r+
yeet

PR #1192, opened 2018-11-28
Fix CameraOrtho when coordinates are upside-down

### LINKED ISSUE #1190: CameraNormalizeMode::Contain loses space when bottom > top
version: master (77285f6a26c046ce145de69d5abc37504f1b768d) with nightly enabled

code:
```rust
.with(CameraOrtho {
    mode: CameraNormalizeMode::Contain,
    world_coordinates: CameraOrthoWorldCoordinates {
        left: 0.0,
        top: 0.0,
        right: GAME_WIDTH,
        bottom: GAME_HEIGHT,
    }
})
```

expected behaviour: any aspect ratio besides the original should have "black bars", no space should be lost

actual behaviour: when ratio is "tall", space is lost

wide window with side black bars, one tall window missing some space on either side

CosineP:
Upside-down coordinates would be when `bottom` is a higher value than
`top`. This is valuable since many people coming from 2D gamedev are
used to this coordinate system and want to keep it.

I have no idea if this is an actual use-case because for me, when i realized id have to flip all my assets and stuff, i just decided to get my headspace into bottom=0 mode.

Fixes #1190

### DIFF
--- a/amethyst_utils/src/ortho_camera.rs
+++ b/amethyst_utils/src/ortho_camera.rs
@@ -173,7 +173,11 @@ impl CameraNormalizeMode {
         desired_coordinates: &CameraOrthoWorldCoordinates,
     ) -> (f32, f32, f32, f32) {
         // If bottom is higher than top (common in 2D graphics), we flip the offset
-        let sign = if desired_coordinates.bottom > desired_coordinates.top { -1.0 } else { 1.0 };
+        let sign = if desired_coordinates.bottom > desired_coordinates.top {
+            -1.0
+        } else {
+            1.0
+        };
         let offset = (desired_coordinates.width() / window_aspect_ratio
             - desired_coordinates.height())
             / 2.0
@@ -240,7 +244,7 @@ mod test {
     // TODO: Disabled until someone fixes the formula (if possible).
     /*#[test]
     fn near_far_from_camera() {
-    	use amethyst_core::cgmath::{Ortho, Matrix4};
+        use amethyst_core::cgmath::{Ortho, Matrix4};
         let mat4 = Matrix4::from(Ortho {
             left: 0.0,
             right: 1.0,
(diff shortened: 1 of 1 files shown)


Xaeroxe:
Not super sure I'm on board with making `bottom` and `top` ambiguous like this.  Perhaps we should just be flipping our coordinate system here entirely?

These assets you speak of, how did you make them?  Is our engine not accommodating common tooling?

CosineP:
anyone who's used a 2d-first engine is probably used to a coordinate system like this:

```
0,0 ---------------------------> width
|
|
|
|
V
height                           (width, height)
```

to answer your second question, my attempt at doing this was doing an ortho with `bottom: height, top: 0.0`, buuuut, that meant that all my sprites were upside down. so i think i would have to `flip_vertical: true` them all. they're all just normal pixel-art images made in GIMP

what would "flipping our coordinate system here entirely" mean?

Xaeroxe:
Ok it seems I misunderstood the issue a bit.  I'm going to do some more reading and thinking and then I'll get back to you on this.

Xaeroxe:
By the way I just wanted to say your github username is adorable and I love it :)

CosineP:
thank you!! i use it everywhere (or `cosine` when i can)

Xaeroxe:
Can you run `cargo fmt` on this?  Apparently rustfmt doesn't like this formatting https://concourse.subnetzero.io/builds/950#L5bc8056a:8

CosineP:
woah, that's weird, our two versions of `cargo fmt` disagree fsr? wonder if its cause im on nightly... ill fix it manually

Xaeroxe:
Ah yeah we use stable.

Xaeroxe:
2D games like 🙃

torkleyy:
bors r=Xaeroxe,LucioFranco,torkleyy

PR #1195, opened 2018-11-29
Improve documentation for fly control functionality

torkleyy:


amethyst_controls/src/bundles.rs
```diff
@@ -8,10 +8,27 @@ use amethyst_core::{
...
+/// * `B`: This is the key the `InputHandler` is using for actions. Often, this is a `String`.
+///
+/// # Systems
+///
+/// This bundle adds the following systems:
+///
+/// * `FlyMovementSystem`
+/// * `FlyRotationSystem`
+/// * `MouseFocusUpdateSystem`
+/// * `CursorHideSystem`
```
udoprog: Add link to systems?

amethyst_controls/src/bundles.rs
```diff
@@ -8,10 +8,27 @@ use amethyst_core::{
...
-/// You might want to add "fly_movement" and "free_rotation" as dependencies of the TransformSystem.
+///
+/// Note: Will not actually create a moving entity. It will only register the needed resources and
+/// systems. The generic parameters `A` and `B` are the ones used in `InputHandler<A,B>`.
+///
+/// You might want to add `"fly_movement"` and `"free_rotation"` as dependencies of the
+/// `TransformSystem` in order to apply changes made by these systems in the same frame.
 /// Adding this bundle will grab the mouse, hide it and keep it centered.
+///
+/// # Generic parameters
```
azriel91: I'd go with "Type parameters", based on the [Rust API guidelines](https://rust-lang-nursery.github.io/api-guidelines/naming.html). I think they're intentional about not calling them "generics".

Ditto for the other file.

amethyst_controls/src/bundles.rs
```diff
@@ -8,10 +8,27 @@ use amethyst_core::{
...
+///
+/// * `A`: This is the key the `InputHandler` is using for axes. Often, this is a `String`.
```
azriel91: Should we encourage the use of `String`?
(I feel this will spawn a doc ticket for "how to make your own input axes / action IDs)

amethyst_controls/src/bundles.rs
torkleyy: Yes, I'd also like to see more cross referencing but I've also seen it break in some places due to no standard way of linking and re-exports. Maybe that's fixed by now, I'm probably not up to date here.

udoprog:
Ok, merging since comments were addressed!

bors r=udoprog, jojolepro

Issue #1152, opened 2018-11-15
About the simplicity issue


torkleyy:
It's a rather broad opinion that Amethyst could be simpler to use. This makes our engine less appealing for smaller games. I'm hoping to bring in everybody who shares that concern, and to start off a discussion about how we can make Amethyst simple to use, while keeping the flexibility aspect.

This thread emerged from [a discussion on Twitter](https://twitter.com/lucio_d_franco/status/1063115213954785280).

---

## The issue

A major goal of Amethyst is to be highly customizable for the individual needs of a project. This is what I just use "flexibility" for most of the time. This includes things like

* replacing a part of the engine
* opting out of default behaviour
* structuring your game different, maybe without states, or with very different ones, etc.

That's one side of this. The other is: people want to make simple games, like Pong, Tetris, etc. Dealing with generics, multiple lifetimes, etc. is not what is expected for this.

## A bit of history

Amethyst used to be a just a few crates (mostly just the main crate) a long time ago. At some point, the sub-modules mostly moved into their individual crates. It was clear at that point that any `Application` wrapper would most likely not be able to cover all possible use cases, and we hoped to make the main crate an optional facility that gets you started quickly, but is simple, so that bigger projects should use the individual crates and not the main one.

Now, I think this was not communicated very clearly, and I understand it just makes sense that you want to continue using the main crate to have a certain feature. This has led to more flexibility being added to what used to be rather simple abstractions, and as such, the main crate became rather complex. We are providing a bunch of simplified variants for some things, but that creates its own problems, such as

* being too simple, and thus not allowing important features
* additional maintenance work
* the need for more documentation and potential confusion because of all the options available

---

I think this is a hard discussion and needs some thorough thinking. The solution for this is most likely far from trivial, but please don't hesitate to add your ideas here!

jojolepro:
Related issues:

https://github.com/amethyst/amethyst/issues/1148
https://github.com/amethyst/amethyst/issues/1114
https://github.com/amethyst/amethyst/issues/1101
https://github.com/amethyst/amethyst/issues/1086 (2D render unification fixes this)
https://github.com/amethyst/amethyst/issues/1080
https://github.com/amethyst/amethyst/issues/1044 (Priority)
https://github.com/amethyst/amethyst/issues/989
https://github.com/amethyst/amethyst/issues/986 (Priority)
https://github.com/amethyst/amethyst/issues/951
https://github.com/amethyst/amethyst/issues/762
[...]

PR #1198, opened 2018-11-29
Remove lifetime from `SimpleState` and `SimpleTrans`

### LINKED ISSUE #1178: Consider using Dispatcher with static lifetime only
There is close to no use case for having a lifetime in a system field, so we should consider exporting the dispatcher with only `'static` as lifetimes. That would also allow us to remove the lifetimes from the state implementations.

torkleyy:
Fixes #1178 

Related: #1152

### DIFF
--- a/amethyst_core/src/frame_limiter.rs
+++ b/amethyst_core/src/frame_limiter.rs
@@ -25,7 +25,7 @@
 //! use amethyst::core::frame_limiter::FrameRateLimitStrategy;
 //!
 //! # struct GameState;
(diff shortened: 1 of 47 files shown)


torkleyy:
Ah, changelog entry needed.

torkleyy:
Eh, sorry for the merge commit.
bors r+

torkleyy:
Now, one last time:
bors r+

jojolepro:
Gogogo! :)

PR #1199, opened 2018-11-30
Separate states and callbacks

udoprog:
Hey.

Sorry for the rather length PR, this is the result of a rather [large amount of experimentation](https://community.amethyst-engine.org/t/request-for-comments-simplified-state-machine/143/3) from my part.
My hopes is that at least some of this can be re-used as documentation if this turns out to be palatable.

I _don't_ expect this to be merged. Or at least not as-is, but I'm hoping to gather more feedback and hopefully convince some folks that this is a worthwhile mid-to-long term endeavour. And hopefully receive more feedback.
Either way, doing this has taken me pretty deep into the architecture of Amethyst, so I personally have learned _a lot_.

**Note:** I'm keeping all the existing code around behind a feature flag (`dynamic_app`) to make this easier to rebase. This artificially inflates the **+lines** a fair amount.

EDIT: Moved to amethyst/rfcs#9


bgourlie:
Simply from an ergonomics point of view, this looks very nice!

udoprog:
Moved to amethyst/rfcs#9

PR #1205, opened 2018-12-01
GameDataBuilder: reverse dependencies, API improvements

jjedelsky:
When creating game with large amount of systems I need to split building `GameData` to multiple functions. Otherwise I would end up with unmaintainable code. Lots of time I want system to name it's dependencies including reverse dependencies. This RP tries to add support for such workflow.

Changes:
- Add reverse dependencies to `GameDataBuilder::with`
- `GameDataBuilder` evaluates dependencies when `DataInit::build` is called, so it is possible to reference system which has not been added yet.
- Introduce `SimpleDispatcherBuilder` to abstract over `DispatcherBuilder` and `GameDataBuilder`
- `SystemBundle::build` takes `SimpleDispatcherBuilder` instead of `DispatcherBuilder`
- Add by-ref API to GameDataBuilder
- Introduce `AutoAddSystem` to simplify adding system where it's dependencies are always the same.
- `GameDataBuilder` takes dependencies as iterator instead of slice

### DIFF
--- a/amethyst_animation/src/bundle.rs
+++ b/amethyst_animation/src/bundle.rs
@@ -8,35 +8,35 @@ use crate::{
     },
 };
 
-use amethyst_core::{
-    specs::prelude::{Component, DispatcherBuilder},
-    Result, SystemBundle,
-};
+use amethyst_core::{specs::prelude::Component, Result, SimpleDispatcherBuilder, SystemBundle};
 
 /// Bundle for vertex skinning
 ///
 /// This registers `VertexSkinningSystem`.
 /// Note that the user must make sure this system runs after `TransformSystem`
 #[derive(Default)]
-pub struct VertexSkinningBundle<'a> {
-    dep: &'a [&'a str],
+pub struct VertexSkinningBundle<'a: 'b, 'b> {
+    dep: &'b [&'a str],
 }
 
-impl<'a> VertexSkinningBundle<'a> {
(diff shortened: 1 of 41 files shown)


azriel91:
Hiya, apologies for the long wait to have a review.
First, thanks for explaining the changes succinctly, it made following the code much easier.
While the changes are clean + high quality, there are some existing thoughts that interact in this area that conflict:

* https://github.com/amethyst/amethyst/issues/1152 - Simplicity of the API.
* https://github.com/amethyst/rfcs/pull/9 - RFC that separates the identification of a state from its behaviour. This one in particular will see the `GameData` (and builder) be removed.

In terms of API cleanliness, I think it's better to not have both normal dependencies and reverse dependencies &ndash; opening up the possibility of specifying "B says it depends on A, and A says it must run before B" (redundancy) is going to make it harder to track issues (even if ever so slightly). It could mean `A`'s API is tied to `B` in case they're separate crates, which is potential for error.

ah gotta go, sorry my comment is cut short

jjedelsky:
Hi,
I have look at https://github.com/amethyst/rfcs/pull/9 and I am gonna close/postpone this and maybe try some other design for reverse dependencies later once https://github.com/amethyst/rfcs/pull/9 is implemented.

About API cleanliness: Yes, there are problems related to reverse dependencies, but it is easier to use.
Sometimes it makes more sense to say "A should run before B" than "B should run after A", especially when such order is required by A.

Issue #1204, opened 2018-12-01
Paused systems cause memory leaks in `EventBuffer`


torkleyy:
Events must be read every frame, but paused systems don't do that and will thus cause the `EventChannel` growing.

This is a blocking issue for 0.10.

udoprog:
So I'm gonna clarify why I don't think this is a blocking issue.

This is a known footgun of using EventChannel: passive `ReaderId`'s causes the channel to grow indefinitely.

To me there is nothing special about a paused system that causes this behavior. Paused systems are specialized systems that "do nothing". In the absence of this mechanism users wanting to build something that "pauses systems" would be facing the same set of problems. If they forget to consume the `ReaderId` they would be leaking memory.

AFAICT, this is also the case for the conventional way of pausing systems today: using a separate dispatcher without dispatching it.

So for this reason I don't think we should treat Pausable systems as a blocker for 0.10.

But mid-term we should come up with a pattern for building event channels that doesn't have these properties. But I still don't think this is a blocker since it's a known caveat already.

torkleyy:
> AFAICT, this is also the case for the conventional way of pausing systems today: using a separate dispatcher without dispatching it.

Okay, that makes sense. I'm fine with closing this issue as soon as there's documentation in place pointing this out.

fhaynes:
Closing this as it is done

PR #1206, opened 2018-12-01
Add warning about using EventChannel with pausable systems

udoprog:
CC #1204

### DIFF
--- a/amethyst_core/src/system_ext.rs
+++ b/amethyst_core/src/system_ext.rs
@@ -10,6 +10,15 @@ use specs::prelude::{Read, System};
 pub trait SystemExt {
     /// Make a system pausable by tying it to a specific value of a resource.
     ///
+    /// When the value of the resource differs from `value` the system is considered "paused",
+    /// and the `run` method of the pausable system will not be called.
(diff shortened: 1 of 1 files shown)


amethyst_core/src/system_ext.rs
```diff
@@ -10,6 +10,15 @@ use specs::prelude::{Read, System};
 pub trait SystemExt {
     /// Make a system pausable by tying it to a specific value of a resource.
     ///
+    /// When the value of the resource differs from `value` the system is considered "paused",
```
TimonPost: Nice, nitpick incomming: "paused" => 'paused'

udoprog: Thanks, how come single quotes is correct here?

azriel91:
bors r=torkleyy,TimonPost,azriel91

amethyst_core/src/system_ext.rs
TimonPost: Double quote is for marking a statement said by someone. If you just want to highlight a word you should use single quote.

PR #1207, opened 2018-12-01
Clarify pong chapter introduction

### LINKED ISSUE #1194: Rewrite pong tutorial about `application_root_dir()`
I tried writing and running the pong tutorial https://www.amethyst.rs/book/latest/pong-tutorial.html , and error message appeared as follow.

```
error[E0425]: cannot find function `application_root_dir` in this scope                                               
  --> src/main.rs:28:3                                                                                                
   |                                                                                                                  
28 |         application_root_dir()                                                                                   
   |         ^^^^^^^^^^^^^^^^^^^^ not found in this scope                                                             
help: possible candidates are found in other modules, you can import them into scope                                  
   |                                                                                                                  
6  | use amethyst::utils::app_root_dir::application_root_dir;                                                         
   |                                                                                                                  
6  | use amethyst::utils::application_root_dir;  
```
And I found this commit https://github.com/amethyst/amethyst/commit/8288591fc3085a3b34bc3930f7165ba03f9ce113 .

So That, I think the pong tutorial must be rewritten about `application_root_dir()` an example this page https://www.amethyst.rs/book/latest/pong-tutorial/pong-tutorial-03.html#modifying-the-transform .

Thanks in advance.

torkleyy:
Fixes #1194

### DIFF
--- a/book/src/pong-tutorial.md
+++ b/book/src/pong-tutorial.md
@@ -15,7 +15,9 @@ The main difference between real game code and the example code is where the
 
 For instance, in the pong_tutorial_01 example we have:
 
-```rust,ignore
+```rust,norun
+use amethyst::utils::application_root_dir;
+
 let app_root = application_root_dir();
 
 let path = format!(
@@ -26,9 +28,17 @@ let path = format!(
 
 But for your own project you'll probably want something like this:
 
-```rust,ignore
+```rust,norun
 let path = "./resources/display_config.ron";
 ```
+
+The reason `application_root_dir` is used in our examples is that the
+required resources are not in the working directory, but rather somewhere
+in the `examples` folder. However, in an actual game it would be bad to
+hardcode the path of the resources to the directory that was used to compile
+the application, because then the game cannot be moved (or else it wouldn't
+find the resources). That's why you should stick to a relative path for your
+game, as shown above.
 
 [pong]: https://github.com/amethyst/amethyst/tree/master/examples/pong
 
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ even if the game is running on a massive 64 cores processor.
 
 ### Community
 
-While we may not be feature-packed (yet!), we all strongly believe that the community-oriented side of amethyst will thrive as we move forward!
+While we may not be feature-packed (yet!), we all strongly believe that the community-oriented side of Amethyst will thrive as we move forward!
 
 ## Features
 


book/src/pong-tutorial.md
```diff
@@ -26,9 +28,15 @@ let path = format!(
...
-```rust,ignore
+```rust,norun
 let path = "./resources/display_config.ron";
 ```
 
+The reason `application_root_dir` is used in our examples is that the
+required resources are not in the working directory, but rather somewhere
+in the `examples` folder. However, in an actual game it would be bad to
+hardcode the path of the resources to the directory that was used to compile
+the application.
```
udoprog: Could you clarify _why_ it would be bad? What would happen?

torkleyy: Yeah, I somehow assumed the last step is just logical. Really gotta learn explaining things better.

udoprog: I'm super happy that you are working over deficiencies in the docs and book as you see them!

udoprog:
@jojolepro While there is some subtlety involved, sometimes being completely accurate get's in the way of learning. I think the added section sufficiently describes why something was done in this context. And if we wish to elaborate that it should probably go into the API docs.

jojolepro:
What I meant is that using a relative path instead of app root dir it completely breaks on mac or when executing from a different folder than were the executable is.

udoprog:
Ok. So here's what I'm thinking.

This is what `application_root_dir` does:

* It points to `CARGO_MANIFEST_DIR`
* Otherwise it points to the directory of the binary.

None of these allows a user to easily start working with custom assets in the existing examples project. Using a relative path is not perfect, but it gets them going and is relatively (no pun intended) straightforward to understand.

Once they do get to the point of working on their own project (or distribute a binary!) it is worthwhile to learn how to correctly specify an assets path. But that might even vary somewhat based on need.

jojolepro:
> None of these allows a user to easily start working with custom assets in the existing examples project.

How do you do that with relative paths?

The only option I see is cloning amethyst and using ../amethyst/examples/assets

But in that case you can also do approotdir/../amethyst/examples/assets

Or do you mean cloning amethyst and modifying the examples in place?

udoprog:
> Or do you mean cloning amethyst and modifying the examples in place?

Yes, that's the part that works with relative paths. But more generally, relative paths works _quite well_.

I think part of this could also be that we don't have a solid game distribution story yet either, so `application_root_dir` exists as an "ok" solution to that.

torkleyy:
Handled by #1213

PR #1213, opened 2018-12-04
Don't panic in application_root_dir and make use of std::path

udoprog:
This improves diagnostics on errors, and encourages people to use the existing filesystem abstraction, which we probably should.

Also making use of canonicalize (typically a single system call) instead of recursively resolving links.

### DIFF
--- a/examples/sprite_camera_follow/main.rs
+++ b/examples/sprite_camera_follow/main.rs
@@ -157,8 +157,7 @@ fn main() -> amethyst::Result<()> {
     let game_data = GameDataBuilder::default()
         .with_bundle(TransformBundle::new())?
         .with_bundle(
-            InputBundle::<String, String>::new()
-                .with_bindings_from_file(root.join("input.ron"))?,
+            InputBundle::<String, String>::new().with_bindings_from_file(root.join("input.ron"))?,
         )?
         .with(MovementSystem, "movement", &[])
         .with_bundle(


amethyst_utils/src/app_root_dir.rs
```diff
@@ -1,21 +1,28 @@
...
-                .to_str()
-                .expect("Failed to get valid string from path to the executable."),
-        )
-    })
+pub fn application_root_dir() -> Result<path::PathBuf, io::Error> {
+    if let Some(manifest_dir) = env::var_os("CARGO_MANIFEST_DIR") {
+        let manifest_dir = manifest_dir.into_string().map_err(|_| {
+            io::Error::new(
+                io::ErrorKind::Other,
+                "`CARGO_MANFIEST_DIR` exists, but contains non-utf8 data",
```
torkleyy: That should not be a requirement though.

book/src/pong-tutorial/pong-tutorial-01.md
```diff
@@ -124,12 +124,13 @@ In `main()` in `main.rs`, we will load the configuration from the file:
 # extern crate amethyst;
 # use amethyst::prelude::*;
 # use amethyst::renderer::DisplayConfig;
-# fn main() {
+# fn main() -> amethyst::Result<()> {
 use amethyst::utils::application_root_dir;
 
-let path = format!("{}/resources/display_config.ron", application_root_dir());
+let path = application_root_dir()?.join("resources/display_config.ron");
```
torkleyy: Would be nice to have `application_dir(AsRef<Path>)`

amethyst_utils/src/app_root_dir.rs
udoprog: I re-implemented the old behavior, but I agree that it shouldn't be a requirement.

amethyst_utils/src/app_root_dir.rs
```diff
@@ -1,21 +1,21 @@
...
+        return Ok(path::PathBuf::from(manifest_dir));
+    }
+
+    if let Some(parent) = env::current_exe()?.canonicalize()?.parent() {
+        return Ok(parent.to_owned());
+    }
+
+    Err(io::Error::new(
+        io::ErrorKind::Other,
+        "failed to find an application root",
```
jojolepro:
```suggestion
        "Failed to find an application root",
```

udoprog:
@jojolepro Manual testing scales rather poorly, I'd prefer identifying the variable factors and testing those:

1, and 2 isn't really a dependency on running cargo in any particular way, but the `CARGO_MANIFEST_DIR` environment variable.

`env::current_exe()` has behavior which is guaranteed by Rust's stability guarantee and 3, 4, and 5 all inherit from that. Or put in other words: There's no difference in clicking the executable, running it in a terminal. Nor doing any of that after moving it into a different folder. I.e. testing one of them should warrant the behavior of the rest.

That reads like two test cases for me?

jojolepro:
:+1: 

We had issues with mac 4 months ago, so I added more test for this one time just to be sure we don't have to fix this method again. Finder would set the running directory to /Users/YourName for some reason :P

book/src/pong-tutorial/pong-tutorial-01.md
udoprog: Ah, Isn't this technically an API addition? I'd rather do that separately if that's all right by you.

udoprog:
@torkleyy Commented back, I thought it was optional. Sorry!

book/src/pong-tutorial/pong-tutorial-01.md
torkleyy: Yeah it is, but it can be fixed here. It's a simple function, so I think the friction for another PR is a bit high, but I won't block this.

Once the book section is updated, feel free to r+ this.

udoprog: I've added it now.

udoprog:
Changelog updated, examples were already up-to-date. Hopefully everything still works!

udoprog:
bors r=torkleyy,azriel91,jojolepro

udoprog:
bors retry

PR #1219, opened 2018-12-04
UI hidpi fix

jojolepro:
This fixes some of the logic issues related to hidpi. It also disables automatic text scaling. This will be replaced soon by automatic text resizing and proper layouting techniques, which should be prefered to hidpi anyway.

Breaking change: ScreenDimensions takes hidpi into account. (logical size instead of previous physical size).

### DIFF
--- a/amethyst_input/src/input_handler.rs
+++ b/amethyst_input/src/input_handler.rs
@@ -63,7 +63,12 @@ where
     ///
     /// The Amethyst game engine will automatically call this if the InputHandler is attached to
     /// the world as a resource with id 0.
-    pub fn send_event(&mut self, event: &Event, event_handler: &mut EventChannel<InputEvent<AC>>) {
+    pub fn send_event(
+        &mut self,
+        event: &Event,
+        event_handler: &mut EventChannel<InputEvent<AC>>,
+        hidpi: f64,
+    ) {
         match *event {
             Event::WindowEvent { ref event, .. } => match *event {
                 WindowEvent::ReceivedCharacter(c) => {
@@ -221,11 +226,11 @@ where
                 } => {
                     if let Some((old_x, old_y)) = self.mouse_position {
                         event_handler.single_write(CursorMoved {
-                            delta_x: x - old_x,
-                            delta_y: y - old_y,
+                            delta_x: x * hidpi - old_x,
+                            delta_y: y * hidpi - old_y,
                         });
                     }
-                    self.mouse_position = Some((x, y));
+                    self.mouse_position = Some((x * hidpi, y * hidpi));
                 }
                 WindowEvent::Focused(false) => {
                     self.pressed_keys.clear();
--- a/amethyst_input/src/system.rs
+++ b/amethyst_input/src/system.rs
@@ -6,8 +6,9 @@ use winit::Event;
 
 use amethyst_core::{
     shrev::{EventChannel, ReaderId},
-    specs::prelude::{Read, Resources, System, Write},
+    specs::prelude::{Read, ReadExpect, Resources, System, Write},
 };
+use amethyst_renderer::ScreenDimensions;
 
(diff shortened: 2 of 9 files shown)


amethyst_input/src/input_handler.rs
```diff
@@ -219,13 +224,14 @@ where
                     position: LogicalPosition { x, y },
                     ..
                 } => {
+                    println!("Reported position winit {}x{}", x, y);
```
LucioFranco: oopsie woopsie 😄

amethyst_ui/src/text.rs
```diff
@@ -316,10 +316,17 @@ impl<'a> System<'a> for UiKeyboardSystem {
                     ..
                 } => {
                     let hidpi = screen_dimensions.hidpi_factor() as f32;
+                    info!(
```
LucioFranco: Does this need to be info or can be it moved to trace?

amethyst_ui/src/text.rs
```diff
@@ -316,10 +316,17 @@ impl<'a> System<'a> for UiKeyboardSystem {
...
+                        screen_dimensions.width(),
+                        screen_dimensions.height(),
+                        position
+                    );
                     self.mouse_position = (
                         position.x as f32 * hidpi,
-                        (screen_dimensions.height() - position.y as f32) * hidpi,
+                        (screen_dimensions.height() - position.y as f32 * hidpi),
                     );
+                    info!("Mouse Position: {:?}", self.mouse_position);
```
LucioFranco: Same here

jojolepro:
Thanks ;)

torkleyy:
I'm not sure about this. According to winit docs, we should only ever get
logical pixels. Why do we need to handle this?

Also, I'm not quite sure if it's a good idea to return this factor like
this. I don't think the existing methods should be modified.

jojolepro:
Because we get physical pixels in ScreenDimensions somehow. I can add methods get_physical_* to it though.

jojolepro:
https://github.com/amethyst/amethyst/blob/master/amethyst_renderer/src/system.rs#L130

Having the physical size in ScreenPosition makes sense since the use might want to resize, so we need to expose that value.

Its hard to know if the window itself is 1080p in my 4k screen or if its 4k but pretends to be 1080p to get the scaling effect...

jojolepro:
Ah the renderer is casting the logical size to physical size implicitely. I think this is correct, but it does explain why this was an issue to start with.

@Xaeroxe is there any reason someone would want to resize using the physical size instead of the logical size?

torkleyy:
We should use logical pixels everywhere. I suspect that we're mixing that currently, and I don't think we should fix it by using physical pixels.

amethyst_input/src/system.rs
```diff
@@ -58,16 +60,22 @@ where
         Read<'a, EventChannel<Event>>,
         Write<'a, InputHandler<AX, AC>>,
         Write<'a, EventChannel<InputEvent<AC>>>,
+        ReadExpect<'a, ScreenDimensions>,
```
happenslol: Does this _need_ to be `ReadExpect`? Are there _any_ circumstances where this could hurt us? It would be simple to just pass `screen_dimensions.map(|it| it.hidpi_factor()).unwrap_or(1.0)` below, but if there's no way this can be `None` then this is fine.

happenslol: Alright

LucioFranco:
@torkleyy do you still want to block this PR or should we move forward with it?

jojolepro:
I'm slowly working on it

jojolepro:
Done. I didn't get stuck for two hours because I had self.w; instead of self.w = w;......................

LucioFranco:
bors r=happenslol,LucioFranco

Issue #1214, opened 2018-12-04
Support causes in amethyst::Error


udoprog:
Our existing error type is sparse and doesn't include any context of where an error happened.

We should include causes with contexts, and backtraces to our error type to improve our diagnostic capabilities both to users and developers.

#### Causes

The [`failure`](https://docs.rs/failure) allows doing this by adding "contexts":

```rust
use failure::{Error, ResultExt};

type Result<T> = Result<T, Error>;

fn ring_bell(sound: &str) -> Result<()> {
  bail!("I don't feel like ringing the bell")
}

fn something() -> Result<()> {
  ring_bell("ding").with_context(|_| format_err!("Failed to ding"))?;
  ring_bell("dong").with_context(|_| format_err!("Failed to dong"))?;
  Ok(())
}
```

Causal chains are useful because errors can be quite generic (i.e. `I don't feel like ringing the bell`). But in this case an important piece of information is: did it fail when dinging or donging?

#### Backtraces

I think most folks should be familiar with back traces, but I'll just provide some brief thoughts on why they should be included.

Backtraces provide additional context, but can be quite low level. Panics by default include a backtrace. This is something we will be losing out on unless we include it in our error types. 

Causes without backtraces could be rendered like this, and should be highly useful for many error cases:

```rust
Failed to ding
Caused by: I don't feel like ringing the bell.
```

Including backtraces provides much more detailed information:

```
Failed to ding
stack backtrace:
   0:     0x5616b3933aa6 - backtrace::backtrace::libunwind::trace::hebc3cbfc0bad42f5
                        at /home/udoprog/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.9/src/backtrace/libunwind.rs:53
                         - backtrace::backtrace::trace::h662d733ae29451d7
                        at /home/udoprog/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.9/src/backtrace/mod.rs:42
   1:     0x5616b392e053 - backtrace::capture::Backtrace::new_unresolved::hbefd36b5dad230fc
                        at /home/udoprog/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.9/src/capture.rs:88
   2:     0x5616b392cb8f - failure::backtrace::internal::InternalBacktrace::new::h47fcb9ce8ccc42f5
[...]

PR #1220, opened 2018-12-05
Introduce amethyst_error

udoprog:
Following #1214 

This introduces the minimal API necessary to convert existing Error types (some which includes causes).

It currently mimics the `failure::Error` API, with methods named according to the ["fix Error" RFC](https://github.com/rust-lang/rfcs/blob/master/text/2504-fix-error.md).

That means:
* We include the `format_err!` macro and the `ResultExt` trait with an implementation of `with_context` from `failure`.
* The error type includes backtraces directly (instead of wrapping it in through e.g. `Context<D>`) in comparison to `failure`.
* We include a blanket `From<E> where E: std::error::Error` implementation to convert any type that implements `std::error::Error` into `Error`.
  * Note: this has the side-effect that this `Error` can't implement `std::error::Error`.
* We are using `source` from `std::error::Error` instead of `cause`.
* We don't implement `std::error::Error` directly, but we do implement some methods which are similar (`source`). We implement `as_error` to access a reference to the underlying `std::error::Error`. Similarly to how `failure::Error` doesn't implement `Fail` directly but provides `as_fail`.

Note that since this RFC is so sparse, there's really not much to implement, and most other methods which are used to actually build errors are local to this crate (e.g. `ResultExt::with_context`).

I have a branch where I've tried to apply this crate here:
https://github.com/udoprog/amethyst/commit/amethyst-error-full

I think I might have accidentally botched some of the `From<T>` implementations, but it should still give an idea of how this type is used.

#### Composite APIs

Composite APIs are APIs which potentially make use of local error information. Today these tend to lead to "stringy" errors where most of the underlying error information is lost and the errors undergoes a conversion. These are candidates to make use of `amethyst_error::Error` directly right now.

The traits I've so far identified that has a tendency to lead to this are:

* [`amethyst_assets::Source`](/amethyst/amethyst/blob/master/amethyst_assets/src/source/mod.rs)
* [`amethyst_assets::PrefabData`](/amethyst/amethyst/blob/master/amethyst_assets/src/prefab/mod.rs)
* [`amethyst_assets::SimpleFormat`](/amethyst/amethyst/blob/master/amethyst_assets/src/asset.rs)
  * Here's an example where SimpleFormat "stringifies" errors instead of propagating them to work around error constraints: https://github.com/amethyst/amethyst/blob/master/amethyst_ui/src/prefab.rs#L736
* [`amethyst_renderer::Pass/Pipeline` stuff](/amethyst/amethyst/blob/master/amethyst_renderer/src/pipe/pass.rs). Unsure because it spans over to UI and I don't know how "end user" this is.

#### Open Discussions

* Dropping the `Sync` constraint ([comment](https://github.com/amethyst/amethyst/pull/1220#discussion_r239030384)).
* ~~Either drop `no_std` support or improve it ([comment 1](https://github.com/amethyst/amethyst/pull/1220#discussion_r239052724), [comment 2](https://github.com/amethyst/amethyst/pull/1220#discussion_r243899686)).~~ backtrace is no longer optional.

### DIFF
--- a/amethyst_error/Cargo.toml
+++ b/amethyst_error/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+name = "amethyst_error"
+version = "0.1.0"
+authors = ["John-John Tedro <udoprog@tedro.se>"]
+readme = "README.md"
+description = """
+Internal error handling for Amethyst.
+"""
+keywords = ["game", "error", "amethyst"]
+categories = ["error"]
+
+documentation = "https://www.amethyst.rs/doc/master/doc/amethyst/"
+homepage = "https://www.amethyst.rs/"
+repository = "https://github.com/amethyst/amethyst"
+
+[badges]
+appveyor = { repository = "amethyst/amethyst", branch = "master" }
+travis-ci = { repository = "amethyst/amethyst" }
+
+[dependencies]
+backtrace = {version = "0.3.9", optional = true}
+
+[features]
+default = ["backtrace"]
--- a/amethyst_error/Cargo.toml
+++ b/amethyst_error/Cargo.toml
@@ -20,8 +20,5 @@ travis-ci = { repository = "amethyst/amethyst" }
 [dependencies]
 backtrace = {version = "0.3.9", optional = true}
 
-[dev-dependencies]
-failure = {version = "0.1"}
-
 [features]
 default = ["backtrace"]
--- a/amethyst_error/Cargo.toml
+++ b/amethyst_error/Cargo.toml
@@ -19,7 +19,4 @@ appveyor = { repository = "amethyst/amethyst", branch = "master" }
 travis-ci = { repository = "amethyst/amethyst" }
 
 [dependencies]
-backtrace = {version = "0.3.9", optional = true}
-
-[features]
-default = ["backtrace"]
+backtrace = "0.3.13"
--- a/amethyst_error/Cargo.toml
+++ b/amethyst_error/Cargo.toml
@@ -20,5 +20,8 @@ travis-ci = { repository = "amethyst/amethyst" }
 [dependencies]
 backtrace = {version = "0.3.9", optional = true}
 
+[dev-dependencies]
+failure = {version = "0.1"}
+
 [features]
 default = ["backtrace"]
--- a/amethyst_error/Cargo.toml
+++ b/amethyst_error/Cargo.toml
@@ -3,6 +3,7 @@ name = "amethyst_error"
 version = "0.1.0"
 authors = ["John-John Tedro <udoprog@tedro.se>"]
 readme = "README.md"
+edition = "2018"
 description = """
 Internal error handling for Amethyst.
 """
--- a/amethyst_error/README.md
+++ b/amethyst_error/README.md
@@ -14,5 +14,5 @@ MIT/Apache-2.
 
 ## License
 
-`amethyst_locale` is distributed under the terms of both the MIT
+`amethyst_error` is distributed under the terms of both the MIT
 license and the Apache License (Version 2.0).
--- a/amethyst_error/README.md
+++ b/amethyst_error/README.md
@@ -1,0 +1,18 @@
+# amethyst_error
+
+This crate contains types used internally and externally to handle and communicate errors.
+
+## Contribution
+
+Contribution is highly welcome! If you'd like another
+feature, just create an issue. You can also help
+out if you want to; just pick a "help wanted" issue.
+If you need any help, feel free to ask!
+
+All contributions are assumed to be dual-licensed under
+MIT/Apache-2.
+
+## License
+
+`amethyst_locale` is distributed under the terms of both the MIT
+license and the Apache License (Version 2.0).
--- a/amethyst_error/src/lib.rs
+++ b/amethyst_error/src/lib.rs
@@ -433,4 +433,14 @@ mod tests {
             .find(|n| *n == "a_really_unique_name_42")
             .is_some());
     }
+
+    #[test]
+    fn test_backtrace_disabled() {
+        use super::BACKTRACE_ENABLED;
+        use std::sync::atomic;
+
+        BACKTRACE_ENABLED.store(1, atomic::Ordering::Relaxed);
+
+        assert!(Error::from_string("an error").backtrace().is_none());
+    }
 }
--- a/amethyst_error/src/lib.rs
+++ b/amethyst_error/src/lib.rs
@@ -5,16 +5,17 @@
 //! type as an opaque [`std::error::Error`].
 //!
 //! [`std::error::Error`]: https://doc.rust-lang.org/std/error/trait.Error.html
-use self::internal::{Backtrace, HAS_BACKTRACE};
+
+#[cfg(all(feature = "backtrace", feature = "std"))]
+extern crate backtrace;
+
+use self::internal::Backtrace;
 use std::borrow::Cow;
-use std::env;
 use std::error;
-use std::ffi;
 use std::fmt;
 use std::result;
-use std::sync::atomic;
 
-/// Private
+/// Internal parts of `Error`.
 enum ErrorImpl {
     Message(Cow<'static, str>),
     Error(Box<dyn error::Error + Send + Sync>),
@@ -54,6 +55,7 @@ impl From<Box<dyn error::Error + Send + Sync>> for ErrorImpl {
 pub struct Error {
     error_impl: ErrorImpl,
     source: Option<Box<Error>>,
+    #[cfg(all(feature = "backtrace", feature = "std"))]
     backtrace: Option<Backtrace>,
 }
 
@@ -68,7 +70,8 @@ impl Error {
         Self {
             error_impl: ErrorImpl::Error(Box::new(error)),
             source: None,
-            backtrace: new_backtrace(),
+            #[cfg(all(feature = "backtrace", feature = "std"))]
+            backtrace: self::internal::bt::new(),
         }
     }
 
@@ -80,13 +83,22 @@ impl Error {
         Self {
             error_impl: ErrorImpl::Message(message.into()),
             source: None,
-            backtrace: new_backtrace(),
+            #[cfg(all(feature = "backtrace", feature = "std"))]
+            backtrace: self::internal::bt::new(),
         }
     }
 
     /// Get backtrace.
     pub fn backtrace(&self) -> Option<&Backtrace> {
-        self.backtrace.as_ref()
+        #[cfg(all(feature = "backtrace", feature = "std"))]
+        {
+            self.backtrace.as_ref()
+        }
+
+        #[cfg(not(all(feature = "backtrace", feature = "std")))]
+        {
+            None
+        }
     }
 
     /// Get the source of the error.
@@ -292,66 +304,54 @@ where
     }
 }
 
-/// Test if backtracing is enabled.
-fn is_backtrace_enabled<F: Fn(&str) -> Option<ffi::OsString>>(get_var: F) -> bool {
-    match get_var(RUST_BACKTRACE) {
-        Some(ref val) if val != "0" => true,
-        _ => false,
-    }
-}
-
-fn new_backtrace() -> Option<Backtrace> {
-    if !HAS_BACKTRACE {
-        return None;
-    }
-
-    static ENABLED: atomic::AtomicUsize = atomic::ATOMIC_USIZE_INIT;
-
-    match ENABLED.load(atomic::Ordering::SeqCst) {
-        0 => {
-            let enabled = is_backtrace_enabled(|var| env::var_os(var));
-
-            ENABLED.store(enabled as usize + 1, atomic::Ordering::SeqCst);
-
-            if !enabled {
-                return None;
-            }
-        }
-        1 => return None,
-        _ => {}
-    }
-
-    Some(Backtrace::new())
-}
-
-const RUST_BACKTRACE: &str = "RUST_BACKTRACE";
-
-#[cfg(all(feature = "backtrace", feature = "std"))]
-mod internal {
-    pub const HAS_BACKTRACE: bool = true;
-
-    pub use backtrace::Backtrace;
-}
-
 #[cfg(not(all(feature = "backtrace", feature = "std")))]
 mod internal {
-    pub const HAS_BACKTRACE: bool = false;
-
     use std::fmt;
 
     /// Fake internal representation.
     pub struct Backtrace(());
 
-    impl Backtrace {
-        pub fn new() -> Backtrace {
-            Backtrace(())
-        }
-    }
-
     impl fmt::Debug for Backtrace {
         fn fmt(&self, _fmt: &mut fmt::Formatter) -> fmt::Result {
             Ok(())
         }
+    }
+}
+
+#[cfg(all(feature = "backtrace", feature = "std"))]
+mod internal {
+    pub use backtrace::Backtrace;
+    use std::{env, ffi, sync::atomic};
+
+    const RUST_BACKTRACE: &str = "RUST_BACKTRACE";
+
+    /// Test if backtracing is enabled.
+    pub fn is_backtrace_enabled<F: Fn(&str) -> Option<ffi::OsString>>(get_var: F) -> bool {
+        match get_var(RUST_BACKTRACE) {
+            Some(ref val) if val != "0" => true,
+            _ => false,
+        }
+    }
+
+    /// Constructs a new backtrace, if backtraces are enabled.
+    fn new() -> Option<Backtrace> {
+        static ENABLED: atomic::AtomicUsize = atomic::ATOMIC_USIZE_INIT;
+
+        match ENABLED.load(atomic::Ordering::SeqCst) {
+            0 => {
+                let enabled = is_backtrace_enabled(|var| env::var_os(var));
+
+                ENABLED.store(enabled as usize + 1, atomic::Ordering::SeqCst);
+
+                if !enabled {
+                    return None;
+                }
+            }
+            1 => return None,
+            _ => {}
+        }
+
+        Some(Backtrace::new())
     }
 }
 
--- a/amethyst_error/src/lib.rs
+++ b/amethyst_error/src/lib.rs
@@ -382,11 +382,11 @@ mod internal {
     fn new() -> Option<Backtrace> {
         static ENABLED: atomic::AtomicUsize = atomic::ATOMIC_USIZE_INIT;
 
-        match ENABLED.load(atomic::Ordering::SeqCst) {
+        match ENABLED.load(atomic::Ordering::Relaxed) {
             0 => {
                 let enabled = is_backtrace_enabled(|var| env::var_os(var));
 
-                ENABLED.store(enabled as usize + 1, atomic::Ordering::SeqCst);
+                ENABLED.store(enabled as usize + 1, atomic::Ordering::Relaxed);
 
                 if !enabled {
                     return None;
--- a/amethyst_error/src/lib.rs
+++ b/amethyst_error/src/lib.rs
@@ -318,11 +318,11 @@ static BACKTRACE_STATUS: atomic::AtomicUsize = atomic::ATOMIC_USIZE_INIT;
 
 /// Constructs a new backtrace, if backtraces are enabled.
 fn new_backtrace() -> Option<Backtrace> {
-    match BACKTRACE_ENABLED.load(atomic::Ordering::Relaxed) {
+    match BACKTRACE_STATUS.load(atomic::Ordering::Relaxed) {
         0 => {
             let enabled = is_backtrace_enabled(|var| env::var_os(var));
 
-            BACKTRACE_ENABLED.store(enabled as usize + 1, atomic::Ordering::Relaxed);
+            BACKTRACE_STATUS.store(enabled as usize + 1, atomic::Ordering::Relaxed);
 
             if !enabled {
                 return None;
@@ -410,10 +410,10 @@ mod tests {
 
     #[test]
     fn test_backtrace() {
-        use super::BACKTRACE_ENABLED;
+        use super::BACKTRACE_STATUS;
         use std::sync::atomic;
 
-        BACKTRACE_ENABLED.store(2, atomic::Ordering::Relaxed);
+        BACKTRACE_STATUS.store(2, atomic::Ordering::Relaxed);
 
         #[inline(never)]
         #[no_mangle]
@@ -439,10 +439,10 @@ mod tests {
 
     #[test]
     fn test_backtrace_disabled() {
-        use super::BACKTRACE_ENABLED;
+        use super::BACKTRACE_STATUS;
         use std::sync::atomic;
 
-        BACKTRACE_ENABLED.store(1, atomic::Ordering::Relaxed);
+        BACKTRACE_STATUS.store(1, atomic::Ordering::Relaxed);
 
         assert!(Error::from_string("an error").backtrace().is_none());
     }
--- a/amethyst_error/src/lib.rs
+++ b/amethyst_error/src/lib.rs
@@ -1,0 +1,402 @@
+//! Contains the `Error` type and company as used by Amethyst.
+//!
+//! **Note:** This type is not intended to be used outside of Amethyst.
+//! If you are integrating a crate from amethyst to use this, it is recommended that you treat this
+//! type as an opaque [`std::error::Error`].
+//!
+//! [`std::error::Error`]: https://doc.rust-lang.org/std/error/trait.Error.html
+use self::internal::{Backtrace, HAS_BACKTRACE};
+use std::borrow::Cow;
+use std::env;
+use std::error;
+use std::ffi;
+use std::fmt;
+use std::result;
+use std::sync::atomic;
+
+/// Private
+enum ErrorImpl {
+    Message(Cow<'static, str>),
+    Error(Box<dyn error::Error + Send + Sync>),
+}
+
+impl fmt::Debug for ErrorImpl {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            ErrorImpl::Message(ref message) => fmt::Debug::fmt(message, fmt),
+            ErrorImpl::Error(ref e) => fmt::Debug::fmt(e, fmt),
+        }
+    }
+}
+
+impl From<&'static str> for ErrorImpl {
+    fn from(message: &'static str) -> Self {
+        ErrorImpl::Message(Cow::from(message))
+    }
+}
+
+impl From<String> for ErrorImpl {
+    fn from(message: String) -> Self {
+        ErrorImpl::Message(Cow::from(message))
+    }
+}
+
+impl From<Box<dyn error::Error + Send + Sync>> for ErrorImpl {
+    fn from(error: Box<dyn error::Error + Send + Sync>) -> Self {
+        ErrorImpl::Error(error)
+    }
+}
+
+/// The error type used by Amethyst.
+///
+/// Wraps error diagnostics like messages and other errors, and keeps track of causal chains and
+/// backtraces.
+pub struct Error {
+    error_impl: ErrorImpl,
+    source: Option<Box<Error>>,
+    backtrace: Option<Backtrace>,
+}
+
+impl Error {
+    /// Default constructor for our error types.
+    ///
+    /// Wraps anything that is an error in a box.
+    pub fn new<E>(error: E) -> Self
+    where
+        E: 'static + error::Error + Send + Sync,
+    {
+        Self {
+            error_impl: ErrorImpl::Error(Box::new(error)),
+            source: None,
+            backtrace: new_backtrace(),
+        }
+    }
+
+    /// Construct a new error from a string.
+    pub fn from_string<M>(message: M) -> Self
+    where
+        M: Into<Cow<'static, str>>,
+    {
+        Self {
+            error_impl: ErrorImpl::Message(message.into()),
+            source: None,
+            backtrace: new_backtrace(),
+        }
+    }
+
+    /// Get backtrace.
+    pub fn backtrace(&self) -> Option<&Backtrace> {
+        self.backtrace.as_ref()
+    }
+
+    /// Get the source of the error.
+    ///
+    /// # Examples
+    ///
+    /// The only way to set the source is through [`ResultExt`](ResultExt) using
+    /// [`with_context`](ResultExt::with_context).
+    ///
+    /// ```rust
+    /// use amethyst_error::{Error, ResultExt};
+    /// use std::io;
+    ///
+    /// let e = io::Error::new(io::ErrorKind::Other, "wrapped");
+    /// let a = Error::new(e);
+    ///
+    /// let res = Result::Err::<(), Error>(a).with_context(|_| Error::from_string("top"));
+    /// let e = res.expect_err("no error");
+    ///
+    /// assert_eq!("top", e.to_string());
+    /// assert_eq!("wrapped", e.source().expect("no source").to_string());
+    /// ```
+    pub fn source(&self) -> Option<&Error> {
+        self.source.as_ref().map(|e| &**e)
+    }
+
+    /// Iterate over all causes, including this one.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use amethyst_error::{Error, ResultExt};
+    ///
+    /// fn failing_function() -> Result<(), Error> {
+    ///     Err(Error::from_string("failing"))
+    /// }
+    ///
+    /// fn other_function() -> Result<(), Error> {
+    ///     Ok(failing_function().with_context(|_| Error::from_string("other"))?)
+    /// }
+    ///
+    /// let e = other_function().expect_err("no error");
+    ///
+    /// let messages = e.causes().map(|e| e.to_string()).collect::<Vec<_>>();
+    /// assert_eq!(vec!["other", "failing"], messages);
+    /// ```
+    pub fn causes(&self) -> Causes {
+        Causes {
+            current: Some(self),
+        }
+    }
+}
+
+/// Blanket implementation.
+///
+/// Encapsulate errors which are Send + Sync.
+impl<T> From<T> for Error
+where
+    T: 'static + error::Error + Send + Sync,
+{
+    fn from(value: T) -> Error {
+        Error::new(value)
+    }
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        match self.error_impl {
+            ErrorImpl::Message(ref message) => message.fmt(fmt),
+            ErrorImpl::Error(ref e) => e.fmt(fmt),
+        }
+    }
+}
+
+impl fmt::Debug for Error {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        fmt.debug_struct("Error")
+            .field("error_impl", &self.error_impl)
+            .field("source", &self.source)
+            .finish()
+    }
+}
+
+/// Extra convenience functions for results based on core errors.
+pub trait ResultExt<T>
+where
+    Self: Sized,
+{
+    /// Provide a context for the result in case it is an error.
+    ///
+    /// The context callback is expected to return a new error, which will replace the given error
+    /// and set the replaced error as its [`source`](Error::source).
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use amethyst_error::{Error, ResultExt};
+    ///
+    /// fn failing_function() -> Result<(), Error> {
+    ///     Err(Error::from_string("failing"))
+    /// }
+    ///
+    /// fn other_function() -> Result<(), Error> {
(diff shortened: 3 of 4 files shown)


amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+        ErrorImpl::Error(error)
+    }
+}
+
+/// The error type used by Amethyst.
+///
+/// Wraps error diagnostics like messages and other errors, and keeps track of causal chains and
+/// backtraces.
+pub struct Error {
+    error_impl: ErrorImpl,
```
torkleyy: Hmm, this means the error is bigger than 3 times pointer size.

I think the best strategy is, independent of how std will adopt failure, to have two kinds of errors:

* cheap to construct enums for things that may fail but can easily be recovered from
* allocated, pointer sized errors for things that usually don't fail as to not slow down the succesful execution

I'll explain this more later today.

amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+use self::internal::{Backtrace, HAS_BACKTRACE};
+use std::borrow::Cow;
+use std::env;
+use std::error;
+use std::ffi;
+use std::fmt;
+use std::result;
+use std::sync::atomic;
+
+/// Private
```
TimonPost: Why this comment, no really needed.

amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+}
+
+/// The error type used by Amethyst.
+///
+/// Wraps error diagnostics like messages and other errors, and keeps track of causal chains and
+/// backtraces.
+pub struct Error {
+    error_impl: ErrorImpl,
+    source: Option<Box<Error>>,
+    backtrace: Option<Backtrace>,
```
DoumanAsh: Do not include `Backtrace` at all if it is not enabled by compile feature (extra size for no needs)

amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+        _ => false,
+    }
+}
+
+fn new_backtrace() -> Option<Backtrace> {
+    if !HAS_BACKTRACE {
+        return None;
+    }
+
+    static ENABLED: atomic::AtomicUsize = atomic::ATOMIC_USIZE_INIT;
```
DoumanAsh: What is this atomic for used?

amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+use std::error;
+use std::ffi;
+use std::fmt;
+use std::result;
+use std::sync::atomic;
+
+/// Private
+enum ErrorImpl {
+    Message(Cow<'static, str>),
+    Error(Box<dyn error::Error + Send + Sync>),
```
DoumanAsh: Current `ametyst::Error` is not `Sync`, why requiring it here?

amethyst_error/src/lib.rs
DoumanAsh: To add to what torkleyy says.

It might be better to have common `Error` as

```rust
pub enum Error {
    Thin(PlainError),
    Fat(FatError)
}
```

Where `FatError` could contains extra information while `PlainError` would be cheap enum (most likely used by `amethyst` APIs)

udoprog: The cheap errors are better built locally in the crate that uses them, instead of having one centralized type.

torkleyy: Yes, definitely.

torkleyy: That is, if by "cheap" you mean enums. Both variants are cheap in their own way.

amethyst_error/src/lib.rs
udoprog: It's an important diagnostic tool. It will probably be enabled. See the linked issue at the top.

torkleyy: I think what @DoumanAsh meant is to feature flag this field as to not make the struct any bigger than necessary.

DoumanAsh: Yes, since we control it by feature, you should also exclude it from `Error` as it is overhead for users like me who wouldn't use backtrace

amethyst_error/src/lib.rs
udoprog: Accidentally resolved. But it makes life easier, and I'm mimicking the behavior of failure.

amethyst_error/src/lib.rs
DoumanAsh: Ok I thought this crate would introduce general error for all `amethyst` crates, if cheap errors will remain in other crates then I don't see this bloated error as bad thing

omni-viral: But this `Error` is even fatter than `FatError`
```
pub enum Error {
    Thin(PlainError),
    Fat(FatError)
}
```

DoumanAsh: It replaces heap allocations at cost of single integer

omni-viral:
@DoumanAsh 
> I hope amethyst is not going to adopt awful failure like errors

What are issues with `failure`?

amethyst_error/src/lib.rs
udoprog: Ah, sure!

udoprog:
@omni-viral

> What the purpose of this crate?

Provide an implementation of `failure::Error`-like error type, that doesn't implement `failure::Fail` that can be used in Amethyst.

> If you mimicking failure then why not just use it?

Because the approach it is being based on (an external trait fit feeling with errors) is experimental and is being deprecated in favor of fixing `std::error::Error`.
We are stuck in a catch 22 where we can't use failure because it's being deprecated, and we can't use std Error because it's not done, and there are no libraries that support building those kinds of errors yet.

This breaks that by explicitly only providing an internal API that we can control. This provides the minimum amount of guarantees to our users (`impl std::error::Error`).

It is designed to support the future std trait once it becomes available, and can hopefully be phased out once better alternatives are. One thing that is guaranteed to happen is the std:ification of backtrace, for example.

Now, there is no std:ification happening w.r.t. error compositing. Which is why that API mimicks the one provided by failure, minus implementing `Fail`.

> If you think failure has some serious issues then why not fix them in failure?

The only problem is that the approach of having split Error traits above turned out to cause ecosystem issues. So it's advisable in the mid term to avoid exporting it in APIs.

amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+        1 => return None,
+        _ => {}
+    }
+
+    Some(Backtrace::new())
+}
+
+const RUST_BACKTRACE: &str = "RUST_BACKTRACE";
+
+#[cfg(all(feature = "backtrace", feature = "std"))]
```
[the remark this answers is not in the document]
udoprog: This is for `no_std` support which is enabled through an attribute. I'm not sure you can make flags depend on it but I'll give it a shot!

amethyst_error/src/lib.rs
torkleyy: @udoprog To explain my suggestion more, I'd imagine that `Erorr` is only a `Box` around `ErrorImpl<E>` with `E: Error + ?Sized`. At least that's the pattern I prefer for this kind of things, because that way all the error data is on the heap, while at the same time there's only one allocation (per fat error).

amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+impl fmt::Debug for ErrorImpl {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            ErrorImpl::Message(ref message) => fmt::Debug::fmt(message, fmt),
+            ErrorImpl::Error(ref e) => fmt::Debug::fmt(e, fmt),
+        }
+    }
+}
+
+impl From<&'static str> for ErrorImpl {
```
torkleyy: I definitely see the motivation for this, but I think we should discourage using strings anywhere in the engine.

amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+    }
+}
+
+/// The error type used by Amethyst.
+///
+/// Wraps error diagnostics like messages and other errors, and keeps track of causal chains and
+/// backtraces.
+pub struct Error {
+    error_impl: ErrorImpl,
+    source: Option<Box<Error>>,
```
torkleyy: Can we call this `cause`?

amethyst_error/src/lib.rs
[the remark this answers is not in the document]
torkleyy: Why would Amethyst need `no_std`?

amethyst_error/src/lib.rs
DoumanAsh: I'm not sure why static strings should be discouraged?

DoumanAsh:
>What are issues with failure?

it is poor man replacement for exceptions that carry tons of information

amethyst_error/src/lib.rs
DoumanAsh: Oh maybe @torkleyy approach would be better, if we assume that errors are rare anyway

amethyst_error/src/lib.rs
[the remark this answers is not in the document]
udoprog: Probably wouldn't. This is a requirement of the backtrace lib, but I can relax it.

The difference is that anything depending on no_std and amethyst_error _will_ blow up during compilation instead of degrading to not have backtraces.

amethyst_error/src/lib.rs
udoprog: Working around `!Sync` is a lot harder than `Sync`. Enforcing `Sync` is IMO the more conservative option, and I didn't encounter any instance in the current codebase where we needed `!Sync`.

In the future we can loosen the constraint if we find a good case for it. But I'd prefer to wait until we have one.

udoprog: This has also kinda become the norm in error types now because of the popularity of failure.

amethyst_error/src/lib.rs
LucioFranco: Discourages low effort error handling. Strings should be embedded at the lower leaf level of errors, not on the high level type.

I agree with @torkleyy I would like to see the `Message` part dropped imo.

amethyst_error/src/lib.rs
udoprog: We could... but the rfc uses `source` since `cause` is occupied. It's not a big deal though since I think we'll build done kind of error handling abstraction in utils which will most likely be the only one using this right now.

amethyst_error/src/lib.rs
LucioFranco: I kinda agree, I would like to drop the `Sync` constraint. I think it will cause way more pain then it will benefit.

udoprog:
@DoumanAsh it is very important to distinguish what the _purpose_ of such an error type is.

One is used to communicate simple local errors in your API. For this is perfectly acceptable to use an efficient error type. The requirement on these are either that a) they can be turned into a string, or b) that they can somehow be converted into a boxable std::error::Error. For this anything works, basically.

The second use-case is higher level compositing of various APIs. Here it just simply doesn't work very well to dictate a single concrete error type. That is where types like failure::Error and this comes in. See #1214 for a simplified example of this.

amethyst_error/src/lib.rs
udoprog: This can certainly be done once we've refactored all instances where we use them, but not before. If like to punt that as out of scope for now since it would be a significant undertaking.

amethyst_error/src/lib.rs
udoprog: @torkleyy I'm gonna try to wrap my head around what you are proposing. I'll poke you on Discord.

amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+}
+
+fn new_backtrace() -> Option<Backtrace> {
+    if !HAS_BACKTRACE {
+        return None;
+    }
+
+    static ENABLED: atomic::AtomicUsize = atomic::ATOMIC_USIZE_INIT;
+
+    match ENABLED.load(atomic::Ordering::SeqCst) {
```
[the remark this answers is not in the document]
udoprog: I agree, changed. Thank you!

torkleyy:
https://community.amethyst-engine.org/t/error-handling-discussion/294

amethyst_error/src/lib.rs
omni-viral: Is there real use case for writing `!Sync` error type?

amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+    backtrace: Option<Backtrace>,
+    error: T,
+}
+
+/// The error type used by Amethyst.
+///
+/// Wraps error diagnostics like messages and other errors, and keeps track of causal chains and
+/// backtraces.
+pub struct Error {
+    inner: Box<Inner<dyn error::Error + Send + Sync>>,
```
omni-viral: This way `Error` size is two pointer.
Considering that wrapped errors should be rare it make sense to limit its side to one pointer.
Even at the cost of additional allocation e.g. `Box<Box<UnsizedType>>`.
Because this cost only applied to error path, while happy path gets smaller return type.

amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+    /// Wraps anything that is an error in a box.
+    pub fn new<E>(error: E) -> Self
+    where
+        E: 'static + error::Error + Send + Sync,
+    {
+        Self {
+            inner: Box::new(Inner {
+                source: None,
+                #[cfg(all(feature = "backtrace", feature = "std"))]
+                backtrace: self::internal::bt::new(),
```
omni-viral: So each time I wrap it collects brand new backtrace.
```
fn bar() -> Result<(), Error> {
  foo().with_context(|_| Error::from_string("foo failed in `bar`"))
}
```
Now `foo` returns an error with backtrace and `bar` captures one more redundant backtrace.

amethyst_error/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+
+[badges]
+appveyor = { repository = "amethyst/amethyst", branch = "master" }
+travis-ci = { repository = "amethyst/amethyst" }
+
+[dependencies]
+backtrace = {version = "0.3.9", optional = true}
+
+[features]
+default = ["backtrace"]
```
omni-viral: Code won't capture backtraces without missing `std` feature.

amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+//! If you are integrating a crate from amethyst to use this, it is recommended that you treat this
+//! type as an opaque [`std::error::Error`].
+//!
+//! [`std::error::Error`]: https://doc.rust-lang.org/std/error/trait.Error.html
+
+// Parts copied from failure:
+// https://github.com/rust-lang-nursery/failure
+
+#![warn(missing_docs, rust_2018_idioms, rust_2018_compatibility)]
+
```
omni-viral: `#![cfg_attr(not(feature = "std"), no_std)]` should be here to make this compatible with `no_std`.

amethyst_error/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+readme = "README.md"
+description = """
+Internal error handling for Amethyst.
+"""
+keywords = ["game", "error", "amethyst"]
+categories = ["error"]
+
+documentation = "https://www.amethyst.rs/doc/master/doc/amethyst/"
+homepage = "https://www.amethyst.rs/"
+repository = "https://github.com/amethyst/amethyst"
```
omni-viral: It's time for new edition :)

amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+use self::internal::Backtrace;
+use std::borrow::Cow;
+use std::error;
+use std::fmt;
+use std::result;
+
+/// Internal parts of `Error`.
+#[derive(Debug)]
+struct Inner<T: ?Sized> {
+    source: Option<Box<Error>>,
```
omni-viral: Usually, after top-level error, the most interesting one is root cause.

amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+            1 => return None,
+            _ => {}
+        }
+
+        Some(Backtrace::new())
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
omni-viral: We should add test to check backtrace functionality.

amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+        {
+            None
+        }
+    }
+
+    /// Get the source of the error.
+    ///
+    /// # Examples
+    ///
+    /// The only way to set the source is through [`ResultExt`](ResultExt) using
```
omni-viral: We should provide a way to create `Error` with source without `ResultExt`.
Simple `with_context` method that works same way as `ResultExt` would do.

amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+        }
+    }
+
+    /// Access the internal `std::error::Error` as a trait.
+    ///
+    /// This can be useful for integrating with systems that operate on `std::error::Error`.
+    ///
+    /// **Warning:** This erases most diagnostics in favor of returning only the top error.
+    /// `std::error::Error` is expanded further.
+    pub fn as_error(&self) -> &dyn error::Error {
```
omni-viral: Similar to `std::error::Error` this should return `&dyn (Error + 'static)`

amethyst_error/src/lib.rs
udoprog: I was going to consider dropping `no_std` completely based on [a comment above](https://github.com/amethyst/amethyst/pull/1220#discussion_r239069562), what would your input be on this?

amethyst_error/src/lib.rs
udoprog: :+1:

amethyst_error/src/lib.rs
udoprog: Good call!

amethyst_error/src/lib.rs
udoprog: So is your proposal is to only capture the backtrace for the root error? How would we do this?

amethyst_error/Cargo.toml
udoprog: :+1:

amethyst_error/src/lib.rs
udoprog: Added `Error::with_source`.

amethyst_error/src/lib.rs
udoprog: I'm not convinced. A `2*usize` return value is already tiny and doesn't automatically hold its own weight compared to the extra indirection another pointer would for me.

amethyst_error/src/lib.rs
derekdreery: From the nomicron:

> Major exceptions include:
>
>  - raw pointers are neither Send nor Sync (because they have no safety guards).
>  - UnsafeCell isn't Sync (and therefore Cell and RefCell aren't).
>  - Rc isn't Send or Sync (because the refcount is shared and unsynchronized).

So the error type would have to have some kind of shared ownership, basically one of `RefCell., `Cell`, `Rc`, or `*(const|mut)`. I'd be very suprised if this happens in practice, and I think it would denote a design error - `std::error::Error` is when something's gone wrong and you're probably just going to print it to the user and give up (it implements `Display` and has a `source`) - if you're doing actual game logic in it use a custom type.

amethyst_error/src/lib.rs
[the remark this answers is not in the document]
derekdreery: specs isn't going to work without an allocator.

amethyst_error/src/lib.rs
udoprog: It guards testing for the `RUST_BACKTRACE` environment variable, to be more efficient on subsequent tests.

amethyst_error/src/lib.rs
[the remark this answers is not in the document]
udoprog: backtrace is no longer optional, thanks for the feedback!

amethyst_error/Cargo.toml
udoprog: Backtrace is no longer optional.

amethyst_error/src/lib.rs
udoprog: Done!

amethyst_error/src/lib.rs
udoprog: I would expect most error reporters to print the whole causal chain. Plus, errors are supposed to be "less frequent", so taking a performance hit when looking for the root cause doesn't seem overly severe.

amethyst_error/src/lib.rs
udoprog: Backtraces are no longer optional!

derekdreery:
I'm just going to write a bit of a blurb about how I see error handling in rust atm. It's fairly subjective, and there are plenty of things to argue with.

Rust has some great abstractions for handling failure situations, where some I/O or something else hasn't worked out. You have 3 choices when something hasn't worked.

 1. Ignore the failure and continue anyway
 2. (If you're in a function) return a `Result<T, E>` instead of `T`
    1. the type `E` implements `std::error::Error`
    2. the type `E` has no constraints
 3. `panic!`

I would try to avoid (1) at all costs. For me, there are some instances of this in amethyst - I think there are systems that just do nothing if some resource or component isn't present that should be. This indicates a programmer error and they should panic (with a message explaining the problem). 

It's also not true IMO that panicking is strictly worse than `Result`, sometimes it is more appropriate (where there is mis-configuration or programmer error), as long as the invariants required to avoid it are well documented.

If you're using a `Result`, then you need to decide what the error type will be. I would argue that `std::error::Error` is really only for errors that will be reported to a person as text. Its main functionality is that it can be printed (using `Display`) and can contain another error (using `source`). If you're not using this functionality, you're essentially implementing `Error` just as a marker, which is not worth doing. Any error you're going to recover from is probably not worth implementing `Error` for.

This brings me to my next point - sometimes it's OK for errors to be strings of text: specifically when the error is unrecoverable and only going to be consumed by a human. Say you have a function that can error, and there is only 1 way you can error. You'll probably do something like the following

```rust
#[derive(Debug)]
struct MyError;

impl Display for MyError {
    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
        f.write_str("this is the error message")
    }
}
```

That's a lot of lines to just give a string error, and since there is only one way it can fail, a machine can just ignore the value of the error.

Of course there are gray areas, mostly when you have an error that you *may* recover from. For example, if a texture isn't available, you might want to continue and display an error message in-game, because that helps modders more than crashing out. In that case you would use an enum error, and implement `Display` as a courtesy, so the game developer can decide what to do.

Another issue is how to deal with `io::Error`, and it demonstrates the value, and limitations, of error chains. It may be that there are some kinds I expect more - maybe that a file is missing, or that I don't have permission to write a file I need to. You can test for these kinds and convert them to your own error kind, but if you encounter an error (say `InvalidData` for example) that you didn't expect, you want to give the user as much information as possible. On the other hand, all the information you got from the OS was *invalid data* that's not going to be much more use than *an unexpected error occurred*.

# Summary

 1. Avoid failing silently, a.k.a. allowing code paths that you haven't really thought about, and where you don't know what's going to happen.
 1. Use `panic!` to indicate to a developer using your crate that he has not upheld some invariant you insist upon.
 2. Use `Result<T, E>` for things that might fail otherwise.
 3. Always implement `std::error::Error` when an error might be used in an error chain. 
 4. Maybe implement `std::error::Error` at other times (it's not 100% essential, and you're basically just doing it for the `Display` impl and the marker).
 4. Keep the happy path fast and lean.
 5. `Box` your error if it contains a lot of data. `Result` is as wide as its widest variant, and you're only going to pay the cost of indirection once something has already gone wrong (i.e. performance probably isn't so important any more). I think it's perfectly acceptable to return `Result<T, Box<MyError>>` when you want to keep the size down. As always with performance, benchmark. :)
 5. Trait objects make everything more complicated. You start having issues with `Send/Sync/'static`. Most of the time you don't need them, you can just use `cause: ConcreteInnerErrorType`. The only time you use `dyn Error (+ 'static + Send + Sync + ...)` is when you're implementing `source`.
 6. Sometimes it makes sense to wrap the underlying error (e.g. IO error), sometimes it makes sense to subsume it (e.g. unit struct error). Use your judgment.
 7. Don't be afraid of just implementing the trait in the way that makes sense for your error type. You don't need to use a procedural macro and shouldn't unless it's literally just to save some boilerplate.

# Example

```rust
use std::{
    error::Error,
    fmt,
    fs::File,
    io,
    path::{Path, PathBuf},
};

pub fn load_image(image_path: &str) -> Result<Image, MyError> {
    let mut img_src =
        File::open(image_path).map_err(|e| MyError::error_opening_file(image_path, e))?;
    let img = process_image(&mut img_src).map_err(|e| MyError::error_in_lib(image_path, e))?;
    Ok(img)
}

// Error types

/// The error type for this function. Could be re-used if there are other functions with exactly
/// the same error possibilities.
#[derive(Debug, Clone)]
pub struct MyError {
    // These could be public, but using accessor methods leaves open the possibilities of
    // reorganising things in the future.
    /// The path to the image we tried to load
    image_path: PathBuf,
    /// The type of error we encountered
    kind: MyErrorKind,
}

#[derive(Debug)]
pub enum MyErrorKind {
    /// Error opening file stream.
    OpeningFile(io::Error),
    // here we've chosen to consume the unerlying error. We could decide to wrap it, but I think
    // that would be more complicated.
    /// Error reading data from the file stream.
    ReadingData(io::Error),
    /// The magic bytes were wrong
    Magic([u8; 4]),
    /// Data in the stream was invalid
    Invalid(usize, Option<String>),
    /// Some error we didn't expect.
    // By capturing the error this way, we can still use `Display` to print out anything in it.
    // It's the only time you want to be storing it as a trait object.
    UnknownLib(Box<dyn Error + Send + Sync>),
}

// Impls

impl MyError {
    /// What type of error occurred.
    pub fn kind(&self) -> &MyErrorKind {
        &self.kind
    }

    /// The image file we were trying to work with when this error occurred.
    pub fn image_path(&self) -> &Path {
        self.image_path.as_ref()
    }

    /// Wrap an io::Error the way we want. Helper method
    fn error_opening_file(image_path: impl Into<PathBuf>, e: io::Error) -> Self {
        let kind = MyErrorKind::OpeningFile(e); // io::Error isn't too wide, so no indrection.
        MyError::from_parts(image_path, kind)
    }

    /// Process a library error. Helper method
    fn error_in_lib(image_path: impl Into<PathBuf>, e: ImageError) -> Self {
        let kind = match e {
            ImageError::Io(e) => MyErrorKind::ReadingData(e),
            ImageError::Magic(bytes) => MyErrorKind::Magic(bytes),
            ImageError::Invalid(pos, msg) => MyErrorKind::Invalid(pos, msg),
            e => MyErrorKind::UnknownLib(Box::new(e) as Box<dyn Error + Send + Sync>),
        };
        MyError::from_parts(image_path, kind)
    }

    /// Boilerplate for creating the struct. Helper method
    fn from_parts(image_path: impl Into<PathBuf>, kind: MyErrorKind) -> Self {
        MyError {
            image_path: image_path.into(),
            kind,
        }
    }
}

impl fmt::Display for MyError {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        let image_path = self.image_path.display();
        match self.kind() {
            MyErrorKind::OpeningFile(e) => {
                let cannot_open = format!("cannot open file at \"{}\" as an image because", image_path);
                match e.kind() {
                    io::ErrorKind::NotFound => write!(f, "{} it could not be found", cannot_open),
                    io::ErrorKind::PermissionDenied => write!(f, "{} this process doesn't have permission to read it", cannot_open),
                    _ => write!(f, "{} of an unexpected i/o error: {}", cannot_open, e),
                }
            }
            // If we've already opened the file - something wierd is happening if we can't read it,
            // so just print the i/o error:
            MyErrorKind::ReadingData(e) => write!(f, "cannot read from file at \"{}\" because of an unexpected i/o error: {}", image_path, e),
            MyErrorKind::Magic(bytes) => write!(f, "while reading image at \"{}\", expected the magic bytes [1, 2, 3, 4] but found {:?}", image_path, bytes),
            MyErrorKind::Invalid(pos, msg) => {
                let invalid_offset = format!("image at \"{}\" has an invalid byte at offset {}", image_path, pos);
                match msg {
                    Some(msg) => write!(f, "{}: {}", invalid_offset, msg),
                    None => write!(f, "{}", invalid_offset),
                }
            }
            MyErrorKind::UnknownLib(e) => write!(
                f,
                "while decoding image at \"{}\", received the error \"{}\"",
                self.image_path.display(), e
            ),
        }
    }
}

// We can't derive this because io::Error doesn't implement clone. We're just going to discard any
// inner error here.
impl Clone for MyErrorKind {
    fn clone(&self) -> MyErrorKind {
        match self {
            // Discard what we have to to make this cloneable.
            MyErrorKind::OpeningFile(e) => MyErrorKind::OpeningFile(e.kind().into()),
            MyErrorKind::ReadingData(e) => MyErrorKind::ReadingData(e.kind().into()),
            // Next things we can clone
            MyErrorKind::Magic(bytes) => MyErrorKind::Magic(*bytes),
            MyErrorKind::Invalid(pos, msg) => MyErrorKind::Invalid(*pos, msg.clone()),
            // The error might not be clone, so just convert it to text.
            MyErrorKind::UnknownLib(e) => MyErrorKind::UnknownLib(Box::new(StringError::from(
                e.to_string(),
            ))
                as Box<dyn Error + Send + Sync>),
        }
    }
}

// I really think Error should be impl for String
#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Default)]
#[repr(transparent)]
pub struct StringError(String);

impl fmt::Display for StringError {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        fmt::Display::fmt(&self.0, f)
    }
}

impl Error for StringError {}

impl From<String> for StringError {
    fn from(s: String) -> StringError {
        StringError(s)
    }
}

impl From<StringError> for String {
    fn from(e: StringError) -> String {
        e.0
    }
}

// External

// Imagine ImageError looks like this. It's good to think how you would wrap all kinds of errors,
// since different crate authors will do different things.

// Because ImageError is Send + Sync, changing this would be backwards-incompatible.
// That's why we can use `dyn Error + Send + Sync` above. This isn't always true.
#[derive(Debug)]
pub enum ImageError {
    /// An i/o error occurred.
    Io(io::Error),
    /// The magic 4 bytes were wrong.
    Magic([u8; 4]),
    /// A byte in the file was invalid for its position. Some more information is given as text.
    Invalid(usize, Option<String>),
    #[doc(hidden)]
    __NonExhaustive,
}

impl fmt::Display for ImageError {
    fn fmt(&self, _f: &mut fmt::Formatter) -> fmt::Result {
        unimplemented!()
    }
}

impl Error for ImageError {}

pub fn process_image(_r: &mut impl io::Read) -> Result<Image, ImageError> {
    unimplemented!()
}

pub type Image = ();
```

udoprog:
@derekdreery I appreciate the post! It does seem more suitable for the [thread linked by @torkleyy earlier](https://community.amethyst-engine.org/t/error-handling-discussion/294) to try and limit this one to the technical aspects of the PR. Would you mind moving it?

udoprog:
I've rebased the branch I used PoC:ing with amethyst_error (linked above):
https://github.com/udoprog/amethyst/commit/amethyst-error-full

If anyone is curious about how many stringy errors we have in there, you can now grep for `format_err`. These are potential candidates to be replaced with crate-local error types that can be constructed easily using something like @torkleyy's [`err-derive`](https://crates.io/crates/err-derive).

The traits I've identified so far that has awkward error depenendencies which are candidates for using `amethyst_error` in their APIs are:

* [`amethyst_assets::Source`](/amethyst/amethyst/blob/master/amethyst_assets/src/source/mod.rs)
* [`amethyst_assets::PrefabData`](/amethyst/amethyst/blob/master/amethyst_assets/src/prefab/mod.rs)

And this is a maybe:

* [`amethyst_renderer::Pass/Pipeline` stuff](/amethyst/amethyst/blob/master/amethyst_renderer/src/pipe/pass.rs). Unsure because it spans over to UI.

udoprog:
@LucioFranco @omni-viral - you have blocking reviews and I think I've fixed or responded to your feedback. Can you look again?

amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+
+/// Test if backtracing is enabled.
+fn is_backtrace_enabled<F: Fn(&str) -> Option<ffi::OsString>>(get_var: F) -> bool {
+    match get_var(RUST_BACKTRACE) {
+        Some(ref val) if val != "0" => true,
+        _ => false,
+    }
+}
+
+static BACKTRACE_ENABLED: atomic::AtomicUsize = atomic::ATOMIC_USIZE_INIT;
```
torkleyy: Could you rename this?

```suggestion
// 0: unchecked
// 1: disabled
// 2: enabled
static BACKTRACE_STATUS: atomic::AtomicUsize = atomic::ATOMIC_USIZE_INIT;
```

amethyst_error/src/lib.rs
omni-viral: Return type size can make significant difference in generated instructions. And additional indirection occur only in case of error, when, as you stated, performance is not of concern.

But we probably should benchmark it first.

udoprog: Since this doesn't affect the API I suggest resolving this discussion. We can experiment (and change) this without much effort if we find the need to in the future.

amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,449 @@
...
+}
+
+impl<T, E> ResultExt<T> for result::Result<T, E>
+where
+    E: Into<Error>,
+{
+    fn with_context<C, D>(self, chain: C) -> Result<T, Error>
+    where
+        C: FnOnce(&Error) -> D,
+        D: Into<Error>,
```
omni-viral: `D` should impl `Into<Error>` but `error::Error` instead.

amethyst_error/src/lib.rs
```diff
@@ -0,0 +1,449 @@
...
+            inner: Box::new(Inner {
+                source: None,
+                backtrace: new_backtrace(),
+                error: Box::new(error),
+            }),
+        }
+    }
+
+    /// Update the source of an error.
+    pub fn with_source<S>(mut self, source: S) -> Self
```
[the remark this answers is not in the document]
udoprog: At the moment I don't agree that we shouldn't capture all backtraces. In particular I don't think we can guarantee that we capture all important context to troubleshoot an error.

I suggest opening a discussion about it on discourse where you can elaborate more on why you think it's important, and we can fix it in a future PR.

amethyst_error/src/lib.rs
udoprog: Converting into `Error` is to guarantee capturing a backtrace at these composition points. Similarly to how `with_context` returns `Context<E>` in failure.

amethyst_error/src/lib.rs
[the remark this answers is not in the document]
udoprog: Moved here: https://community.amethyst-engine.org/t/error-handling-discussion/294/19?u=udoprog

amethyst_error/README.md
```diff
@@ -0,0 +1,18 @@
+# amethyst_error
+
+This crate contains types used internally and externally to handle and communicate errors.
```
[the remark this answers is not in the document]
LucioFranco: Great, this sounds like a good thing to create a post on the forums for 😄

amethyst_error/src/lib.rs
omni-viral: Capturing a backtrace is exactly what I want to avoid in this function.

fhaynes:
@omni-viral we still good to merge this one?

udoprog:
I believe this is ready to merge. I'm holding off since I'm pretty busy this week (Work + GDQ).

OTOH, If you want to merge it before I can keep an eye on it, go ahead!

LucioFranco:
Seems the mdbook curl call failed so lets retry

bors r=omni-viral,torkleyy,LucioFranco

LucioFranco:
I've reproduced this failure on my macbook

```
---- tests::test_backtrace stdout ----
thread 'tests::test_backtrace' panicked at 'assertion failed: frame_names.iter().find(|n| *n == "a_really_unique_name_42").is_some()', amethyst_error/src/lib.rs:434:9
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

@udoprog any idea why this might be coming up? if not ill take a deeper look tonight.

udoprog:
bors retry

Frame formatting isn't identical across platforms, I'm trying to relax the comparison in the test a bit.

udoprog:
bors retry

udoprog:
bors retry

udoprog:
bors r=omni-viral,torkleyy,LucioFranco

udoprog:
Looks like a test in amethyst_network breaks sometimes:
https://travis-ci.org/amethyst/amethyst/jobs/477278701#L1744

@LucioFranco do you know what's up?

udoprog:
bors retry

since #1335 was merged

udoprog:
Thank you bors <3

PR #1221, opened 2018-12-05
Add config file for GitLab CI/CD

fhaynes:
Since we now have access to the Gold level features of GitLab, this PR adds a basic config for it. Auto-mirroring is already setup in GitLab, but we need the config file in the repo, otherwise the mirroring removes it. This CI/CD file has been tested and is mostly ported from the Concourse config.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,34 @@
+image: 'rust:latest'
+
+stages:
+  - fmt
+  - test
+
+variables:
+  CARGO_HOME: $CI_PROJECT_DIR/cargo
+  APT_CACHE_DIR: $CI_PROJECT_DIR/apt
+  RUSTFLAGS: "-D warnings"
+
+before_script:
+  - apt-get update -yq
+  - rustup component add rustfmt
+  - apt-get install -o dir::cache::archives="$APT_CACHE_DIR" -y alsa-utils libasound2-dev cmake git libasound2-dev libx11-xcb-dev libssl-dev cmake libfreetype6-dev libexpat1-dev libxcb1-dev python3 build-essential libsdl2-dev
+
+check formatting:
+  stage: fmt
+  script:
+    - cargo fmt --version
+    - cargo fmt --all -- --check
+
+run tests:
+  stage: test
+  script:
+    - rustc --version
+    - cargo --version
+    - cargo test --verbose --all --all-features
+
+cache:
+  paths:
+    - apt/
+    - cargo/
+    - target/


.gitlab-ci.yml
```diff
@@ -0,0 +1,34 @@
...
+  script:
+    - cargo fmt --version
+    - cargo fmt -- --check
+
+run tests:
+  stage: test
+  script:
+    - rustc --version
+    - cargo --version
+    - cargo test --verbose --features sdl_controller,profiler
```
torkleyy: I think we need to test json and profiler, too.

fhaynes: profiler is in there, I think. or is it a different profiler?

torkleyy: Um, yeah, missed it.

torkleyy: I think it could just be `--all-features`.

fhaynes:
@torkleyy This will eventually replace the Concourse server, which will let us get rid of one server.

torkleyy:
Ah okay! Do you know how the performance compares? Or are we just gonna try it out?

fhaynes:
https://gitlab.com/amethyst-engine/amethyst/-/jobs has details. 35m to test everything.

.gitlab-ci.yml
fhaynes: Changed.

.gitlab-ci.yml
```diff
@@ -0,0 +1,34 @@
...
+  - test
+
+variables:
+  CARGO_HOME: $CI_PROJECT_DIR/cargo
+  APT_CACHE_DIR: $CI_PROJECT_DIR/apt
+  RUSTFLAGS: "-D warnings"
+
+before_script:
+  - apt-get update -yq
+  - rustup component add rustfmt-preview
```
LucioFranco: I believe this is changing soon to `rustfmt` no longer with the `preview`.

.gitlab-ci.yml
```diff
@@ -0,0 +1,34 @@
...
+  - apt-get install -o dir::cache::archives="$APT_CACHE_DIR" -y alsa-utils libasound2-dev cmake git libasound2-dev libx11-xcb-dev libssl-dev cmake libfreetype6-dev libexpat1-dev libxcb1-dev python3 build-essential libsdl2-dev
+
+check formatting:
+  stage: fmt
+  script:
+    - cargo fmt --version
+    - cargo fmt -- --check
```
LucioFranco: `cargo fmt --all -- --check`

.gitlab-ci.yml
```diff
@@ -0,0 +1,34 @@
+image: 'rust:latest'
```
LucioFranco: Does this guarantee that we are on stable? How do toolchain updates get propagated?

.gitlab-ci.yml
fhaynes: Does it work now without the preview? If not, we'll have to wait to change it

LucioFranco: I don't think so yet, but it will happen soon, we can open an issue for it.

.gitlab-ci.yml
fhaynes: The latest tag is a relative tag. It always has the latest stable.

.gitlab-ci.yml
fhaynes: Opened #1223 for it

fhaynes: #1223

.gitlab-ci.yml
LucioFranco: Do we depend now on docker to update it or rustup? I would rather install the toolchain each time that way we can reduce possible issues.

fhaynes: The rust team updates the docker image. https://hub.docker.com/_/rust/

I'm ok with not using the latest tag and specifying the latest stable, but not manually installing rustup and configuring all that when it is already done in their official image.

LucioFranco: Sure, I would say tho if the toolchain is already installed and rustup is also already installed the `rustup install stable` should be very quick.

fhaynes: But to what end? When a new Rust stable version comes out, we just bump the image tag.

LucioFranco: You would rather do that manually? I would like this to be automated and on release days its not a thing we have to think about. Im fine though keeping the image.

fhaynes: No, I would rather use the latest tag which does exactly what you are asking for. =P

LucioFranco: Works for me!

fhaynes:
I'm just going to redo this one from a fresh fork.

PR #1232, opened 2018-12-08
Add some docs for actions.

derekdreery:
Explains the structure of `Bindings::action`.

### DIFF
--- a/amethyst_input/src/bindings.rs
+++ b/amethyst_input/src/bindings.rs
@@ -18,10 +18,10 @@ where
     AC: Hash + Eq,
 {
     pub(super) axes: HashMap<AX, Axis>,
-    // The inner array here is for button combinations, the other is for different possibilities.
-    //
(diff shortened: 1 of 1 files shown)


amethyst_input/src/bindings.rs
```diff
@@ -18,6 +18,10 @@ where
     AC: Hash + Eq,
 {
     pub(super) axes: HashMap<AX, Axis>,
+    // The inner array here is for button combinations, the other is for different possibilities.
+    //
+    // So for example if you want to quit by either "Esc" or "Ctrl+q" you would have
+    // `[[Esc], [Ctrl, Q]]`.
```
udoprog: Even though this is a field with limited visibility (`pub(super)`) we should be using doc comments `///` (three slashes instead of two).

Xaeroxe:
Perfect!

bors r+

PR #1236, opened 2018-12-10
Add missing dependancy Flipped to Pong tutorial

AndrewJakubowicz:
As I was working through the tutorial I got blocked because `Flipped` isn't included in the initial dependency dump.

In the example that `Flipped` is used, there isn't an explicit import.
With this change the user going through the tutorial won't have to hunt for the dependency.


happenslol:
Thanks, totally missed the `Flipped` import there when I introduced that component. If you squash the commits, we can merge this imo!

AndrewJakubowicz:
Thanks for the reviews. You guys are doing excellent work and I look forward to hopefully contributing more in the future.

Squashed as requested by @happenslol

TimonPost:
Going to merge this since it is squashed an approved.

bors r=udoprog, Xaeroxe,happenslol,TimonPost

PR #1237, opened 2018-12-10
Derive `Copy`, `PartialEq`, `Eq`, `Serialize`, `Deserialize` on `Flipped` component.

azriel91:


azriel91:
Should rebase over #1239 instead of doing a sneaky changelog move.

PR #1240, opened 2018-12-10
Replace gitter chat w/ discord in CONTRIBUTING.md

distransient:
Gitter was also missing its link so this fixes that as well.


Xaeroxe:
Since this is unlikely to be controversial I'm going to bend the rules a bit, and merge this before the 24 hours is up.

bors r+

PR #1242, opened 2018-12-10
Maintenance/update vergen

azriel91:
`vergen` doesn't have docs on docs.rs for version `3.0.0~4` for some reason, I checked the source code for the changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,7 +91,7 @@ genmesh = "0.6"
 ron = "0.4"
 
 [build-dependencies]
-vergen = "2.0"
+vergen = "3.0"
 
 [[example]]
 name = "hello_world"
--- a/build.rs
+++ b/build.rs
@@ -1,14 +1,8 @@
-use vergen::{ConstantsFlags, Vergen};
+use vergen::{self, ConstantsFlags};
 
 fn main() {
-    let vergen = Vergen::new(ConstantsFlags::all()).unwrap_or_else(|e| {
-        panic!(
-            "Vergen crate failed to generate version information! {:?}",
-            e
-        );
-    });
+    vergen::generate_cargo_keys(ConstantsFlags::all())
+        .unwrap_or_else(|e| panic!("Vergen crate failed to generate version information! {}", e));
 
-    for (k, v) in vergen.build_info() {
-        println!("cargo:rustc-env={}={}", k.name(), v);
-    }
+    println!("cargo:rerun-if-changed=build.rs");
 }


docs/CHANGELOG.md
```diff
@@ -13,10 +13,14 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 
 ### Changed
 
+* Bumped version of `vergen` to `3.0`. ([#1242])
+
```
torkleyy: This is an internal change without any impact, not sure if it's worth listing here.

Cargo.toml
```diff
@@ -91,7 +91,8 @@ genmesh = "0.6"
 ron = "0.4"
 
 [build-dependencies]
-vergen = "2.0"
+failure = "0"
```
torkleyy: 0?

azriel91: ya, so `vergen` uses `failure` under the hood with that "version". It doesn't re-export the crate, but uses the `Fallible<T>` type as a return type (shorthand for `Result<T, failure::Error>`).

I could PR to `vergen` first to re-export `failure`, or use an `.expect("..")` in `build.rs` if you prefer

docs/CHANGELOG.md
azriel91: true that, I did want to revisit that changelog discussion of what should be in/out, communicating breakages better etc.

Cargo.toml
Xaeroxe: I for one would just prefer an expect.

torkleyy: Yeah, that seems to be common practice in build scripts.

azriel91: Yay, we need https://github.com/rust-lang-nursery/failure/pull/279 for a better error message

Xaeroxe:
@azriel91 why is this blocked?

azriel91:
There ~~is~~ was a chain of things, in reverse order:

1. To get a nice error message, I'd like to include the `vergen` error: `format!("{}", e)` has to work.
2. `failure` couldn't build due to ambiguity in choosing a formatting implementation in `backtrace`. [`failure` PR](https://github.com/rust-lang-nursery/failure/pull/279) fixes the ambiguity.
3. That PR is pending a release of [`backtrace`](https://github.com/alexcrichton/backtrace-rs) for Rust 1.25 compatibility.

Hm. However since `backtrace` had a release (0.3.12) that no longer needs disambiguation, then this shouldn't be blocked anymore. Shall carry on!

docs/CHANGELOG.md
azriel91: Took it out from the changelog

Xaeroxe:
bors try

azriel91:
Ah I locked the `backtrace` crate earlier, taking that version restriction away.

Xaeroxe:
bors r= @torkleyy, @Xaeroxe

PR #1243, opened 2018-12-11
Removed unnecessary code.

TimonPost:
Removed some unnecessary code.

### DIFF
--- a/amethyst_network/src/network_socket.rs
+++ b/amethyst_network/src/network_socket.rs
@@ -187,12 +187,10 @@ where
         }
 
         for raw_event in self.rx.try_iter() {
-            let mut matched = false;
             // Get the NetConnection from the source
             for net_connection in (&mut net_connections).join() {
                 // We found the origin
                 if net_connection.target == raw_event.source {
-                    matched = true;
                     // Get the event
                     let net_event = deserialize_event::<E>(raw_event.data.as_slice());
                     match net_event {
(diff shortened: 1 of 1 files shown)


amethyst_network/src/network_socket.rs
```diff
@@ -204,9 +202,8 @@ where
                             e, raw_event.source
                         ),
                     }
-                }
-                if !matched {
-                    println!("Received packet from unknown source");
+                }else {
```
azriel91: missing a blank space

TimonPost: yea I noticed, will be fixing it soon.

amethyst_network/src/network_socket.rs
```diff
@@ -172,27 +172,25 @@ where
...
-                    .expect("Unreachable: Channel will be alive until a stop event is sent");
-            } else if net_connection.state == ConnectionState::Disconnected {
+                {
+                    self.tx
+                        .send(InternalSocketEvent::SendEvents {
+                            target,
+                            events: net_connection.send_buffer_early_read().cloned().collect(),
+                        })
+                        .expect("Unreachable: Channel will be alive until a stop event is sent");
+                } else if net_connection.state == ConnectionState::Disconnected {
```
udoprog: Formatting mishap?

TimonPost:
Ah that's the reason, strange to have meaningless code and expect others to know the filter is supposed to be there. We should probably move this into an issue.

Xaeroxe:
I disagree regarding the issue.  By the time we implement filtering it might not take this form, and leaving this doesn't really save anyone that much work.

TimonPost:
I would just say merge it and by the time we need to implement filters we add it back. It are just a few lines of code.

Xaeroxe:
@jojolepro requesting your feedback on our feedback :P

PR #1248, opened 2018-12-12
Add example Ron configs to docs

Sogomn:


amethyst_input/src/bindings.rs
```diff
@@ -10,6 +10,28 @@ use super::{Axis, Button};
...
+///             neg: Key(Down)
+///         ),
+///         "leftright": Emulated(
+///             pos: Key(Right),
+///             neg: Key(Left)
+///         )
+///     },
+///     actions: {
+///         "fire": [ [Mouse(Left)], [Key(X)] ], // Up to four bindings for one action
+///         "reload": [ [Key(LControl), Key(R)] ] // Combinations of two bindings possible
```
Xaeroxe: Please remove "two" and "four" here.  We optimize for those values as they're very common, but they're not hard limits.  You can have more.

Xaeroxe:
bors r= @Xaeroxe , @LucioFranco , @jojolepro @udoprog , @azriel91

Issue #951, opened 2018-09-20
[Suggestions] Doc Improvements


Acadaniel:
A PR to gather some improvements that can be made to the docs, as I didn't find a fitting issue (neither open nor closed).

For the [`Loader.load...`](https://www.amethyst.rs/doc/develop/doc/amethyst_assets/struct.Loader.html#method.load) functions, it may be worthwhile to add, that loading failure is only noticeable if a ProgressCounter is used.

Telzhaak:
- [ ] Explanation for `Transparency` on sprites, which is currently only a sticky on discord
- [ ] An example for `ProgressCounter` usage in a State for asset loading
- [ ] A broken use-statement in [pong tut 02](https://www.amethyst.rs/book/master/pong-tutorial/pong-tutorial-02.html) under the `Drawing` section where the texture coordinates are defined, the first line.

azriel91:
Original comment is covered by #869, fixed in #892.

torkleyy:
> that loading failure is only noticeable if a ProgressCounter is used.

It should be printed to the console.

jojolepro:
@torkleyy I think it does

torkleyy:
I know, that's what I wanted to point out :sweat_smile:

Telzhaak:
*Bump*

An architectural graph of the `amethyst crates` and a list of their `features` would be really nice for `version = 0.10.0`

Is it possible to do something like that using `cargo magic`?

jojolepro:
Which is why we need people to finish the book. Also be sure you are reading the "master" book and not the 0.8 one.

jakelacey2012:
Appreciate the help @jojolepro and I'll try and PR where I can to help the project looks interesting. Just to clarify though the released create is `0.8` but the stable documentation is on the master branch? How come this hasn't been deployed if its the recommended one?

torkleyy:
Well, the examples on the `master` branch are only compatible with the code on master, which is not the released one. We used to refer people to the respective release branch, but we don't have release branches anymore..

jakelacey2012:
So I should expect some compilation errors when following the pong tutorial on the master branch whilst on version `0.8`. Sorry if this seems like a silly question, I'm just getting a little confused where I should start.

torkleyy:
> So I should expect some compilation errors when following the pong tutorial on the master branch whilst on version 0.8.

Yes, definitely.

torkleyy:
[...]

PR #1251, opened 2018-12-14
Added "How To" guides for assets.

azriel91:
Issue #951

---
This change is [Reviewable](https://reviewable.io/reviews/amethyst/amethyst/1251)


azriel91:
r? @amethyst/documentation-team

Btw, the "Prefabs in Amethyst" page was the existing "Prefab" page which I just moved down a level, because I wanted the concept explained first.

book/src/assets/how_to_define_custom_assets.md
```diff
@@ -0,0 +1,279 @@
...
+    # extern crate serde_derive;
+    #
+    use amethyst::{
+        assets::Handle,
+        ecs::VecStorage,
+    };
+
+    /// Custom asset representing an energy blast.
+    #[derive(Clone, Debug, Default, PartialEq, Eq)]
+    pub struct EnergyBlast {
```
khionu: I think this is a bit of a poor example type. If using an example that's just multimedia is too restrictive, then I would suggest at least supplementing this example with some.

book/src/assets/how_to_use_assets.md
```diff
@@ -0,0 +1,221 @@
...
+    #           .clear_target([0.; 4], 0.)
+    #           .with_pass(DrawUi::new()),
+    #   );
+        // ..
+
+        let game_data = GameDataBuilder::default()
+            .with_bundle(UiBundle::<String, String>::new())?
+            .with_bundle(
+                RenderBundle::new(pipeline, Some(display_config))
+                    .with_sprite_sheet_processor()
```
khionu: Given Sprites are covered separately, I think we should use a different type here.

book/src/assets/how_to_define_custom_assets.md
azriel91: Just to make sure I understand correctly:

* What do you mean by multimedia?
* Do you mean it's simpler with multimedia as an asset?

    Thought this was simpler &ndash; struct with 2 fields &ndash; unless you meant "this is too simple", and I should use an image / sound

book/src/assets/how_to_use_assets.md
azriel91: I could remove this one specifically, since `UiBundle` already demonstrates a bundle with a `Processor` &ndash; was thinking this is a good example for "sometimes you need to configure the bundle". Would you prefer this be removed?

book/src/assets/how_to_define_custom_assets.md
khionu: I meant that, as is, I think it's a little too simple. The restrictive remark was in regards to the potential concern that only using visuals/sounds in an Asset would be misleading as to the scope of an Asset. As is, however, I feel it could confuse people into thinking that that isn't what an Asset is for.

book/src/assets/how_to_use_assets.md
khionu: I don't think we need to worry about representing things that are outside the scope of the given section of the book, or overlap with other sections. Bundles is more of a bigger picture detail, not something for the scope of this section.

book/src/assets/how_to_define_custom_assets.md
azriel91: > as is, I think it's a little too simple

So, from past work experience, the simple example is actually valuable:

* When it was made simple, people said "this is too simple, we want something closer to real world usage".
* When we gave them an example that was slightly more difficult, without the simple case, we hear back "there's too much of a conceptual leap, show us something easier first".

We couldn't "win", though the true win is actually the simple case &ndash; more people understood and learned &ndash; so we settle on that. Then if necessary we also added another more difficult example.

Also, if the opening page is read, then I think people can infer that "oh this `EnergyBlast` can be replaced with my audio / image type". Is that assumption alright? See second half of this comment.

---

> only using visuals/sounds in an Asset would be misleading as to the scope of an Asset

I believe the explanation on the [`assets`](https://github.com/amethyst/amethyst/blob/e5eed5b223af2ba179ac7523102c3d0bcff18686/book/src/assets.md) page should be enough for people to realize "it can be any *configuration* that your application needs". If the opening sentence isn't clear enough, I can include an example of [a serialized asset](https://github.com/amethyst/amethyst/pull/1251/files#diff-1909f34eb356799930dc82795b0d59c2R82), but I prefer people to not "use code to explain code", and rely on their understanding of "textures, sounds, and game level configuration".

book/src/assets/how_to_use_assets.md
azriel91: Re:original question, do you think the `UiBundle` is enough, or what else would you suggest?

---

Bundles should be mentioned here though &ndash; imagine a reader who wants to use an asset:

1. I need to tell Amethyst where my assets live &ndash; assets directory. Easy enough.
2. Oh, the book says I need to tell Amethyst to process them, otherwise they are never loaded.
3. ...

In step 2, we could simply tell them "add the `Processor<A>`". However, Amethyst also provides these systems in bundles &ndash; it's a convention / common practice. If we do not also write this here, they wouldn't know to look (unless we link it, but we don't have somewhere to link it to yet), so I think it's valuable to keep.

book/src/assets/how_to_define_custom_assets.md
khionu: I'm not keen on relying on people making conclusion jumps. I can get your point on not "using code to explain code", so maybe a little extra wording to make the "it can be anything" point explicit?

book/src/assets/how_to_use_assets.md
khionu: I think, then, this falls prey to your "let's not use code to explain code" point.

book/src/assets/how_to_define_custom_assets.md
azriel91: yeaps shall have a think!
also, maybe i'm just paranoid of the `.ron` file being "code" &ndash; trying to avoid the "let's explain `git` using terms like 'commits' and 'branches'" scenario.

azriel91: > a little extra wording to make the "it can be anything" point explicit

[here we go](https://github.com/amethyst/amethyst/compare/e5eed5b223af2ba179ac7523102c3d0bcff18686..2c58febb2ebca31aaff230fa94b2a500e475da58#diff-df2bed129120ea43546d75d912837f44)

book/src/assets.md
```diff
@@ -0,0 +1,13 @@
+# Assets
+
+**Assets** are configuration that are loaded by a game when it is run. These may be textures, sounds, game level scripts, and so on. In fact, any non-player-specific information loaded by the executable can be called an asset. These are usually stored as files, and distributed alongside the game.
```
distransient: Instead of "Assets are configuration that are..." maybe, "Assets are data that's loaded by a game when it is run."

distransient: "In fact, any non-player-specific information loaded by the executable can be called an asset." it is unclear here what "non-player-specific information" is, my first thought for this sentence would maybe be "...any data that is loaded at runtime may be considered an asset."

book/src/assets/how_to_use_assets.md
```diff
@@ -0,0 +1,221 @@
...
+    #   game.run();
+    #   Ok(())
+    }
+    ```
+
+3. Use the [`Loader`][doc_loader] resource to load the asset.
+
+    ```rust,no_run,noplaypen
+    # extern crate amethyst;
+    use amethyst::{
```
distransient: To be consistent with the rest of the examples here, this should probably be commented out while imo `LoadingState` would be helpful to show to the reader, although at what point that should be done I'm not sure, since the type is introduced in Step 1.

book/src/assets/how_to_use_assets.md
```diff
@@ -0,0 +1,221 @@
...
+* amethyst_audio
+* amethyst_gltf
+* amethyst_locale
+* amethyst_ui
+
+## Steps
+
+1. Instantiate the Amethyst application with the assets directory.
+
+    ```rust,no_run,noplaypen
```
distransient: Is there a reason to no_run these examples? They'd make for helpful doctests.

book/src/assets/how_to_define_custom_assets.md
```diff
@@ -0,0 +1,279 @@
+# How to Define Custom Assets
+
+This guide explains how to define a new asset type to be used in an Amethyst application. If you are defining a new asset type that may be useful to others, please send us a PR!
```
distransient: Let's add a link here to either the PR submission page or Contribution page!

book/src/prefabs.md
```diff
@@ -0,0 +1,50 @@
...
+
+* Position
+* Velocity
+* Texture
+* Health points
+* Attack damage
+
+It is certainly possible to define the values for each monster in code. However, if the components are initialized in code, then the executable needs to be recompiled whenever the component values are changed. Waiting a number of minutes to every time a small change is made is both inefficient and frustrating.
+
+In data-oriented design (DoD), instantiating monsters is logic and is part of the executable, but the values to use for the components is data. The executable can instantiate any monster using the data for the monsters read from a *prefab* file like the following:
```
distransient: Unless the abbreviation "DoD" here is used subsequently in the text I believe it's unnecessary.

book/src/assets.md
khionu: +1

khionu: +1

book/src/assets/how_to_define_custom_assets.md
khionu: I agree with Kel's feedback on that.

book/src/assets.md
azriel91: I can roll with that; was trying to disambiguate "information that is common to all users" from "information that changes per user".

book/src/assets/how_to_use_assets.md
azriel91: Oh yes, shall comment the imports and uncomment `LoadingState` here, since it's the focus of this step

book/src/assets/how_to_use_assets.md
azriel91: 1. could be uncommented, but doesn't really do much.
2. uses `RenderBundle` which would fail on headless CI
3. would attempt to load the non-existent asset `"player.png"`, which wouldn't hard error, but may be noisy.
4. would wait for a progress tracker that never ends

I can uncomment 1. and 3., but i think the value isn't worth the cost

distransient: It's fine for this PR I think but I'd really like if we looked into making all the examples in the book runnable for doctests with mocks for things like RenderBundle.

PR #1253, opened 2018-12-14
Remove SmallVec from Bindings APIs

Xaeroxe:
We should never be exposing a `SmallVec` in public APIs.  This fixes that.

### DIFF
--- a/amethyst_renderer/src/resources.rs
+++ b/amethyst_renderer/src/resources.rs
@@ -33,7 +33,7 @@ impl<'a> PrefabData<'a> for AmbientColor {
     }
 }
 
-/// This specs resource with id 0 permits sending commands to the
+/// This specs resource permits sending commands to the
 /// renderer internal window.
 #[derive(Default)]
 pub struct WindowMessages {
--- a/amethyst_utils/src/fps_counter.rs
+++ b/amethyst_utils/src/fps_counter.rs
@@ -10,7 +10,7 @@ use crate::circular_buffer::CircularBuffer;
 
 /// The FPSCounter resource needed by the FPSCounterSystem.
 ///
-/// Add it to your resources with id 0 to be able to use the FPSCounterSystem.
+/// Add it to your resources to be able to use the FPSCounterSystem.
 ///
 /// ## Usage:
 /// Get the FPSCounter resource from the world then call either `frame_fps` or `sampled_fps` to


torkleyy:
There it is: http://troubles.md/posts/improving-smallvec/

Xaeroxe:
Counterpoint: https://github.com/servo/rust-smallvec/issues/3#issuecomment-99254932

happenslol:
I think it's still worth to use `SmallVec` in places where locality is important - i.e. in components or similar places that are constantly iterated over. I do agree though that in most places, it's probably not worth using it.

Xaeroxe:
Given the conflicting info available on this I decided to bench it myself.  Here's my bench, feel free to play with it

```rust
use std::time::Instant;

use smallvec::{SmallVec, smallvec};

fn main() {
    let iter_count = 1;
    let vec: Vec<Vec<Vec<i32>>> = vec![vec![vec![1, 2]; 4]; 20];
    let smallvec: Vec<SmallVec<[SmallVec<[i32; 2]>; 4]>> =  vec![smallvec![smallvec![1, 2]; 4]; 20];
    let start = Instant::now();
    let result = (0..iter_count).all(|_| vec.iter().all(|v| v.iter().all(|v| v.iter().all(|&i| i > 0 && i < 3))));
    let duration = start.elapsed();
    println!("Result: {}, Time to compare vec: {:?}", result, duration);
    let start = Instant::now();
    let result = (0..iter_count).all(|_| smallvec.iter().all(|v| v.iter().all(|v| v.iter().all(|&i| i > 0 && i < 3))));
    let duration = start.elapsed();
    println!("Result: {}, Time to compare smallvec: {:?}", result, duration);
}
```

What I determined from this is that smallvec is a bit slower for extremely large collections where we don't benefit from the cache locality, however smallvec is about 33% faster when I tune the numbers to what I expect them to be for our use case.

Xaeroxe:
Here's some example output:

```
Result: true, Time to compare vec: 3.316µs
Result: true, Time to compare smallvec: 2.296µs
```

Xaeroxe:
Oh and another thing I learned from the bench, more iterations == more in favor of vec.

I take it this is because iterating over a smallvec comes with a bit more overhead than iterating over a vec.  Retrieving the vec simply takes a bit longer.

Xaeroxe:
Since this accounts for 0.00625% of our frame time though we should probably dedicate our brilliant minds elsewhere though 😉

PR #1259, opened 2018-12-17
Removed concourse CI/CD

fhaynes:
The server running concourse is going away, so we need to remove this.


LucioFranco:
Why is gitlab not running and concourse is 🤔

Xaeroxe:
Not merged yet.  I don't think it takes effect until master is updated.

LucioFranco:
As far as I know if gitlab accepts the webhooks it shouldn't pull from master unless its setup in the gitlab settings to only pull from master. This will cause issues anyways as the only way then to verify changes to the CI script will have to be done on master, which isn't a good idea.

fhaynes:
GitLab mirrors everything; master, all branches

fhaynes:
Ah, it is hung because there is no more .concourse.

PR #1263, opened 2018-12-19
Added amethyst network metrics abstraction.

TimonPost:
What is the metric abstraction about?

This will abstract away different ways of outputting the metrics.

You could take for example CSV, text, InfluxDb, excel etc.
The idea is that amethyst provides some abstractions and the user is able to implement his own.

how it will look for people enabling the metrics: 

```
.with_bundle(
  NetworkBundle::<()>::new( "127.0.0.1:3456".parse().unwrap(), Vec::new())
       .with_metric(From::from(ConsoleMetrics::new(Duration::from_millis(500)))))
       .with_metric(From::from(CsvMetric::new(File::open("test.csv"), Duration::from_millis(500)))));
```

If the user passed a `MetricsObserver` the system will monitor:

1. Incoming packets
2. Outgoing packets
3. Throughput 

More can be implemented later.

Currently, only `ConsoleMetrics` is implemented `CsvMetrics` and `InfluxDbMetrics` will be implemented later and I will create an issue for that if this gets merged.

The networking team has a userstorie for this abstraction. We have to figure out how we are going to link those tho GitHub.


amethyst_network/src/metrics/console_metrics.rs
```diff
@@ -0,0 +1,36 @@
...
+    last_send: Instant,
+    // interval rate
+    update_rate: Duration,
+}
+
+impl ConsoleMetrics {
+    /// Create a new instance of `ConsoleMetrics` by passing in a interval at which the metrics will be written to screen.
+    pub fn new(update_rate: Duration) -> ConsoleMetrics {
+        ConsoleMetrics {
+            last_send: Instant::now(),
```
LucioFranco: This should probably be an option to represent this properly since, technically you haven't sent anything so how can the last_send be now?  😄

amethyst_network/src/metrics/console_metrics.rs
```diff
@@ -0,0 +1,36 @@
...
+            update_rate,
+        }
+    }
+}
+
+impl MetricObserver<NetworkMetricObject> for ConsoleMetrics {
+    fn update(&mut self, metrics_object: &NetworkMetricObject) {
+        if self.last_send.elapsed() >= self.update_rate {
```
LucioFranco: wow how have I never found this `elapsed` function 🤦‍♂️

amethyst_network/src/metrics/console_metrics.rs
```diff
@@ -0,0 +1,36 @@
...
+            self.write(metrics_object);
+        }
+        self.last_send = Instant::now();
+    }
+
+    fn write(&self, metrics_object: &NetworkMetricObject) {
+        println!("{:?}", metrics_object);
```
LucioFranco: I would switch this to a display and then have the impl be

```rust
impl MetricObserver<M> for ConsoleMetrics
where M: NetworkMetricObject + fmt::Display
```

amethyst_network/src/metrics/csv_metrics.rs
```diff
@@ -0,0 +1,39 @@
...
+
+/// This will handle changing metrics and write them to a CSV file.
+/// CSV write interval could be managed by passing in a `Duration` at creation time.
+pub struct CsvMetrics {
+    // the file handle to the CSV file
+    csv_file: File,
+    // this will be used for the writing interval to the console.
+    last_send: Instant,
+    // interval rate
+    update_rate: Duration,
```
LucioFranco: so it seems that this updated logic is quite repetitive, maybe we can move it to a trait?

amethyst_network/src/metrics/csv_metrics.rs
```diff
@@ -0,0 +1,39 @@
...
+            last_send: Instant::now(),
+            update_rate,
+        }
+    }
+}
+
+impl MetricObserver<NetworkMetricObject> for CsvMetrics {
+    fn update(&mut self, metrics_object: &NetworkMetricObject) {
+        if self.last_send.elapsed() >= self.update_rate {
+            self.write(metrics_object);
```
LucioFranco: will this block and how do we communicate that to the user of these metrics?

amethyst_network/src/metrics/csv_metrics.rs
```diff
@@ -0,0 +1,39 @@
...
+        }
+    }
+
+    fn write(&self, metrics_object: &NetworkMetricObject) {
+        // todo:
+        // 1) serialize object to csv
```
LucioFranco: this will also block so we need to think about how that will be handled, also is there a reason you are accepting a trait object?

amethyst_network/src/metrics/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+pub use self::console_metrics::ConsoleMetrics;
+pub use self::csv_metrics::CsvMetrics;
+pub use self::influxdb_metrics::InfluxDBMetrics;
+pub use self::network_metrics::NetworkMetrics;
+
+/// Metrics observer which could be implemented to observe metrics changes.
+/// This trait could be implemented for handling metrics a certain way.
+pub trait MetricObserver<T: Sync + Send> {
+    /// This method will be called to when a change in some metric has occurred.
+    fn update(&mut self, metrics_object: &T);
```
LucioFranco: This could take a `delta` of the elapsed time since the last update?

amethyst_network/src/metrics/mod.rs
```diff
@@ -0,0 +1,59 @@
+//! Module for managing network metrics.
```
LucioFranco: I would like to start providing good module docs. I think even before we write a bunch of code there should be a good paragraph here explaining and an example. Not needed right now but I would like to start doing that.

amethyst_network/src/metrics/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+pub trait MetricObserver<T: Sync + Send> {
+    /// This method will be called to when a change in some metric has occurred.
+    fn update(&mut self, metrics_object: &T);
+    /// This method could be used to write the object to the implemented metric output.
+    fn write(&self, metrics_object: &T);
+}
+
+/// Object containing network metrics.
+#[derive(Debug)]
+pub struct NetworkMetricObject {
```
LucioFranco: `Object` is triggering me! Can we just name this `NetworkMetric`?

amethyst_network/src/metrics/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+    /// This will be reset to 0 after one second elapsed
+    pub throughput_counter: u32,
+    /// The total incoming packets
+    pub total_incoming_packets: u32,
+    /// The total outgoing packets
+    pub total_outgoing_packets: u32,
+    /// The counter which will be used to monitor incoming packets during 1 second
+    pub throughput_monitor: Instant,
+    /// The elapse time for monitoring throughput
+    throughput_elapse: Duration,
```
LucioFranco: `throughput_elasped`? also should this be public?

amethyst_network/src/metrics/network_metrics.rs
```diff
@@ -0,0 +1,51 @@
...
+    /// Register an outgoing packet and notify all metrics observers.
+    pub fn add_outgoing_packets(&mut self, count: u16) {
+        self.metric.total_outgoing_packets += count as u32;
+        self.notify_observers();
+    }
+
+    /// Notify all metrics observers.
+    fn notify_observers(&mut self) {
+        for observer in self.observers.iter_mut() {
+            observer.update(&mut self.metric);
```
LucioFranco: I have a feeling this might be quite heavy perf wise 😞

examples/net_server/main.rs
```diff
@@ -5,7 +5,10 @@ extern crate log;
 use amethyst::{
     core::frame_limiter::FrameRateLimitStrategy,
     ecs::{Join, System, WriteStorage},
-    network::*,
+    network::{
+        metrics::{ConsoleMetrics, MetricObserver, NetworkMetricObject},
+        *,
```
LucioFranco: can we remove this star import?

LucioFranco:
Ideally I would like the API to look like

```rust
.with_metric(ConsoleSink::new())
.with_metric(InfluxDB::new())
```

I want to think about how we can drop the trait objects. Don't know if it will be possible but at least drop some of the boxes.

amethyst_network/src/metrics/console_metrics.rs
TimonPost: I was thinking about that. However, if I use options I need to unwrap the option which I don't like since last_seen will always be used. So I figured just having the time instance there to initialize it would be fine.

amethyst_network/src/metrics/csv_metrics.rs
TimonPost: jup was thinking about that. I'll take it a shot.

amethyst_network/src/metrics/console_metrics.rs
LucioFranco: You can do 
```rust
if let Some(last_send) = self.last_send {
	if last_send.elapsed() >= self.update_rate { ... }
} else {
	// send here
}

amethyst_network/src/metrics/csv_metrics.rs
TimonPost: Actually, maybe we want some thread where you can schedule you're a metrics package. Indeed waiting could be problematic here same goes for network calls. How are you making a separate issue for this?

amethyst_network/src/metrics/mod.rs
TimonPost: Interesting I'll take a look

amethyst_network/src/metrics/mod.rs
TimonPost: Sure, I did this since I was not sure about somethings yet.

amethyst_network/src/metrics/console_metrics.rs
TimonPost: Yea, I know how that work. But it is more about the logic needed to unwrap the option. Since last seen is not really optional but not known at initializing time.

TimonPost:
Oke, working on moving stuff into its own thread now so that a system would not be blocked if an `update()` call is taking longer than expected.

fhaynes:
I'll review this in more detail later, but one thing is that sending metrics should *never* block anything. It is usually done in its own thread for this reason, amongst others.

TimonPost:
Jup and therefore I am creating a thread now as noted above.

jojolepro:
If you put the data into a struct that derives Serialize, you can use all the different serde-compatible crates. 
Example: export to ron, json, binary, csv + probably a lot more I didn't find yet.

amethyst_network/src/metrics/csv_metrics.rs
LucioFranco: > How are you making a separate issue for this?

What do you mean?

LucioFranco: we should probably provide metrics gathering utlities in a separate crate.

TimonPost:
Closing this now, @LucioFranco, @fhaynes, @TimonPost decided to take this PR a level higher and make metrics available for whole amethyst

PR #1267, opened 2018-12-19
Correct pixel perfect texture coordinates using from_pixel_values

Frizi:
Texture coordinates were miscalculated, resulting in artifacts when rendering in situations different from perfect, that is when any texture filtering is present or framebuffer pixel grid was not exactly the same density as sprite's, which is common when hidpi factor is different than 1.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 ### Changed
 
 * Make `application_root_dir` return a `Result<Path>` instead of a `String` ([#1213])
+* Remove unnecessary texture coordinates offset in `Sprite::from_pixel_values` ([#1267])
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


azriel91:
> I'm curious how @azriel91 feels about it.

very happy 😄

docs/CHANGELOG.md
```diff
@@ -17,7 +17,7 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 ### Changed
 
 * Make `application_root_dir` return a `Result<Path>` instead of a `String` ([#1213])
-
+* Remove unnecessary texture coordinates offset in `Sprite::from_pixel_values` ([#1267])
```
azriel91: ah, you also need to add this to the list below:

```
[#1267]: https://github.com/amethyst/amethyst/pull/1267
```

also, could you add a newline after this line

Frizi: done

azriel91:
bors r=Xaeroxe,azriel91,jojolepro

PR #1271, opened 2018-12-20
Add backers to README

onbjerg:
I had a little trouble finding a way to donate, so I used the donation box in the announcement post.

I think having backers in the README will make the process of finding the Amethyst OpenCollective a bit easier.


khionu:
We're wanting to promote DonorBox over OpenCollective, as the latter isn't as straightforward when it comes to taxes. I'm not familiar with the specifics of this, @Moxinilian was the one who said this on the Discord.

LucioFranco:
@khionu sure, but this will need to be a larger discussion and we already will have to deal with open collective and taxes anyways. So I think this should still go through.

Issue #1193, opened 2018-11-28
3D Camera aspect ratio changes.


jojolepro:
# Task
Add a `System` that automatically adjust the camera aspect ratio depending on the current window size.
Should only apply on camera with a `AutoFov` component.

```rs
pub struct AutoFov {
    base_horizontal_fov: f32,
    fov_grow_ratio: f32, // needed?
    base_aspect_ratio, // The ratio at which the base fov is applied without any transformation (ie 16/9 or 4/3)
}
```

JameeKim:
Can you elaborate on what the system does with each field of the `AutoFov` component? This feature is exactly what I need, but I don't get how I should approach this problem.

jojolepro:
base_aspect_ratio is the screen dimension ratio (16/9, 4/3, etc) for which the base_horizontal_fov is meant to apply. When the aspect ratio grows (horizontal stretch), you also grow the horizontal fov by a factor of 

delta_fov = fov_grow_ratio / (current_aspect_ratio - base_aspect_ratio)
new_camera_fov = base_horizontal_fov + delta_fov

Also you'll want min_fov and max_fov fields on the AutoFov struct, because otherwise it would be possible to see 360 degrees by making the window very large but not high.

Does that answer all your questions?

JameeKim:
Thank you so much! I'll start working on this :)

JameeKim:
Maybe this issue can be closed?

torkleyy:
Yep, thanks!

PR #1281, opened 2018-12-24
add AutoFov for automatic resizing of camera

JameeKim:
issue #1193 
Currently, when the screen aspect ratio exceeds the point where the horizontal fov reaches its specified maximum, the rendered view stretches to fit the screen. I added a `TODO` label to handle that bc I couldn't figure out what to do with it. Any ideas? Or would it be just fine to leave it like that?

### DIFF
--- a/amethyst_utils/src/auto_fov.rs
+++ b/amethyst_utils/src/auto_fov.rs
@@ -1,0 +1,251 @@
+//! Utility to adjust the aspect ratio of cameras automatically
+
+use amethyst_assets::{PrefabData, PrefabError};
+use amethyst_core::specs::{
+    Component, Entity, HashMapStorage, Join, ReadExpect, ReadStorage, System, WriteStorage,
+};
+use amethyst_renderer::{Camera, ScreenDimensions};
+
+/// A component describing the behavior of the camera in accordance with the screen dimensions
+#[derive(Clone, Deserialize, PrefabData, Serialize)]
+#[prefab(Component)]
+#[serde(default)]
+pub struct AutoFov {
+    /// The horizontal FOV value at the aspect ratio in the field `base_aspect_ratio`
+    base_fovx: f32,
+
+    /// The factor determining how sensitive the FOV change should be
+    fovx_growth_rate: f32,
+
+    /// If the FOV grow rate specified in the field `fovx_growth_rate` should be applied as-is
+    fixed_growth_rate: bool,
+
+    /// The aspect ratio when the camera's horizontal FOV is identical to `base_fovx`
+    base_aspect_ratio: (usize, usize),
+
+    /// The minimum value the horizontal FOV can have
+    min_fovx: f32,
+
+    /// The maximum value the horizontal FOV can have
+    max_fovx: f32,
+}
+
+impl AutoFov {
+    /// Creates a new instance with the default values for all fields
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    /// The horizontal FOV value at the aspect ratio in the field `base_aspect_ratio`
+    ///
+    /// This value should be between the `min_fovx` and `max_fovx` values. Value in radians.
+    /// Defaults to `1.861684535`, which is the horizontal FOV value when the vertial FOV and aspect
+    /// ratio for a camera is set to `1.0471975512`(60 deg) and `16/9`, respectively.
+    pub fn base_fovx(&self) -> f32 {
+        self.base_fovx
+    }
+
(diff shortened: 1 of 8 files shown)


LucioFranco:
Hi! thanks for the PR! Sorry it has taken this long!

r? @amethyst/engine-devs @amethyst/render-devs

torkleyy:
> I'd say stretching is the correct behaviour here.
So the only thing I'd do is specify it in the API documentation.

JameeKim:
> So the only thing I'd do is specify it in the API documentation.

By this you mean the changelog, right?

torkleyy:
I meant the doc comment, which you adjusted. Thanks @JameeKim!

r? @jojolepro @udoprog

jojolepro:
Will review in the next few hours

amethyst_utils/src/auto_fov.rs
```diff
@@ -0,0 +1,248 @@
...
+    ///
+    /// This value should be between the `min_fovx` and `max_fovx` values. Value in raidans.
+    /// Defaults to `1.861684535`, which is the horizontal FOV value when the vertial FOV and aspect
+    /// ratio for a camera is set to `1.0471975512`(60 deg) and `16/9`, respectively.
+    pub base_fovx: f32,
+
+    /// The factor determining how sensitive the FOV change should be
+    ///
+    /// Defaults to `1.0`.
+    pub fovx_grow_rate: f32,
```
happenslol: I think "growth rate" is the more common term to use here

amethyst_utils/src/auto_fov.rs
```diff
@@ -0,0 +1,248 @@
...
+};
+use amethyst_renderer::{Camera, ScreenDimensions};
+
+/// A component describing the behavior of the camera in accordance with the screen dimensions
+#[derive(Clone, Deserialize, Serialize)]
+#[serde(default)]
+pub struct AutoFov {
+    /// The horizontal FOV value at the aspect ratio in the field `base_aspect_ratio`
+    ///
+    /// This value should be between the `min_fovx` and `max_fovx` values. Value in raidans.
```
happenslol: `raidans` => `radians`

amethyst_utils/src/auto_fov.rs
```diff
@@ -0,0 +1,248 @@
...
+
+impl Default for AutoFov {
+    fn default() -> Self {
+        AutoFov {
+            base_fovx: 1.861684535,
+            fovx_grow_rate: 1.0,
+            fixed_grow_rate: false,
+            base_aspect_ratio: (16, 9),
+            min_fovx: 0.1,
+            max_fovx: ::std::f32::consts::PI,
```
happenslol: The leading `::` are not needed in 2018 edition iirc

JameeKim:
```suggestion
            max_fovx: std::f32::consts::PI,
```

JameeKim:
@happenslol Thank you for the feedback!
As I was intending to provide the methods to validate the values of the fields, I guess it would be more appropriate to change the fields private and provide getter/setter methods. And your point about using `assert!` is also right, so I'll change it to use `error!` instead.
I'll start as soon as we also hear from other members :)

udoprog:
@happenslol 

> 1. The way that the AutoFov is "incrementally" constructed with the values you want for it. Returning the struct as-is with a new value added looks very much like a builder pattern - however, you're not using a builder, but offer the methods on the struct itself. [..]

Taking it by-value is actually kind of nice since it's a fairly good indicator that these methods should only be used during initialization. I would agree with using a separate builder for other languages though which doesn't have Rust's ownership rules where it's more ambiguous.

> 2. I'm not sure about the use of assert in the "builder" methods. I think there's 2 valid uses for asserts in game engine code: [..]

Assert is just a convenient way for panicking. As long as it's clearly documented, and it's a reasonable panic it's OK. I think this usage qualifies.

udoprog:
> As I was intending to provide the methods to validate the values of the fields, I guess it would be more appropriate to change the fields private and provide getter/setter methods.

:+1: This seems like a good idea since we want those invariants enforced.

khionu:
@happenslol have your issues been addressed?

happenslol:
Yes, thank you!

JameeKim:
done :D

khionu:
bors +r

khionu:
bors +r

khionu:
bors r+

Third time is a charm. I swear, I'm fine.

PR #1282, opened 2018-12-24
Add support for rendering entities with a simple RGBA value.

Xaeroxe:
I wanted to programatically control a sprite's alpha value and I realized there wasn't an easy way to do that.  I figured while I was at it, why not also allow all four RGBA channels to be controlled like this?

### DIFF
--- a/amethyst_renderer/src/pass/shaders/vertex/sprite.glsl
+++ b/amethyst_renderer/src/pass/shaders/vertex/sprite.glsl
@@ -15,7 +15,13 @@ in float depth;
 in vec2 u_offset;
 in vec2 v_offset;
 
-out vec2 tex_uv;
+in vec4 color;
+
+
+out VertexData {
+    vec2 tex_uv;
+    vec4 color;
+} vertex;
 
 const vec2 positions[6] = vec2[](
     // First triangle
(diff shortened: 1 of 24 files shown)


amethyst_renderer/src/pass/shaders/vertex/sprite.glsl
```diff
@@ -39,8 +45,8 @@ void main() {
     float tex_v = positions[gl_VertexID][1];
 
     vec2 uv = pos + tex_u * dir_x + tex_v * dir_y;
-    tex_uv = texture_coords(vec2(tex_u, tex_v), u_offset, v_offset);
-
+    vertex.tex_uv = texture_coords(vec2(tex_u, tex_v), u_offset, v_offset);
+    vertex.color = color;
     vec4 vertex = vec4(uv, depth, 1.0);
```
azriel91: does glsl support shadowing?
I thought it would clash

Xaeroxe: This isn't technically shadowing, as `vertex.color` is in a different namespace from `color`.

Xaeroxe: But to your other comment, I tested the output of all these shaders.

Xaeroxe:
@azriel91 changelog entry added, thanks for the reminder!

amethyst_renderer/src/pass/shaders/vertex/sprite.glsl
azriel91: ah cool

Xaeroxe:
I'm doubtful we're going to get a request for changes in the next 3 hours, so I'm just going to merge this.  Happy holidays everyone!

bors r+

PR #1287, opened 2018-12-25
Use Option::get_or_insert_with in UI example

Havvy:
[Option::get_or_insert_with](https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert_with)

There's no need to `.unwrap()` or manually wrap the `Some()` when there's a method specifically for this pattern of using `Option`. This also makes it easier to see the Amethyst specific stuff.

(Since I'm submitting this by editing from github's tools, I haven't done Cargo build/fmt, although I have something similar in my own local code that I have built)

### DIFF
--- a/examples/ui/main.rs
+++ b/examples/ui/main.rs
@@ -123,12 +123,10 @@ impl<'a> System<'a> for UiEventHandlerSystem {
     type SystemData = Write<'a, EventChannel<UiEvent>>;
 
     fn run(&mut self, mut events: Self::SystemData) {
-        if self.reader_id.is_none() {
-            self.reader_id = Some(events.register_reader());
-        }
+        let reader_id = self.reader_id.get_or_insert_with(|| events.register_reader());
 
         // Reader id was just initialized above if empty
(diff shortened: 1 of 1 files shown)


examples/ui/main.rs
```diff
@@ -123,12 +123,10 @@ impl<'a> System<'a> for UiEventHandlerSystem {
     type SystemData = Write<'a, EventChannel<UiEvent>>;
 
     fn run(&mut self, mut events: Self::SystemData) {
-        if self.reader_id.is_none() {
-            self.reader_id = Some(events.register_reader());
-        }
+        let reader_id = self.reader_id.get_or_insert_with(|| ui_events.register_reader());
```
Xaeroxe: pretty sure `ui_events` isn't correct here, you probably want `events` instead.

Xaeroxe:
Formatting test failed.  We'll need to run `cargo fmt --all` on this before we can merge it.

Xaeroxe:
Ok, let's try again

bors r+

PR #1289, opened 2018-12-26
Parallel Packet Processing

TimonPost:
Removed any dependencies to specs by abstracting away all socket logic to the `Server` type. 

The current network library does send and receiving in one thread. This could cause the program to not receive some packets at high server load when the thread is in the send loop. I moved the receiving logic and sending logic into separate threads so that we can receive and send in parallel.

_A diagram of the implemented code:_
untitled diagram

### DIFF
--- a/amethyst_network/src/bundle.rs
+++ b/amethyst_network/src/bundle.rs
@@ -7,21 +7,36 @@ use amethyst_core::{
     shred::DispatcherBuilder,
 };
 
-use crate::{filter::NetFilter, NetSocketSystem};
+use crate::{filter::NetFilter, server::ServerConfig};
+
+use super::NetSocketSystem;
 
 /// A convenience bundle to create the infrastructure needed to send and receive network messages.
 pub struct NetworkBundle<T> {
-    /// Local socket address
-    addr: SocketAddr,
+    /// the configuration used for the networking crate.
+    config: ServerConfig,
 
     /// The filters applied on received network events.
     filters: Vec<Box<dyn NetFilter<T>>>,
 }
 
 impl<T> NetworkBundle<T> {
-    /// Creates a new NetworkBundle that connects to the `addr`.
-    pub fn new(addr: SocketAddr, filters: Vec<Box<dyn NetFilter<T>>>) -> Self {
-        NetworkBundle { addr, filters }
+    /// Creates a new NetworkBundle.
+    ///
+    /// `receive_addr`: this is the address on which we will receive incoming packets.
+    /// `send_addr`: this is the address from which we will send outgoing packets.
+    pub fn new(
+        receive_addr: SocketAddr,
+        send_addr: SocketAddr,
+        filters: Vec<Box<dyn NetFilter<T>>>,
+    ) -> Self {
+        let config = ServerConfig {
+            udp_recv_addr: receive_addr,
+            udp_send_addr: send_addr,
+            max_throughput: 5000,
+        };
+
+        NetworkBundle { config, filters }
     }
 }
 
@@ -29,8 +44,9 @@ impl<'a, 'b, T> SystemBundle<'a, 'b> for NetworkBundle<T>
 where
     T: Send + Sync + PartialEq + Serialize + Clone + DeserializeOwned + 'static,
 {
+    /// Build the networking bundle by adding the networking system to the application.
     fn build(self, builder: &mut DispatcherBuilder<'_, '_>) -> Result<()> {
-        let socket_system = NetSocketSystem::<T>::new(self.addr, self.filters)
+        let socket_system = NetSocketSystem::<T>::new(self.config, self.filters)
             .chain_err(|| "Failed to open network system.")?;
 
         builder.add(socket_system, "net_socket", &[]);
--- a/amethyst_network/src/connection.rs
+++ b/amethyst_network/src/connection.rs
@@ -14,8 +14,10 @@ use crate::NetEvent;
 #[derive(Serialize)]
 #[serde(bound = "")]
 pub struct NetConnection<E: 'static> {
-    /// The remote socket address of this connection.
-    pub target: SocketAddr,
+    /// The target remote socket address who is listening for incoming packets.
+    pub target_receiver: SocketAddr,
+    /// The target remote socket address who is sending packets to us.
+    pub target_sender: SocketAddr,
     /// The state of the connection.
     pub state: ConnectionState,
     /// The buffer of events to be sent.
@@ -31,12 +33,13 @@ pub struct NetConnection<E: 'static> {
 
 impl<E: Send + Sync + 'static> NetConnection<E> {
     /// Construct a new NetConnection. `SocketAddr` is the address that will be connected to.
-    pub fn new(target: SocketAddr) -> Self {
+    pub fn new(target_receiver: SocketAddr, target_sender: SocketAddr) -> Self {
         let mut send_buffer = EventChannel::new();
         let send_reader = send_buffer.register_reader();
 
         NetConnection {
-            target,
+            target_receiver,
+            target_sender,
             state: ConnectionState::Connecting,
             send_buffer,
             receive_buffer: EventChannel::<NetEvent<E>>::new(),
@@ -58,7 +61,10 @@ impl<E: Send + Sync + 'static> NetConnection<E> {
 
 impl<E> PartialEq for NetConnection<E> {
     fn eq(&self, other: &Self) -> bool {
-        self.target == other.target && self.state == other.state
+        self.target_receiver == other.target_receiver
+            && self.state == other.state
+            && self.target_sender == other.target_sender
+            && self.state == other.state
     }
 }
 
--- a/amethyst_network/src/error.rs
+++ b/amethyst_network/src/error.rs
@@ -1,0 +1,74 @@
+//! Module containing error handling logic.
+
+use std::{
+    fmt::{self, Display, Formatter},
+    io,
+    sync::mpsc,
+};
+
+use crate::server::ServerSocketEvent;
+
+/// The `amethyst_network` result type.
+pub type Result<T> = std::result::Result<T, ErrorKind>;
+
+/// Wrapper for all errors who could occur in `amethyst_network`.
+#[derive(Debug)]
+pub enum ErrorKind {
+    /// Error that could occur on the UDP-socket
+    UdpError(laminar::error::NetworkError),
+    /// Error that could occur whit IO.
+    IoError(io::Error),
+    /// Error that could occur when serializing whit `bincode`
+    SerializeError(bincode::ErrorKind),
+    /// Error that could occur when sending an `ServerSocketEvent` to some channel.
+    ChannelSendError(mpsc::SendError<ServerSocketEvent>),
+    #[doc(hidden)]
+    __Nonexhaustive,
+}
+
+impl std::error::Error for ErrorKind {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match *self {
+            ErrorKind::IoError(ref e) => Some(e),
+            ErrorKind::ChannelSendError(ref e) => Some(e),
+            ErrorKind::SerializeError(ref e) => Some(e),
+            _ => None,
+        }
+    }
+}
+
+impl Display for ErrorKind {
+    fn fmt(&self, fmt: &mut Formatter<'_>) -> fmt::Result {
+        match *self {
+            ErrorKind::UdpError(_) => write!(fmt, "UDP-error occurred"),
+            ErrorKind::IoError(_) => write!(fmt, "IO-error occurred"),
+            ErrorKind::ChannelSendError(_) => write!(fmt, "Channel send error occurred"),
+            ErrorKind::SerializeError(_) => write!(fmt, "Channel send error occurred"),
+            _ => write!(fmt, "Some error has occurred"),
+        }
+    }
+}
+
+impl From<io::Error> for ErrorKind {
+    fn from(e: io::Error) -> ErrorKind {
+        ErrorKind::IoError(e)
+    }
+}
+
+impl From<mpsc::SendError<ServerSocketEvent>> for ErrorKind {
+    fn from(e: mpsc::SendError<ServerSocketEvent>) -> ErrorKind {
+        ErrorKind::ChannelSendError(e)
+    }
+}
+
+impl From<laminar::error::NetworkError> for ErrorKind {
+    fn from(e: laminar::error::NetworkError) -> ErrorKind {
+        ErrorKind::UdpError(e)
+    }
+}
+
+impl From<Box<bincode::ErrorKind>> for ErrorKind {
+    fn from(e: Box<bincode::ErrorKind>) -> ErrorKind {
+        ErrorKind::SerializeError(*e)
+    }
+}
--- a/amethyst_network/src/lib.rs
+++ b/amethyst_network/src/lib.rs
@@ -6,37 +6,45 @@ extern crate serde;
 
 mod bundle;
 mod connection;
+mod error;
 mod filter;
 mod net_event;
 mod network_socket;
+mod server;
 mod test;
 
 pub use crate::{
     bundle::NetworkBundle,
     connection::{ConnectionState, NetConnection, NetIdentity},
+    error::Result,
     filter::{FilterConnected, NetFilter},
     net_event::NetEvent,
     network_socket::NetSocketSystem,
+    server::{Host, ServerConfig, ServerSocketEvent},
 };
 
-use bincode::{deserialize, serialize, ErrorKind};
-use laminar::{net::UdpSocket, Packet};
+use bincode::{deserialize, serialize};
+use laminar::Packet;
 use log::error;
 use serde::{de::DeserializeOwned, Serialize};
 use std::net::SocketAddr;
+use std::sync::mpsc::SyncSender;
 
 /// Sends an event to the target NetConnection using the provided network Socket.
 /// The socket has to be bound.
-pub fn send_event<T>(event: &NetEvent<T>, addr: &SocketAddr, socket: &mut UdpSocket)
+pub fn send_event<T>(event: NetEvent<T>, addr: SocketAddr, sender: &SyncSender<ServerSocketEvent>)
 where
     T: Serialize,
 {
-    let ser = serialize(event);
+    let ser = serialize(&event);
     match ser {
         Ok(s) => {
             let slice = s.as_slice();
             // send an unreliable `Packet` from laminar which is basically just a bare UDP packet.
-            match socket.send(&Packet::unreliable(*addr, slice.to_vec())) {
+            match sender.send(ServerSocketEvent::Packet(Packet::unreliable(
+                addr,
+                slice.to_owned(),
+            ))) {
                 Ok(_qty) => {}
                 Err(e) => error!("Failed to send data to network socket: {}", e),
             }
@@ -45,10 +53,10 @@ where
     }
 }
 
-/// Attempts to deserialize an event from the raw byte data.
-fn deserialize_event<T>(data: &[u8]) -> Result<NetEvent<T>, Box<ErrorKind>>
+// Attempts to deserialize an event from the raw byte data.
+fn deserialize_event<T>(data: &[u8]) -> Result<NetEvent<T>>
 where
     T: DeserializeOwned,
 {
-    deserialize::<NetEvent<T>>(data)
+    Ok(deserialize::<NetEvent<T>>(data)?)
 }
--- a/amethyst_network/src/net_event.rs
+++ b/amethyst_network/src/net_event.rs
@@ -46,7 +46,7 @@ impl<T> NetEvent<T> {
     /// Tries to convert a NetEvent to a custom event type.
     pub fn custom(&self) -> Option<&T> {
         if let NetEvent::Custom(ref t) = self {
-            Some(t)
+            Some(&t)
         } else {
             None
         }
--- a/amethyst_network/src/network_socket.rs
+++ b/amethyst_network/src/network_socket.rs
@@ -2,18 +2,27 @@
 
 use std::{
     clone::Clone,
-    io::{Error, ErrorKind},
     net::SocketAddr,
-    sync::mpsc::{channel, Receiver, Sender},
+    sync::{
+        mpsc::{self, Receiver, Sender},
+        Arc, Mutex,
+    },
     thread,
 };
 
 use amethyst_core::specs::{Join, Resources, System, SystemData, WriteStorage};
-use laminar::{error, net::UdpSocket, DeliveryMethod, NetworkConfig, Packet};
+
+use laminar::Packet;
 use log::{error, warn};
 use serde::{de::DeserializeOwned, Serialize};
 
-use crate::{deserialize_event, send_event, ConnectionState, NetConnection, NetEvent, NetFilter};
+use super::{
+    deserialize_event,
+    error::Result,
+    send_event,
+    server::{Host, ReceiveHandler, SendHandler, ServerConfig, ServerSocketEvent},
+    ConnectionState, NetConnection, NetEvent, NetFilter,
+};
 
 enum InternalSocketEvent<E> {
     SendEvents {
@@ -21,34 +30,6 @@ enum InternalSocketEvent<E> {
         events: Vec<NetEvent<E>>,
     },
     Stop,
-}
-
-struct RawEvent {
-    pub byte_count: usize,
-    pub data: Vec<u8>,
-    pub source: SocketAddr,
-    pub delivery_method: DeliveryMethod,
-}
-
-impl From<Packet> for RawEvent {
-    fn from(packet: Packet) -> Self {
-        RawEvent {
-            byte_count: packet.payload().len(),
-            data: packet.payload().to_vec(),
-            source: packet.addr(),
-            delivery_method: packet.delivery_method(),
-        }
-    }
-}
-
-impl From<RawEvent> for Packet {
-    fn from(raw_event: RawEvent) -> Self {
-        Packet::new(
-            raw_event.source,
-            raw_event.data.into_boxed_slice(),
-            raw_event.delivery_method,
-        )
-    }
 }
 
 // If a client sends both a connect event and other events,
@@ -69,9 +50,11 @@ where
 {
     /// The list of filters applied on the events received.
     pub filters: Vec<Box<dyn NetFilter<E>>>,
-
-    tx: Sender<InternalSocketEvent<E>>,
-    rx: Receiver<RawEvent>,
+    // sender on which you can queue packets to send to some endpoint.
+    transport_sender: Sender<InternalSocketEvent<E>>,
+    // receiver from which you can read received packets.
+    transport_receiver: Receiver<Packet>,
+    config: ServerConfig,
 }
 
 impl<E> NetSocketSystem<E>
@@ -79,79 +62,83 @@ where
     E: Serialize + PartialEq + Send + 'static,
 {
     /// Creates a `NetSocketSystem` and binds the Socket on the ip and port added in parameters.
-    pub fn new(addr: SocketAddr, filters: Vec<Box<dyn NetFilter<E>>>) -> Result<Self, Error> {
-        if addr.port() < 1024 {
+    pub fn new(config: ServerConfig, filters: Vec<Box<dyn NetFilter<E>>>) -> Result<Self> {
+        if config.udp_recv_addr.port() < 1024 {
             // Just warning the user here, just in case they want to use the root port.
             warn!("Using a port below 1024, this will require root permission and should not be done.");
         }
 
-        let mut socket = UdpSocket::bind(addr, NetworkConfig::default())
-            .map_err(|x| Error::new(ErrorKind::Other, x.to_string()))?;
+        let server = Host::run(&config)?;
 
-        socket.set_nonblocking(true).map_err(|_| {
-            Error::new(
-                ErrorKind::Other,
-                "Unable to set `UdpSocket` to non-blocking mode",
-            )
-        })?;
+        let udp_send_handle = server.udp_send_handle();
+        let udp_receive_handle = server.udp_receive_handle();
 
-        // this -> thread
-        let (tx1, rx1) = channel();
-        // thread -> this
-        let (tx2, rx2) = channel();
+        let server_sender = NetSocketSystem::<E>::start_sending(udp_send_handle);
+        let server_receiver = NetSocketSystem::<E>::start_receiving(udp_receive_handle);
 
-        thread::spawn(move || {
-            //rx1,tx2
-            let send_queue = rx1;
-            let receive_queue = tx2;
-            let mut socket = socket;
+        Ok(NetSocketSystem {
+            filters,
+            transport_sender: server_sender,
+            transport_receiver: server_receiver,
+            config,
+        })
+    }
 
-            'outer: loop {
-                // send
-                for control_event in send_queue.try_iter() {
-                    match control_event {
-                        InternalSocketEvent::SendEvents { target, events } => {
-                            for ev in events {
-                                send_event(&ev, &target, &mut socket);
-                            }
+    /// Start a thread to send all queued packets.
+    fn start_sending(sender: Arc<SendHandler>) -> Sender<InternalSocketEvent<E>> {
+        let (tx, send_queue) = mpsc::channel();
+
+        thread::spawn(move || loop {
+            for control_event in send_queue.try_iter() {
+                match control_event {
+                    InternalSocketEvent::SendEvents { target, events } => {
+                        for ev in events {
+                            send_event(ev, target, &sender.get_sender());
                         }
-                        InternalSocketEvent::Stop => break 'outer,
                     }
-                }
-
-                // receive
-                loop {
-                    match socket.recv() {
-                        // Data received
-                        Ok(Some(packet)) => {
-                            if let Err(_) = receive_queue.send(RawEvent::from(packet)) {
-                                error!("`NetworkSocketSystem` was dropped");
-                                break 'outer;
-                            }
-                        }
-                        Err(e) => match e.kind() {
-                            error::NetworkErrorKind::IOError(io_error) => {
-                                if io_error.kind() == ErrorKind::WouldBlock {
-                                    break;
-                                } else {
-                                    error!("Could not receive datagram: {}", e);
-                                }
-                            }
-                            _ => {
-                                error!("Could not receive datagram: {:?}", e);
-                            }
-                        },
-                        Ok(None) => {}
+                    InternalSocketEvent::Stop => {
+                        break;
                     }
                 }
             }
         });
 
-        Ok(NetSocketSystem {
-            filters,
-            tx: tx1,
-            rx: rx2,
-        })
+        tx
+    }
+
+    /// Starts a thread which receives incoming packets and sends them onto the 'Receiver' channel.
+    fn start_receiving(receiver: Arc<Mutex<ReceiveHandler>>) -> Receiver<Packet> {
+        let (receive_queue, rx) = mpsc::channel();
+
+        thread::spawn(move || {
+            // take note that this lock will be there as long this thread lives.
+            // We only have one receiver, if one tries to have two receivers the program should panic.
+            // Why is there one receiver? This is because how a channel works. Once we read the messages it will be removed from the channel.
+            match receiver.try_lock() {
+                Ok(receiver) => loop {
+                    for event in receiver.iter() {
+                        match event {
+                            ServerSocketEvent::Packet(packet) => {
+                                if let Err(error) = receive_queue.send(packet.clone()) {
+                                    error!(
+                                        "`NetworkSocketSystem` was dropped. Reason: {:?}",
+                                        error
+                                    );
+                                    break;
+                                }
+                            }
+                            ServerSocketEvent::Error(error) => error!("{:?}", error),
+                            _ => error!("Event not supported"),
+                        }
+                    }
+                },
+                Err(_) => {
+                    panic!("Two packet receivers can't run at the same time");
+                }
+            }
+        });
+
+        rx
     }
 }
 
@@ -161,50 +148,56 @@ where
 {
     type SystemData = (WriteStorage<'a, NetConnection<E>>);
 
-    fn setup(&mut self, res: &mut Resources) {
-        Self::SystemData::setup(res);
-    }
-
     fn run(&mut self, mut net_connections: Self::SystemData) {
         for net_connection in (&mut net_connections).join() {
-            let target = net_connection.target;
+            let target = net_connection.target_receiver;
 
             if net_connection.state == ConnectionState::Connected
                 || net_connection.state == ConnectionState::Connecting
             {
-                self.tx
+                self.transport_sender
                     .send(InternalSocketEvent::SendEvents {
                         target,
                         events: net_connection.send_buffer_early_read().cloned().collect(),
                     })
                     .expect("Unreachable: Channel will be alive until a stop event is sent");
             } else if net_connection.state == ConnectionState::Disconnected {
-                self.tx
+                self.transport_sender
                     .send(InternalSocketEvent::Stop)
                     .expect("Already sent a stop event to the channel");
             }
         }
 
-        for raw_event in self.rx.try_iter() {
+        for (counter, raw_event) in self.transport_receiver.try_iter().enumerate() {
             // Get the NetConnection from the source
             for net_connection in (&mut net_connections).join() {
-                // We found the origin
-                if net_connection.target == raw_event.source {
+                if net_connection.target_sender == raw_event.addr() {
                     // Get the event
-                    let net_event = deserialize_event::<E>(raw_event.data.as_slice());
-                    match net_event {
+                    match deserialize_event::<E>(raw_event.payload()) {
                         Ok(ev) => {
                             net_connection.receive_buffer.single_write(ev);
                         }
                         Err(e) => error!(
                             "Failed to deserialize an incoming network event: {} From source: {:?}",
-                            e, raw_event.source
+                            e,
+                            raw_event.addr()
                         ),
                     }
                 } else {
                     warn!("Received packet from unknown source");
                 }
             }
+
+            // this will prevent our system to be stuck in the iterator.
+            // After 10000 packets we will continue and leave the other packets for the next run.
+            // eventually some congestion prevention should be done.
+            if counter >= self.config.max_throughput as usize {
+                break;
+            }
         }
+    }
+
+    fn setup(&mut self, res: &mut Resources) {
+        Self::SystemData::setup(res);
     }
 }
--- a/amethyst_network/src/server/config.rs
+++ b/amethyst_network/src/server/config.rs
@@ -1,0 +1,25 @@
+use std::net::SocketAddr;
+
+#[derive(Clone, Debug)]
+/// The configuration used for the networking system.
+pub struct ServerConfig {
+    /// Address at which the UDP server will listen for incoming packets.
+    pub udp_recv_addr: SocketAddr,
+    /// Address from which the UDP server will be sending packets.
+    pub udp_send_addr: SocketAddr,
+    /// Specifies what the maximal packets that could be handled by the server.
+    /// This value is meant for preventing some loops to read infinitely long when many packets are send and received.
+    /// This value is by default 5000.
+    pub max_throughput: u16,
+}
+
+impl Default for ServerConfig {
+    fn default() -> Self {
+        ServerConfig {
+            // by passing in :0 port the OS will give an available port.
+            udp_recv_addr: "0.0.0.0:0".parse().unwrap(),
+            udp_send_addr: "0.0.0.0:0".parse().unwrap(),
+            max_throughput: 5000,
+        }
+    }
+}
--- a/amethyst_network/src/server/host.rs
+++ b/amethyst_network/src/server/host.rs
@@ -1,0 +1,61 @@
+/// This module is about the communication with this host to some other endpoints.
+///
+/// 1. Sending Data
+/// 2. Receiving Data
+/// 3. Broadcasting
+use crate::error::Result;
+use crate::server::{
+    ReceiveHandler, SendHandler, ServerConfig, ServerSocketEvent, UdpReceiver, UdpSender,
+};
+use laminar::Packet;
+use std::sync::{Arc, Mutex};
+
+/// 'Host' abstracts TCP and UDP sockets away.
+pub struct Host {
+    // Handler to access the internals of the UDP receiving thread
+    udp_receiver: Arc<Mutex<ReceiveHandler>>,
+    // Handler to access the internals of the UDP sender thread
+    udp_sender: Arc<SendHandler>,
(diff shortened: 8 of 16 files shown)


amethyst_network/src/network_socket.rs
```diff
@@ -71,88 +45,95 @@ where
...
+                                    error!("`NetworkSocketSystem` was dropped");
                                     break;
-                                } else {
-                                    error!("Could not receive datagram: {}", e);
                                 }
                             }
-                            _ => {
-                                error!("Could not receive datagram: {:?}", e);
+                            ServerSocketEvent::Error(error) => {
+                                error!("[ERROR NETWORKING SYSTEM] {:?}", error)
```
jojolepro: error! already shows from which file it is called in the logs

amethyst_network/src/net_event.rs
```diff
@@ -45,10 +45,10 @@ pub enum NetEvent<T> {
...
-        if let NetEvent::Custom(ref t) = self {
-            Some(t)
+        if let &NetEvent::Custom(ref t) = self {
+            Some(&t)
         } else {
             None
         }
     }
-}
+}
```
jojolepro: Missing new line after this one

examples/net_client/main.rs
```diff
@@ -72,32 +75,3 @@ impl<'a> System<'a> for SpamSystem {
         }
     }
 }
-
-/// A simple system reading received events.
-/// Used to see events sent by the net_echo_server example.
-struct ReaderSystem {
```
jojolepro: We can't read received packets from `System`s now?

TimonPost: Wasn't used so I figured to remove it. The net_server does not send packets back.

TimonPost:
@jojolepro yeah, we had some discussions about it. We might move away this code to its own crate some time. For now it could be placed in this crate.

amethyst_network/src/server/server.rs
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+
+    /// This will start the server.
+    /// 1: Fire up a TCP-sender and TCP-receiver if enabled in config.
+    /// 2: Fire up a UDP-sender and UDP-receiver.
+    /// 3: Set up some `channels` to communicate with underlying threads.
+    ///
+    /// The method uses the config provided when creating a `Server` instance.
+    pub fn run(&mut self) {
```
LucioFranco: from what I noticed above we could probably just merge run with new? This seems like an OOP style which is not really needed.

amethyst_network/src/server/server.rs
```diff
@@ -0,0 +1,105 @@
...
+
+        // setup a UDP-receiver which will receive packets from any endpoint.
+        self.udp_receiver = Some(UdpReceiver::run(self.config.udp_recv_addr));
+
+        // setup the UDP-sender which will send packets to an certain endpoint.
+        self.udp_sender = Some(UdpSender::run(self.config.udp_send_addr));
+    }
+
+    /// Get the handle to the internals of the UDP-receiving threat.
+    pub fn get_udp_receive_handle(&self) -> &Option<ReceiveHandler> {
```
LucioFranco: I would drop the `get_` and just have it `udp_receive_handle` since the get part can be infered from the return type not being a `&mut`.

amethyst_network/src/server/server.rs
```diff
@@ -0,0 +1,105 @@
...
+        &self.udp_receiver
+    }
+
+    /// Get the handle to the internals of the UDP-sending thread.
+    pub fn get_udp_send_handle(&self) -> &Option<SendHandler> {
```
LucioFranco: same here.

amethyst_network/src/server/server.rs
```diff
@@ -0,0 +1,105 @@
...
+        &self.udp_sender
+    }
+
+    /// Schedule a TCP-packet for sending.
+    pub fn send_tcp(&mut self, payload: &[u8]) -> NetworkResult<()> {
+        unimplemented!()
```
LucioFranco: we should try and avoid putting these into master so I think if this PR isn't going to include this we should have an issue and remove this function. aka I don't want these to slip out into an actual release.

amethyst_network/src/server/server.rs
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /// Schedule a UDP-packet for sending.
+    pub fn send_udp(&mut self, packet: Packet) -> NetworkResult<()> {
+        match &self.udp_sender {
+            None => {
+                error!("Trying to send UDP while it is not enabled");
+            }
+            Some(ref send_handle) => {
+                send_handle.send(ServerSocketEvent::Packet(packet)).unwrap();
```
LucioFranco: Can we use `?` instead of unwrap?

amethyst_network/src/server/server.rs
```diff
@@ -0,0 +1,105 @@
...
+        };
+
+        Ok(())
+    }
+
+    /// Shutdown the server and destroy all related resources.
+    fn shutdown(&mut self) {
+        // 1. shutdown all connections udp/tcp
+        // 2. clean up resources by closing thread's and such.
+        unimplemented!()
```
LucioFranco: these two should also be removed and put into an issue.

amethyst_network/src/server/server.rs
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /// Broadcast a TCP-packet to all connected clients.
+    fn broad_cast_tcp(&mut self, payload: &[u8], addr: SocketAddr) -> NetworkResult<()> {
```
LucioFranco: I think broadcast is generally one word 😄

amethyst_network/src/server/server_socket_event.rs
```diff
@@ -0,0 +1,44 @@
...
+    Error(NetworkError),
+    /// events that can happen with a client
+    ClientEvent(ClientEvent),
+    /// event used for defaults
+    Empty,
+}
+
+impl Clone for ServerSocketEvent {
+    fn clone(&self) -> Self {
+        ServerSocketEvent::Empty
```
LucioFranco: This doesn't seem right, seems like a default impl but this is for clone?

amethyst_network/src/server/server_socket_event.rs
```diff
@@ -0,0 +1,44 @@
...
+/// Event that could occur with a client.
+pub enum ClientEvent {
+    /// represents a connecting client
+    Connected,
+    /// represents a disconnecting client
+    Disconnected,
+    /// represents a client who is timing out
+    Timedout,
+    /// represents an default value for this enum.
+    None,
```
LucioFranco: Can you explain in what situations you might get an event but its none? that seems redudant

amethyst_network/src/server/udp.rs
```diff
@@ -0,0 +1,125 @@
...
+        let mut udp_receiver = UdpReceiver::new(addr);
+
+        // channel used for communicating about received packets.
+        let (tx, rx) = mpsc::channel();
+
+        let thread_handle = thread::spawn(move || {
+            udp_receiver.start_receiving(tx);
+        });
+
+        ReceiveHandler::new(5000, rx, thread_handle)
```
LucioFranco: where does this 5000 come from?

amethyst_network/src/server/udp.rs
```diff
@@ -0,0 +1,125 @@
...
+                }
+                Err(e) => {
+                    if let Err(e) = tx.send(ServerSocketEvent::Error(e)) {
+                        warn!("Send channel error. Reason: {:?}", e)
+                    }
+                }
+            };
+
+            // iter over client events occurred from within laminar.
+            self.socket.events().into_iter().map(|e| {
```
LucioFranco: nit but a for_each is better here since it describes that you are not mapping the values but actually just using the values.

LucioFranco:
about moving this to another crate, I think for the moment its fine, but I imagine we could provide some abstractions on top of laminar for some of this. But I think amethyst is a good playground in a sense to find what works.

amethyst_network/src/net_event.rs
```diff
@@ -45,10 +45,10 @@ pub enum NetEvent<T> {
 impl<T> NetEvent<T> {
     /// Tries to convert a NetEvent to a custom event type.
     pub fn custom(&self) -> Option<&T> {
-        if let NetEvent::Custom(ref t) = self {
-            Some(t)
+        if let &NetEvent::Custom(ref t) = self {
```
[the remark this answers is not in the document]
TimonPost: Not sure

amethyst_network/src/server/server.rs
```diff
@@ -0,0 +1,105 @@
+use super::{ReceiveHandler, SendHandler, ServerConfig, ServerSocketEvent, UdpReceiver, UdpSender};
+use laminar::{error::NetworkResult, DeliveryMethod, Packet};
+use std::net::SocketAddr;
+use std::sync::Arc;
+
+/// 'Server' abstracts TCP and UDP sockets away.
```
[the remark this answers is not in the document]
TimonPost: Yes I like that

amethyst_network/src/server/server.rs
TimonPost: Read my PR description 😉

LucioFranco: should of put that at the top!!! hahaha totally missed that with the diagram 🤦‍♂️

TimonPost:
According to all unimplemented macro's I'll make sure all of them are getting issues bound to it. I forgot to mention it in the PR description.

LucioFranco:
@TimonPost I think it would be fine to group them into one big tracking issue with checkboxes etc.

amethyst_network/src/server/udp.rs
TimonPost: Don't know actually, and I'm not sure at how many packets we should break the receiving. This would be eventually done with congestion prevention.

amethyst_network/src/bundle.rs
```diff
@@ -7,32 +7,56 @@ use amethyst_core::{
...
+
 use crate::filter::NetFilter;
 
 use super::NetSocketSystem;
 
 /// A convenience bundle to create the infrastructure needed to send and receive network messages.
 pub struct NetworkBundle<T> {
     /// Local socket address
-    addr: SocketAddr,
+    pub config: ServerConfig,
```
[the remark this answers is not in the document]
TimonPost: can be private actually

amethyst_network/src/connection.rs
```diff
@@ -32,12 +34,13 @@ pub struct NetConnection<E: 'static> {
 
 impl<E: Send + Sync + 'static> NetConnection<E> {
     /// Construct a new NetConnection. `SocketAddr` is the address that will be connected to.
-    pub fn new(target: SocketAddr) -> Self {
+    pub fn new(target_receiver: SocketAddr, target_sender: SocketAddr) -> Self {
```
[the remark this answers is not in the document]
TimonPost: I leave it since the connection created heer is one that simulates some connection to some endpoint. By having `target` it sounds clearer to me that you want to specify the target.

amethyst_network/src/lib.rs
```diff
@@ -12,26 +12,27 @@ mod connection;
...
+mod server;
 mod test;
 
 pub use crate::{
     bundle::NetworkBundle,
     connection::{ConnectionState, NetConnection, NetIdentity},
     filter::{FilterConnected, NetFilter},
     net_event::NetEvent,
-    network_socket::NetSocketSystem,
+    network_socket::*,
```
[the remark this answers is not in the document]
TimonPost: I had a reason for this when debugging some things, thanks for pointing out.

amethyst_network/src/server/server_socket_event.rs
TimonPost: I was using that for events currently not implemented. But now I took a look at it I saw that it was only used for `QualityChange` event from laminar. I changed `None` to `QualityChange` now.

amethyst_network/src/lib.rs
```diff
@@ -40,7 +41,10 @@ where
         Ok(s) => {
             let slice = s.as_slice();
             // send an unreliable `Packet` from laminar which is basically just a bare UDP packet.
-            match socket.send(&Packet::unreliable(*addr, slice.to_vec())) {
+            match sender.send(ServerSocketEvent::Packet(Packet::unreliable(
```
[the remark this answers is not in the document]
TimonPost: What do you mean by that?

amethyst_network/src/server/receive_handler.rs
```diff
@@ -0,0 +1,55 @@
+use crate::server::ServerSocketEvent;
+use std::thread::JoinHandle;
+use std::sync::{
+    mpsc::Receiver
+};
+
+/// Handler to access the internals of a receiving socket.
+pub struct ReceiveHandler {
+    /// handle that should be used for reading received packets from a given socket
+    pub receiver: Receiver<ServerSocketEvent>,
```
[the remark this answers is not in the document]
TimonPost: I marked them private because public it is not needed for now

amethyst_network/src/net_event.rs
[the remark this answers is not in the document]
TimonPost: changed it back.

amethyst_network/src/server/server.rs
```diff
@@ -0,0 +1,105 @@
...
+pub struct Server {
+    config: ServerConfig,
+    // Handler to access the internals of the UDP receiving thread
+    udp_receiver: Option<ReceiveHandler>,
```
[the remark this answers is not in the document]
TimonPost: Removed the Option because Udp is always used. Tcp is still an option

amethyst_network/src/server/server_socket_event.rs
TimonPost: Clone is used in a vector allocation: `

 let mut buffer = vec![ServerSocketEvent::Empty; receiver.max_throughput];`

to this buffer, incoming events will be written.

amethyst_network/src/server/udp.rs
TimonPost: I will make it work via the `ServerConfig`

examples/net_client/main.rs
TimonPost: I added it again

amethyst_network/src/server/send_handler.rs
```diff
@@ -0,0 +1,32 @@
...
+        }
+    }
+
+    /// Send an event on the internal channel, by doing this it will be scheduled for sending.
+    pub fn send(&self, event: ServerSocketEvent) -> Result<(), SendError<ServerSocketEvent>> {
+        self.sender.send(event)
+    }
+}
+
+unsafe impl Sync for SendHandler {}
```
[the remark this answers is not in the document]
LucioFranco: Let's take a look at it tomorrow together if you'd like.

amethyst_network/src/net_event.rs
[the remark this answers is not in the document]
torkleyy: Consider using `*self` instead (or you'll get a Clippy warning :wink:)

amethyst_network/src/error.rs
```diff
@@ -0,0 +1,85 @@
...
+    UdpError(laminar::error::NetworkError),
+    /// Error that could occur whit IO.
+    IoError(io::Error),
+    /// Error that could occur when serializing whit `bincode`
+    SerializeError(bincode::ErrorKind),
+    /// Error that could occur when sending an `ServerSocketEvent` to some channel.
+    ChannelSendError(mpsc::SendError<ServerSocketEvent>),
+}
+
+impl StdError for NetworkErrorKind {
```
LucioFranco: honestly, I here would do `std::error::Error` I think that is better than renaming.

TimonPost: Oke

amethyst_network/src/error.rs
```diff
@@ -0,0 +1,85 @@
...
+    error::Error as StdError,
+    fmt::{Display, Formatter, Result as FmtResult},
+    io,
+    sync::mpsc,
+};
+
+use crate::server::ServerSocketEvent;
+
+/// The `amethyst_network` result type.
+pub type NetworkError<T> = Result<T, NetworkErrorKind>;
```
LucioFranco: This should be ` pub type Result<T> = Result<T, Error>;`

We should follow how `std::io::Error` does it. https://doc.rust-lang.org/src/std/io/error.rs.html#66-68

So just create:

```rust
pub struct Error {
  kind: ErrorKind,
}
```

and rename `NetworkErrorKind` to `ErrorKind`, since we have modules we can namespace via that and they can be renamed downstream as needed. This is how std does its errors.

LucioFranco:
@udoprog @torkleyy If either of you have time to checkout https://github.com/amethyst/amethyst/pull/1289/commits/40fa2f29ab6ae5a2d36af2cee5be22358b2129a6 related to error handling.

amethyst_network/src/error.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+}
+
+impl From<laminar::error::NetworkError> for NetworkErrorKind {
+    fn from(e: laminar::error::NetworkError) -> NetworkErrorKind {
+        NetworkErrorKind::UdpError(e)
+    }
+}
+
+impl From<Box<bincode::ErrorKind>> for NetworkErrorKind {
```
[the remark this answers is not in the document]
TimonPost: No this could not be done. `bincode::deserialize` is returning an error of type: `Result<T, Box<ErrorKind>`. I need to have a conversion from that to my `NetworkErrorKind`.

amethyst_network/src/error.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// Error that could occur when serializing whit `bincode`
+    SerializeError(bincode::ErrorKind),
+    /// Error that could occur when sending an `ServerSocketEvent` to some channel.
+    ChannelSendError(mpsc::SendError<ServerSocketEvent>),
+}
+
+impl StdError for NetworkErrorKind {
+    fn description(&self) -> &str {
+        match *self {
+            NetworkErrorKind::UdpError(_) => "An UDP error occurred",
```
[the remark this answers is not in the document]
LucioFranco: though the error should still implement display so you can use that for now.

amethyst_network/src/lib.rs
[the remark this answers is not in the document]
LucioFranco: Break out some of the parts to construct the type so 

```rust
let packet = Packet::unreliable(...);
let event = ServerSocketEvent::Packet(...);
match sender.send(event) {
...
}
```

amethyst_network/src/server/server_socket_event.rs
LucioFranco: I don't understand what you mean here? Clone should be to kinda copy the item not construct a new one?

amethyst_network/src/server/udp.rs
LucioFranco: 👍 we should add this to some sort of config.

amethyst_network/src/error.rs
```diff
@@ -0,0 +1,85 @@
...
+    IoError(io::Error),
+    /// Error that could occur when serializing whit `bincode`
+    SerializeError(Box<bincode::ErrorKind>),
+    /// Error that could occur when sending an `ServerSocketEvent` to some channel.
+    ChannelSendError(mpsc::SendError<ServerSocketEvent>),
+}
+
+impl std::error::Error for ErrorKind {
+    fn description(&self) -> &str {
+        use std::error::Error;
```
torkleyy: `description` is deprecated, so I would remove this.

amethyst_network/src/error.rs
```diff
@@ -0,0 +1,85 @@
...
+        match *self {
+            ErrorKind::UdpError(_) => "An UDP error occurred",
+            ErrorKind::IoError(ref error) => error.description(),
+            ErrorKind::ChannelSendError(ref error) => error.description(),
+            ErrorKind::SerializeError(ref error) => error.description(),
+        }
+    }
+
+    fn cause(&self) -> Option<&dyn std::error::Error> {
```
torkleyy: `cause` is deprecated; `source` is used instead. This implementation will create a `source` that returns `None`.

amethyst_network/src/error.rs
```diff
@@ -0,0 +1,85 @@
...
+        match *self {
+            ErrorKind::IoError(ref error) => Some(error),
+            ErrorKind::SerializeError(ref error) => Some(error),
+            ErrorKind::ChannelSendError(ref error) => Some(error),
+            _ => None,
+        }
+    }
+}
+
+impl Display for ErrorKind {
```
torkleyy: I'd recommend using `err-derive` for deriving this and the `source` implementation.

amethyst_network/src/error.rs
```diff
@@ -0,0 +1,85 @@
...
+use std::{
+    fmt::{self, Display, Formatter},
+    io,
+    sync::mpsc,
+};
+
+use crate::server::ServerSocketEvent;
+
+/// The `amethyst_network` result type.
+pub type Error<T> = Result<T, ErrorKind>;
```
torkleyy: This is confusing; please call it `Result`.

torkleyy: Consequently, `ErrorKind` can be called `Error`. `ErrorKind` is usually used for tags-only.

amethyst_network/src/error.rs
```diff
@@ -0,0 +1,85 @@
...
+            _ => None,
+        }
+    }
+}
+
+impl Display for ErrorKind {
+    fn fmt(&self, fmt: &mut Formatter<'_>) -> fmt::Result {
+        match *self {
+            ErrorKind::UdpError(ref error) => write!(fmt, "UDP-error occurred {}", error),
+            ErrorKind::IoError(ref error) => write!(fmt, "IO-error occurred {}", error),
```
torkleyy: One last thing, printing an error that's also the `source()` is discouraged since that will duplicate all errors when printing the chain.

amethyst_network/src/error.rs
TimonPost: Thanks for pointing out it should be indeed Result, must have accidentally renamed it to Error

amethyst_network/src/error.rs
```diff
@@ -0,0 +1,85 @@
...
+
+/// Wrapper for all errors who could occur in `amethyst_network`.
+#[derive(Debug)]
+pub enum ErrorKind {
+    /// Error that could occur on the UDP-socket
+    UdpError(laminar::error::NetworkError),
+    /// Error that could occur whit IO.
+    IoError(io::Error),
+    /// Error that could occur when serializing whit `bincode`
+    SerializeError(Box<bincode::ErrorKind>),
```
udoprog: This has an alias, `bincode::Error`, that's probably what we'd want to use.

amethyst_network/src/error.rs
```diff
@@ -0,0 +1,85 @@
...
+};
+
+use crate::server::ServerSocketEvent;
+
+/// The `amethyst_network` result type.
+pub type Error<T> = Result<T, ErrorKind>;
+
+/// Wrapper for all errors who could occur in `amethyst_network`.
+#[derive(Debug)]
+pub enum ErrorKind {
```
udoprog: Make use of the [`__NonExhaustive` pattern](https://github.com/rust-lang/rfcs/blob/master/text/2008-non-exhaustive.md#how-we-do-this-today) so we guard against compatibility issues in case someone pattern matches over the type.

amethyst_network/src/error.rs
TimonPost: I will take a look at it, because deserialize from bincode returns a boxed error kind. Maybe I can try do dereference it.

udoprog: No need to deref, using the alias directly works (assuming it's available in our version of bincode):
https://docs.rs/bincode/1.0.1/bincode/type.Error.html

TimonPost:
Any other people who want to review those changes, I like this PR to merged.

amethyst_network/src/error.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// Error that could occur when sending an `ServerSocketEvent` to some channel.
+    ChannelSendError(mpsc::SendError<ServerSocketEvent>),
+    #[doc(hidden)]
+    __Nonexhaustive,
+}
+
+impl std::error::Error for ErrorKind {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        Some(self)
+    }
```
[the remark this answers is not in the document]
udoprog: Right, that means that you should remove `cause` completely and _only_ implement `source`.

The default implementation of `source` is returning `None`: https://doc.rust-lang.org/src/std/error.rs.html#206
The default implementation of `cause` calls `source`: https://doc.rust-lang.org/src/std/error.rs.html#143

amethyst_network/src/lib.rs
```diff
@@ -50,9 +56,9 @@ where
 }
 
 /// Attempts to deserialize an event from the raw byte data.
-fn deserialize_event<T>(data: &[u8]) -> Result<NetEvent<T>, Box<ErrorKind>>
+fn deserialize_event<T>(data: &[u8]) -> Result<NetEvent<T>>
```
fhaynes: Non-public function so can use `//` not `///`

amethyst_network/src/network_socket.rs
```diff
@@ -71,88 +46,88 @@ where
...
-                // receive
-                loop {
-                    match socket.recv() {
-                        // Data received
-                        Ok(Some(packet)) => {
-                            if let Err(_) = receive_queue.send(RawEvent::from(packet)) {
+                for event in buffer[0..received].iter() {
+                    match event {
+                        ServerSocketEvent::Packet(packet) => {
+                            if let Err(_) = receive_queue.send(packet.clone()) {
```
fhaynes: Can we not throw the error away and print it? Would be helpful in debugging.

amethyst_network/src/network_socket.rs
```diff
@@ -162,50 +137,60 @@ where
...
+                            raw_event.addr()
                         ),
                     }
                 } else {
                     warn!("Received packet from unknown source");
                 }
             }
+
+            // this will prevent our system to be stuck in the iterator.
+            // After 10000 packets we will continue and leaf the other packets for the next run.
```
fhaynes: Should be `leave` not `leaf`

amethyst_network/src/network_socket.rs
```diff
@@ -162,50 +137,60 @@ where
...
+            // eventually some congestion prevention should be done.
+            if counter >= 10000 {
```
fhaynes: Can we make this a constant somewhere? Or even better, something we can set via the config?

amethyst_network/src/network_socket.rs
```diff
@@ -162,50 +137,60 @@ where
...
+        for raw_event in self.transport_receiver.try_iter() {
             // Get the NetConnection from the source
             for net_connection in (&mut net_connections).join() {
-                // We found the origin
-                if net_connection.target == raw_event.source {
+                if net_connection.target_sender == raw_event.addr() {
                     // Get the event
-                    let net_event = deserialize_event::<E>(raw_event.data.as_slice());
+                    let net_event = deserialize_event::<E>(raw_event.payload());
                     match net_event {
```
fhaynes: Just to make sure I understand, the idea here is that you add packets until you can deserialize an Event?

amethyst_network/src/server/config.rs
```diff
@@ -0,0 +1,26 @@
...
+    /// Specifies whether the server should listen for TCP packets.
+    pub enable_tcp: bool,
+    /// TCP address at which the server will listen for incoming packets.
+    pub tcp_addr: Option<SocketAddr>,
+}
+
+impl Default for ServerConfig {
+    fn default() -> Self {
+        ServerConfig {
+            // py passing in :0 port the OS will give an available port.
```
fhaynes: Should be 'By' not 'py'

amethyst_network/src/server/host.rs
```diff
@@ -0,0 +1,100 @@
+/// This module is about the communication whit this host to some other endpoints.
```
fhaynes: should be `with` not `whit`

amethyst_network/src/server/host.rs
```diff
@@ -0,0 +1,100 @@
...
+impl Host {
+    /// This will start and return an instance of the host.
+    /// 1: Fire up a TCP-sender and TCP-receiver if enabled in config.
+    /// 2: Fire up a UDP-sender and UDP-receiver.
+    /// 3: Set up some `channels` to communicate with underlying threads.
+    ///
+    /// The method uses the config provided when creating a `host` instance.
+    pub fn run(config: ServerConfig) -> Result<Host> {
+        if config.enable_tcp {
+            // todo: create TCP support.
```
fhaynes: Can we print out a message here so the user doesn't think TCP is working when it really isn't?

amethyst_network/src/server/host.rs
```diff
@@ -0,0 +1,100 @@
...
+        unimplemented!()
+    }
+
+    /// Schedule a UDP-packet for sending.
+    pub fn send_udp(&mut self, packet: Packet) -> Result<()> {
+        self.udp_sender.send(ServerSocketEvent::Packet(packet))?;
+        Ok(())
+    }
+
+    /// Shutdown the server and destroy all related resources.
```
fhaynes: Private function, can do `//`

amethyst_network/src/server/host.rs
```diff
@@ -0,0 +1,100 @@
...
+    fn shutdown(&mut self) {
+        // 1. shutdown all connections udp/tcp
+        // 2. clean up resources by closing thread's and such.
+        unimplemented!()
+    }
+
+    /// Broadcast a TCP-packet to all connected clients.
+    fn broadcast_tcp(&mut self, payload: &[u8], addr: SocketAddr) -> Result<()> {
```
fhaynes: Private function, can do `//`

amethyst_network/src/server/host.rs
```diff
@@ -0,0 +1,100 @@
...
+        // 1. Loop tcp clients.
+        // 2. Send data to all clients.
+        unimplemented!()
+    }
+
+    /// Broadcast a UDP-packet to all connected clients.
```
fhaynes: Private function, can do `//`

amethyst_network/src/server/host.rs
```diff
@@ -0,0 +1,100 @@
...
+    fn broadcast_cast_upd(
```
fhaynes: should be `udp` not `upd`

amethyst_network/src/server/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+pub use self::config::ServerConfig;
+pub use self::host::Host;
+pub use self::receive_handler::ReceiveHandler;
+pub use self::send_handler::SendHandler;
+pub use self::udp::{UdpReceiver, UdpSender};
+
+pub use self::server_socket_event::{ClientEvent, ServerSocketEvent};
+use std::sync::mpsc::{Receiver, Sender};
+
+pub trait PacketReceiving {
```
fhaynes: Comment the trait please since it is public

amethyst_network/src/server/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+    /// Start receiving data.
+    fn start_receiving(&mut self, tx: Sender<ServerSocketEvent>);
+}
+
+pub trait PacketSending {
```
fhaynes: Comment this as it is a public trait please

amethyst_network/src/server/send_handler.rs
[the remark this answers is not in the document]
fhaynes: I don't think we should need unsafe for any of this. Let's please try to get rid of this.

amethyst_network/src/server/udp.rs
```diff
@@ -0,0 +1,128 @@
...
+        let mut udp_receiver = UdpReceiver::new(addr)?;
+
+        // channel used for communicating about received packets.
+        let (tx, rx) = mpsc::channel();
+
+        let thread_handle = thread::spawn(move || {
+            udp_receiver.start_receiving(tx);
+        });
+
+        Ok(ReceiveHandler::new(5000, rx, thread_handle))
```
fhaynes: This should be a constant or configurable, I think

amethyst_network/src/server/udp.rs
```diff
@@ -0,0 +1,128 @@
...
+    fn start_sending(&mut self, rx: Receiver<ServerSocketEvent>) {
+        loop {
+            match rx.recv() {
+                Ok(packet) => match packet {
+                    ServerSocketEvent::Packet(packet) => {
+                        if let Err(e) = self.socket.send(&packet) {
+                            error!("Something went wrong when trying to send a packet with UDP socket. Reason: {:?}", e)
+                        }
+                    }
+                    _ => {}
```
fhaynes: Might be worth printing an `error!` here, just in case

amethyst_network/src/server/server_socket_event.rs
fhaynes: Clone should indeed copy the internal state of whatever is being cloned. I think this would be default, not clone.

amethyst_network/src/network_socket.rs
TimonPost: The idea here is that your packet will be deserialized into a `NetEvent<E>` from the received packet payload.

amethyst_network/src/error.rs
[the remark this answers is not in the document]
udoprog: This is not fixed yet afaics?

amethyst_network/src/lib.rs
udoprog: Should probably still prefer `///` since they are contextual to the function and used by e.g. RLS. Don't know if that works as intended with `//`.

amethyst_network/src/error.rs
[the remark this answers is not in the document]
TimonPost: 😑 I'm totally confused now, I'll take another try. I think I am missing interpreting  your suggestions. Do you have an code exaple?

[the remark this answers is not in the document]
TimonPost: And is this really needed to change, since when the new error handling crate from your PR is merged all this error handling code will probably change.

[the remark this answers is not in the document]
udoprog: This is what it should look like (it should look familiar since it's what you initially did in `cause`):
```rust
    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
        match *self {
            ErrorKind::UdpError(ref e) => Some(e),
            ErrorKind::IoError(ref e) => Some(e),
            ErrorKind::ChannelSendError(ref e) => Some(e),
            ErrorKind::SerializeError(ref e) => Some(e),
            _ => None,
        }
    }
```

This is necessary because you are implementing a custom error type. Not using the dynamic one introduced in #1220.

The reason why your approach would cause an infinite loop is because you walk through the error chain like this:

```rust
let mut e = ...;

println!("Error: {}", e);

while let Some(s) = e.source() {
    println!("Caused by: {}", s);
    e = s;
}
```

And if `source` returns `Some(self)` this is gonna loop forever.

Check out the documentation for source for a more complete example on this:
https://doc.rust-lang.org/std/error/trait.Error.html#method.source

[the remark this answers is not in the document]
TimonPost: Thanks for writing that up!!. I'll see what I can do whit this.

amethyst_network/src/bundle.rs
```diff
@@ -7,32 +7,58 @@ use amethyst_core::{
...
     filters: Vec<Box<dyn NetFilter<T>>>,
 }
 
 impl<T> NetworkBundle<T> {
-    /// Creates a new NetworkBundle that connects to the `addr`.
-    pub fn new(addr: SocketAddr, filters: Vec<Box<dyn NetFilter<T>>>) -> Self {
-        NetworkBundle { addr, filters }
+    /// Creates a new NetworkBundle.
+    ///
+    /// `receive_addr:` this is the address on which we will receive incoming packets.
```
LucioFranco: nit lets move the `:` after the ticks 😄

amethyst_network/src/lib.rs
```diff
@@ -49,10 +55,10 @@ where
...
 
-/// Attempts to deserialize an event from the raw byte data.
-fn deserialize_event<T>(data: &[u8]) -> Result<NetEvent<T>, Box<ErrorKind>>
+// Attempts to deserialize an event from the raw byte data.
+fn deserialize_event<T>(data: &[u8]) -> Result<NetEvent<T>>
 where
     T: DeserializeOwned,
 {
-    deserialize::<NetEvent<T>>(data)
+    Ok(deserialize::<NetEvent<T>>(data)?)
```
LucioFranco: This should be `deserialize::<NetEvent<T>>(data).map_err(Error::from)`

amethyst_network/src/network_socket.rs
```diff
@@ -70,89 +45,90 @@ where
...
-            .map_err(|x| Error::new(ErrorKind::Other, x.to_string()))?;
+        let server = Host::run(config.clone())?;
 
-        socket.set_nonblocking(true).map_err(|_| {
-            Error::new(
-                ErrorKind::Other,
-                "Unable to set `UdpSocket` to non-blocking mode",
-            )
-        })?;
+        let server_sender = NetSocketSystem::<E>::start_sending(&server);
```
LucioFranco: this block isn't clear to me why do you send a reference in one call but not in the other? Does a clone happen on the other side if so I would like this to be moved up to this piece of code.

amethyst_network/src/network_socket.rs
```diff
@@ -70,89 +45,90 @@ where
...
-                // send
-                for control_event in send_queue.try_iter() {
-                    match control_event {
-                        InternalSocketEvent::SendEvents { target, events } => {
-                            for ev in events {
-                                send_event(&ev, &target, &mut socket);
-                            }
+    /// Start a thread to send all queued packets.
+    fn start_sending(server: &Host) -> Sender<InternalSocketEvent<E>> {
+        let (tx, send_queue) = channel();
```
LucioFranco: small nit but can you use `mpsc::channel()` here?

amethyst_network/src/network_socket.rs
```diff
@@ -70,89 +45,90 @@ where
...
+    }
+
+    /// Starts a thread which receives incoming packets and sends them onto the 'Receiver' channel.
+    fn start_receiving(server: Host) -> Receiver<Packet> {
+        let (receive_queue, rx) = channel();
+
+        thread::spawn(move || {
+            let receiver = server.udp_receive_handle();
+
+            let mut buffer = vec![ServerSocketEvent::Empty; receiver.max_throughput];
```
LucioFranco: Nit: I'd like to see this renamed to buffer_size not needed for this PR but just a note.

amethyst_network/src/network_socket.rs
TimonPost: `start_sending` takes the reference and `start_receiving` will eventually own the instance. It is just a way to play whit the borrowing.

LucioFranco: Yeah, I see now this confuses me, to be honest, I find it unclear what is being borrowed and for how long. Since you are passing a channel handle down do you think that this function could take that handle instead?

amethyst_network/src/network_socket.rs
```diff
@@ -70,89 +45,90 @@ where
...
-                        // Data received
-                        Ok(Some(packet)) => {
-                            if let Err(_) = receive_queue.send(RawEvent::from(packet)) {
-                                error!("`NetworkSocketSystem` was dropped");
-                                break 'outer;
+        tx
+    }
+
+    /// Starts a thread which receives incoming packets and sends them onto the 'Receiver' channel.
+    fn start_receiving(server: Host) -> Receiver<Packet> {
```
LucioFranco: I think this should receive the udp receive handle instead of the host, as that is a clearer interface.

amethyst_network/src/network_socket.rs
```diff
@@ -162,50 +138,60 @@ where
...
+        let mut counter = 0;
+        for raw_event in self.transport_receiver.try_iter() {
             // Get the NetConnection from the source
             for net_connection in (&mut net_connections).join() {
-                // We found the origin
-                if net_connection.target == raw_event.source {
+                if net_connection.target_sender == raw_event.addr() {
                     // Get the event
-                    let net_event = deserialize_event::<E>(raw_event.data.as_slice());
+                    let net_event = deserialize_event::<E>(raw_event.payload());
```
LucioFranco: not important but these can be merged.

amethyst_network/src/server/send_handler.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    /// Send an event on the internal channel, by doing this it will be scheduled for sending.
+    pub fn send(&self, event: ServerSocketEvent) -> Result<()> {
+        self.sender.send(event)?;
+        Ok(())
+    }
+
+    /// Get the sending channel of this `SendHandler` to which you can send  
+    pub fn get_sender(&self) -> &SyncSender<ServerSocketEvent> {
+        return &self.sender;
```
LucioFranco: drop this return

amethyst_network/src/server/send_handler.rs
```diff
@@ -0,0 +1,41 @@
...
+    ) -> SendHandler {
+        SendHandler {
+            sender,
+            thread_handle,
+        }
+    }
+
+    /// Send an event on the internal channel, by doing this it will be scheduled for sending.
+    pub fn send(&self, event: ServerSocketEvent) -> Result<()> {
+        self.sender.send(event)?;
```
LucioFranco: these two lines can just be `self.sender.send(event)`

amethyst_network/src/server/udp.rs
```diff
@@ -0,0 +1,136 @@
...
+use laminar::NetworkConfig;
+use std::net::SocketAddr;
+use std::sync::mpsc::{self, Receiver, Sender};
+use std::thread;
+
+/// An UDP receiver, wrapper for starting the UDP-receiving thread.
+pub struct UdpReceiver {
+    // socket used for receiving packets
+    socket: UdpSocket,
+    pub max_troughput: u16,
```
LucioFranco: This either needs a comment or if this can not change during the life of a UdpReceiver then I would move this to a `max_throughput()` function.

amethyst_network/src/error.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+impl From<mpsc::SendError<ServerSocketEvent>> for ErrorKind {
+    fn from(e: mpsc::SendError<ServerSocketEvent>) -> ErrorKind {
+        ErrorKind::ChannelSendError(e)
+    }
+}
+
+impl From<laminar::error::NetworkError> for ErrorKind {
```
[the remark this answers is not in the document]
TimonPost: This should be an issue for laminar, but I agree laminar should have also have `ErrorKind`, `Result`

amethyst_network/src/error.rs
```diff
@@ -0,0 +1,75 @@
...
+    fn from(e: laminar::error::NetworkError) -> ErrorKind {
+        ErrorKind::UdpError(e)
+    }
+}
+
+impl From<Box<bincode::ErrorKind>> for ErrorKind {
```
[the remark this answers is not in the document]
TimonPost: Can't be done to my knowledge. deserialize returns a boxed error and the compiler is searching for `From<Box<ErrorKind>>` implementations to translate this error into our own ErrorKind.

amethyst_network/src/network_socket.rs
TimonPost: could you elaborate on what you mean whit merge?

amethyst_network/src/network_socket.rs
TimonPost: Agree, I have to experiment with this further, I tried some things a few seconds ago but it is a little difficult to borrow those Handles.

amethyst_network/src/network_socket.rs
LucioFranco:
```rust
match deserialize_event::<E>(raw_event.payload()) { ... }
```

amethyst_network/src/network_socket.rs
TimonPost: done, I had to move `SendHandler` into an Arc. Although for a `Receiver` it is more difficult to move it around threads since it does not implement `Sync`. For this, I had to move it into a `Mutex`

TimonPost:
@LucioFranco I removed the `get_events` method with an iterator call, this is IMO much better. Also, I refactored the functions `start_sending` and `start_receiving` to not take the `Host` but instead just the `ReceiveHandler` or `SendHandler` as requested. However, I had to move the receiver into an `Arc<Mutex<Receiver>>` because `Receiver` does not implement `Sync`.

amethyst_network/src/bundle.rs
```diff
@@ -7,30 +7,50 @@ use amethyst_core::{
     shred::DispatcherBuilder,
 };
 
-use crate::{filter::NetFilter, NetSocketSystem};
+use super::server::ServerConfig;
```
LucioFranco: super super minor but I'd like these to be merged if possible.

amethyst_network/src/network_socket.rs
```diff
@@ -69,89 +50,91 @@ where
...
+        tx
+    }
 
-                // receive
+    /// Starts a thread which receives incoming packets and sends them onto the 'Receiver' channel.
+    fn start_receiving(receiver: Arc<Mutex<ReceiveHandler>>) -> Receiver<Packet> {
+        let (receive_queue, rx) = mpsc::channel();
+
+        thread::spawn(move || {
+            if let Ok(receiver) = receiver.lock() {
```
LucioFranco: this lock will never be unlocked unless there is a send error, this seems a bit odd? Though I think this function will only get called once? So I think we should either document that or panic if it gets called twice.

amethyst_network/src/network_socket.rs
```diff
@@ -161,50 +144,56 @@ where
...
             } else if net_connection.state == ConnectionState::Disconnected {
-                self.tx
+                self.transport_sender
                     .send(InternalSocketEvent::Stop)
                     .expect("Already sent a stop event to the channel");
             }
         }
 
-        for raw_event in self.rx.try_iter() {
+        for (counter, raw_event) in self.transport_receiver.try_iter().enumerate() {
```
LucioFranco: 👍 good solution I didn't think of it this way and this is def simpler! Good stuff!

amethyst_network/src/network_socket.rs
TimonPost: Indeed, this will not be possible. 

The whole point why the `Receiver` is not sync is that there could not be different receivers. This is the case here. We have one receiving thread, and so requiring the lock for the receiver there is a valid use case. If someone else tries to receive it is not possible. And if he does he should not be able.

LucioFranco: Yeah what I’m saying is that we should be explicit about this behavior so there should be a comment about it and a panic if it gets called more than once.

TimonPost: Yes, I'll add that

TimonPost:
bors r= LucioFranco, fhaynes

TimonPost:
bors r= LucioFranco, fhaynes

TimonPost:
bors r= LucioFranco, fhaynes

TimonPost:
I hope I don't make bors angry today. I put him at work too much haha.

TimonPost:
bors r= LucioFranco, fhaynes

PR #1300, opened 2018-12-31
Use encoder pattern for DrawFlat2D render pass.

Frizi:


amethyst_renderer/src/pass/flat2d/encode_image.rs
```diff
@@ -0,0 +1,101 @@
...
+    specs::{Component, Join, Read, ReadStorage, System, VecStorage, Write},
+    GlobalTransform,
+};
+
+/// Information for rendering a standalone texture as plain image.
+///
+/// Instead of using a `Mesh` on a `DrawFlat` render pass, we can use a simpler set of shaders to
+/// render textures to quads. This struct carries the information necessary for the `DrawFlat2D` pass.
+#[derive(Clone, Debug)]
+pub struct RenderImageFlat2D(pub Handle<Texture>);
```
termhn: I'm not sure I like this naming of the struct, it seems a bit wonky/cumbersome. Perhaps `SpriteRender` for this (or `IndividualSpriteRender` if we really want to be explicit) and `SpriteSheetRender` for the version to be used with a SpriteSheet?

Frizi: There was some discussion about that before I made that into a PR. I understand your point and it is indeed visually more appealing, but the conclusion was that we currently have some ambiguity with existing type names. I wanted to make it as clear as possible that this type represents some kind of cross-section of `DrawFlat2D` and `Image`. Well, actually it could be `Texture` to make it perfectly consistent.

Frizi: It's important to note that other passes will have separate types even if you want to use the same shape of data with them. This is the only way for the user to specify the intent. L and to not influence another rendering pass by accident.

omni-viral:
First of all I appreciate your desire to make rendering code easier to understand and maintain.
But I'd like to discuss design first before we will move with this.


Renderer is most performance critical system (probably right after flying controlling systems for ✈️ and 🚀) so we should be concern of performance above almost everything else.

So I see a little nits in the design. It has `O(n)` complexity, no way to make lower, ideally `O(k)` where `k` is the number of objects whose rendering relevant parameters changed.
I know that what we have now also runs in `O(n)`, but we want better.

The way this implemented seems not extensible. Any user who would want to draw 2d object with custom shader that needs additional data from user's own component  would need to write own encoder system and pass implementations duplicating 99% of the code from existing ones.

My intent for new render is to make it more data-driven and require less code to add new small feature.
Ideally making object that will be rendered with custom shaders would require 0 lines of code.
If custom shader requires some additional data it should require just few lines of code.

I not saying code in PR is worse than what we have now in `amethyst-renderer`. Current render is a mess 😃 
But I worry we would have to completely rewrite it again soon. And hours of work put here would be better to put into making new renderer delivered 📦

omni-viral:
> dynamically-bound extra constant data can be passed through an asset such as `Handle<Material>`

I'm saying about additional information that is contained in user-provided `Component`s required for custom shaders. Implemented `Encoder` can't know how to read that and send to the render pass and has no means for extensions.

kabergstrom:
> Implemented Encoder can't know how to read that and send to the render pass and has no means for extensions.

I see what you mean. I think this is a question of core Amethyst values though. Is Amethyst a library, an extendable framework or an off-the-shelf game creation environment? For some, it is all three, the question is at which levels of abstraction and which parts of the engine do we design for one of these three?

By saying this code should be extendable rather than replaceable(composable), you are saying we should aim to be a rendering framework in this case, right?

omni-viral:
I'm not sure if Amethyst should be a framework or collection of libraries, yet I'm sure it should be able to serve needs of different users with different projects.
Tools that Amethyst provides must be flexible. User shouldn't completely rewrite parts that not entirely fit into requirements.

This solution seems to me not composable in a good way. It can be composed in a sense that user writes his own `WibblyWobbly2DEncoder` and `DrawWibblyWobbly2D` and use it to render wibbly-wobbly 2d objects, while still using out-of-the-box `DrawFlat2D` pass for simple sprites and images.
But well composable it would be if user could add wibbly-wobblyness without rewriting common 2d parts.

Amethyst is data-driven engine. I think that we should aim renderer to be data-driven too.
This means that `Component`s attached to the `Entity` must control how it is get drawn, without myriads of additional entities like encoders and passes.


This PR unifies rendering of similar objects (sprites and bare images), that's great, but can actually hurt performance as it puts some calculations and data updates onto CPU whereas GPU could do it lot faster.

For instance picking a sprite out of spritesheet for particular frame of the particular animation can be done directly on GPU, CPU can update data in GPU buffer only when animation changes, setting timestamp of animation start. In its turn the vertex shader would check current timestamp (updated each frame for all shaders that would use it in single location) and calculate which key-frame in spriteshit should be rendered :tada: we transfer 0 bytes each frame per sprite and basically just resubmit the commands recorder previously.

Considering `O(n)` vs `O(k)`  I'd like to note that `k` usually is order of magnitude lower than `n` and in most cases never gets high. Think of all static objects that never moves or changes, or just stuck in endless loop of single animation. Most objects are like this in games. Of course to get use of that we should share data like camera transform between all draw calls which we are not doing right now.

amethyst_renderer/src/bundle.rs
```diff
@@ -116,6 +129,36 @@ impl<'a, 'b, 'c, B: PipelineBuild<Pipeline = P>, P: 'b + PolyPipeline> SystemBun
                 &["parent_hierarchy_system"],
             );
         }
+
+        // TODO: Extract DrawFlat2D specific settings, possibly to a separate bundle.
+        if let Some(deps) = self.drawflat2d_encoders_deps {
+            builder.add(Flat2DImageEncoder::default(), "flat2d_image_encoder", deps);
```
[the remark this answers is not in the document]
Frizi: That was decided that way by rust_fmt for some reason.

kabergstrom:
Attempt to design somewhat data-driven approach to `World` extraction.

We can boil down World data extraction to the following problem statement.
1. We know the layout of Components at compile-time
2. We don't know the layout of Pipeline buffers that Component data will be written to at compile-time (without current Pass type for specific object kind + known shader struct)

If we can at compile time know that a specific set of `Component` types mean they should be rendered (for example, RenderSpriteFlat2D + GlobalTransform), and that a single `Pipeline` can be extracted from them, we can do the following:
- Define "render field attribute" in components, similar to how `serde` does it; `#[serde(rename= "name")]`. The names would be used to link to `Pipeline` offsets. 
- Use macros that take a set of components and generates three different functions.
- 1. Extract a single non-optional `Pipeline` from the set of components.
- 2. const fn: Get an ordered list of the "render field names" for the set of components.
- 3. Extract the `World` data. Takes a ptr and a list of offsets (corresponding to ordering of fn #2) and writes each extracted field to the appropriate offset. Each offset would be "optional" in the generated code, essentially being a conditional move/memcpy.

To render:
- For each entity with this set of components,
- Extract the `Pipeline`
- Find the common set of "render field names" between the set of components and the `Pipeline` buffers and generate the ordered field offset list
- Call the `World` extraction function with the field offset list

I think this would be very optimized since there is no conditional/indirect load or other load that may induce a cache miss. Conditional writes do not induce latency in the CPU pipeline so should be fast.

We could probably also support custom `write` functions similar to `#[serde(serialize_with = "path")]` that take the set of components and an offset to write to.

amethyst_renderer/src/bundle.rs
```diff
@@ -61,9 +59,22 @@ where
         self
     }
 
-    /// Enable transparent sprite sorting, with the given dependencies
-    pub fn with_sprite_visibility_sorting(mut self, dep: &'a [&'a str]) -> Self {
-        self.sprite_visibility_sorting = Some(dep);
+    /// Enable encoders for `DrawFlat2D` render pass.
+    ///
+    /// This has to be enabled for the pass to render anything on the screen.
+    pub fn with_drawflat2d_encoders(mut self, dep: &'a [&'a str]) -> Self {
```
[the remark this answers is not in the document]
Frizi: I fixed the doc part, names are kept the same for now. If anybody has better idea, i'm happy to change.

omni-viral:
> For each entity with this set of components,

What set of components? `RenderSpriteFlat2D` + `GlobalTransform`? What about not 2d, and not sprites?

Sets of components that defines `Pipeline` should contain shaders in some way.

kabergstrom:
We spread out this discussion among a lot of different places, let's focus it in the forums: https://community.amethyst-engine.org/t/rendering-runtime-api-types-rethink/298/24?u=kabergstrom

termhn:
I think this PR is now outdated as the encoder system has been completely rewritten. Closing.

PR #1302, opened 2018-12-31
Fix the "json" feature for 2018 edition

randomPoison:
Looks like we missed a bit of feature-gated code that needed to updated when switching to the 2018 edition. The fix is simple, but it would be good to also test make sure we're testing the "json" feature under CI. If someone could point me to the right place to do that, I'll add it to this PR.

### DIFF
--- a/amethyst_assets/src/lib.rs
+++ b/amethyst_assets/src/lib.rs
@@ -21,8 +21,6 @@ extern crate log;
 
 #[macro_use]
 extern crate serde;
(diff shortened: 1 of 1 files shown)


randomPoison:
Alright here goes:

bors r+

randomPoison:
rustfmt won't install on my machine for some reason so, uh, guess I'm just gonna let bors tell me what I did wrong 😓 

bors r+

PR #1304, opened 2019-01-01
Fix typo.

yuvallanger:


yuvallanger:
@TimonPost How about now?

book/src/concepts/state.md
```diff
@@ -224,7 +224,7 @@ impl State<(), MyEvent> for GameplayState {
 # fn main() {}
 ```
 
-To make `Application` aware of the change to which events to send to the state, you also need to supply both the 
+To make `Application` aware of the change, and decide to which events the state should be send, you need to supply both the
```
jojolepro:
```suggestion
 To make `Application` aware of the change, and decide which events to send to the state, you need to supply both the 
```

yuvallanger: This GitHub-centric patch-on-a-patch-on-a-patch workflow is new to me. I hope I won't blow it all up.

khionu:
bors +r

khionu:
bors r+

I'm going to start typing this over and over until I have it in muscle memory.

Moxinilian:
@yuvallanger Could you please remove the leading space and squash so we can merge?
bors r-

PR #1306, opened 2019-01-02
Fix the SRV mipmap range

kvark:
The old range was causing a panic on gfx-device-gl 0.15.5 since it was out of range of the source texture.


jojolepro:
@omni-viral

Moxinilian:
This should be merged then applied to the 0.10 release to be release as 0.10.2 to avoid future errors if people cargo update.

azriel91:
in my opinion the fix is critical enough to speed up the review process, and:

* the change is small, so risk of unexpected side effects is low to zero
* it's by `@kvark`
* tested

bors r+

PR #1307, opened 2019-01-02
Change the recommended donation website

Moxinilian:
This PR changes the donation link on the README to go to the donation page on the new website instead of OpenCollective, as the DonorBox platform available there is the recommended donation platform (as it takes a smaller cut on the donation).

Do we also want to modify/remove the OpenCollective backer avatars under it? I personally think we should just not refer to the OpenCollective platform as it is only here for donators that prefer it over anything else.

Also, is it okay to link to the new website or should we wait it has undergone all the rewording and refactoring? @happenslol

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ you can look through the [issue tracker][it].
 
 ## Backers
 
-Thank you to all our backers! 🙏 [Become a backer](https://opencollective.com/amethyst#backer)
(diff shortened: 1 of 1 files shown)


happenslol:
It's fine to link it, it's definitely usable in the current state. And at least the donation page offers more information than was possible on the old page (e. g. Donor wall).

PR #1311, opened 2019-01-02
Correct the stated intentions in the sentence.

yuvallanger:
We want to check if the code successfully runs, not just successfully compiles.

### DIFF
--- a/book/src/pong-tutorial/pong-tutorial-02.md
+++ b/book/src/pong-tutorial/pong-tutorial-02.md
@@ -280,7 +280,7 @@ This is all the information Amethyst needs to track and move the paddles in our
 virtual world, but we'll need to do some more work to actually *draw* them.
 
 As a sanity check, let's make sure the code for initialising the paddles
-compiles. Update the `on_start` method to the following:
+successfully runs. Update the `on_start` method to the following:
 
 ```rust,no_run,noplaypen
 # extern crate amethyst;


yuvallanger:
Oh boy. Sorry about that. Will do! Should I cancel the other pull request
and add the changes here?

Moxinilian:
Yes that would be best. Thanks!

yuvallanger:
@Moxinilian Is there any way of pushing more commits to this branch without firing up the CI pipeline? Should I maybe accumulate more and more changes, then squash them all before pushing?

Moxinilian:
It seems like CI is broken anyway so don't bother lol
But making it into one PR still makes it easier for us to review so go ahead.

TimonPost:
Thanks for the work! You could just close this PR and reopen when you're ready with all the sentence improvements. This will prevent the CI to trigger.

book/src/pong-tutorial/pong-tutorial-02.md
```diff
@@ -280,7 +280,7 @@ This is all the information Amethyst needs to track and move the paddles in our
 virtual world, but we'll need to do some more work to actually *draw* them.
 
 As a sanity check, let's make sure the code for initialising the paddles
-compiles. Update the `on_start` method to the following:
+successfully runs. Update the `on_start` method to the following:
```
torkleyy: Shouldn't it be "runs successfully"?

khionu: Not necessarily. In "I occasionally drink soda", the adverb is before the verb. However, "I drink soda, occasionally" is also valid. This is getting into really nit-picky grammar rules that I don't fully understand, but, while "This code runs successfully" is accepted English, I don't think it's correct by the actual rules. In any case, I don't think we should fuss about rules this deep. Language is flexible because of the demands of human nature.

OvermindDL1: As a note, an english teacher here (I work at a college in the central USA) says that the comma in `"I drink soda, occasionally"` is incorrect and it should be `"I drink soda occasionally"`.  A comma in that position is an antiquated form that is not recommended for modern usage unless you are trying to imply an actual pause (in which case using `"I drink soda...occasionally"` is more common to show that pause; however, that is not strictly correct).

PR #1315, opened 2019-01-02
lib.rs Formatting

SkylineP1g3on:
So, I know that rustfmt is supposed to be used for this but why? This order seems to make more sense then what was there previous. 

My order:
-> #![doc( ... 
-> #![warn( ...
-> #![macro_use( ... 
-> pub use ...
-> use ...
-> pub use crate::{...
-> pub use self::{...
-> mod ...
-> public mod ...

 amethyst_core/src/lib.rs is a good example, in general I tried to logically group stuff together.

### DIFF
--- a/amethyst_config/src/lib.rs
+++ b/amethyst_config/src/lib.rs
@@ -1,5 +1,4 @@
 //! Loads RON files into a structure for easy / statically typed usage.
-//!
 
 #![crate_name = "amethyst_config"]
 #![doc(html_logo_url = "https://www.amethyst.rs/assets/amethyst.svg")]
@@ -7,7 +6,6 @@
 
 #[macro_use]
 extern crate log;
-use ron;
 
 use std::{
     error::Error,
@@ -15,6 +13,7 @@ use std::{
     path::{Path, PathBuf},
 };
 
+use ron;
 use ron::{de::Error as DeError, ser::Error as SerError};
 use serde::{Deserialize, Serialize};
 
--- a/amethyst_network/src/lib.rs
+++ b/amethyst_network/src/lib.rs
@@ -1,17 +1,9 @@
 //! Provides a client-server networking architecture to amethyst.
 
 #![warn(missing_docs, rust_2018_idioms, rust_2018_compatibility)]
+
 #[macro_use]
 extern crate serde;
-
-mod bundle;
-mod connection;
-mod error;
-mod filter;
-mod net_event;
-mod network_socket;
-mod server;
-mod test;
 
 pub use crate::{
     bundle::NetworkBundle,
@@ -23,12 +15,21 @@ pub use crate::{
     server::{Host, ServerConfig, ServerSocketEvent},
 };
 
+use std::{net::SocketAddr, sync::mpsc::SyncSender};
+
 use bincode::{deserialize, serialize};
 use laminar::Packet;
 use log::error;
 use serde::{de::DeserializeOwned, Serialize};
-use std::net::SocketAddr;
-use std::sync::mpsc::SyncSender;
+
+mod bundle;
+mod connection;
+mod error;
+mod filter;
+mod net_event;
+mod network_socket;
+mod server;
+mod test;
 
 /// Sends an event to the target NetConnection using the provided network Socket.
 /// The socket has to be bound.
--- a/amethyst_ui/src/lib.rs
+++ b/amethyst_ui/src/lib.rs
@@ -3,53 +3,23 @@
 #![doc(html_logo_url = "https://www.amethyst.rs/assets/amethyst.svg")]
 #![warn(missing_docs, rust_2018_idioms, rust_2018_compatibility)]
 
-use amethyst_assets;
-use amethyst_audio;
-use amethyst_core;
-
-use amethyst_renderer;
-use clipboard;
 #[macro_use]
 extern crate derivative;
+
 #[macro_use]
 extern crate derive_new;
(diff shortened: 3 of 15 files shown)


Xaeroxe:
Sorry formatting that we can't automate is kind of a non-starter for our repository. If you can make rustfmt.toml follow your guidelines then we can consider it.

Xaeroxe:
Ok, so this might actually be workable. Here's the output from `cargo fmt --all -- --check`

```
> cargo fmt --all -- --check
Diff in \\?\C:\...\amethyst\amethyst_core\src\lib.rs at line 51:

 /// A rayon thread pool wrapped in an `Arc`. This should be used as resource in `World`.
 pub type ArcThreadPool = Arc<rayon::ThreadPool>;
+
Diff in \\?\C:\...\amethyst\amethyst_input\src\lib.rs at line 10:

 use amethyst_core;
 use smallvec;
-use winit;
 use std::iter::Iterator;
+use winit;
 use winit::VirtualKeyCode;

 #[cfg(feature = "sdl_controller")]
Diff in \\?\C:\...\amethyst\amethyst_ui\src\lib.rs at line 30:
 use gfx_glyph;
 use glsl_layout;

+use clipboard;
 use fnv;
 use hibitset;
 use ron;
Diff in \\?\C:\...\amethyst\amethyst_ui\src\lib.rs at line 36:
-use clipboard;
 use shred;
 use winit;

Diff in \\?\C:\...\amethyst\amethyst_assets\src\lib.rs at line 31:
 #[macro_use]
 extern crate shred_derive;

-use shred;
 use amethyst_core;
+use shred;

 pub use crate::{
     asset::{Asset, Format, FormatValue, SimpleFormat},
Diff in \\?\C:\...\amethyst\amethyst_renderer\src\lib.rs at line 65:

 use gfx;
 use gfx_core;
-use winit;
 use rayon;
 use shred;
 use wavefront_obj;
Diff in \\?\C:\...\amethyst\amethyst_renderer\src\lib.rs at line 72:
+use winit;

 pub use crate::{
     blink::{Blink, BlinkSystem},
```

Fix these and we might have something we can merge.

torkleyy:
Also I'd like to adjust the order a bit; from

> @SkylineP1g3on's order:
> -> #![doc( ...
> -> #![warn( ...
> -> #![macro_use( ...
> -> pub use ...
> -> use ...
> -> pub use crate::{...
> -> pub use self::{...
> -> mod ...
> -> public mod ...

to

@torkleyy's order:
* pub external crates
* external crates
* pub use
* use
* pub mod
* mod

uses are ordered as follows:

* std
* external crates
* workspace crates
* local crate
* local mod (`self`)

I don't think an attribute on the `extern crate` should change its location; please just leave that alphabetical.

---

While we're at it, we should add this to a style guidelines document.

cc @fhaynes

Xaeroxe:
@torkleyy btw we migrated to Rust 2018 which discourages any kind of "extern crate". So maybe remove those from your order.

Xaeroxe:
We do, and they should be removed.

SkylineP1g3on:
What actually makes something an "extern crate".

Xaeroxe:
The literal text "extern crate" imo.

amethyst_ui/src/lib.rs
```diff
@@ -1,56 +1,26 @@
 //! Provides components and systems to create an in game user interface.
 
-#![doc(html_logo_url = "https://www.amethyst.rs/assets/amethyst.svg")]
 #![warn(missing_docs, rust_2018_idioms, rust_2018_compatibility)]
+#![doc(html_logo_url = "https://www.amethyst.rs/assets/amethyst.svg")]
```
azriel91: yo, I think you moved this by accident (based on the order mentioned in the first comment)

amethyst_network/src/lib.rs
```diff
@@ -23,12 +15,22 @@ pub use crate::{
     server::{Host, ServerConfig, ServerSocketEvent},
 };
 
+use std::net::SocketAddr;
+use std::sync::mpsc::SyncSender;
```
azriel91: since you're moving these, could you group them into a `std::{net::SocketAddr, sync::mpsc::SyncSender};`? Thanks!

SkylineP1g3on:
I put the spaces there for readability as my eyes kept missing a line when they were all bunched up. 

Also I don't really need to make concourse chug for those two changes?

amethyst_network/src/lib.rs
SkylineP1g3on: Done. :)

azriel91:
don't really have to, but I don't think we have a "hey concourse, ignore building this commit" tag / flag

forgot a `git push`?

Xaeroxe:
Thanks for your work! I'll be happy to merge this as soon as we squash the commits into one commit using `git rebase -i master`.

amethyst_config/src/lib.rs
```diff
@@ -1,20 +1,19 @@
...
 extern crate log;
-use ron;
 
 use std::{
     error::Error,
     fmt, io,
     path::{Path, PathBuf},
 };
 
+use ron;
```
LucioFranco: I thought we could drop these now with rust 2018?

khionu: 2018 allows us to remove `extern crate`s when the crate in question is 2018 or doesn't need `#[macro_use]`. 2015 crates that export macros still require `extern crate`s

PR #1318, opened 2019-01-04
Minor fixes to resolve compile errors.

Toqozz:
Absolute path is required for using the pong module, otherwise there's a compile error on rust stable (1.31.1); https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html

Ideally my previous commit would have included these cases, sorry about that.


Xaeroxe:
Crashing on a network test, which this didn't touch.

So

bors r+

again I guess.

Xaeroxe:
I overrode bors. We gotta figure out why those networks tests are unreliable.

torkleyy:
I think doing anything that relies on any sort of performance, threading, etc. is bad to test on Travis. Those build systems are simply not reliable for that.

PR #1320, opened 2019-01-04
Fix #1319

### LINKED ISSUE #1319: [BUG] Ortho camera utils do not adjust the camera aspect ratio if screen is not resized
## Description
Using `OrthoCamera` component and `CameraOrthoSystem` system successfully adjusts the rendered view to the screen size when the window is resized, but not so if the window is not resized at all. This is critical when the game is run in fullscreen mode.

## Your Environment

- Version of Amethyst = 0.10.0 (but also applies to the current master branch)

## Additional Context
The cause of this problem seems to be [this part](../blob/master/amethyst_utils/src/ortho_camera.rs#L222) in the [`CameraOrthoSystem`](../blob/master/amethyst_utils/src/ortho_camera.rs#L207).
```rust
pub struct CameraOrthoSystem {
    aspect_ratio_cache: f32,
}

impl<'a> System<'a> for CameraOrthoSystem {
    ...
    fn run(&mut self, (dimensions, mut cameras, ortho_cameras): Self::SystemData) {
        let aspect = dimensions.aspect_ratio();

        if aspect != self.aspect_ratio_cache {
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
            self.aspect_ratio_cache = aspect;

            for (mut camera, ortho_camera) in (&mut cameras, &ortho_cameras).join() { ... }
        }
}
```

Xaeroxe:
Fixes #1319

@JameeKim Try this branch

### DIFF
--- a/amethyst_utils/src/ortho_camera.rs
+++ b/amethyst_utils/src/ortho_camera.rs
@@ -349,6 +349,7 @@ mod test {
                 top: 0.0,
                 bottom: 1.0,
             },
+            aspect_ratio_cache: 0.0,
         };
         assert_eq!((0.0, 1.0, 1.5, -0.5), cam.camera_offsets(aspect));
     }
@@ -364,6 +365,7 @@ mod test {
                 top: 2.0,
                 bottom: 0.0,
             },
+            aspect_ratio_cache: 0.0,
         };
         assert_eq!((0.0, 2.0, 0.0, 2.0), cam.camera_offsets(aspect));
     }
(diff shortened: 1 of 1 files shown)


JameeKim:
Worked like a charm! Thanks <3

Xaeroxe:
Woops, forgot to update the tests. Fixing that now.

azriel91:
bors r=JameeKim,jojolepro,azriel91

PR #1326, opened 2019-01-07
Update links to new docs URL

khionu:
Waiting to hear if this is what we're ready to do, but this was a simple `sed` replace across all files in the repo. I reviewed the diff, there doesn't appear to be any bad edits, but there might be missing edits.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ exclude = ["examples/*", "book/*"]
 keywords = ["game", "engine", "sdk", "amethyst"]
 categories = ["game-engines"]
 
-documentation = "https://www.amethyst.rs/doc/master/doc/amethyst"
+documentation = "https://www.amethyst-engine.org/doc/latest/doc/amethyst/"
 homepage = "https://www.amethyst.rs/"
 repository = "https://github.com/amethyst/amethyst"
 
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 [s1]: https://travis-ci.org/amethyst/amethyst.svg?branch=master
 [s2]: https://img.shields.io/crates/v/amethyst.svg
 [docs-badge]: https://img.shields.io/badge/docs-website-blue.svg
-[docs]: https://www.amethyst.rs/doc/
+[docs]: https://www.amethyst-engine.org/doc/
 [s3]: https://img.shields.io/badge/license-MIT%2FApache-blue.svg
 [s4]: https://img.shields.io/discord/425678876929163284.svg?logo=discord
 [s6]: https://tokei.rs/b1/github/amethyst/amethyst?category=code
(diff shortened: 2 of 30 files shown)


Cargo.toml
```diff
@@ -8,7 +8,7 @@ exclude = ["examples/*", "book/*"]
 keywords = ["game", "engine", "sdk", "amethyst"]
 categories = ["game-engines"]
 
-documentation = "https://www.amethyst.rs/doc/master/doc/amethyst"
+documentation = "https://www.amethyst-engine.org/doc/latest/doc/amethyst/"
```
LucioFranco: Are we moving to this domain? Has it been setup to build from `amethyst/website`?

khionu: It was said that it was an eventuality, and with the sudden outage, I figured it would be prudent to prepare.

That, and I was testing TC's PR building.

And I'm not sure if it regularly builds/deploys from it, but it is currently usable.

LucioFranco: Can we update the links but remain on amethyst.rs? At least for now, though the links need to be updated to point to latest for release versions.

khionu: I'm not sure. This PR is a preparation for something I know has been planned for, but is not something I can say we're ready for.

fhaynes: @LucioFranco yes, we have to move to this domain, because we have no control over amethyst.rs, nor can we get it. Before merging, I'd like a checklist of everything we need to do to migrate to a different domain. This can wait until then.

fhaynes:
This actually did run in TC: https://teamcity.amethyst-engine.org/viewType.html?buildTypeId=Amethyst_Build&branch_Amethyst=pull%2F1326&tab=buildTypeStatusDiv

Now to get it to post messages back...

Cargo.toml
LucioFranco: Right so what I was suggesting is that we use this to update the cargo and the links to actually point to latest instead of master. I know we will have to move but currently amethyst.rs works and as far as I can see amethyst-engine.org isn't fully setup yet.

LucioFranco: so I'd like to move this forward or I can open another that will make similar changes because we need to update the links anyways but we should still remain on the old domain until the new one is ready.

fhaynes:
Hey, TeamCity posted the results! Yay!

Cargo.toml
fhaynes: Close this one for now, and do another one with the old links. Moving domains is going to require some coordination.

fhaynes:
And now we're re-opening! What fun!

Xaeroxe:
About 66% of this PR is now merge conflicts. Might be easier to just rewrite it if and when we decide to do this.

khionu:
I'll leave this for someone else to do.

PR #1328, opened 2019-01-07
compare system font extension as lower case

verzuz:
Windows might return the default system font with the extension "TTF" (uppercase), so before comparing to "ttf" it should be converted to lowercase (for the comparison only).

### DIFF
--- a/amethyst_ui/src/font/default.rs
+++ b/amethyst_ui/src/font/default.rs
@@ -18,11 +18,16 @@ pub fn get_default_font(loader: &Loader, storage: &AssetStorage<FontAsset>) -> F
         Ok(handle) => match handle {
             FontKitHandle::Path { path, .. } => {
                 if let Some(file_extension) = path.extension() {
-                    let format = if file_extension == "ttf" || file_extension == "otf" {
-                        Some(TtfFormat)
-                    } else {
-                        error!("System font '{}' has unknown format", path.display());
-                        None
+                    let format = match file_extension.to_str() {
+                        Some(ext) => {
+                            let ext = ext.to_lowercase();
+                            if ext == "ttf" || ext == "otf" {
+                                Some(TtfFormat)
+                            } else {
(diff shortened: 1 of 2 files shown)


docs/CHANGELOG.md
```diff
@@ -41,6 +41,9 @@ it is attached to. ([#1282])
 [#1281]: https://github.com/amethyst/amethyst/pull/1281
 [#1302]: https://github.com/amethyst/amethyst/pull/1302
 
+* Fixed default system font loading to accept uppercase extension ("TTF") ([#1328])
```
azriel91: Hiya, thanks for this! Please could you move this line to the `Fixed` section above?
Right now it [doesn't render the link correctly](https://github.com/amethyst/amethyst/blob/56c60c235569898eae557edc9ffdc91fd96a0f42/docs/CHANGELOG.md#fixed)

amethyst_ui/src/font/default.rs
```diff
@@ -18,11 +18,16 @@ pub fn get_default_font(loader: &Loader, storage: &AssetStorage<FontAsset>) -> F
...
             FontKitHandle::Path { path, .. } => {
                 if let Some(file_extension) = path.extension() {
-                    let format = if file_extension == "ttf" || file_extension == "otf" {
-                        Some(TtfFormat)
-                    } else {
-                        error!("System font '{}' has unknown format", path.display());
-                        None
+                    let format = match file_extension.to_str() {
+                        Some(ext) => {
+                            let ext = ext.to_lowercase();
```
udoprog: Note: this could use [`eq_ignore_ascii_case`](https://doc.rust-lang.org/std/primitive.str.html#method.eq_ignore_ascii_case) to avoid the copies used in `to_lowercase`.

verzuz: Interesting! I'll try that.

PR #1329, opened 2019-01-07
ReaderId<winit::Event> doesn't make sense. 

SkylineP1g3on:
Also seems bit odd to be calling out to renderer to handle close events here, in the theme of re-usability and all.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -8,8 +8,6 @@ use rayon::ThreadPoolBuilder;
 
 #[cfg(feature = "profiler")]
 use thread_profiler::{register_thread_with_profiler, write_profile};
-use winit::Event;
-
 use crate::{
     assets::{Loader, Source},
     callback_queue::CallbackQueue,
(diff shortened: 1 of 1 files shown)


jojolepro:
Okay, so I thought your removal was pretty funny, but I took the time to test it and I see where the confusion came from.

First of all, the piece of code you remove is the "last defense" at closing the app when the user closes the window.

That is fine, it is the expected behavior.

However, I also found out that we are doing the exact same thing in two other different places.
The first one is here: https://github.com/amethyst/amethyst/blob/47bcf8a2e5e86a4af6b8abfc189e8f0690cbd581/src/state.rs#L139 (and here) https://github.com/amethyst/amethyst/blob/47bcf8a2e5e86a4af6b8abfc189e8f0690cbd581/src/state.rs#L246

The second one, is that in some examples, we are checking AGAIN for the window closed events. Like here: https://github.com/amethyst/amethyst/blob/47bcf8a2e5e86a4af6b8abfc189e8f0690cbd581/examples/custom_game_data/main.rs#L89
In this last case, we are also checking for the `Escape` key.

There's improvement that can be done, for example by re-adding the code you remove, and instead remove the code at the two previous locations. That would require adding a opt-in handling in app.rs for the `Escape` key, which would be used by some of the examples.

Also, I do agree that app.rs isn't the best place to put all of this in. :)

Good luck!

SkylineP1g3on:
Glad that someone understands my humor around here :) . Thank you @jojolepro, for looking.  I understood the expected behavior was to catch the window-close. Is it possible to move this to the .with_bundle(Render... , as we don't really need winit and renderer::{} in the top level generic app? ( Headless server )

PR #1334, opened 2019-01-09
Transform Refactor

jojolepro:
As I discussed earlier, I made some pretty significant changes to how `Transform` and `GlobalTransform` work.

The fork is located here: https://github.com/jojolepro/amethyst/tree/transform

Here's a quick recap:
- `GlobalTransform` has been removed, and the fields were moved into `Transform`
- `Transform` is now generic over the N: Real bound (f32, f64).

Here's what is left to do:
- Fix the broken code
- Fix the broken docs

Here's some numbers:
- There is 92 mentions of "GlobalTransform" in the whole codebase (includes book).
- There is 840 mentions of "Transform".

Here are the crates that will need to be adapted to support this change and tested. I would suggest picking between two and four of them for each person wanting to participate.

When you are in charge of a crate, your job is to make it compile, along with the tests, and adapt the documentation where applicable.

Thanks!

- [x] amethyst_animation 
- [x] amethyst_config    
- [x] amethyst_derive 
- [x] amethyst_locale   
- [x] amethyst_ui     
- [x] amethyst_assets     
- [x] amethyst_controls  
- [x] amethyst_gltf    
- [x] amethyst_network   
- [x] amethyst_utils
- [x] amethyst_audio      
- [x] amethyst_core      
- [x] amethyst_input   
- [x] amethyst_renderer
- [x] book
- [x] examples
- [x] amethyst_test


amethyst_core/src/transform/systems.rs
```diff
@@ -38,25 +33,14 @@ impl<'a> System<'a> for TransformSystem {
     type SystemData = (
         Entities<'a>,
         ReadExpect<'a, ParentHierarchy>,
-        ReadStorage<'a, Transform>,
+        WriteStorage<'a, Transform3<f32>>,
```
Frizi: If systems are just going to hardcode the generic, what's the utility of having it generic at all? I see many potential problems of different systems trying to use either f32 of f64, which will end up being a totally separate components.

Frizi: We would need to propagate the generic with bounds to every single system that uses transforms (which will probably be a lot). It will add a lot of noise and possible source of bugs, yet I'm not sure what are the upsides to counter that.

jojolepro: that's me being dumb, I forgot to update it correctly.
What I'm trying to go for is forcing all systems to have this bound:

```
N: Real = f32
```
That way you don't need additional code if you want the default f32, and you can still enable f64 easily when creating the systems.

Frizi: You will have to remember to register the right component system types to the ECS though. It's not really obvious to newcommers that all of those types must match, otherwise things will silently fail to see each other.

Frizi: There isn't. That's the point. You might also actually have both at the same time if you so desire. I'm just in general not sure what's the point of having support for `f64` at all. I might be not up to date. Were there any usecases presented for that?

I fear that this will push people into unnecessarily implementing their own systems in a generic way. Just because hardcoding the type might feel like a tech debt, people will often do it. But in reality it might be totally unnecessary to care about generics here at all. This problem might probably be solved by good documentation. It just feels like a bunch of unnecessary noise to me anyway.

Frizi: Also note that this might have an unitentional implact on other teams down the road. I suspect that this `Transform` component will be used to position things in the editor for example. If you want to be oblivious to the chosen type, you have to somehow support both. What if an entity has both at the same time? Also any other components that end up being often combined with Transform might end up needing the same generic in order to effectively do math between them without precision problems. It is just very "infectious".

Another example is the system I'm currently working on - encoders. I would have to implement the encoder in generic way and explain that the registration needs the correct Real type in order to take any effect. Now supporting `Transform<f64>` actually requires changes to the shader source code - all my matrices must use double precision. This quickly gets out of hands.

kabergstrom: I haven't read through everything here yet, just wanted to chime in on the aspect of supporting `f64`. Star Citizen's team went as far as to say 
> 64-bit world space is one of the driving elements behind making Star Citizen possible.
[Source](https://www.gamersnexus.net/gg/2622-star-citizen-sean-tracy-64bit-engine-tech-edge-blending)

CBenoit:
If it's okay, I'm willing to help. Maybe two crates as I'm not very experimented with amethyst internals.

jojolepro:
Cool! Feel free to pick those you feel the most comfortable with :)

CBenoit:
I'll start with amethyst_network then!

jojolepro:
Does amethyst network even use transforms at all? :)

CBenoit:
Check! Not even one. :rofl:  I'll do `amethyst_animation` instead!

jojolepro:
amethyst_core now compiles (excluding tests). That way, it will be easier to fix the other crates. ;)

khionu:
Why does this PR add CI changes and a PR Template?

jojolepro:
I probably messed up a merge somewhere

CBenoit:
You can check amethyst_controls (https://github.com/jojolepro/amethyst/pull/14), amethyst_input and amethyst_ui.

amethyst_core/Cargo.toml
```diff
@@ -16,7 +16,9 @@ appveyor = { repository = "amethyst/amethyst" }
 travis-ci = { repository = "amethyst/amethyst" }
 
 [dependencies]
-nalgebra = { version = "0.17", features = ["serde-serialize", "mint"] }
+#nalgebra = { version = "0.17", features = ["serde-serialize", "mint"] }
+nalgebra = { git = "http://github.com/jojolepro/nalgebra", branch = "finite", features = ["serde-serialize", "mint"] }
```
CBenoit: We might want to revert this.

jojolepro: We don't use N::is_finite(value) anymore?

CBenoit: Ah no, probably still used

CBenoit:
`amethyst_test` already compiles and examples in the `book` compiles too.

'
```diff
@@ -0,0 +1,285 @@
+use std::marker::PhantomData;
```
distransient: This file appears to be erroneously included.

amethyst_core/Cargo.toml
termhn: A git dependency will mean we can't release 0.11 until nalgebra releases with the needed feature.

jojolepro: nalgebra was release a little while ago

distransient: Yeah this was fixed in cc955c9. I'm just going to resolve.

amethyst_core/src/transform/systems.rs
distransient: The original issue in this thread has also been fixed.

jojolepro:
bors r=fhaynes,distransient,CBenoit

distransient:
Should be able to r=fhaynes,distransient now

jojolepro:
bors r=fhaynes,distransient,CBenoit

jojolepro:
bors retry

PR #1338, opened 2019-01-10
Add unit tests to input handler

Xaeroxe:
Since winit permits this now I'm going to be slowly adding unit tests to the input handler. This is the first PR of several.

### DIFF
--- a/amethyst_input/src/button.rs
+++ b/amethyst_input/src/button.rs
@@ -10,7 +10,7 @@ pub enum Button {
     Key(VirtualKeyCode),
 
     /// Scan code from keyboard, use this when the position of the key matters
-    /// more than letter on the key.
+    /// more than the letter on the key.
     ScanCode(u32),
 
     /// Mouse buttons
--- a/amethyst_input/src/input_handler.rs
+++ b/amethyst_input/src/input_handler.rs
@@ -644,11 +644,10 @@ mod tests {
         T: PartialEq<T> + Debug,
     {
         let mut ret = a.len() == b.len();
-        
+
         if ret {
-            let mut a = a.iter().collect::<Vec<_>>();
             let mut b = b.iter().collect::<Vec<_>>();
-            while let Some(a) = a.pop() {
+            for a in a.iter() {
                 if let Some(i) = b.iter().position(|b| a == *b) {
                     b.swap_remove(i);
                 } else {
@@ -656,7 +655,6 @@ mod tests {
                     break;
                 }
             }
-            assert_eq!(a.len(), 0);
             assert_eq!(b.len(), 0);
         };
         if !ret {
@@ -670,7 +668,11 @@ right: `{:?}`",
     }
 
     #[test]
-    fn action_response() {
+    fn key_action_response() {
+        // Register an action triggered by a key
+        // Press the key and check for a press event of both the key and the action.
+        // Release the key and check for a release event of both the key and the action.
+
         let mut handler = InputHandler::<String, String>::new();
         let mut events = EventChannel::<InputEvent<String>>::new();
         let mut reader = events.register_reader();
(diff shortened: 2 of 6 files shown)


Xaeroxe:
Note for me: [1, 1, 2, 3] and [1, 2, 2, 3] eval as equal in sets_equal. This should be corrected.

Xaeroxe:
Over 500 lines now. If we go much more that'll hamper the review process so I'm opening this up for review.

amethyst_input/src/input_handler.rs
```diff
@@ -628,3 +628,514 @@ where
...
+            let mut b = b.iter().collect::<Vec<_>>();
+            for a in a.iter() {
+                if let Some(i) = b.iter().position(|b| a == *b) {
+                    b.swap_remove(i);
+                } else {
+                    ret = false;
+                    break;
+                }
+            }
+            assert_eq!(b.len(), 0);
```
azriel91: Ah, this would prevent us from getting the nice error message below when there are unequal types &ndash; we'd get

```
left: `2` right: `0`
```

or similar. If you remove this line then we'd get what we should.

amethyst_input/src/button.rs
```diff
@@ -10,7 +10,7 @@ pub enum Button {
     Key(VirtualKeyCode),
 
     /// Scan code from keyboard, use this when the position of the key matters
```
fhaynes: What does position of the key mean, exactly?

amethyst_input/src/input_handler.rs
```diff
@@ -628,3 +628,514 @@ where
...
+            },
+        );
+        assert_eq!(handler.axis_value("test_axis"), Some(0.0));
+        handler.send_event(
+            &Event::WindowEvent {
+                window_id: unsafe { WindowId::dummy() },
+                event: WindowEvent::KeyboardInput {
+                    device_id: unsafe { DeviceId::dummy() },
+                    input: KeyboardInput {
+                        scancode: 112,
```
azriel91: do you want to make this `104` so as to not overlap with the `Down` key? same in line 1037, 1081, and 1103

amethyst_input/src/button.rs
azriel91: "geographical" location of the key (er, not sure a better word for this).
Like, on US keyboard layouts, there's `qwerty`. Some other keyboards might be `azerty`, but you want the top left key

amethyst_input/src/input_handler.rs
```diff
@@ -628,3 +628,514 @@ where
...
+        let mut events = EventChannel::<InputEvent<String>>::new();
+        let mut reader = events.register_reader();
+        handler.bindings.insert_action_binding(
+            String::from("test_key_action"),
+            [Button::Key(VirtualKeyCode::Up)].iter().cloned(),
+        );
+        assert_eq!(handler.action_is_down("test_key_action"), Some(false));
+        handler.send_event(
+            &Event::WindowEvent {
+                window_id: unsafe { WindowId::dummy() },
```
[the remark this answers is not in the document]
azriel91: yeaps, in real life, these are generated by `winit`

amethyst_input/src/input_handler.rs
udoprog: Agreed. Something like this is probably better:
```rust
assert_eq!(a.iter().collect::<HashSet<_>>(), b.iter().collect::<HashSet<_>>());
```

amethyst_input/src/input_handler.rs
```diff
@@ -628,3 +628,514 @@ where
...
+#[cfg(test)]
+mod tests {
+    use std::fmt::Debug;
+
+    use super::*;
+    use winit::{
+        DeviceId, ElementState, Event, KeyboardInput, ModifiersState, WindowEvent, WindowId,
+    };
+
+    /// Compares two sets for equality, but not the order
```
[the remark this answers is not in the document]
fhaynes: @udoprog I thought so, then someone chastised me =). If I understand you correctly, all comments should be `///`?

Xaeroxe:
@fhaynes @azriel91 @udoprog Ok I've incorporated review feedback. I think this should be ready for an r+, if any of you agree feel free to give it.

Issue #1249, opened 2018-12-13
Add ignored tests for the input handler


Xaeroxe:
We can't run these tests in a headless environment which means we can't run them on CI, but there is still value in being able to test the code locally where we can make a window.

Xaeroxe:
Just tried to add this, and we still have a problem in that we can't generate device IDs manually.  So, as soon as [this PR](https://github.com/tomaka/winit/pull/738) is merged and released to cargo then we can make unit tests for this and we don't even need to have them be "ignored".

azriel91:
thank you for enabling this 🙇!

PR #1344, opened 2019-01-11
Extracted some common code in input handler tests.

azriel91:
Issue #1249

* moved the `sets_are_equal` function down (so that the first thing you see is a `#[test]`)
* abstracted the event creation logic, should make it easier to see what the test is doing
* made a `HIDPI` constant (no longer have to check what `1.0` means)


Xaeroxe:
I found myself with some free time to add another unit test. Hope you don't mind me slipping a commit into this PR.

Xaeroxe:
Ok, I screwed that up a bit, for a bit, but I fixed it. Let's just agree Jake should stop pushing commits to other people's forks for a bit.

azriel91:
@Xaeroxe by all means, hijack this! 😄
I don't mind PR hijackings as long as we're doing the same thing ⚔️

Xaeroxe:
@azriel91 hijacked away :) I added some more unit tests.

azriel91:
✅ to @Xaeroxe's changes (can't approve a self-opened PR).

also,

> I don't mind if you hijack the car I'm driving, as long as you're a better driver and we're going to the same destination 😛

Xaeroxe:
Heh thanks for cleaning up my mess, though it seems we missed `ControllerButtonReleased` docs.

PR #1346, opened 2019-01-11
Separate initialization of `AudioBundle` from creation of `DjSystem`

### LINKED ISSUE #1322: [MAINTENANCE] Deprioritize DjSystem
## Note

This type of request is for things like `unwrap()`s that snuck into production. They aren't likely to cause problems and aren't really a bug, but should be updated. Other examples include formatting updates, dependency updates, or code simplification. 

## Description

Right now the `DjSystem` is front and center when I create an `AudioBundle`. I can apparently pass a no-op closure if I don't want to use it, but if I create an `AudioBundle`, I get it whether I want it or not. I think it should be optional. Just as display initialization doesn't make any huge assumptions about my configuration or what components I need, the audio system shouldn't saddle me with a music system I may not need.

## Reason

First, what is music?

 * Game music is almost certainly music.
 * Is the cockpit ambience of my spaceship also music? It shares many of the characteristics--loops infinitely, not spatialized--but the DjSystem apparently only handles a single track at once, so I'd essentially need to create a new system that duplicates many of `DjSystem`'s characteristics.
 * Is my non-spatialized weapon shot also music? It doesn't loop, but will probably also be non-spatialized so will need its own system. Maybe `DjSystem` could be expanded to handle non-localized non-looping sounds, but then it quickly becomes unwieldy and general.

Further, while this system seems to support the simplest of cases--playing a single music track on infinite loop and occasionally changing it--I feel like I'd quickly hit its limitations and have to write my own system for handling music, non-spatialized level ambience, etc. And, while such a system should probably exist and may even be a refinement of the current `DjSystem`, it probably shouldn't be so integral that it must be initialized with `AudioBundle::new()`.

I think that `DjSystem` should either be initialized independently, or should be initialized via another method on `AudioBundle`. It shouldn't be required. It may also need improvements, but I think such improvements are a separate discussion from whether or not I should be able to completely opt out of it and build my audio system up from emitters and listeners.

I plan on being a heavy user of the audio subsystem, and am happy to take this change on. I just want a temperature check before going ahead with it.

## Impact

My understanding is that the audio system isn't heavily used, and it definitely isn't documented, so the only impact would likely be updating examples and crate documentation.

ndarilek:
Fixes #1322

### DIFF
--- a/amethyst_audio/src/bundle.rs
+++ b/amethyst_audio/src/bundle.rs
@@ -30,17 +30,23 @@ use crate::{source::*, systems::DjSystem};
 pub struct AudioBundle<'a, F, R> {
     dep: &'a [&'a str],
     marker: PhantomData<R>,
-    picker: F,
+    picker: Option<F>,
 }
 
 impl<'a, F, R> AudioBundle<'a, F, R> {
-    /// Create a new DJ bundle
-    pub fn new(picker: F) -> Self {
+    /// Create a new audio bundle
+    pub fn new() -> Self {
         AudioBundle {
             dep: &[],
             marker: PhantomData,
-            picker,
+            picker: None,
         }
+    }
+
+    /// Initialize the DJ system
+    pub fn with_dj_system(mut self, picker: F) -> Self {
+        self.picker = Some(picker);
+        self
     }
 
     /// Set dependencies for the `DjSystem`
@@ -57,8 +63,8 @@ where
 {
     fn build(self, builder: &mut DispatcherBuilder<'a, 'b>) -> Result<()> {
         builder.add(Processor::<Source>::new(), "source_processor", &[]);
-        if default_output_device().is_some() {
-            builder.add(DjSystem::new(self.picker), "dj_system", self.dep);
+        if default_output_device().is_some() && self.picker.is_some() {
+            builder.add(DjSystem::new(self.picker.unwrap()), "dj_system", self.dep);
         }
         Ok(())
     }
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -19,12 +19,14 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 * `amethyst_renderer::Rgba` is now a `Component` that changes the color and transparency of the entity
 it is attached to. ([#1282])
 * `AutoFov` and `AutoFovSystem` to adjust horizontal FOV to screen aspect ratio. ([#1281])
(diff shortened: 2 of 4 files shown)


ndarilek:
One thing I'm not clear on, should this PR bump the crate minor version since it adds new features?

Xaeroxe:
It would actually bump the major version since this isn't a backwards compatible change.

However, we usually don't have contributors worry about that. Assessing release needs and timing is left to the team for when they perform the rlease.

Xaeroxe:
Doh, look at me I'm misusing my vocabulary.

major.minor.patch

If that's the scheme then yes we bump minor version.

ndarilek:
Got it. It's also pre-1.0, so while it technically *is* a breaking 
change, my understanding is that almost anything goes if the semver 
version is below 1.0.


But in any case, glad it isn't my responsibility. :)

ndarilek:
Fixed a minor issue in the changelog--left out the footnote link to the 
PR, which should now be added.


Also, sorry for the meta discussion in this PR. I'm a bit confused by 
how Amethyst is 0.10 but this crate is 0.5, yet I assume its changes are 
tracked in the main changelog. The changelog also lists fixes paired 
with the Amethyst version. If we bump this crate to, say, 0.6, then list 
this change paired with Amethyst 0.11, how do I know which version of 
amethyst_audio a change is associated with?


Or should amethyst_audio have the same version number as amethyst?


I get that this isn't my responsibility, but I'm making this fix because 
right now I don't know how to just use the audio system without the DJ 
system, so I'll need this for my games. Since the amethyst_audio version 
is different, it seems like its release cycle should be different too, 
but the changelog doesn't show individual components with separate versions.


Thanks.

azriel91:
**Re:versioning**
It's true it's not easy to track what version of each of the sub crates corresponds to which version of the main crate, since there is only one change log. Currently, how it goes in Amethyst is, a sub crate starts at 0.1, and increments by one minor each time Amethyst releases.
That's how it existed when I encountered Amethyst, and I just maintain status quo. Personally I go with "all sub crates use the same version as the main crate" for simplicity in managing versions.

---

**Re: PR**
I think the pong example (and tutorial) need to be updated -- sound may not play

ndarilek:
How else do they need to be upgraded? I ran the main Pong example with 
the diff as applied and got audio. Did I miss something?


Thanks for the explanation on versioning.

azriel91:
oh, i thought the tutorial had a "let's add music" part which hadn't been updated, but actually it doesn't yet exist 😅

PR #1359, opened 2019-01-16
`AudioListener` now retrieves `Output` as a resource, and implements `Default` for setting ear coordinates.

ndarilek:


Rhuagh:
I believe it's better to retrieve the `Output` optionally and keep the check in the system. That way audio output selection can be deferred from game initialisation if needed.

Rhuagh:
So that would be `Option<Read<'a, Output>>`.

ndarilek:
I'm not clear on what the practical difference would be. Here we perform 
the check, then fail fast if no device is available. I think what you're 
proposing is a map/unwrap on every `run` call which, while not heavy 
when isolated, might cost more if done again and again.


FWIW, #1361 makes some significant changes to how audio is initialized.


Anyhow, I appreciate your thoughts and explanations. I'd like to do more 
with the audio system over time, so they're helpful.

Rhuagh:
Unwrap is a nop after compile

ndarilek:
If someone is initializing an `AudioSystem`, that implies to me that 
they expect their game to run with sound. I'd think the burden is on the 
developer to not initialize an `AudioSystem` if sound is optional, 
either via command line flags, config files, etc. or to otherwise fail 
the initialization and not reach `run` to begin with.


Even if the unwrap is a no-op, it feels a bit odd to have to perform 
that check every single time the system runs. I feel like I 
inadvertently took things a bit out of order, and this PR should be 
evaluated concurrently with #1361. Maybe we can come up with a more 
robust initialization process in the system/bundle that lets developers 
handle lack of audio devices on initialization, while ensuring that 
`run` either never runs because there isn't audio hardware, or the 
device is initialized and added so the `Read` won't fail?

Rhuagh:
The problem is that bundles are applied before you can know if an audio device exists. There are also scenarios where the audio device will be configured later (during system setup or in a later state), but you want to use audio at some point, and hence want to add the audio bundle so you get the processor. Ultimately, we can't assume we know all these scenarios up front, so internally in the engine we need to design functionality to be fault tolerant to the furthest extent.

ndarilek:
Ah, gotcha, so the game as a whole can only know that audio is available 
when a system runs, not necessarily when a bundle is created? I wonder 
if that also invalidates my work in #1361, since I am adding audio 
device resources in bundle initialization.

Rhuagh:
Not sure tbh, I just think that if we have a potential failure scenario it's better to have good handling rather than just panic. That panic is likely not gonna occur on a developers machine, but in a user, and we'll have an angry dev screaming at us.

PR #1361, opened 2019-01-17
Refactor audio initialization to be bundle-centric.

ndarilek:
I haven't yet written the changelog entry for this because this is my biggest change so far, and I'd like feedback on its implementation so I'm not rewriting the changelog entries.

This change addresses the following points of confusion for me:

 * We have a thing called an `AudioSystem` which seems to only manage spatial audio. There's another thing called a `DjSystem` that manages background music.
 * Device initialization seems to happen outside of either system. It isn't enough to add a single bundle to get sound support, and to look to that bundle's entrypoints for guidance on how to do that. Instead, you initialize the device, and it isn't clear if I even *can* select something other than the default output device.
 * There is an `AudioBundle`, but it really doesn't seem to add very much value.

This PR gets rid of the separate functionality for initializing audio devices, moving it into `AudioBundle`. The `Default` implementation of this bundle initializes a default `AudioSystem` which, in turn, initializes the default audio device. Other devices can be passed to `AudioSystem::new(device)`, or to `AudioBundle::new(...)` which, in turn, propagates it to the system.

It's possible that `AudioSystem` should be broken into `SpatialAudioSystem` with `AudioSystem` only managing sound hardware more conveniently. It's also possible that `AudioSystem` may eventually manage both non-spatial and spatial audio. I'd like to eventually see a more seamless joining of spatial audio, non-spatial audio, and music. Technically, music is just non-spatial audio, as are some ambiences, UI sounds, etc. and the only difference is that you've got a `SoundEmitter` without a `Transform`.

But for now I'm just trying to get a handle on cleaning up initialization, and would like to punt potentially splitting systems to future work. It just makes sense to me that, just as the `RenderSystem` manages display devices, the `AudioSystem` should be the single source of truth for audio hardware, and the entrypoint for picking a device should be `AudioBundle`. It was the first place I looked when trying to figure things out.

Thanks.

### DIFF
--- a/amethyst_audio/src/bundle.rs
+++ b/amethyst_audio/src/bundle.rs
@@ -6,7 +6,11 @@ use amethyst_core::{
     specs::prelude::DispatcherBuilder,
 };
 
-use crate::source::*;
+use crate::{
+    systems::AudioSystem,
+    output::Output,
+    source::*
+};
 
 /// Audio bundle
 ///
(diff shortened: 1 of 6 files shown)


ndarilek:
Closing in favor of #1388.

PR #1363, opened 2019-01-17
Set width and height of Pong Paddles

tammoippen:
I am not sure, whether this is taken care somewhere else / at some different time, but the master version of the book already changed this (see line 182ff [here](https://github.com/amethyst/amethyst/blob/master/book/src/pong-tutorial/pong-tutorial-02.md)):

In the example, the paddle come with a dimension of 1x1:
```rust
impl Paddle {
    fn new(side: Side) -> Paddle {
        Paddle {
            side: side,
            width: 1.0,
            height: 1.0,
        }
    }
}
```

The `BounceSystem` later use these width / height to compute collisions ... so only exact hits in the center of the paddle give a collision.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -34,6 +34,7 @@ it is attached to. ([#1282])
 
 * Fixed the "json" feature for amethyst_assets. ([#1302])
 * Fixed default system font loading to accept uppercase extension ("TTF"). ([#1328])
+* Set width and height of Pong Paddles ([#1363])
 
 [#1114]: https://github.com/amethyst/amethyst/pull/1114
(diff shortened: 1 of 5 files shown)


azriel91:
CI is failing on Rust stable introducing warnings for certain things. @Xaeroxe is fixing them, have to get this rebased once that one's done. I suspect TeamCity passed because the stable toolchain likely hasn't been updated on those agents yet.

Xaeroxe:
Ok if you rebase this on current master it should pass CI now.

tammoippen:
@Xaeroxe rebased

azriel91:
bors r=Rhuagh,azriel91

PR #1367, opened 2019-01-19
Remove OpenCollective from README

Moxinilian:
OpenCollective changed the way their donation image integration renders, making it a more obnoxious "Donate" button. As we want to discourage donation through OpenCollective (higher fees, jankier taxes organization, less control), I believe we can remove it from the landing README.

Eventually, it could be useful to include it on the donation page on amethyst-engine.org, but at least after the user gets to see DonorBox.


Moxinilian:
This is what I meant by "Donate" sorry.

erlend-sh:
Lemme try contact them first to see if they could offer a paste-in snippet that doesn’t include the donation button.

fhaynes:
Merging this. My experience so far has been...challenging, and I think we need to discuss OC further.

PR #1383, opened 2019-01-30
[Fix] Ui performance on windows + nvidia gpu

jojolepro:
Fixes the issue where drawing text would take a really long amount of time.

This replaces the current version by the one on github. Once the github version of glyph-brush is published, another PR will be made to update to that version.

This PR should be merged as soon as possible. Thanks!


jojolepro:
bors r+

I'll take the blame if I break something ^u^

PR #1387, opened 2019-01-30
Initialize `AudioSystem` as part of `AudioBundle`.

ndarilek:
I think this was an oversight of the initial `DjSystem` removal. The Pong example wasn't even updated to add a separate `AudioSystem`, and it doesn't make sense to me that `AudioBundle` shouldn't do this automatically.


ndarilek:
I'm not sure if this needs a changelog entry. It doesn't change anything as such, just modifies what I expect was an oversight in removing `DjSystem` from `AudioBundle` initialization. As such, nothing in that change description is invalidated. Further, without this change, `AudioBundle` doesn't initialize an `AudioSystem`, which is a bit surprising to say the least. :) So I feel like a changelog entry that says "revert regression that never made it into a release" would be a bit confusing as well. Happy to write one if needed, though.

ndarilek:
Oopse, sorry, should be done now. Thanks for the heads-up.

PR #1388, opened 2019-01-31
Make audio initialization more bundle-centric

ndarilek:


ndarilek:
Just saw, and hopefully fixed, the build failure on this PR. Is there any way to get alerts when builds fail? Kind of surprised it happened silently, and I'd have fixed it days back if I'd gotten a notification.

ndarilek:
I'm now actually attempting to use this API. Previously I wasn't storing the `Output` as a `Resource` so sound wasn't playing. That's now fixed and a simple spatial sound now plays.

This is probably ready to merge, assuming Concourse passes it.

torkleyy:
Concourse failure seems spurious, but there‘s a changelog conflict.

amethyst_audio/src/output.rs
```diff
@@ -22,6 +22,14 @@ pub struct Output {
     pub(crate) device: Device,
 }
 
+impl Default for Output {
+    fn default() -> Self {
+        default_output_device()
+            .map(|re| Output { device: re })
+            .expect("No default output device")
```
torkleyy: When does this fail? If it actually fails, I‘m not sure if it should have a Default impl.

ndarilek:
I imagine it would fail if there is no sound hardware available. I 
understand the idea of not creating a `Default` implementation for 
something that can fail, but given that every device I own, including my 
$35 Raspberry Pis and Asus Tinkerboard has audio hardware, it seems like 
a bit of an edge case. If you're targeting your game at any modern 
platform, then having sound hardware (even if it's just embedded in the 
motherboard and is disconnected or not used) seems like a reasonable 
expectation. And if you're putting forth the effort to target 
non-standard platforms that may lack sound, then you're probably already 
working around special cases and have an escape hatch of not using 
`Default`.

ndarilek:
Sure thing, will do soon!

ndarilek:
All requested changes should be resolved. Thanks for the feedback!

Xaeroxe:
bors r=@Torkleyy, @Xaeroxe

PR #1389, opened 2019-02-01
Fixing incorrect yaw rotation in FreeRotationSystem

jchitel:
The way this was before, when you would try to turn left or right, it would appear to roll the camera instead of rotating the yaw. You could orient yourself differently to get it to work properly, but only when facing that direction.

This fix makes it so that the yaw rotation behaves as expected regardless of the current camera orientation.

### DIFF
--- a/amethyst_controls/src/systems.rs
+++ b/amethyst_controls/src/systems.rs
@@ -170,7 +170,7 @@ where
                     if let DeviceEvent::MouseMotion { delta: (x, y) } = *event {
                         for (transform, _) in (&mut transform, &tag).join() {
                             transform.pitch_local((-y as f32 * self.sensitivity_y).to_radians());
-                            transform.yaw_global((-x as f32 * self.sensitivity_x).to_radians());
+                            transform.yaw_local((-x as f32 * self.sensitivity_x).to_radians());
                         }
                     }
                 }


jojolepro:
I assume you have issues because the camera entity is a child of some other entity? This type of camera is meant for debugging, and shouldn't be use as a child of another entity (which normally would be the physics body). I have controllers that do that in the amethyst-extra crate (github unpublished version) if you need then. I'll eventually add them in amethyst, once we stabilize physics.

jojolepro:
Is the fly_camera example working for you?

JameeKim:
I'd say that the behavior is intended. If the camera yaws locally when the mouse moves left or right, the camera angle is going to be screwed up and unfixable after a while. For example, consider a situation where you move the mouse up, right, then down. Then even if the camera comes back to the original position, it would be tilted clockwise. If the camera yaws globally, then it might seem to roll when it is pointing straight up or down, but it is guaranteed that it remains untilted however it rotates. I think it would have been better if the FreeRotation thing had constraints on the pitch value so that the camera doesn't point too much up or too much down.

jojolepro:
Ah okay that way it makes sense. I'll test it out when I get time.

JameeKim:
Oh and about the renderable example, the camera there is not in its default rotation. What I mean by default rotation is +y upward and -z forward, but in the example the upward direction is set to +z as far as I remember. I also had some trouble understanding the axes directions conventions with the renderable example. Maybe it needs a fix...

jojolepro:
Yeah it probably does

robert-w-gries:
@JameeKim Is it suggested to remove the `rotation` value from the renderable prefab file so that we can force the "default" rotation?  You can find the prefab's camera entity definition [here](https://github.com/amethyst/amethyst/blob/master/examples/assets/prefab/renderable.ron#L117-L128).

I still observed strange camera behavior after removing the `rotation` value, but I assume it's because the `FreeRotation` system should be using quaternion rotation instead of `yaw_global` and `pitch_local`.  It looks like the correct code is in @jojolepro's extras crate [here](https://github.com/jojolepro/amethyst-extra/blob/master/src/lib.rs#L951-L966):

```rust
// Camera
if let Some(tr) = transforms.get_mut(entity) {
    *tr.rotation_mut() = UnitQuaternion::from_euler_angles(
        rotation_control.mouse_accum_y,
        0.0,
        0.0,
    );
}

// Player collider
if let Some(tr) = transforms.get_mut(parent.entity) {
    *tr.rotation_mut() = UnitQuaternion::from_euler_angles(
        0.0,
        rotation_control.mouse_accum_x,
        0.0,
    );
}
```

JameeKim:
@robert-w-gries The `FreeRotationSystem` is working correctly as long as its original intention is considered and except for some edge cases. The reason the renderable example is still not working is because all the meshes are still having +z as their upward direction. Though I think the name "free rotation" is quite misleading. What it is currently doing is appropriate for a character walking/running on a surface, so it is constrained in some way. If you are making a first-person shooting game where the player freely floats in the space, for example, then it would make more sense to use `yaw_local` rather than the global one. But in that case it would also be necessary to provide a way to roll the camera for the player, bc a truly *free* rotation system tends to mess up the roll angle.

jojolepro:
Yeah you are right, the name is misleading. Its more of a "as unconstrained as possible, while blocking the camera from rolling."

I think we should make the free rotation system to be roll enabled (this pr), make a copy of the original one with roll disabled as FlyCameraRotation.

Does this make sense?

JameeKim:
Yes, it would be a bit of a breaking change, but it will significantly reduce the confusion.

jchitel:
Ok, I now understand what's going on here, and I think the system is actually fine the way it is (i.e. before this PR). The only "issue" is that it depends on the camera being in a specific initial orientation relative to the world. Assuming the scene you use with the fly control is oriented correctly, it will work as expected.

Now that I've been playing around with the system as implemented with this PR, I don't think that such a "free" rotation system is at all practical. Even if you're floating in space, you'd expect the left-right rotation to be fixed relative to the user's current "up" orientation, so that the camera *doesn't* roll when you move around.

Perhaps what we can do is just add the ability to explicitly roll (using keys like Q and E for example) and leave the yaw rotation as yaw_global(). Does that sound good?

JameeKim:
@jchitel Take a look at [this repo](/JameeKim/floating-in-space) that I created. What I mean by "space" in the phrase "floating in space" is the actual space, or universe, where there is no absolute guidance on which direction is up or down.

> Even if you're floating in space, you'd expect the left-right rotation to be fixed relative to the user's current "up" orientation, so that the camera _doesn't_ roll when you move around.

I agree, but the problem arises when the player character is not having its up direction line up with the world's up direction. One way to solve the problem would be to make the camera entity a child of the player character entity, but that approach too has its own drawbacks.

fhaynes:
Closing this due to drift. Please resolve the conflicts and re-open if still relevant.

PR #1390, opened 2019-02-01
Implement widget structs and traits

happenslol:
This is up for preliminary review, as I'm still very unsure about the specific implementation.

Basically this aims to do the following:
* Implement structs for each widget type which contain all entities belonging to the widget, effectively "tracking" the parts of the widget,
* Provide a way to easily see which components a widget definitely or potentially has, and get each of them, as well as
* Have an easy way to compose widget from different entities with components.

As an example, I've implemented these things for `UiButton` for now, to see how viable my solutions are. Please see the `widgets.rs` module as well as the `button/mod.rs` file, I've added comments with explanations.

Here's what's still missing to complete this PR:
* [x] Fix a weird generic/lifetime error in the builders for `UiLabel` and `UiWidget`
* [x] Implement `Widget`/`WidgetId` support for prefabs
* [x] Implement custom derive for `WidgetId`
* [x] ~Add examples and documentation for using `Widgets`~

### DIFF
--- a/amethyst_derive/src/lib.rs
+++ b/amethyst_derive/src/lib.rs
@@ -8,6 +8,7 @@ use syn::{parse_macro_input, DeriveInput};
 
 mod event_reader;
 mod prefab_data;
+mod widget_id;
 
 #[proc_macro_derive(EventReader, attributes(reader))]
 pub fn event_reader_derive(input: TokenStream) -> TokenStream {
@@ -23,5 +24,17 @@ pub fn event_reader_derive(input: TokenStream) -> TokenStream {
 pub fn prefab_data_derive(input: TokenStream) -> TokenStream {
     let ast = parse_macro_input!(input as DeriveInput);
     let gen = prefab_data::impl_prefab_data(&ast);
+    gen.into()
+}
+
+/// This allows the use of an enum as an ID for the `Widgets` resource. One
+/// variant has to be marked as the default variant with `#[widget_id_default]
+/// and will be used when a `Widget` is added to the resource without an
+/// explicit ID. Note that when using `Widgets::add`, this will overwrite
+/// an existing widget with the same default id!
+#[proc_macro_derive(WidgetId, attributes(widget_id))]
+pub fn widget_id_derive(input: TokenStream) -> TokenStream {
+    let ast = parse_macro_input!(input as DeriveInput);
+    let gen = widget_id::impl_widget_id(&ast);
     gen.into()
 }
--- a/amethyst_ui/src/button/builder.rs
+++ b/amethyst_ui/src/button/builder.rs
@@ -5,7 +5,7 @@ use smallvec::{smallvec, SmallVec};
 use amethyst_assets::{AssetStorage, Loader};
 use amethyst_audio::SourceHandle;
 use amethyst_core::{
-    ecs::prelude::{Entities, Entity, Read, ReadExpect, World, WriteStorage},
+    ecs::prelude::{Entities, Entity, Read, ReadExpect, World, WriteExpect, WriteStorage},
     Parent,
 };
 use amethyst_renderer::{Texture, TextureHandle};
@@ -15,7 +15,7 @@ use crate::{
     Anchor, FontAsset, FontHandle, Interactable, Selectable, Stretch, UiButton, UiButtonAction,
     UiButtonActionRetrigger,
     UiButtonActionType::{self, *},
-    UiPlaySoundAction, UiSoundRetrigger, UiText, UiTransform,
+    UiPlaySoundAction, UiSoundRetrigger, UiText, UiTransform, WidgetId, Widgets,
 };
 
 use std::marker::PhantomData;
@@ -29,7 +29,7 @@ const DEFAULT_TXT_COLOR: [f32; 4] = [0.0, 0.0, 0.0, 1.0];
 
 /// Container for all the resources the builder needs to make a new UiButton.
 #[derive(SystemData)]
-pub struct UiButtonBuilderResources<'a, G: PartialEq + Send + Sync + 'static> {
+pub struct UiButtonBuilderResources<'a, G: PartialEq + Send + Sync + 'static, I: WidgetId = u32> {
     font_asset: Read<'a, AssetStorage<FontAsset>>,
     texture_asset: Read<'a, AssetStorage<Texture>>,
     loader: ReadExpect<'a, Loader>,
@@ -39,7 +39,7 @@ pub struct UiButtonBuilderResources<'a, G: PartialEq + Send + Sync + 'static> {
     parent: WriteStorage<'a, Parent>,
     text: WriteStorage<'a, UiText>,
     transform: WriteStorage<'a, UiTransform>,
-    button: WriteStorage<'a, UiButton>,
+    button_widgets: WriteExpect<'a, Widgets<UiButton, I>>,
     sound_retrigger: WriteStorage<'a, UiSoundRetrigger>,
     button_action_retrigger: WriteStorage<'a, UiButtonActionRetrigger>,
     selectables: WriteStorage<'a, Selectable<G>>,
@@ -47,8 +47,8 @@ pub struct UiButtonBuilderResources<'a, G: PartialEq + Send + Sync + 'static> {
 
 /// Convenience structure for building a button
 #[derive(Debug, Clone)]
-pub struct UiButtonBuilder<G> {
-    name: String,
+pub struct UiButtonBuilder<G, I: WidgetId> {
+    id: Option<I>,
     x: f32,
     y: f32,
     z: f32,
@@ -75,10 +75,13 @@ pub struct UiButtonBuilder<G> {
     _phantom: PhantomData<G>,
 }
 
-impl<G> Default for UiButtonBuilder<G> {
+impl<G, I> Default for UiButtonBuilder<G, I>
+where
+    I: WidgetId,
+{
     fn default() -> Self {
         UiButtonBuilder {
-            name: "".to_string(),
+            id: None,
             x: 0.,
             y: 0.,
             z: DEFAULT_Z,
@@ -105,15 +108,22 @@ impl<G> Default for UiButtonBuilder<G> {
     }
 }
 
-impl<G: PartialEq + Send + Sync + 'static> UiButtonBuilder<G> {
+impl<'a, G: PartialEq + Send + Sync + 'static, I: WidgetId> UiButtonBuilder<G, I> {
     /// Construct a new UiButtonBuilder.
     /// This allows easy use of default values for text and button appearance and allows the user
-    /// to easily set other UI-related options.
-    pub fn new<N: ToString, S: ToString>(name: N, text: S) -> UiButtonBuilder<G> {
+    /// to easily set other UI-related options. It also allows easy retrieval and updating through
+    /// the appropriate widgets resouce, see [`Widgets`](../../struct.Widgets.html).
+    pub fn new<S: ToString>(text: S) -> UiButtonBuilder<G, I> {
         let mut builder = UiButtonBuilder::default();
-        builder.name = name.to_string();
         builder.text = text.to_string();
         builder
+    }
+
+    /// Sets an ID for this widget. The type of this ID will determine which `Widgets`
+    /// resource this widget will be added to, see see [`Widgets`](../../struct.Widgets.html).
+    pub fn with_id(mut self, id: I) -> Self {
+        self.id = Some(id);
+        self
     }
 
     /// Add a parent to the button.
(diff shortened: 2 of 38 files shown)


amethyst_ui/src/button/mod.rs
```diff
@@ -2,35 +2,66 @@ mod actions;
...
+use amethyst_renderer::TextureHandle;
 
 /// A clickable button, this must be paired with a `TextureHandle`
 /// and this entity must have a child entity with a `UiText`.
-#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+#[derive(Debug, Clone)]
 pub struct UiButton {
-    /// Default text color
-    text_color: [f32; 4],
+    name: String,
```
jojolepro: If the `Widgets` resource is a HashMap, you don't need to have the key here (name). Also, not having the key here allows to not require generics on UiButton.

UiButton<K> { id: K, image_entity.....}

amethyst_ui/src/button/mod.rs
```diff
@@ -2,35 +2,66 @@ mod actions;
...
+    }
 }
+
+// This allows us to define what components can be attached to a widget.
+// We get better typing as well as autocompletion here, and the system code
+// will look a lot cleaner if it uses the get/get_maybe methods.
+
+// The one big problem I see here is that due to trait typing, it's not possible
+// for sub-entities to have these methods implemented for them.
+// I.e. you can either implement WidgetHasComponent<UiButton, UiTransform> to reveice
```
jojolepro: reveice -> receive

amethyst_ui/src/button/text_color.rs
```diff
@@ -0,0 +1,9 @@
+use amethyst_core::specs::prelude::{Component, DenseVecStorage};
+
+pub struct UiButtonTextColor {
+    pub text_color: [f32; 4],
```
jojolepro: Technically this information is already available in the text entity, in the UiText component. Is this needed?

amethyst_ui/src/widgets.rs
```diff
@@ -0,0 +1,98 @@
+use amethyst_core::specs::prelude::{Component, ReadStorage, WriteStorage};
+use derivative::Derivative;
+use std::{
+    ops::Deref,
+    collections::HashMap,
+};
+
+pub trait Widget {
+    fn get_id(&self) -> &str;
```
jojolepro: As explained in another comment, this shouldn't be needed.

amethyst_ui/src/widgets.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+#[derive(Derivative)]
+#[derivative(Default(bound = ""))]
+pub struct Widgets<T: Widget>(HashMap<String, T>);
```
jojolepro: type Widgets<K, W> = HashMap<K, W>;
type ButtonWidgets<K> = Widgets<K, UiButton>;

jojolepro:
I'd like to merge this soon. Do you have time to work on it? If not, I can probably spend some time on it.

amethyst_ui/src/button/builder.rs
```diff
@@ -362,15 +364,29 @@ impl<G: PartialEq + Send + Sync + 'static> UiButtonBuilder<G> {
...
             .insert(
                 image_entity,
-                UiButton {
+                UiButtonTextColor {
                     text_color: self.text_color,
                 },
             )
             .expect("Unreachable: Inserting newly created entity");
 
+        let widget_id = if self.name.len() > 0 {
```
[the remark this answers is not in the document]
happenslol: So in this case, does every entity _need_ an `EntityId` that's user defined? Or can the user choose to omit it, in which case it will be auto generated? For now, I'll fall back on using the entity handle as the identifier, we can replace that with the `EntityId` component as soon as that's done.

amethyst_ui/src/button/mod.rs
```diff
@@ -2,35 +2,66 @@ mod actions;
...
+            image_entity,
+            text_entity,
+        }
     }
 }
 
-impl Component for UiButton {
-    type Storage = DenseVecStorage<Self>;
+impl Widget for UiButton {
+    fn get_id(&self) -> &str {
```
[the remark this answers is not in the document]
happenslol: See comment below

amethyst_ui/src/button/text_color.rs
happenslol: The `UiButton` component had this information previously, so I moved it into its own component since the `UiButton` is now a `Widget` and not a `Component` anymore. I can remove it if it's not needed?

amethyst_ui/src/widgets.rs
happenslol: I've changed this so that the widget id type is now an associated type on `Widget`. The problem with this approach is the following:
Imagine there is an external library which adds lots of UI components. They define ButtonWidgets as `Widgets<String, UiButton>`. Now you have already created a few of your own button widgets before, but you used them as `Widgets<u32, UiButton>`. Now there are 2 instances, and it's possible that libraries conflict between each other or with your code in this way.

amethyst_ui/src/button/builder.rs
```diff
@@ -247,9 +252,20 @@ impl<G: PartialEq + Send + Sync + 'static> UiButtonBuilder<G> {
...
     /// Build this with the `UiButtonBuilderResources`.
-    pub fn build(mut self, mut res: UiButtonBuilderResources<'_, G>) -> Entity {
-        let mut id = self.name.clone();
+    pub fn build(mut self, mut res: UiButtonBuilderResources<'a, G, I>) -> (I, UiButton) {
         let image_entity = res.entities.create();
+        let text_entity = res.entities.create();
+        let widget = UiButton::new(image_entity, text_entity);
+
+        let id = {
+            let widget = widget.clone();
```
jojolepro: This method is double cloning when self.id is Some

amethyst_ui/src/button/builder.rs
```diff
@@ -39,16 +40,16 @@ pub struct UiButtonBuilderResources<'a, G: PartialEq + Send + Sync + 'static> {
...
     sound_retrigger: WriteStorage<'a, UiSoundRetrigger>,
     button_action_retrigger: WriteStorage<'a, UiButtonActionRetrigger>,
     selectables: WriteStorage<'a, Selectable<G>>,
 }
 
 /// Convenience structure for building a button
 #[derive(Debug, Clone)]
-pub struct UiButtonBuilder<G> {
-    name: String,
+pub struct UiButtonBuilder<G, I: WidgetId> {
```
jojolepro: I thought the point of using predefined implementations with strings and u32 was to avoid adding a bound like this on the builder by hardcoding a u32. Is it just for automatic generation in this case?

amethyst_ui/src/button/mod.rs
```diff
@@ -9,28 +9,22 @@ pub use self::{
...
+use amethyst_renderer::TextureHandle;
 
-/// A clickable button, this must be paired with a `TextureHandle`
-/// and this entity must have a child entity with a `UiText`.
-#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
-pub struct UiButton {
-    /// Default text color
-    text_color: [f32; 4],
-}
+define_widget!(UiButton =>
```
jojolepro: I like this a lot :)

amethyst_ui/src/button/builder.rs
happenslol: Yes, this makes sure that when the user doesn't provide any ids at all in his prefab (also important for backwards compat), he will automatically get `u32`s and can still iterate over the `Widgets` resource.

happenslol:
I'm done with this. I've decided against making an example and a book chapter for this PR, since it would probably completely blow up the scope - and the widgets trait/resource is not implemented for every Widget anyways yet. Imo, those can be done in smaller work packages since this sets up most of the needed groundwork.

amethyst_ui/src/prefab.rs
```diff
@@ -359,6 +356,7 @@ where
 /// - `TF`: `Format` used for loading `Texture`s
 /// - `AF`: `Format` used for loading sounds
 /// - `FF`: `Format` used for loading fonts
+/// - `WID`: Type used for Widget IDs
```
jojolepro: Why not just W? ;)

happenslol: Was trying to keep it consistent with the others, I have no idea why they are different from all others either tbh =P I'll change them all to be in line with the rest of the file.

amethyst_derive/src/lib.rs
```diff
@@ -25,3 +26,10 @@ pub fn prefab_data_derive(input: TokenStream) -> TokenStream {
     let gen = prefab_data::impl_prefab_data(&ast);
     gen.into()
 }
+
+#[proc_macro_derive(WidgetId, attributes(widget_id))]
+pub fn widget_id_derive(input: TokenStream) -> TokenStream {
```
LucioFranco: are there docs around this macro?

docs/CHANGELOG.md
```diff
@@ -50,12 +51,13 @@ it is attached to. ([#1282])
 * `FrameRateLimitConfig` has a `new` constructor, and its fields are made public. ([#1436])
 * Derive `Deserialize, Serialize` for `MaterialPrimitive` and `SpriteRenderPrimitive`, remove
 extra bounds from `AnimatablePrefab` and `AnimationSetPrefab` ([#1435])
-* Renamed `amethyst_core::specs` to `amethyst_core::ecs` and `amethyst_core::nalgebra` to `amethyst_core::math`. ([#1410])
+* Renamed `amethyst_core::ecs` to `amethyst_core::ecs` and `amethyst_core::nalgebra` to `amethyst_core::math`. ([#1410])
```
torkleyy: "Renamed `amethyst_core::ecs` to `amethyst_core::ecs`"?

amethyst_ui/src/widgets.rs
```diff
@@ -0,0 +1,208 @@
...
+    /// for it.
+    pub fn add(&mut self, widget: T) -> I {
+        let id = I::generate(&self.last_key);
+        self.items.insert(id.clone(), widget);
+        id
+    }
+
+    /// Adds a widget with a specified ID. This will overwrite any
+    /// existing widget of the same type with the same ID.
+    pub fn add_with_id(&mut self, id: I, widget: T) -> I {
```
torkleyy: Since the ID is passed by the user, I'd return `Option<T>` here so you can check if it did overwrite the entry.

amethyst_ui/src/widgets.rs
```diff
@@ -0,0 +1,208 @@
...
+        self.items.insert(id.clone(), widget);
+        id
+    }
+
+    /// Mutably retrieves a widget by its ID.
+    pub fn get_mut(&mut self, id: I) -> Option<&mut T> {
```
torkleyy: Why is there no immutable variant?

docs/CHANGELOG.md
happenslol: oof, seems to be a squash messup.

amethyst_ui/src/widgets.rs
happenslol: So, your suggestion is that this returns `(I, Option<T>)` and that `Option<T>` will be `Some(widget)` containing the overwritten widget if it was already there? I feel like that leads to a weird pattern of users always feeling like they need to check the return value... Maybe we can add another method that does `add_or_replace` which can have your suggested signature?

amethyst_ui/src/widgets.rs
happenslol: The immutable variant can be retrieved through the `Index` trait. Now that you mention it though, that interface is a little different.. I'll add one.

amethyst_ui/src/widgets.rs
torkleyy: No, I think it should _only_ return `Option<T>`, since the user can clone the ID themselves.

happenslol: Ah, I see. To be honest I would prefer keeping the interface as-is, so that it stays the same as `add`, and also since this consumes the ID I think this would be easier to use. Would you be OK with my above suggestion of adding an explicit non-overwrite method?

amethyst_derive/src/lib.rs
happenslol: Not yet. My next PR will focus on providing detailed docs and examples on how to use Widgets (and probably even a small section in the book), as well as implementing them for every existing widget, so I wanted to put it off until then. If this has a high priority though, I can add a short description.

happenslol: Added a short description.

amethyst_ui/src/widgets.rs
torkleyy: Yeah that makes sense. Hmm, but that doesn't cover the case where you do want to overwrite, but acquire ownership over the old widget, does it?

I'm also happy to leave this up to a future PR as written in the TODO comment.

happenslol: Nevermind, now that I think about it, doing it your way mimics the behavior of `HashMap` and is probably more familiar to most people. And there's no way to avoid a clone with my method if the user doesn't care about keeping the ID, so yours is more flexible.

happenslol:
bors retry

happenslol:
bors retry

jojolepro:
bors retry

jojolepro:
image

happenslol:
bors retry

PR #1392, opened 2019-02-03
Store MeshData when loading objects

seppo0010:
I'm not sure if this is the right approach, but I'm opening the PR to get some feedback. I wanted to get the MeshData from a mesh obj file after loading it. gltf sees to have this implemented.


torkleyy:
This doesn't seem like the right way to do it. Can't we expose the mesh data at an earlier stage and allow to just load the data (without uploading to the GPU)?

torkleyy:
You should be able to just use `Format::import` to get the data. If you still want to upload it afterwards, `Loader::load_from_data` can be used.

PR #1400, opened 2019-02-07
Remove unnecessary string allocs

Xaeroxe:
## Description

`.expect(format!(...))` is a pattern shown to have adverse performance impacts due to string allocations that are only used on a cold path. This PR removes those string allocations and retains the error messages.

## Additions

- None

## Removals

- None

## Modifications

- Slight performance improvement

## PR Checklist

- [ ] **N/A** Updated the content of the book if this PR would make the book outdated.
- [ ] **N/A** Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.
- [ ] **N/A** Added unit tests for new APIs if any were added in this PR.

### DIFF
--- a/amethyst_derive/src/event_reader.rs
+++ b/amethyst_derive/src/event_reader.rs
@@ -30,8 +30,9 @@ pub fn impl_event_reader(ast: &DeriveInput) -> TokenStream {
         };
     }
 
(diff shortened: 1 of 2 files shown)


Xaeroxe:
bors r=@Azriel91

PR #1404, opened 2019-02-08
Reduce logging verbosity

### LINKED ISSUE #1403: [MAINTENANCE] Increase default log level of child crates to `warn`
## Description

Users commonly complain about their logs being polluted with information such as `[INFO][gfx_device_gl::factory] Created one zillion buffers`. We should default these log levels to a `warn` level when they are not set manually by the user so that logging done by the user and more important log info such as warnings and errors are more immediately apparent.

## Implementation

This change would likely need to be made in the `src/logger.rs` file. At first glance, of note would be the `LoggerConfig` type, its `Default` impl, and the `env_var_override` function. #1316 is relevant here since this relates to the logging environment variables.

Xaeroxe:
## Description

Fixes #1403 

This crate raises the default log level to "info" and by default filters out info logs from the crate "gfx_device_gl" which is... quite verbose, to say the least.

## Additions

- Added LoggerConfig::log_gfx_device_level

## Removals

- None

## Modifications

- Set default log level to Info

## PR Checklist

- [ ] Added unit tests for new APIs if any were added in this PR. (No great way to unit test as I'm not gfx_device_gl)

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -27,15 +27,18 @@ pub struct LoggerConfig {
     pub log_file: Option<PathBuf>,
     /// If set, allows the config values to be overriden via the corresponding environmental variables.
     pub allow_env_override: bool,
+    /// If false gfx_device_gl won't be logged.
+    pub log_gfx_device: bool,
 }
 
 impl Default for LoggerConfig {
     fn default() -> LoggerConfig {
         LoggerConfig {
             stdout: StdoutLog::Colored,
-            level_filter: LevelFilter::Debug,
+            level_filter: LevelFilter::Info,
             log_file: None,
             allow_env_override: true,
+            log_gfx_device: false,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/logger.rs
```diff
@@ -86,6 +89,12 @@ impl Logger {
             StdoutLog::Off => {}
         }
 
+        if !config.log_gfx_device {
+            logger.dispatch = logger
+                .dispatch
+                .filter(|m| !m.target().starts_with("gfx_device_gl"));
```
distransient: This is quite nuclear, couldn't you do perhaps
```rust
pub gfx_device_level: Option<LevelFilter>,
```
in the struct with a default of `Warn` and then
```rust
if let Some(level) = config.gfx_device_level {
    logger.dispatch = logger.dispatch.level_for("gfx_device_gl", level);
}
```
here?

Xaeroxe: That's way better, thanks!

Xaeroxe:
bors r=@distransient

PR #1405, opened 2019-02-08
Allow setting a window icon programatically

seppo0010:
## Description

Setting a window icon from a file path in the config file is already
allowed. Now the icon can be set from an image generated
programatically, for example in the new example a red square is painted.

image

## Additions

- Add `loaded_icon` to `DisplayConfig`. This property must be set programatically if wanted, cannot be set in a ron file.

## PR Checklist


Xaeroxe:
bors r=@moxinilian, @jojolepro

PR #1406, opened 2019-02-08
Remove non-opengl stubs and fix dpi on OSX

Xaeroxe:
## Description

Fix a DPI issue on initialization and remove stubs for rendering backends we don't plan on supporting.

## Additions

- None

## Removals

- Removed stubs for backends we don't support. Weren't exposed anyways.

## Modifications

- Fix window initialization for renderer where DPI != 1.

## PR Checklist

- [ ] Added unit tests for new APIs if any were added in this PR. (Non-trivial to unit test)


jojolepro:
Did you ask omni if those stubs can be useful to them once they integrate rendy?

Xaeroxe:
Nope, so let's ask @omni-viral

Xaeroxe:
Since the fix doesn't actually work this PR doesn't provide much, so I'm going to close it.

PR #1411, opened 2019-02-12
Moved gltf from dev-only workspace

jojolepro:
## Description

It was in a separated workspace to cut down on build times.
However, the way it was done was inconvenient, because you had to manually import it separately from the main amethyst crate in external projects. Also it had some crate syncing issues when using git-based dependencies during development and testing.

## Modifications

- Moved amethyst_gltf from development workspace to be like the other amethyst_* subcrates and re-exported by amethyst as amethyst::gltf.

## PR Checklist

By placing an x in the boxes I certify that I have:

- [X] Ran `cargo test --all` locally if this modified any rs files.
- N/A Ran `cargo +stable fmt --all` locally if this modified any rs files.
- N/A Updated the content of the book if this PR would make the book outdated.
- [X] Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.
- N/A Added unit tests for new APIs if any were added in this PR.
- [X] Acknowledged that by making this pull request I release this code under an MIT/Apache 2.0 dual licensing scheme.

PR #1412, opened 2019-02-12
Basic Feature Gates

jojolepro:
## Description

This adds optional feature gates which can be used to cut compilation time down.
More will be added in the future (those that require more work).

## Additions

- Optional feature gates which will reduce compilation times when used.

## PR Checklist

- N/A Updated the content of the book if this PR would make the book outdated.

### DIFF
--- a/.concourse/tasks/linux-test.yml
+++ b/.concourse/tasks/linux-test.yml
@@ -16,9 +16,9 @@ run:
   - |
     rustc --version
     export RUSTFLAGS="-D warnings"
-    cargo check --all --features sdl_controller,profiler
+    cargo check --all --features sdl_controller,profiler,gltf
     cargo check --all --no-default-features
     export CARGO_TARGET_DIR="../target"
-    cargo test --all
+    cargo test --all --all-features
 caches:
 - path: target
--- a/.concourse/tasks/linux-test.yml
+++ b/.concourse/tasks/linux-test.yml
@@ -19,6 +19,6 @@ run:
     cargo check --all --features sdl_controller,profiler,gltf
     cargo check --all --no-default-features
     export CARGO_TARGET_DIR="../target"
-    cargo test --all --all-features
+    cargo test --all --features gltf
 caches:
 - path: target
--- a/.concourse/tasks/linux-test.yml
+++ b/.concourse/tasks/linux-test.yml
@@ -17,6 +17,7 @@ run:
     rustc --version
     export RUSTFLAGS="-D warnings"
     cargo check --all --features sdl_controller,profiler
+    cargo check --all --no-default-features
     export CARGO_TARGET_DIR="../target"
     cargo test --all
 caches:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,25 @@ license = "MIT/Apache-2.0"
 travis-ci = { repository = "amethyst/amethyst", branch = "master" }
 
 [features]
+default = ["animation", "audio", "gltf", "locale", "network"]
+
+animation = [
+    "amethyst_animation"
+]
+audio = [
+    "amethyst_audio"
+]
+gltf = [
+    "amethyst_gltf",
+    "amethyst_animation"
+]
+locale = [
+    "amethyst_locale"
+]
+network = [
+    "amethyst_network"
+]
+
 profiler = [
     "thread_profiler",
     "thread_profiler/thread_profiler",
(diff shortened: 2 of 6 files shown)


khionu:
Could UI be added to this? I would say it's of the same priority as audio

minecrawler:
I would even add the renderer and input. That way, Amethyst could be run in headless mode, which is exactly what we need for servers or bots.

khionu:
I think that would require more discussion on the purpose of the `amethyst` crate. As is, I was under the assumption that it was meant to be a quick import for a game, rather than needing to specify all the individual crates.

minecrawler:
Then you are questioning this whole PR?

docs/CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ it is attached to. ([#1282])
 * Add `icon` to `DisplayConfig` to set a window icon using a path to a file ([#1373])
 * Added setting to control gfx_device_gl logging level separately, and set it to Warn by default. ([#1404])
 * Add `loaded_icon` to `DisplayConfig` to set a window icon programatically ([#1405])
+* Added optional feature gates which will reduce compilation times when used. ([#1412])
```
torkleyy: Can we change this to "which will reduce compilation time when disabled"?

khionu:
No, I'm not questioning the whole PR. Adding the renderer as optional changes the scope of `amethyst` from "wrapper to help get going" to "the entry point that everything could use". If my understanding of the scope is wrong, then I would go further to say everything should be made optional and the crate refactored to ensure that that's possible. If my understanding is correct, the renderer would be a bit much to add to this PR.

Either way, this PR could be merged without your suggestion, and it could be moved to a follow up discussion/PR.

torkleyy:
One request: can we plase make `gltf` optional? It's known to take very long to compile and I don't think it was enabled before #1411.

fhaynes:
If we're going to do feature gates, please add renderer. Headless servers are a must and need to be in as soon as possible.

fhaynes:
We can do that in a separate PR though

jojolepro:
Ui and rendering: its more complex than the other ones, and they don't do anything if you are not using the bundles. Headless servers are already a thing this way.

I might try to add feature gates for those at a later time.

Also if I make gltf optional it won't get any testing from people using the cargo test commands.

jojolepro:
Actually, if someone is up for it, feel free to open a new pull request after this one for input, ui and rendering. I'll continue to go through my todo list instead of going super deep in each element.

torkleyy:
> Also if I make gltf optional it won't get any testing from people using the cargo test commands.

What about `cargo test --all`? Or `cargo test --all --all-features` (which should be used on CI)?

jojolepro:
cargo test --all --all-features
works, but we have to update the pull request template so people don't forget. ;)

torkleyy:
Accidentally? We want to test everything ;)

jojolepro:
I assume most devs won't have installed the dependencies required for sdl testing.

torkleyy:
I don't really see your point, sorry. If somebody wants to test everything, they just install the dependencies. However, testing is done by CI anyway. There is nothing `gltf` does that we couldn't test on CI.

And if that was the case, you could still do `cargo test --features gltf`.

jojolepro:
we'll see if it works

torkleyy:
Ah, nightly feature seems to be an issue :(

.concourse/tasks/linux-test.yml
```diff
@@ -16,8 +16,9 @@ run:
   - |
     rustc --version
     export RUSTFLAGS="-D warnings"
-    cargo check --all --features sdl_controller,profiler
+    cargo check --all --features sdl_controller,profiler,gltf
```
LucioFranco: `--all-features` here as well?

Cargo.toml
```diff
@@ -19,6 +19,25 @@ license = "MIT/Apache-2.0"
...
+
+animation = [
+    "amethyst_animation"
+]
+audio = [
+    "amethyst_audio"
+]
+gltf = [
+    "amethyst_gltf",
+    "amethyst_animation"
```
LucioFranco: not totally positive but can we not reference the `animation` feature instead of the crate here?

docs/CHANGELOG.md
```diff
@@ -66,6 +69,8 @@ it is attached to. ([#1282])
 [#1404]: https://github.com/amethyst/amethyst/pull/1404
 [#1408]: https://github.com/amethyst/amethyst/pull/1408
 [#1405]: https://github.com/amethyst/amethyst/pull/1405
+[#1411]: https://github.com/amethyst/amethyst/pull/1411
+[#1412]: https://github.com/amethyst/amethyst/pull/1412
```
torkleyy: The entry for 1412 seems to be missing.

jojolepro: * Added optional feature gates which will reduce compilation times when used. ([#1412])

Cargo.toml
jojolepro: doesn't work

.concourse/tasks/linux-test.yml
jojolepro: would not work on nightly. Torkleyy made a change to support it in another PR.

book/src/appendices/c_feature_gates.md
```diff
@@ -0,0 +1,46 @@
...
+
+Enabling the nightly feature adds a bit of debug information when running into runtime issues. To use it, you need to use the nightly rust compiler toolchain.
+
+Here is how to enable it:
+
+```
+cargo (build/test/run) --features nightly
+```
+
+The reason mentionned the most often for using it is to show exactly which resource is missing when a Resources::fetch() fails (or World::read_resource()).
```
azriel91: Maybe:

The most common use of this feature is to find out the type name of the resource that is missing, such as when a `Resources::fetch()` or `World::read_resource()` invocation fails.

jojolepro: This sounds way better than what I have, thanks!

jojolepro:
bors retry

PR #1414, opened 2019-02-12
[WIP] Never again rebase `CHANGELOG.md`

### LINKED ISSUE #1029: New changelog strategy
Our changelog is producing a lot of conflicts, leading to a lot of rebasing fun.

We can

* enforce specific commit formats to generate changelogs from (higher barrier for contributors, not what I recommend)
* continue doing what we are doing
* go through the PRs ourselves and write the changelog
* create a tool using the GitHub API that creates the changelog interactively

The tool would be my favourite. It'll be as simple as

* go through every PR from the selection
* ask if it's an addition (a), change (c) or fix (f)
* ask for a description, default to PR title
* automatically create the link and add it to the log

torkleyy:
## Description

This PR makes each entry of the changelog its own file such that multiple PRs do not need to rebase the `CHANGELOG.md`.

The individual entries will be merged using [`clmerge`](https://github.com/torkleyy/clmerge).

Fixes #1029 

## Modifications

* Changelog strategy changed

## TODO

* [ ] Update contribution guidelines

## PR Checklist

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,4 +1,6 @@
+<!-- This file is auto-generated. Do not edit. -->
 # Change Log
+
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog][kc], and this project adheres to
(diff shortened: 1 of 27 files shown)


LucioFranco:
I like the idea but I'm personally not fond of having a file for each entry. I would much rather see the changelog be set via the PR title. Have you looked into that at all?

LucioFranco:
Also, how is order maintained between changelog entries? I don't know if this is something we need but in the past, we had an order, mostly due to people adding them after each other.

Xaeroxe:
> Also, how is order maintained between changelog entries? I don't know if this is something we need but in the past, we had an order, mostly due to people adding them after each other.

This is in my opinion unnecessary as "chronological" is arbitrary and doesn't come with any benefits for the changelog reader.

Xaeroxe:
> I like the idea but I'm personally not fond of having a file for each entry. I would much rather see the changelog be set via the PR title. Have you looked into that at all?

Not all PRs should get a changelog entry, and even for the ones that should the PR title isn't always a good changelog entry. I'd prefer to give us the opportunity to handcraft a message for the changelog.

LucioFranco:
> This is in my opinion unnecessary as chronological is arbitrary and doesn't come with any benefits for the changelog reader.

Sure this was a minor thing but I do generally like the order. It can be helpful as you can kinda see how things were built.

> Not all PRs should get a changelog entry, and even for the ones that should the PR title isn't always a good changelog entry. I'd prefer to give us the opportunity to handcraft a message for the changelog.

So my thoughts around this were to use labels and PR titles to generate meaningful changelogs. Depending on certain labels you can have that PR be ignored by the changelog. Now from this, we can even enforce that there is a label for the crates that are touched by this PR and this would then allow us to build changelog entries for each of amethyst crates without really having to manually curate. This would mean whoever does the merge has to make sure things look right.

torkleyy:
That‘s all possible, but it‘s work somebody has to do. Both the setup and following all the conventions. The generators I tried so far were very slow and need more time to set up. Do you have a specific alternative in mind?

LucioFranco:
@torkleyy what about that changelog generator you sent me a while ago? That one works quite well. I'd like to explore more options than this personally.

https://github.com/github-changelog-generator/github-changelog-generator

LucioFranco:
second before we merge this, what is the process behind this? When does it get updated? Who does it? Does CI build it every night? etc this to me isn't clear.

torkleyy:
Unfortunately I didn't get that one to work well with Amethyst. Last time I tried, it kept running for hours with no output so I killed it at some point.

LucioFranco:
@torkleyy I have it working and we talk a bit on discord about this, I would like to delay the merging of this until we have found a proper way to automate and build the changelogs.

Frizi:
Could we extend bors to read a specific entry from PR's first post and insert it as a new commit just before merging to master? That way the message is still handcrafted and present only when was explicitly written, it's all in single file in VCS and there are no conflicts about it.

torkleyy:
> Could we extend bors to read a specific entry from PR's first post and insert it into changelog as a new commit just before merging to master?

So far, the bors team stated they don't want to mess with commits / code. And I'm not sure we want to maintain a fork.

Frizi:
Bors is doing the merge anyway, so maybe a clever precommit hook installed on it's repo would do the job without forking.

LucioFranco:
Not a fan of using bors honestly for this, I'd rather move the changelog to the website and have that build every 24h.

fhaynes:
Closing this, as there doesn't seem to be consensus and we have merge conflicts again. Please open an issue for discussion if someone wants to pursue this.

PR #1415, opened 2019-02-13
Improve documentation for logger env variables

### LINKED ISSUE #1316: Amethyst logger ignores wrong env var values
## Reproduction Steps

1. Launch any Amethyst game using the provided logger
2. Set `AMETHYST_LOG_STDOUT` to `this_is_not_allowed`
3. Notice no error is reported

The application prints an error message.

## What Actually Happened

The variable's value was just silently ignored.

torkleyy:
## Description

Improves docs & allows more values as described in #1316.

Fixes #1316

## Additions

* Add "yes" / "no" to allowed values
* Document allowed values
* Add test for env var override of `LoggerConfig`

## PR Checklist

- [x] ~~Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.~~

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -130,8 +130,18 @@ impl Logger {
 ///
 /// Configuration of the logger can also be controlled via environment variables:
 /// * AMETHYST_LOG_STDOUT - determines the output to the terminal
+///     * "no" / "off" disables logging to stdout
+///     * "plain" / "yes" enables logging to stdout
+///     * "colored" enables logging and makes it colored
 /// * AMETHYST_LOG_LEVEL_FILTER - sets the log level
+///     * "off" disables all logging
+///     * "error" enables only error logging
+///     * "warn" only errors and warnings are emitted
+///     * "info" only error, warning and info messages
+///     * "debug" everything except trace
+///     * "trace" everything
 /// * AMETHYST_LOG_FILE_PATH - if set, enables logging to the file at the path
+///     * the value is expected to be a path to the logging file
 pub fn start_logger(config: LoggerConfig) {
     Logger::from_config(config).start();
 }
@@ -139,8 +149,8 @@ pub fn start_logger(config: LoggerConfig) {
 fn env_var_override(config: &mut LoggerConfig) {
     if let Ok(var) = env::var("AMETHYST_LOG_STDOUT") {
         match var.to_lowercase().as_ref() {
-            "off" => config.stdout = StdoutLog::Off,
-            "plain" => config.stdout = StdoutLog::Plain,
+            "off" | "no" => config.stdout = StdoutLog::Off,
+            "plain" | "yes" => config.stdout = StdoutLog::Plain,
             "colored" => config.stdout = StdoutLog::Colored,
             _ => {}
         }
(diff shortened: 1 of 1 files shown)


src/logger.rs
```diff
@@ -130,17 +130,27 @@ impl Logger {
...
 /// Configuration of the logger can also be controlled via environment variables:
 /// * AMETHYST_LOG_STDOUT - determines the output to the terminal
+///     * "no" / "off" disables logging to stdout
+///     * "plain" / "yes" enables logging to stdout
+///     * "colored" enables logging and makes it colored
 /// * AMETHYST_LOG_LEVEL_FILTER - sets the log level
+///     * "off" disables all logging
+///     * "error" enables only error logging
+///     * "warn" only errors and warnings are emitted
+///     * "info" only error, warning and info messages
```
jojolepro: When I tried info in my projects, info + warn would actually NOT show. I'm not sure why that was the case.
Using debug would show info + warn + errors though.

torkleyy: I tried every possible filter, all worked as expected.

src/logger.rs
```diff
@@ -130,17 +130,27 @@ impl Logger {
...
 pub fn start_logger(config: LoggerConfig) {
     Logger::from_config(config).start();
 }
 
 fn env_var_override(config: &mut LoggerConfig) {
     if let Ok(var) = env::var("AMETHYST_LOG_STDOUT") {
         match var.to_lowercase().as_ref() {
-            "off" => config.stdout = StdoutLog::Off,
-            "plain" => config.stdout = StdoutLog::Plain,
+            "off" | "no" => config.stdout = StdoutLog::Off,
```
khionu: Can we add numeric representation as well? 0, 1, and 2

torkleyy:
I think this would be confusing for the log level, so I addressed it for stdout only.

khionu:
How would it be confusing? Each log level includes itself and everything less verbose than itself. In fact, I think associating each level with a number could help keep it straight to the consumer.

I'm not terribly concerned about this, however. I just think it would be more consistent to have both have numeric counterparts than just one.

torkleyy:
> How would it be confusing?

I wouldn't be able to tell what "1" is without looking at the docs. It could be

* the most verbose level (`trace`)
* the least verbose level (`error`)
* "logging enabled" => some level (maybe `info`)

khionu:
I mean, that holds for any enumeration beyond "true/false", including the one you just added. Even "true/false" would have to be inferred through the understanding that 0 is false and 1 is true as a standard.

khionu:
Also, given 0 is off, I would assume 1, which is just above 0, is the minimum logging. Makes no sense to order them as "Off, Max, ..., Min"

torkleyy:
Yeah, the color one is not 100% obvious, either. But really, I wouldn't even know how to implement the numbers for the logging level. And the fact that 0 is false and 1 is true is something we can assume.

torkleyy:
Can we please just focus on what I added? We don't need to get it perfect 🙂

khionu:
https://github.com/amethyst/amethyst/blob/master/src/logger.rs#L149

Here, you would just add an `else if` calling a reimpl of [this private function from `log`](https://docs.rs/log/0.4.6/src/log/lib.rs.html#637). I would also open an issue to make that function public, for serialization purposes.

khionu:
Yes, I can approve what we have. Can discuss this further after.

torkleyy:
Thanks!
bors r=jojolepro,khionu

PR #1419, opened 2019-02-13
Move transparent to default

Xaeroxe:
## Description

Users seem to expect this to be the default, so we're going to make it the default.

## Additions

- Several passes got `disable_transparency` which disables transparency in the pass.
- Several passes got `with_transparency_settings` which changes the transparency settings for the pass.

## Removals

- Several passes had the method `with_transparency` removed as it is now the default.

## Modifications

- `Default::default` now returns a pass with transparency enabled for all applicable passes.

## PR Checklist

### DIFF
--- a/amethyst_renderer/src/pass/flat/interleaved.rs
+++ b/amethyst_renderer/src/pass/flat/interleaved.rs
@@ -19,7 +19,7 @@ use crate::{
     hidden::{Hidden, HiddenPropagate},
     mesh::{Mesh, MeshHandle},
     mtl::{Material, MaterialDefaults},
-    pass::util::{draw_mesh, get_camera, setup_textures, VertexArgs},
+    pass::util::{default_transparency, draw_mesh, get_camera, setup_textures, VertexArgs},
     pipe::{
         pass::{Pass, PassData},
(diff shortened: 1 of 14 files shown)


Xaeroxe:
Hey uh how bout we don't merge this yet as it currently breaks things. I'll investigate more later.

torkleyy:
Can we maybe just add a boolean parameter to `with_transparency`? That would make the breakage easier to fix, is shorter and it's the positive form (which I think is the recommendation for API). Just a quick idea, feel free to ignore if you don't have time for it ;)

torkleyy:
To elaborate: `true` would just set the settings to the default.

Xaeroxe:
@torkleyy the only thing I worry about with that is it might not be clear to users that using that function also alters the transparency settings. Otherwise I agree.

Xaeroxe:
Ok I've picked some better defaults that work for 99% of scenarios.

torkleyy:
@Xaeroxe Yeah, I agree. Although, what we could do is just ignore a `true` value in case there are settings already.

Xaeroxe:
bors r=@torkleyy

PR #1424, opened 2019-02-17
ScreenSpace Component. Draw Sprite as UI Elements!

jojolepro:
## Description

Add the ScreenSpace component to any 2d entity, and it will shown using the screen coordinates instead of the camera's settings.

TODO: Figure out why it ScreenSpace entities always draw underneath non-ScreenSpace entities.

## Additions

- ScreenSpace component

This PR is non-breaking/additive.

## PR Checklist


jojolepro:
tests aren't triggering >_<

amethyst_renderer/src/pass/util.rs
```diff
@@ -333,6 +334,28 @@ pub fn set_view_args(
     effect.update_constant_buffer("ViewArgs", &view_args.std140(), encoder);
 }
 
+pub fn set_view_args_screen(
```
Moxinilian: Could you document this please?

Moxinilian:
Also regarding your ordering issue, I think it's just an implementation detail and should not cause any issue for the user (they should use Z ordering in cases where it matters).

jojolepro:
Its definitely causing issues. Fletcher helped me investigate this a bit, I'll see if its possible to fix it without having a separated pass.

Moxinilian:
What issues specifically?

amethyst_renderer/src/pass/util.rs
jojolepro: done

amethyst_renderer/src/pass/flat2d/interleaved.rs
```diff
@@ -203,27 +220,37 @@ impl Pass for DrawFlat2D {
...
+                        screen_maybe.is_some(),
+                    );
                 }
 
                 // We are free to optimize the order of the opaque sprites.
                 self.batch.sort();
 
                 for entity in &visibility.visible_ordered {
+                    let screen = screens.contains(*entity);
+                    println!("visibility.visible_ordered screen: {}", screen);
```
azriel91: extra code

amethyst_renderer/src/pass/flat2d/interleaved.rs
```diff
@@ -394,20 +444,51 @@ impl TextureBatch {
...
+            TextureBatch::encode_vec(
+                &self.textures,
+                encoder,
+                factory,
+                effect,
+                sprite_sheet_storage,
+                tex_storage,
+            );
+        }
+        if let Some(depth_data) = &effect.data.out_depth {
```
azriel91: can you put this within the `textures_screen` conditional as well? so the depth buffer doesn't clear if it doesn't need to be

amethyst_renderer/src/pass/flat2d/interleaved.rs
```diff
@@ -394,20 +444,51 @@ impl TextureBatch {
...
+                factory,
+                effect,
+                sprite_sheet_storage,
+                tex_storage,
+            );
+        }
+    }
+
+    fn encode_vec(
+        elems: &Vec<TextureDrawData>,
```
azriel91: best rename `elems` to `textures` or something similarly specific

amethyst_renderer/src/pass/util.rs
```diff
@@ -333,6 +335,36 @@ pub fn set_view_args(
...
+/// Sets the view arguments in the constant buffer using the screen dimensions.
+pub fn set_view_args_screen(
+    effect: &mut Effect,
+    encoder: &mut Encoder,
+    screen_dimensions: &ScreenDimensions,
+) {
+    #[cfg(feature = "profiler")]
+    profile_scope!("render_setviewargsscreen");
+
+    let translation = Vector3::<f32>::new(0.0, 0.0, 1000.0);
```
azriel91: does 2000.0 work, and if it does, could you extract this and the one in the `Orthographic3` into a `pub const` (and export all the way up)?
That way people depending on magic numbers can at least have a constant to rely on

amethyst_renderer/src/screen_space.rs
```diff
@@ -0,0 +1,11 @@
+use amethyst_core::specs::{Component, NullStorage};
+
+/// Indicates that the entity on which this is placed should use the coordinates of the window instead of the coordinates relative to the `Camera`.
```
azriel91: think rustfmt would deny line length

amethyst_renderer/src/screen_space.rs
```diff
@@ -0,0 +1,11 @@
...
+/// Mostly used with user interface elements.
+/// Do not use with entities located outside of -1000 <= z < 1000.
```
azriel91: re:magic numbers, it's difficult for people to discover this -- "you don't know until you know" type of thing, which is a gotcha. 

also, is -1000 because camera_z = 1000, and view distance 2000? this makes me prefer having 2 passes (second pass ignores depth)

jojolepro: You can't ignore the depth though.

jojolepro:
@azriel91 r? :)

amethyst_renderer/src/screen_space.rs
azriel91: haha, "make it configurable". I can live with that, can you update this sentence then:

```rust
/// Do not use with entities located outside of the visible z range as defined by
/// `ScreenSpaceSettings`.
```

jojolepro:
which sentence?

azriel91:
[that one](https://github.com/amethyst/amethyst/pull/1424#discussion_r261895453):

```rust
/// Do not use with entities located outside of the visible z range as defined by
/// `ScreenSpaceSettings`.
```

jojolepro:
Fixed! :)

azriel91:
🚀
bors r=Moxinilian,azriel91

omni-viral:
And now we can confuse screen-space entities with world-space entities by forgetting to check another component.

jojolepro:
They are 2d entities with the only difference being "where" they are draw. There is literally no other difference. ^u^

Moxinilian:
Suppose we forget about the ScreenSpace component.
Wouldn't iterating over all sprites create confusion? (as it would include UI elements as well)
It should be easy to iterate over all sprites that are not UI.

omni-viral:
Actually, if there are disjoint sets of entities not related with each other then they could be handled in parallel, but using same component type disallows this.
I propose to use tagged transform type with different tags for different spaces.

PR #1425, opened 2019-02-17
Roadmap fixes

jojolepro:
Please keep the discussion friendly

Justifications:
Window, osx, linux order: It feels very exclusive for linux users to say that sentence, and I have yet to see a windows-only feature in the engine.
UI: We already have basic ui. We are not trying to rewrite it from scratch, we want to improve the base and make a ton of features.
Physics: Well we want physics :P


docs/ROADMAP.md
```diff
@@ -15,13 +15,13 @@
...
+- Windows, macOS, Linux compatibility
 
 ## Planned Features
-
-- Windows, macOS, Linux compatibility (in that order)
 - A scripting language
-- Basic UI library
+- Fully featured and extensible UI Library.
 - An editor
+- Physics
```
azriel91: Should this be expanded (because of the "Basic physics" above)

jojolepro: ah right, we don't have physics :P

PR #1428, opened 2019-02-18
Measure code coverage using GitLab CI

torkleyy:
Integrate GitLab CI + kcov + codecov.io

## Description

* Collect code coverage using `kcov`
* Upload coverage to codecov.io
* Use GitLab CI to do all the work
* Add badge to readme (not yet valid)

## PR Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -44,3 +44,13 @@ run tests:
   only:
     - branches
     - merge_requests
+
+run kcov:
+  stage: test
+  image: docker:latest
+  services:
+    - docker:dind
+  script:
+    - docker --version
+    - ./scripts/kcov.sh
+    - curl -v -s https://codecov.io/bash | bash
--- a/README.md
+++ b/README.md
@@ -2,7 +2,8 @@
 
 # Amethyst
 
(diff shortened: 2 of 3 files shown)


LucioFranco:
I dont even see this running on this PR or does this run only on master? 

cc @fhaynes what should we do about this? Have we made the choice to go with jenkins or go back to gitlab?

torkleyy:
It _should_ only run on master ;)

.gitlab-ci.yml
```diff
@@ -44,3 +44,13 @@ run tests:
...
     - branches
     - merge_requests
+
+run kcov:
+  stage: test
+  image: docker:latest
+  services:
+    - docker:dind
+  script:
+    - docker --version
```
LucioFranco: do we need this?

LucioFranco:
we should probably run this to make sure it all works, prob can do that via bors try?

azriel91:
bors try

LucioFranco:
That didnt actually test this :/

@torkleyy maybe we can try to enable gitlab for just this branch temporarily?

.gitlab-ci.yml
torkleyy: Well, it‘s additional debug info. Just a habit to print version numbers ^^

fhaynes:
Closing this as we are not using GitLab CI/CD atm

PR #1436, opened 2019-02-27
Feature/expose frame rate limit config constructor

azriel91:
## Description

`FrameRateLimitConfig` has a `new` constructor, and its fields are made public, so that users can construct these as `const` fields.

## Modifications

* `FrameRateLimitConfig` has a `new` constructor, and its fields are made public, so that users can construct these as `const` fields.

## PR Checklist


azriel91:
bors r=jojolepro,distransient

PR #1441, opened 2019-03-03
Fixed outdated code in gamedata book page

Pisti01:
## Description

The example code was outdated in the Custom GameData chapter

## PR Checklist

### DIFF
--- a/book/src/game-data.md
+++ b/book/src/game-data.md
@@ -15,7 +15,10 @@ are essential (like rendering, input and UI).
 
 Let's start by creating the `GameData` structure:
 
-```rust,ignore
+```rust,no_run,noplaypen
+# extern crate amethyst;
+# use amethyst::ecs::prelude::Dispatcher;
+#
 pub struct CustomGameData<'a, 'b> {
     core_dispatcher: Dispatcher<'a, 'b>,
     running_dispatcher: Dispatcher<'a, 'b>,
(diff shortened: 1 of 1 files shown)


Pisti01:
I also found this in the same file:
```rust
.with_running::<ExampleSystem>(ExampleSystem, "example_system", &[])
```
which could be written as
```rust
.with_running(ExampleSystem, "example_system", &[])
```
Should I replace that as well?

jojolepro:
Yes you can! ;)

Moxinilian:
This PR is good as is, but I think this chapter would deserve the same level of automated testing we have in other chapters (with hidden context). Would you have time to do it or do you prefer we do it later?

Pisti01:
I can try to do it but I'm unsure how can I test it

Moxinilian:
1. Install mdbook with cargo install
2. Build Amethyst at least once
3. To test the book, simply run `mdbook test -L target/debug/deps`. If it fails, delete your target folder and rebuild Amethyst. If it still fails, ping me.

Pisti01:
I've been looking at how the code blocks are defined in the books and most places it uses:
`rust,edition2018,no_run,noplaypen`
which causes mdbook test to skip that code.

But if I use:
`rust,no_run,noplaypen` or `rust,edition2018,noplaypen`
mdbook actually tests the code.

How should I define the code blocks?

Moxinilian:
That's surprising. Does it give more information on why it skips it?
In the meantime, you can use `rust,no_run,noplaypen`, it's what we have been using before edition 2018.

Pisti01:
No info, just silently skips it.

Moxinilian:
That's weird. Drop edition then, we'll see if something weird happens later.

Moxinilian:
I think this has been waiting long enough.
bors r=azriel91,Moxinilian

PR #1442, opened 2019-03-04
Add function to transform position

vessd:
## Description

Function to transform position from screen space into world space

## Additions

- `fn screen_to_world`

### DIFF
--- a/amethyst_renderer/src/cam.rs
+++ b/amethyst_renderer/src/cam.rs
@@ -3,11 +3,14 @@
 use amethyst_assets::PrefabData;
 use amethyst_core::{
     ecs::prelude::{Component, Entity, HashMapStorage, Write, WriteStorage},
-    math::{Matrix4, Orthographic3, Perspective3},
+    math::{Matrix4, Orthographic3, Perspective3, Point2, Point3},
+    GlobalTransform,
 };
 use amethyst_error::Error;
 
 use serde::{Deserialize, Serialize};
+
+use crate::ScreenDimensions;
 
 /// The projection mode of a `Camera`.
 ///
@@ -75,6 +78,27 @@ impl Camera {
             width / height,
             std::f32::consts::FRAC_PI_3,
         ))
+    }
+
+    /// Transforms position from screen space to camera space
+    pub fn position_from_screen(
+        &self,
+        screen_position: Point2<f32>,
+        camera_transform: &GlobalTransform,
+        screen_dimensions: &ScreenDimensions,
+    ) -> Point3<f32> {
+        let screen_x = 2.0 * screen_position.x / screen_dimensions.width() - 1.0;
+        let screen_y = 1.0 - 2.0 * screen_position.y / screen_dimensions.height();
+        let screen_point = Point3::new(screen_x, screen_y, 0.0).to_homogeneous();
+
+        let vector = camera_transform.0
+            * self
+                .proj
+                .try_inverse()
+                .expect("Camera projection matrix is not invertible")
+            * screen_point;
+
+        Point3::from_homogeneous(vector).expect("Vector is not homogeneous")
     }
 }
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -27,6 +27,7 @@ it is attached to. ([#1282])
 * Add `SpriteRenderPrefab`. ([#1435])
 * Add `ScreenSpace` component. Draws entities using the screen coordinates. ([#1424])
 * Add `add_removal_to_entity` function. ([#1445])
+* Add `position_from_screen` to `Camera`. Transforms position from screen space to camera space. ([#1442])
 
 ### Changed
 
@@ -92,6 +93,7 @@ extra bounds from `AnimatablePrefab` and `AnimationSetPrefab` ([#1435])
 [#1445]: https://github.com/amethyst/amethyst/pull/1445
 [#1451]: https://github.com/amethyst/amethyst/pull/1451
 [#1454]: https://github.com/amethyst/amethyst/pull/1454
+[#1442]: https://github.com/amethyst/amethyst/pull/1442
 
 ## [0.10.0] - 2018-12
 


amethyst_utils/src/screen_to_world.rs
```diff
@@ -0,0 +1,25 @@
+//! Provides utilities for transform position.
+
+use amethyst_core::nalgebra::{Matrix4, Orthographic3, Vector3};
+use amethyst_renderer::ScreenDimensions;
+
+/// Transforms position from screen space into world space
+pub fn screen_to_world(
+    position: (f64, f64),
```
jojolepro: Can you rename this to screen_position?

amethyst_utils/src/screen_to_world.rs
```diff
@@ -0,0 +1,25 @@
+//! Provides utilities for transform position.
+
+use amethyst_core::nalgebra::{Matrix4, Orthographic3, Vector3};
+use amethyst_renderer::ScreenDimensions;
+
+/// Transforms position from screen space into world space
+pub fn screen_to_world(
+    screen_position: (f64, f64),
+    camera_translation: &Vector3<f32>,
+    camera_proj: &Matrix4<f32>,
```
Moxinilian: Wouldn't it make more sense here to directly take the Orthographic3 struct here?
It makes the user aware of what the parameter is, gives them the responsibility of providing a good projection and saves a cast when they already have the Orthographic3 struct.

vessd: Perhaps. I have little experience in development.
In my case it allows me not to write `use amethyst::core::::nalgebra::Orthographic3`.
And I probably would not understand where to get it, if I saw it the first time.
Right now, it’s enough for the user to pass `camera.proj` without any conversions.

Moxinilian: Well, I think the real issue here is that camera.proj is a Matrix4 and not an Orthographic3, as Orthographic3 anyway is just a wrapper for Matrix4. Let's leave it as is for now, it should be improved in the new renderer.

vessd: May be this should to take a link to the camera object for greater clarity

Moxinilian:
The tests unexpectedly failed, but it probably is a glitch. This will not prevent merging once 24 hours will have passed.

amethyst_utils/src/screen_to_world.rs
Moxinilian: Good idea. What you can do is in the doc just above explain that the matrix is the camera projection matrix field, or an Orthographic3 matrix.

vessd: Do you have an example of how I should format this?

Moxinilian:
Actually, come to thing of it, why don't you make this a function on the Camera component?
So we could do something like `camera.coordinates_from_screen(screen_coords, screen_dim)`.
Maybe that would be more convenient.

jojolepro:
+ the camera global transform as a parameter.

Making it a function of the camera makes sense.

The camera might be a child of another entity, so it would also make sense to use the global transform instead of the translation. However, this will make the math harder, so I don't know if we want that in this PR or not. We don't want to make you work too hard :D

vessd:
Should I remove the translation argument then?

Moxinilian:
Oh right, the translation. Hmmm. Maybe the function could take its own transform or something.
Ah that's a tricky one actually. The reason I would like it to be on the Camera component is because it is an obvious place to look for people coming from non-ECS places. We should look at how Unity is doing it with their ECS.

Moxinilian:
Okay so apparently they're using globals, which is not very good. We should not do that.
I guess we should have it be a method on the Camera, to which we pass the camera's transform, the screen position and the ScreenDimension. What do you think?

amethyst_renderer/src/cam.rs
```diff
@@ -76,6 +79,27 @@ impl Camera {
             std::f32::consts::FRAC_PI_3,
         ))
     }
+
+    /// Transforms position from screen space to camera space
```
jojolepro:
```suggestion
    /// Transforms position from screen space to camera space
    /// This only works if the camera is not a child of another entity and if it is not rotated.
```

Moxinilian: Why not use GlobalTransform then?

amethyst_renderer/src/cam.rs
```diff
@@ -76,6 +79,22 @@ impl Camera {
...
+    pub fn position_from_screen(
+        &self,
+        screen_position: (f64, f64),
+        camera_transform: &GlobalTransform,
+        screen_dimensions: &ScreenDimensions,
+    ) -> (f32, f32) {
+        let screen_x = 2.0 * screen_position.0 as f32 / screen_dimensions.width() - 1.0;
+        let screen_y = 1.0 - 2.0 * screen_position.1 as f32 / screen_dimensions.height();
+        let screen_point = Point3::new(screen_x, screen_y, 0.0).to_homogeneous();
```
vessd: I'm not sure about 0.0, but it works fine for me

Moxinilian: What do you mean?

vessd: I don't know if it's ok to set `0.0` for the z axis for `screen_point`

amethyst_renderer/src/cam.rs
```diff
@@ -76,6 +79,22 @@ impl Camera {
...
+
+        let camera_point = camera_transform.0 * self.proj.try_inverse().unwrap() * screen_point;
```
Moxinilian: Could you make this an `except` instead of an `unwrap`?

amethyst_renderer/src/cam.rs
Moxinilian: I believe it is, considering Z information only resides in the camera. Could you try it with a weird camera, like one with translated and rotated?

vessd: Yes, I tested it with parameters
```
transform.set_xyz(17.0, -2.0, 1.0);
transform.set_rotation_euler(0.0, 0.0, 1.0);
```
It works fine

amethyst_renderer/src/cam.rs
```diff
@@ -76,6 +79,24 @@ impl Camera {
...
+        screen_position: (f64, f64),
+        camera_transform: &GlobalTransform,
+        screen_dimensions: &ScreenDimensions,
+    ) -> (f32, f32) {
+        let screen_x = 2.0 * screen_position.0 as f32 / screen_dimensions.width() - 1.0;
+        let screen_y = 1.0 - 2.0 * screen_position.1 as f32 / screen_dimensions.height();
+        let screen_point = Point3::new(screen_x, screen_y, 0.0).to_homogeneous();
+
+        let camera_point = camera_transform.0
+            * self.proj.try_inverse().expect("Matrix is not invertible")
```
Moxinilian:
```suggestion
            * self.proj.try_inverse().expect("Camera projection matrix is not invertible")
```

Moxinilian:
Actually, while I'm at it, why not return a Vector3 instead of just X and Y? It can be useful when your camera is not on the Z = 0 plane. Also, why are the input coordinates f64? It could be a Vector2, even.

vessd:
> Also, why are the input coordinates f64? It could be a Vector2, even.

Because the `InputHandler::mouse_position()` returns the `Option<(f64, f64)>`.

> Also, why are the input coordinates f64? It could be a Vector2, even.

Hmm, I don’t understand how it works clearly, do I need to take `Vec3` for this?
I mean, is it ok that `screen_point` is always equal to` 0.0` for the `z` plane for this? Or do i need to accept it as a parameter?
I didn't test the `z` coordinate, since I have no idea how to interpret it in my 2D game

Moxinilian:
I think introducing a Z coordinate here as input would just shift how "deep" the click is made. It should not really matter here, we only are converting click position, not raycasting.

However, the Z output can be useful if your camera is in 3D space. As we are doing an inverted camera projection, we are effectively taking the screen plane at Z=0 and applying it to the 3D plane orthogonal to the camera's direction. That plane can be a Z=0 plane, but not necessarily, if the camera is in a weird position.

vessd:
@Moxinilian ok, should I leave a f64 as an input argument?

Moxinilian:
I think you should do as little casting as possible in the body of the function so the user does not have to do useless casting if they already work with f32. Set it to f32 in the parameter, and let the user do the casting.

vessd:
I think it is ready to merge.
r? @Moxinilian @jojolepro @azriel91

amethyst_renderer/src/cam.rs
```diff
@@ -76,6 +79,27 @@ impl Camera {
...
+        &self,
+        screen_position: Vector2<f32>,
+        camera_transform: &GlobalTransform,
+        screen_dimensions: &ScreenDimensions,
+    ) -> Vector3<f32> {
+        let screen_x = 2.0 * screen_position.x / screen_dimensions.width() - 1.0;
+        let screen_y = 1.0 - 2.0 * screen_position.y / screen_dimensions.height();
+        let screen_point = Point3::new(screen_x, screen_y, 0.0).to_homogeneous();
+
+        let matrix = camera_transform.0
```
Moxinilian: Isn't this already a Vector3 or Point3?

amethyst_renderer/src/cam.rs
```diff
@@ -76,6 +79,27 @@ impl Camera {
...
+            * self
+                .proj
+                .try_inverse()
+                .expect("Camera projection matrix is not invertible")
+            * screen_point;
+
+        Vector3::new(matrix.x, matrix.y, matrix.z)
```
Moxinilian: I said Vector3 earlier, but I forgot nalgebra has a Point3 type which might make more sense here. Same applies for the input Vector2 with Point2.

amethyst_renderer/src/cam.rs
vessd: It's a `Matrix1x4`.
I haven't found an easy way to convert it to `Vector3` or `Point3`.

vessd: Hm, I think it should be `Point3::from_homogeneous`

Moxinilian: Ah you are correct, as it's a 1x4 we can't just cast it. I think your initial solution is good then.

azriel91:
bors r=jojolepro,Moxinilian,azriel91

PR #1439, opened 2019-03-02
Pong bounce simplify

Pisti01:
## Description

Removed some duplicate code from pong's bounce system by merging 2 if checks into a single one.

## PR Checklist

By placing an x in the boxes I certify that I have:

- [x] Ran `cargo test --all` locally if this modified any rs files.
- [x] Ran `cargo +stable fmt --all` locally if this modified any rs files.
- [x] Updated the content of the book if this PR would make the book outdated.
- [ ] Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.
- [x] Added unit tests for new APIs if any were added in this PR.
- [x] Acknowledged that by making this pull request I release this code under an MIT/Apache 2.0 dual licensing scheme.

PR #1444, opened 2019-03-04
Add missing changelog entry from #1439

distransient:


distransient:
Eh I approved it without one so I figured it'd be good to add it myself anyways. Now the real question is should we add a changelog entry for adding this changelog entry... 

image

distransient:
bors r=@jojolepro

jojolepro:
bors r+
totally

PR #1446, opened 2019-03-04
fix #780, make frame_limit::do_sleep calculate how long to sleep instead of sleep(0) in a loop

### LINKED ISSUE #780: High CPU usage in empty project
When running an empty amethyst application (i.e. one with no systems registered and no extra logic happening in the game state's `update` method) I've noticed unexpectedly high CPU usage on my Windows 10, somewhere between 15%-25% with a 6 core CPU. I would expect to only see ~1% usage, since the application should be effectively waking up briefly at the beginning of each frame, and then immediately going back to sleep until the next frame. Note that this is limiting to 60 fps with a sleep-only limiting strategy, which should be the strategy the has the lowest CPU-usage.

I profiled the application, and found that most of the CPU time was being spent in rayon's worker thread logic. As best I can tell, what's happening is that rayon has a grace period where worker threads will spin-wait before sleeping them when there's no work available. Once more work is added, all sleeping worker threads are woken back up. Since amethyst adds new work every frame, the worker threads are being woken back up each frame, and once all the work for the frame is done, every worker thread has to spin-wait for some amount of time. This results in an (apparently) large amount of extra CPU time being eaten each frame.

I'm not able to test this behavior on MacOS/Linux, but I'd expect the result is the same on Unix-like platforms as it is on Windows.

It seems like this could be solved if amethyst could notify rayon that all the work for the thread is done, allowing rayon to immediately put all worker threads to sleep until the next frame. I don't know if this is something rayon supports, or if it's the optimal solution.

## How To Get Started

The biggest source of CPU usage when Amethyst is done with all work for a frame is coming from Rayon. The way Amethyst uses Rayon is triggering worst-case behavior in how Rayon handles putting worker threads to sleep, and the only real solution is for Rayon to add better support for quickly sleeping worker threads at the end of the frame.

A good first step for fixing this would be to build a demo that uses rayon the way that Amethyst does so that the Rayon devs can reproduce the issue and test fixes. Specifically, the demo should do the following:

1. Create a thread pool using Rayon. Ideally it should use a setup similar to what Amethyst does in terms of how many worker threads it creates.
2. Run in a loop that spawns a fixed amount of work (a good amount would be 5-10 milliseconds, bonus points for making the amount configurable) at the beginning of every loop iteration.
3. Each iteration, wait for work spawned on the worker thread to finish and then have the main thread wait until enough time has passed to start the next "frame". 60 fps is a good test case, so the demo should attempt to have the body of the loop take 16.6 ms each iteration.

Once we have this demo, we can make a PR to Rayon adding it to their demo projects, and then discuss with the Rayon team about how to best solve the problem.

estokes:
## Description

This fixes PR #780, tested on linux. Passing 0 to sleep just results in busy waiting, instead we should calculate the amount of time we want to wait, and pass that to sleep.

## PR Checklist

- [x ] Ran `cargo test --all` locally if this modified any rs files.
- [x ] Ran `cargo +stable fmt --all` locally if this modified any rs files.
- [x ] Updated the content of the book if this PR would make the book outdated.
- [x ] Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.
- [x ] Added unit tests for new APIs if any were added in this PR.

### DIFF
--- a/amethyst_core/src/frame_limiter.rs
+++ b/amethyst_core/src/frame_limiter.rs
@@ -225,8 +225,13 @@ impl FrameLimiter {
 
     fn do_sleep(&self, stop_on_remaining: Duration) {
         let frame_duration = self.frame_duration - stop_on_remaining;
-        while Instant::now() - self.last_call < frame_duration {
-            sleep(ZERO);
+        loop {
+            let elapsed = Instant::now() - self.last_call;
+            if elapsed >= frame_duration {
+                break;
+            } else {
+                sleep(frame_duration - elapsed);
+            }
         }
     }
 }
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -29,6 +29,7 @@ it is attached to. ([#1282])
 
 ### Changed
 
+* Make `frame_limiter::do_sleep` calculate the amount of time to sleep instead of calling `sleep(0)` ([#780])
 * Make `application_root_dir` return a `Result<Path>` instead of a `String` ([#1213])
 * Remove unnecessary texture coordinates offset in `Sprite::from_pixel_values` ([#1267])
 * Changed `ActiveCamera` to have the `Option` inside. ([#1280])
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -58,6 +58,7 @@ extra bounds from `AnimatablePrefab` and `AnimationSetPrefab` ([#1435])
 * Set width and height of Pong Paddles ([#1363])
 * Fix omission in `PosNormTangTex` documentation. ([#1371])
 
+[#780]: https://github.com/amethyst/amethyst/pull/780
 [#1114]: https://github.com/amethyst/amethyst/pull/1114
 [#1213]: https://github.com/amethyst/amethyst/pull/1213
 [#1237]: https://github.com/amethyst/amethyst/pull/1237
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -29,7 +29,7 @@ it is attached to. ([#1282])
 
 ### Changed
 
-* Make `frame_limiter::do_sleep` calculate the amount of time to sleep instead of calling `sleep(0)` ([#780])
+* Make `frame_limiter::do_sleep` calculate the amount of time to sleep instead of calling `sleep(0)` ([#1446])
 * Make `application_root_dir` return a `Result<Path>` instead of a `String` ([#1213])
 * Remove unnecessary texture coordinates offset in `Sprite::from_pixel_values` ([#1267])
 * Changed `ActiveCamera` to have the `Option` inside. ([#1280])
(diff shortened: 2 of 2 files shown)


distransient:
Should this instead use a platform detection attribute then?

Xaeroxe:
@distransient that is probably the single most productive comment I've seen in the entire year or more of discussion that's been had on this issue. That's a great idea!

termhn:
We should think about using this https://github.com/alexheretic/spin-sleep or at least taking inspiration from it for our own implementation

distransient:
The direct use of the winapi sleep stands out to me here: https://github.com/alexheretic/spin-sleep/blob/master/src/lib.rs#L100-L109

The sleep/spin strategy otherwise seems similar to me as what already occurs in the "SleepAndYield" strategy https://github.com/alexheretic/spin-sleep/blob/master/src/lib.rs#L143-L153

estokes:
I would like to test this on Windows and see if there is a solution to the time quantum problem. I do agree that if there is no way ensure that we wake up before the quantum, which us usually 15ms, then we must busy wait on windows. I did some reading a few days ago, and found conflicting versions of what sleep(0) actually DOES in Windows. The official documentation states that if no other thread is ready to run, then it's a noop, which is what we want, however an MVP blog source stated that it always gives up the remainder of the timeslice, which could result in missing the frame deadline. I have a windows dev box, and I will work today on getting this all to build on it so I can test.

estokes:
Following up, I have tested on the latest stable version of windows 10. I had to jump through some hoops to actually get vsync to turn off, but once I successfully did that I found that the code works equally well on windows and linux. I configured a mostly empty game with a frame limit of 144 fps (7ms frame time), and then used delta_time to record any frames that missed 8ms. With SleepAndYield(2) there were a few at startup, and after that it was extremely rare, in 10 minutes of running maybe 5 additional frames missed 8ms, and there was only 1 that got to 9.5ms, none bigger than that.

So my conclusion is that Windows is just as happy to wake a thread up after a couple of ms as Linux is, and there is no reason I can see why the code should be different for Windows. That isn't the result I expected, and it's a pleasant surprise, clearly they've been busy in Redmond ...

As for the numerous versions of Windows released before 10, I can't say, and I don't have the resources to test any of them. To kind of close this out I'd say that if you REALLY care about never missing a frame deadline, then you MUST use Unlimited or Yield, moreover you must set your process to use whatever realtime scheduler your OS supports, and even then you are only going to get soft realtime behavior, with likely quite complex guarantees as to when you can rely on it. For any game I've ever played, SleepAndYield as implemented is just fine.

distransient:
Personally, I think this just needs the missing CHANGELOG url and then it's ready to be merged.

estokes:
What is a changelog url? Is that something I need to do?

distransient:
when you write `[#1446]` that means you are writing a link, which needs its url specified below like so:

image

estokes:
I added it in the fixed section with all the other ones, does that look correct now?

distransient:
Sorry, you linked the relevant issue, and it should be this PR 😅, aka [#1446].

estokes:
ok sorry, I think I see what you want now.

distransient:
bors r=azriel91, distransient, Moxinilian

PR #1449, opened 2019-03-05
Correct euler angle argument names and documentation in Transform

termhn:
## Modifications

- Changes the `set_euler_angles` method to use `roll, pitch, yaw` arguments instead of `x, y, z` and correct the documentation to match. The old arguments and documentation was incorrect, as x =/= roll, y =/= pitch, z =/= yaw.

## PR Checklist

### DIFF
--- a/amethyst_core/src/transform/components/local_transform.rs
+++ b/amethyst_core/src/transform/components/local_transform.rs
@@ -345,11 +345,11 @@ impl Transform {
         self
     }
 
-    /// Set the rotation using Euler roll, pitch, yaw. 
+    /// Set the rotation using Euler roll, pitch, yaw.
     ///
     /// Note that the *order* of arguments is different than you may be used
     /// to thinking about Euler angle transforms, and does not directly match
-    /// the customary x,y,z axis ordering.
+    /// the customary x, y, z axis ordering.
     ///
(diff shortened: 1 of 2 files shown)


amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -345,15 +345,19 @@ impl Transform {
         self
     }
 
-    /// Set the rotation using Euler x, y, z.
+    /// Set the rotation using Euler roll, pitch, yaw. 
+    ///
+    /// Note that the *order* of arguments is different than you may be used
+    /// to thinking about Euler angle transforms, and does not directly match
+    /// the customary x,y,z axis ordering.
```
azriel91: btw can you add spaces after the commas, my `englishfmt` returns 1

termhn:
👍

amethyst_core/src/transform/components/local_transform.rs
```diff
@@ -345,15 +345,19 @@ impl Transform {
         self
     }
 
-    /// Set the rotation using Euler x, y, z.
+    /// Set the rotation using Euler roll, pitch, yaw.
```
azriel91: there's a trailing space, that's why rustfmt cried 😭

termhn:
This is actually incorrect... making new issue now with more detailed information.

PR #1452, opened 2019-03-06
Bugfix/subtract failed assets from loading count

azriel91:
## Description

When `tracker.fail(..)` is called in `AssetStorage#process()` [[1], [2]], we don't requeue. Therefore, when an asset fails, we should not only add 1 to number of failed assets, but also subtract 1 from `ProgressCounterTracker#num_loading`. Currently we do the first half [[3]].

[1]: <https://github.com/amethyst/amethyst/blob/v0.10.0/amethyst_assets/src/storage.rs#L195>
[2]: <https://github.com/amethyst/amethyst/blob/v0.10.0/amethyst_assets/src/storage.rs#L230>
[3]: <https://github.com/amethyst/amethyst/blob/v0.10.0/amethyst_assets/src/progress.rs#L154>

## Modifications

* `ProgressCounter#num_loading()` no longer includes failed assets.

## PR Checklist

### DIFF
--- a/amethyst_assets/src/progress.rs
+++ b/amethyst_assets/src/progress.rs
@@ -80,9 +80,9 @@ impl ProgressCounter {
         self.num_loading.load(Ordering::Relaxed)
     }
 
-    /// Returns the number of assets this struct is tracking.
+    /// Returns the number of assets that have successfully loaded.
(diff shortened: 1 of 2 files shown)


azriel91:
Thank you both!
bors r=Moxinilian,distransient

PR #1474, opened 2019-03-20
Solid and gradient colour sprite sheets generated at runtime

azriel91:
## Description

Provide code to generate sprite sheets programmatically.

## Additions

* `ColourSpriteSheetGen::solid` and `ColourSpriteSheetGen::gradient` can be used to generate sprite sheets at runtime.

## PR Checklist

- **deferred** Updated the content of the book if this PR would make the book outdated.


amethyst_renderer/src/sprite/colour_sprite_sheet.rs
```diff
@@ -0,0 +1,682 @@
...
+        // Pixel count.
+        let padding_pixels = if padded { 1 } else { 0 };
+        let sprite_w_pad = sprite_w + padding_pixels;
+        let sprite_h_pad = sprite_h + padding_pixels;
+        let image_width = sprite_w_pad as usize * column_count;
+        let image_height = sprite_h_pad as usize * row_count;
+        let pixel_count = image_width * image_height;
+
+        // Element count.
+        let pixel_width = 4; // 4 channels per pixel
```
jojolepro: could be a const at the top of the file

amethyst_renderer/src/sprite/colour_sprite_sheet.rs
```diff
@@ -0,0 +1,682 @@
...
+    }
+
+    /// Returns a `SpriteRender` that holds a reference to a gradient spritesheet.
+    ///
+    /// This generates a sprite for each colour between `colour_begin` and `colour_end` (inclusive).
+    /// The number of sprites in the sprite sheet is equal to the `sprite_count` parameter.
+    ///
+    /// # Parameters
+    ///
+    /// * `colour_sprite_gen_data`: System data needed to load colour sprites.
```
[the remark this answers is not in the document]
azriel91: ah we use `color`, just checked `UiButtonBuilder`. Shall swap english mode

amethyst_renderer/src/sprite/colour_sprite_sheet.rs
```diff
@@ -0,0 +1,682 @@
...
+        if sprite_count < 2 {
+            panic!(
+                "`sprite_count` must be at least 2, received: `{}`.",
+                sprite_count
+            );
+        }
+
+        let sprite_sheet_handle = {
+            let column_count = sprite_count.integer_sqrt();
+            let row_count = {
```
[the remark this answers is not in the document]
azriel91: clever 💡

amethyst_renderer/src/sprite/colour_sprite_sheet.rs
```diff
@@ -0,0 +1,682 @@
...
+        let channel_steps =
+            Self::channel_steps(sprite_count, colour_begin, colour_end, pixel_width);
+
+        (0..row_count).for_each(|sprite_row| {
+            // 1. Build up a row of pixels
+            // 2. Duplicate the row `sprite_h` times
+            // 3. Add padding pixels if necessary
+            // 4. Repeat
+
+            let capacity = sprite_w_pad as usize * column_count * pixel_width;
```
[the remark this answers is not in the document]
azriel91: yeap

azriel91:
> Nice work. What do you think of a spritesheet with growing size sprites? Like 0 is 1 pixel wide, 2 is the same coordinates but 2 pixels wider, etc.

Ah, yeap that can be done ⚔️ -- split the texture generation from the sprite sheet generation. Should be relatively easy. Shall defer it to another PR -- this one's already 1k lines

> Is it possible to do gradients with multiple color points in them? For example, red to yellow to green to transparent?

Haha, I wanted multiple colour stops as well (and positioning of those stops). Shall get to it later on

amethyst_renderer/src/sprite/sprite_sheet_gen.rs
```diff
@@ -0,0 +1,354 @@
...
+        let pixel_right = (pixel_left + sprite_w) as f32;
+        let pixel_bottom = (pixel_top + sprite_h) as f32;
+        let pixel_left = pixel_left as f32;
+        let pixel_top = pixel_top as f32;
+
+        // Texture coordinates are expressed as fractions of the position on the image.
+        // Y axis texture coordinates start at the bottom of the image, so we have to invert them.
+        //
+        // For pixel perfect result, the sprite border must be rendered exactly at
+        // screen pixel border or use nearest-neighbor sampling.
```
Frizi: This isn't true. This logic breaks down on non-unit display pixel ratios.

What could be instead done is to also shift `image_w` and `image_h` down, so the texture would be trimmed, but not stretched.

Frizi:
I am a bit sceptical about introducing such a huge change to old amethyst renderer, but in this case it looks like it's mostly about supporting special format. We might end up putting most of that upstream to rendy during integration, but we will see.

Another problem is how this is going to be loaded with future asset pipeline? I believe that currently it would be generated during import (that is ahead of time) and copied straight as buffer to gpu texture buffers. I wonder if we would prefer to instead keep procedural generation at runtime.

amethyst_renderer/src/sprite/sprite_sheet_gen.rs
azriel91: hm, am having trouble understanding non-unit display pixel ratios [[1], [2]]. I need a less text, more picture explanation `@_@` (or a revisit of those pages when my brain can take it).

[1]: https://stackoverflow.com/questions/8785643/what-exactly-is-device-pixel-ratio
[2]: http://weibeld.net/webdesign/device-pixel-ratio.html

jojolepro:
@azriel91 consider using PartialFunction with x values from 0 to 1 for complex gradients. You can code a function to do the interpolation for you that way.

Moxinilian:
@Frizi both runtime and import time have use cases. If your gradient is just a health bar, it can be generated in the pipeline. But if it depends on user input, you have to do it at runtime.

amethyst_renderer/src/sprite/sprite_sheet_gen.rs
Frizi: Basically think about that this way: Your sprite has size (W,H), while your part of texture that you map to this sprite has size (W-2p, H-2p) where p is your padding. This obviously forces it to stretch. You don't see that stretch on nearest-neighbour if it happens to cover exactly (W,H) physical pixels on your display, because how else could it display it? Once you add more pixels (your display is denser than your texture), then that stretching becomes visible. To counteract that, you have to somehow prevent stretching. You can shrink down the sprite itself to prevent stretching and it should work, but then on denser displays you must expect that sprites will actually have visible space between them.

No matter how you put it, shrinking the UVs generates more problems than it solves. There is no valid way to do it without actually modifying texture or preparing your assets accordingly.

Usually what's done to counteract that is every sprite has added 1px border outside of it boundaries that matches exactly the pixels in the sprite. In case of repeating sprites, this border has "continue" the pattern, basically copying the 1px row/column from opposite side. In that case you just continue to use standard `0..1` UVs and it will all work correctly.

Frizi:
I agree, we probably want both. Though in case of gradients that responds to user input, doing it in shaders would be more preferred. Working with images on CPU and uploading them to GPU every frame will very quickly become a bottleneck.

Moxinilian:
By user input I mean potentially having a color attributed to each user, generated at load time from a remote database.

fhaynes:
> I am a bit sceptical about introducing such a huge change to old amethyst renderer, but in this case it looks like it's mostly about supporting special format. We might end up putting most of that upstream to rendy during integration, but we will see.

@Frizi Until the integration actually happens, the existing renderer can continue to be updated. It isn't the old one, it is the one we have that is working. An indefinite freeze on it isn't viable.

amethyst_renderer/src/sprite/sprite_sheet_gen.rs
azriel91: Understand the shrinking's problems for an object's sprite sheet; the shrinking UVs is okay for the single-colour sprites right?
I can come up with another PR that adds the repeated pixels around each sprite to avoid shrinking the UVs

azriel91:
rebased over `master`

@Frizi 
for single pixel sprites, i think having the half pixel offset is alright -- we don't need to generate an extra pixel border with the sprite's edge pixels. Could implement the extending of each sprite with a 1 pixel perimeter, copying the edge pixel's values in another PR

Frizi:
I think we've actually covered parts of that in rendy integration, implemented as part of the prefab system.
https://github.com/Frizi/amethyst/blob/a36e886bd2e8402c012039d2e0c5f667faa42960/amethyst_rendy/src/formats/texture.rs#L54

You are able to load a four-corner texture with chosen colors, and apply linear filtering to it with GPU texture sampler (which is configurable as part of texture prefab as well). Now because SpriteSheetPrefab is using a texture prefab, all of that also applies to sprites.

The limitation of course is that you are able to have only linear gradients with at most of 4 distinct colors in the corners, and this applies to the whole spritesheet texture (not individual sprites). Still, this can be easily lifted by adding more specific implementations as additional generator variants.

azriel91:
cool, adding more detail from chat:

> The idea is that you generate a texture of four corner pixels with your chosen colors. This is what the prefab code does for you.
> Then, you use that texture with bilinear filtering. This is done by specifying the sampler info, also in the ron file.
>
> Relevant parts:
>
> * https://github.com/Frizi/amethyst/blob/a36e886bd2e8402c012039d2e0c5f667faa42960/amethyst_rendy/src/formats/texture.rs#L54
> * https://github.com/Frizi/amethyst/blob/a36e886bd2e8402c012039d2e0c5f667faa42960/amethyst_rendy/src/formats/texture.rs#L99-L108
> * https://github.com/Frizi/amethyst/blob/a36e886bd2e8402c012039d2e0c5f667faa42960/examples/assets/prefab/animation.ron#L75-L82
>
> In the last one, you would specify `Linear` instead of `Nearest` to get a gradient.

when `rendy` and `atelier` are integrated, should open another PR that uses the linked logic to generate the texture, and:

1. write a custom importer to extend each sprite's border by 1px
2. pass the gradient sprite texture through that
3. generate the sprite sheet coordinates at the edges of sprites

PR #1479, opened 2019-03-24
Use paddle constants to avoid paddle size bug

Reificator:
## Description

Fix the last remaining paddle size bug in the pong example, where the paddle's bounding box is initialized as smaller than its sprite.  I encountered this bug today while following the `latest` release of the book, but it looks like someone has mostly cleaned up `master` since then.

## Modifications

- Replace 1.0 width/height for paddle with defined constants.

## PR Checklist

- [ ] Ran `cargo +stable fmt --all` locally if this modified any rs files. **Rustup claims the required package does not exist for x86_64-pc-windows-msvc.  Also, I did not modify formatting**


azriel91:
Thanks!
bors r=jojolepro,happenslol

PR #1480, opened 2019-03-24
Add functionality to keep the aspect ratio when stretching a UiTransform

0x6273:
## Description

Makes it possible for `UiTransform`s to keep their aspect ratio when stretched to fit their parent. 
This is a breaking change, because it adds a new field to the `Stretch::XY` struct.

## Modifications

- Changed the `Stretch::XY` enum variant struct, adding a new `keep_aspect_ratio` field.
- Updated the `UiTransform` system to use this new field when stretching transforms.
- Updated all the places where `Stretch::XY` is instantiated, with this new field set to false.

## PR Checklist

- [ ] Added unit tests for new APIs if any were added in this PR. **Not sure how to add a test for this since it involves running a system with entities.**

### DIFF
--- a/amethyst_ui/src/button/builder.rs
+++ b/amethyst_ui/src/button/builder.rs
@@ -341,6 +341,7 @@ impl<G: PartialEq + Send + Sync + 'static> UiButtonBuilder<G> {
                     .with_stretch(Stretch::XY {
                         x_margin: 0.,
                         y_margin: 0.,
+                        keep_aspect_ratio: false,
                     }),
             )
             .expect("Unreachable: Inserting newly created entity");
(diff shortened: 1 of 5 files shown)


0x6273:
> Thanks! Can you add documentation explaining that margins are ignored if you are forcing the aspect ratio? Alternatively, you could add a new enum variant on `Stretch` for this purpose, which would be clearer.

The x/y_margin values are not ignored, but treated as the minimum amount of margin rather than the absolute amount . More margin will be added to either the horizontal or vertical axis when necessary to keep the aspect ratio. I have changed the doc comment to clarify this.

PR #1481, opened 2019-03-24
Fix renderer mesh div by zero

### LINKED ISSUE #1476: [BUG] Attempted to divide by zero with vertex length at 0
## Description

I want to run an algorithm to change a `Mesh` every few frames. Therefore, in my `SimpleState::on_start` I make a dummy entity (That is really only a dummy for a few frames). This is considered a dummy entity because there are 0 vertices attached to its mesh. Unfortunately creating a `Mesh` with 0 vertices causes an error of dividing by zero at [`amethyst_renderer\src\mesh.rs:68`](https://github.com/amethyst/amethyst/blob/18a2855bb4c6f660f6ae4cf62731242910909cf3/amethyst_renderer/src/mesh.rs#L68) because it attempts to calculate the stride of the buffer by dividing. 

## Reproduction Steps

I think that the easiest way to reproduce it is by cloning the [`asset_loading`](https://github.com/amethyst/amethyst/tree/master/examples/asset_loading) example and changing [line 79](https://github.com/amethyst/amethyst/blob/master/examples/asset_loading/main.rs#L79) to this:
```
let mesh: Handle<Mesh> = loader.load_from_data(Vec::<PosNormTex>::new().into(), (), meshes);
```

## What was Expected to Happen

It should've run just fine, ie. it would create an empty mesh.

## What Actually Happened

It attempted to divide by zero.

## The Environment
```
Version: 0.10.0
Platform: x86_64-pc-windows-msvc
Amethyst git commit: bc105a7fc18aea8497eeda8850124404529682d0
Rustc version: 1.34.0-nightly Nightly
Rustc git commit: 097c04cf433048585dd9e3f63b30e03cc3509e4b
```

distransient:
Fixes #1476

## PR Checklist

- ***N/A*** Updated the content of the book if this PR would make the book outdated.
- ***N/A*** Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.
- ***N/A*** Added unit tests for new APIs if any were added in this PR.

### DIFF
--- a/amethyst_renderer/src/mesh.rs
+++ b/amethyst_renderer/src/mesh.rs
@@ -6,6 +6,7 @@ use std::{
 };
 
 use gfx::Primitive;
+use log::warn;
 use serde::{Deserialize, Serialize};
 
 use amethyst_assets::Handle;
@@ -65,7 +66,10 @@ where
 
         let verts = self.0.as_ref();
         let slice = cast_slice(verts);
-        let stride = slice.len() / verts.len();
+        let stride = slice.len().checked_div(verts.len()).unwrap_or_else(|| {
+            warn!("Loading mesh with zero vertices.");
+            0
+        });
         let role = Role::Vertex;
         let bind = Bind::empty();
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -65,6 +65,7 @@ extra bounds from `AnimatablePrefab` and `AnimationSetPrefab` ([#1435])
 * Fixed default system font loading to accept uppercase extension ("TTF"). ([#1328])
 * Set width and height of Pong Paddles ([#1363])
 * Fix omission in `PosNormTangTex` documentation. ([#1371])
+* Fix division by zero in vertex data building ([#1481])
 
 [#1114]: https://github.com/amethyst/amethyst/pull/1114
 [#1213]: https://github.com/amethyst/amethyst/pull/1213
@@ -101,6 +102,7 @@ extra bounds from `AnimatablePrefab` and `AnimationSetPrefab` ([#1435])
 [#1454]: https://github.com/amethyst/amethyst/pull/1454
 [#1442]: https://github.com/amethyst/amethyst/pull/1442
 [#1469]: https://github.com/amethyst/amethyst/pull/1469
+[#1481]: https://github.com/amethyst/amethyst/pull/1481
 
 ## [0.10.0] - 2018-12
 


amethyst_renderer/src/mesh.rs
```diff
@@ -65,7 +65,11 @@ where
 
         let verts = self.0.as_ref();
         let slice = cast_slice(verts);
-        let stride = slice.len() / verts.len();
+        let verts_len = verts.len();
+        if verts_len == 0 {
+            return Err(Error::from_string("Number of vertices in mesh is zero."));
```
Frizi: I think this could be a `log::warn` instead. Hard errors on user controlled data are annoying to deal with. When `verts_len` is 0, `slice.len()` will be 0 as well. The stride can be just set to 0 and it should all work out.

amethyst_renderer/src/mesh.rs
```diff
@@ -65,7 +65,11 @@ where
 
         let verts = self.0.as_ref();
         let slice = cast_slice(verts);
-        let stride = slice.len() / verts.len();
+        let verts_len = verts.len();
```
vessd: Сan be replaced by
```rust
let stride = slice
    .len()
    .checked_div(verts.len())
    .ok_or_else(|| Error::from_string("Number of vertices in mesh is zero."))?;
```

amethyst_renderer/src/mesh.rs
distransient: Why have the Result when we're not going to use it?

distransient: Actually, I think the better question for me to ask would be is it useful to actually build empty vertex buffers? You would know this better I think. I see how the Result is being used now, but I'm not sure whether you care about having empty vertex buffers sent to the GPU

Frizi: Well, actually creating 0-sized buffers seems like a problem, and there isn't much use in empty meshes anyway. The draw call with empty mesh, if at all valid, would result in no primitives drawn.

I'm just a bit worried that this might cause some spontaneous crashes on hot-reloading when somebody will clear up the mesh in their's editor and save the file. We might handle that in asset pipeline though :)

distransient: Gotcha. Gonna use the code suggested by @vessd and instead warn!+return zero in the closure.

amethyst_renderer/src/mesh.rs
distransient: Ty!

Frizi:
bors r=azriel91, frizi

PR #1482, opened 2019-03-24
Added pass for PosColor

norman784:
@fhaynes clean PR, hope it doesn't take too long to review.


amethyst_renderer/src/pass/shaders/fragment/flat_color.glsl
```diff
@@ -0,0 +1,16 @@
+// TODO: Needs documentation.
```
happenslol: Are you gonna add the documentation for this and the vertex shader in this PR? It's not too much code, so I don't think too much documentation is needed.

norman784: Removed

norman784:
I've renamed the structs.

Yes, the pong example and all other examples will need maybe a rewrite or fix after the rendy gets implemented, so there's no point to do it now.

amethyst_renderer/src/pass/shaders/fragment/flat_color.glsl
happenslol: Thanks!

amethyst_renderer/src/pass/flat_colored/separate.rs
```diff
@@ -0,0 +1,240 @@
...
+        ReadStorage<'a, JointTransforms>,
+        ReadStorage<'a, Rgba>,
+    );
+}
+
+impl Pass for DrawFlatColoredSeparate {
+    fn compile(&mut self, effect: NewEffect<'_>) -> Result<Effect, Error> {
+        use std::mem;
+        let mut builder = if self.skinning {
+            create_skinning_effect(effect, FRAG_SRC)
```
omni-viral: Basic vertex shader with skinning uses other vertex attributes - texture coordinates instead of color.

amethyst_renderer/src/pass/flat_colored/separate.rs
```diff
@@ -0,0 +1,240 @@
...
+                }
+            }
+            Some(ref visibility) => {
+                for (joint, mesh, material, global, rgba, _) in (
+                    joints.maybe(),
+                    &mesh,
+                    &material,
+                    &global,
+                    rgba.maybe(),
+                    &visibility.visible_unordered,
```
jojolepro: Make sure it is correct to not have !&hidden here

fhaynes:
@norman784 did you get a chance to review @omni-viral's review?

norman784:
@fhaynes I didn't have time to review the changes because it involves to dig the source code to understand how it works from below, I'm sorry that could not have the time to check it, also saw that they are a lot of commits ahead of my fork so it will be a mess again to try to merge it, so in order to do a clean PR I need to create another one, but I'm a bit busy with other stuff that takes all my free time, so I could not commit to do this right now.

torkleyy:
Unfortunately, this is now out of date due to Rendy integration. I'm going to close this, we'll need a new PR for this.
Thank you @norman784!

PR #1484, opened 2019-03-26
Allows to refer to SpriteSheetPrefab by name. 

vessd:
## Description

Allows to refer to SpriteSheetPrefab by name. This can be useful when you load prefabs from several files.

## Additions

- `name` field to `SpriteSheetPrefab::Sheet`

## PR Checklist

### DIFF
--- a/amethyst_renderer/src/sprite/prefab.rs
+++ b/amethyst_renderer/src/sprite/prefab.rs
@@ -181,6 +181,7 @@ impl SpriteSheetLoadedSet {
 }
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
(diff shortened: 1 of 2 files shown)


vessd:
> dunno if it's a good idea in the long term

I have no idea. I think that the prefab system should be reworked, after which this implementation can be removed, but I need this solution right now.
Tell me, if I can help with the new prefab system, which will allow to split the prefab into several files and cross-references between them.

> Needs a changelog entry 

Change log from unstable to unstable looks bad...
I will change enum to untagged, then there will be no breaking changes.

azriel91:
bors r=jojolepro, azriel91

PR #1495, opened 2019-03-29
Adding a register line in the creating entities section of the book

Usummon:
## Description

The with() function used to create entities section of the book requires that the component is first registered which isn't done so in the book.

I'm new to programming and GitHub so please tell me if I am doing something incorrectly.

## Modifications

- added register line to the concepts\world\creating entities

## PR Checklist

### DIFF
--- a/book/src/concepts/world.md
+++ b/book/src/concepts/world.md
@@ -74,6 +74,7 @@ Please note that **in order to use this syntax, you need to import the ``amethys
 # }
 # fn main() {
 #   let mut world = amethyst::ecs::World::new();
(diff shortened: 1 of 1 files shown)


azriel91:
Yeap this is all good!

Some extra knowledge (and possibly why this bug was there):

* Currently in `specs`, there is auto registration magic -- a type that implements the `SystemData` trait will `register` the component type when the system with that `SystemData` is part of a dispatcher.
* When we aren't using that (like in the snippet you fixed), then we have to do it manually, and it's easy to forget.

azriel91:
bors r=azriel91

PR #1496, opened 2019-03-30
Fix some clippy lint warnings

0x6273:
## Description

This fixes 68 warnings. (Only 121 more to go!)

## Modifications

- Various code style improvements.
- Fix a bug where objects would be needlessly allocated in .unwrap_or() and
  similar functions.

## PR Checklist


amethyst_renderer/src/resources.rs
```diff
@@ -83,8 +83,8 @@ impl ScreenDimensions {
     /// Creates a new screen dimensions object with the given width and height.
     pub fn new(w: u32, h: u32, hidpi: f64) -> Self {
         ScreenDimensions {
-            w: w as f64,
-            h: h as f64,
+            w: f64::from(w),
```
LucioFranco: hmm not sure if this one is much better

0x6273: Yeah it doesn't look as good, but apparently doing it this way is considered best practice. From the [clippy lint docs](https://rust-lang.github.io/rust-clippy/master/index.html#cast_lossless):

> Rust’s as keyword will perform many kinds of conversions, including silently lossy conversions. Conversion functions such as i32::from will only perform lossless conversions. Using the conversion functions prevents conversions from turning into silent lossy conversions if the types of the input expressions ever change, and make it easier for people reading the code to know that the conversion is lossless.

PR #1509, opened 2019-04-13
Track binary files with git-lfs

Frizi:
## Description

Track all binary files with git-lfs

## PR Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,5 @@
 # Set the default behavior, in case people don't have core.autocrlf set.
 * text=auto
-
 # Explicitly declare text files you want to always be normalized and converted
 # to native line endings on checkout.
 *.md text
@@ -9,7 +8,11 @@
 *.yml text
 COPYING text
 LICENSE-* text
-
-# Denote all files that are truly binary and should not be modified.
-*.obj binary
-*.custom binary
+# LFS
+examples/assets/large/*.obj filter=lfs diff=lfs merge=lfs -text
+examples/assets/large/*.png filter=lfs diff=lfs merge=lfs -text
+examples/assets/large/*.ogg filter=lfs diff=lfs merge=lfs -text
+examples/assets/large/*.dds filter=lfs diff=lfs merge=lfs -text
+examples/assets/large/*.glb filter=lfs diff=lfs merge=lfs -text
+examples/assets/large/*.bin filter=lfs diff=lfs merge=lfs -text
+examples/assets/large/*.jpg filter=lfs diff=lfs merge=lfs -text
--- a/README.md
+++ b/README.md
@@ -78,6 +78,14 @@ If you don't understand a part of the documentation, please let us know. Join us
 
 ## Getting started
 
+### Before you begin
+
+This repository uses Git LFS for some files used in examples.
+If you intend to run the examples, make sure you have LFS installed in your system before you clone.
+You can download it and read the installation instructions at [Git LFS home page](https://git-lfs.github.com/).
+
+### Examples
+
 To compile any of the examples run:
 
 ```
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -64,6 +64,7 @@ extra bounds from `AnimatablePrefab` and `AnimationSetPrefab` ([#1435])
 * `amethyst_test` returns the panic message of a failed execution. ([#1499])
 * Rename `NetEvent::Custom` variant to `NetEvent::Unreliable`. ([#1513])
 * Updated laminar to 0.2.0. ([#1502])
+* Large binary files in examples are now tracked with `git-lfs`. ([#1509])
 
 ### Removed
 
@@ -119,6 +120,7 @@ extra bounds from `AnimatablePrefab` and `AnimationSetPrefab` ([#1435])
 [#1501]: https://github.com/amethyst/amethyst/pull/1501
 [#1502]: https://github.com/amethyst/amethyst/pull/1515
 [#1513]: https://github.com/amethyst/amethyst/pull/1513
+[#1509]: https://github.com/amethyst/amethyst/pull/1509
 
 ## [0.10.0] - 2018-12
 
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -42,7 +42,8 @@ and try our best to resolve it.
 
 So, you want to write some code? Great!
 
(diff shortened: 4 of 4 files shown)


0x6273:
I don't think this is a very good idea. The few binary files that are tracked are not very big, and are not expected to change frequently, so there isn't much of benefit to using lfs over regular git blobs.

Frizi:
> I don't think this is a very good idea. The binary files that are tracked are not very big, and are not expected to change frequently, so there isn't much of benefit to using lfs over regular git blobs.

The benefit is that we can add more files from now on, without worrying about repository size.

> Pong example doesn't run.

Can you describe your issue? It does work for me.

> The big files are still present in the git history. Clearing it requires to rewrite the history.

I am aware of this. We don't want to rewrite the history, so we are just going to keep the copy of old files in. Alternatives are just too costly, and not doing anything also prohibits checking in new files. This change is intended as a starting point, so all future binary files will be stored exclusively on lfs. I am about to add a lot of test models with rendy integration, and don't want to bloat the repository.

> Requires an extra dependency to be installed on the computer of everyone developing amethyst.

I am aware of that. This is also stated in the readme/contributing. We would just have to adapt. I don't think this is a big deal, as git-lfs is widely available, very easy to install and don't really require any extra knowledge to use. Standard git commands works as expected.

>The PR description doesn't describe why this change is useful.

I hope this comment makes this cleaner :)

0x6273:
> The benefit is that we can add more files from now on, without worrying about repository size.

If there's no benefit to moving the files we have now to lfs, I say we keep them as is. If we need to store large files in the future, like assets for an example game, we can use lfs for only those files, and not break all book and all the other examples for users that don't have the lfs extension installed.

Frizi:
I think you need to have lfs installed to properly clone the repository. Also having the files tracked or not individually would require listing them all individually in the `.gitattributes`. I've choosen to instead go with the simple route of no exceptions or special cases. You have to have git-lfs installed, that's a hard dependency.

0x6273:
> I think you need to have lfs installed to properly clone the repository. Also having the files tracked or not individually would require listing them all individually in the `.gitattributes`. I've choosen to instead go with the simple route of no exceptions or special cases. You have to have git-lfs installed, that's a hard dependency.

You can clone the repo without lfs installed, but files that are stored in lfs will be missing. We can have the lfs extension be an optional thing required only for some examples with large assets, and have all the other examples and the book work as expected with just a `git clone`.

fhaynes:
> > I think you need to have lfs installed to properly clone the repository. Also having the files tracked or not individually would require listing them all individually in the `.gitattributes`. I've choosen to instead go with the simple route of no exceptions or special cases. You have to have git-lfs installed, that's a hard dependency.
> 
> You can clone the repo without lfs installed, but files that are stored in lfs will be missing. We can have the lfs extension be an optional thing required only for some examples with large assets, and have all the other examples and the book work as expected with just a `git clone`.

@0x6273 are you thinking on a per-example basis, or a per-file basis? The per-file basis will result in an ever-growing .gitattributes file.

jojolepro:
```
[INFO][winit::platform::platform::x11::window] Guessed window DPI factor: 1
[INFO][amethyst::app] Initializing Amethyst...
[INFO][amethyst::app] Version: 0.10.0
[INFO][amethyst::app] Platform: x86_64-unknown-linux-gnu
[INFO][amethyst::app] Amethyst git commit: e256738fda766dd9144d2db7730eab51ad0defce
[INFO][amethyst::app] Rustc version: 1.35.0-nightly Nightly
[INFO][amethyst::app] Rustc git commit: f694222887cf31f51e68927716c25736e62f037f
[ERROR][amethyst_assets::storage] "renderer::Texture": Asset "texture/pong_spritesheet.png" (handle id: Handle { id: 7 }) could not be loaded: Failed to load asset with name "texture/pong_spritesheet.png"
[ERROR][amethyst_assets::progress] Error loading handle 7, renderer::Texture, with name texture/pong_spritesheet.png, caused by: Error { inner: Inner { source: Some(Error { inner: Inner { source: Some(Error { inner: Inner { source: Some(Error { inner: Inner { source: None, backtrace: None, error: FormatError("invalid signature") } }), backtrace: None, error: DecodeImageError } }), backtrace: None, error: Format("PNG") } }), backtrace: None, error: Asset("texture/pong_spritesheet.png") } }
caused by: Error { inner: Inner { source: Some(Error { inner: Inner { source: Some(Error { inner: Inner { source: None, backtrace: None, error: FormatError("invalid signature") } }), backtrace: None, error: DecodeImageError } }), backtrace: None, error: Format("PNG") } }
caused by: Error { inner: Inner { source: Some(Error { inner: Inner { source: None, backtrace: None, error: FormatError("invalid signature") } }), backtrace: None, error: DecodeImageError } }
caused by: Error { inner: Inner { source: None, backtrace: None, error: FormatError("invalid signature") } }
[ERROR][amethyst_assets::progress] Note: to handle the error, use a `Progress` other than `()`
[WARN][amethyst_renderer::pass::flat2d::interleaved] Texture not loaded for texture: `Handle { id: 7 }`.
[WARN][amethyst_renderer::pass::flat2d::interleaved] Texture not loaded for texture: `Handle { id: 7 }`.
[WARN][amethyst_renderer::pass::flat2d::interleaved] Texture not loaded for texture: `Handle { id: 7 }`.
Exiting program because of 1 errors...
0: DecoderError
```
Commands:
```
 1047  cd /tmp
 1048  git clone https://github.com/frizi/amethyst -b lfs
 1049  cd amethyst
 1050  cargo run --example pong
```

0x6273:
> @0x6273 are you thinking on a per-example basis, or a per-file basis? The per-file basis will result in an ever-growing .gitattributes file.

I was thinking just a single line like `examples/assets/large/* filter=lfs diff=lfs merge=lfs -text`.

0x6273:
@jojolepro You need to run `git lfs install` before cloning.

fhaynes:
I think, for now, let's create `examples/assets/large` and put them in there. We can revisit the topic after the integration. I think splitting out examples into their own repo is worth discussing. @Rhuagh made a good point that many of our examples right now are actually integration tests.

Frizi:
I have changed the lfs tracked paths to subdirectory. This means currently no files are being tracked with lfs and all existing files stay as they were. The change is purely for the future additions, and i'm going to use that on rendy branch.

Frizi:
bors r=fletcher, jojolepro

PR #1513, opened 2019-04-15
Allow NetConnection to send reliable OR unreliable events

CosineP:
## Description

`NetEvent::Custom` was split into `NetEvent::Reliable` and `NetEvent::Unreliable`, and based on that when laminar packets are made reliability is chosen

Also, when doing so, I realized that Connect, Disconnect, etc are sent unreliably when (due to the possibility of packet loss) they really should be reliable, so that's changed as well.

## Additions

- NetEvent::Reliable(T)

## Modifications

- NetEvent::Custom(T) renamed to NetEvent::Unreliable(T)
- All current builtin NetEvent types are sent reliably

## PR Checklist

No unit tests in this crate in the first place tbh

### DIFF
--- a/amethyst_network/src/lib.rs
+++ b/amethyst_network/src/lib.rs
@@ -28,6 +28,30 @@ mod network_socket;
 mod server;
 mod test;
 
+/// Each event type is either reliable or unreliable:
+/// Reliable events always reach their destination,
+/// Unreliable events may be lost
+/// For Amethyst-defined events, whether it's reliable is specified in this function,
+/// Otherwise, it's specified by the use of NetEvent::Reliable vs NetEvent::Unreliable
+fn is_reliable<T>(event: NetEvent<T>) -> bool {
+	use NetEvent as NE;
+	match event {
+		// I specify them all explicitly so the typechecker can save
+		// us from the mistake of specifying a builtin that SHOULD be
+		// unreliable, but is assumed to be unreliable like all the rest
+		| NE::Connect {..}
+		| NE::Connected {..}
+		| NE::ConnectionRefused {..}
+		| NE::Disconnect {..}
+		| NE::Disconnected {..}
+		| NE::TextMessage {..}
+		| NE::Reliable(_)
+		=> true,
+		| NE::Unreliable(_)
+		=> false,
+	}
+}
+
 /// Sends an event to the target NetConnection using the provided network Socket.
 /// The socket has to be bound.
 pub fn send_event<T>(event: NetEvent<T>, addr: SocketAddr, sender: &SyncSender<ServerSocketEvent>)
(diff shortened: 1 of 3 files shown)


amethyst_network/src/lib.rs
```diff
@@ -28,6 +28,28 @@ mod network_socket;
 mod server;
 mod test;
 
+/// Each event type is either reliable or unreliable:
+/// Reliable events always reach their destination,
+/// Unreliable events may be lost
+/// For Amethyst-defined events, whether it's reliable is specified in this function,
+/// Otherwise, it's specified by the use of NetEvent::Reliable vs NetEvent::Unreliable
+fn is_reliable<T>(event: NetEvent<T>) -> bool {
```
LucioFranco: this should be moved to an fn under `NetEvent`

LucioFranco:
another thing while im on this, we probably do not want to expose the enum as its not backwards compat? We probably want to provide fns for it and make the inner enum private.

amethyst_network/src/lib.rs
LucioFranco: and just to clairfy this is a fn that takes a `NetEvent` which means to call this it must be imported manually and then called with a netevent. In this case its easier for the user to have a netevent and called `my_netevent.is_reliable()`.

TimonPost:
> Though I think we should have the is_reliable fn and even others under the event enum

Yea, that is a better style.

> another thing while I'm on this, we probably do not want to expose the enum as its not backward compat? 

Yea, for now I think it is fine, I am going to make the other PR soon implementing the ability of the user to order hist packets. It might be better indeed to do it with functions. The same as we do with `Packet` in laminar.

CosineP:
> another thing while I'm on this, we probably do not want to expose the enum as its not backward compat?

i don't understand

amethyst_network/src/net_event.rs
```diff
@@ -54,25 +54,25 @@ impl<T> NetEvent<T> {
...
-	        NE::Unreliable(_) => false,
-	    }
-	}
+    /// Each event type is either reliable or unreliable:
+    /// Reliable events always reach their destination,
+    /// Unreliable events may be lost
+    /// For Amethyst-defined events, whether it's reliable is specified in this function,
+    /// Otherwise, it's specified by the use of NetEvent::Reliable vs NetEvent::Unreliable
+    pub fn is_reliable(&self) -> bool {
+        use NetEvent as NE;
```
LucioFranco: als one protip here you can also just do `use NetEvent::*` and then you can drop all the `NE::`

CosineP: I tried that but within match arms it took them to be variable names, not enumerants

LucioFranco:
@CosineP so what I mean by this is, if we expose the raw enum and we add an item for example, it will break code in the sense that if you have an exhaustive match it will break. So to avoid this, you make the enum private and use functions to expose the functionality.

CosineP:
oh ok, i see what you mean

TimonPost:
bors r=LucioFranco, TimonPost

PR #1515, opened 2019-04-15
Updated laminar to 0.2.0

TimonPost:
## Description

Updated to laminar 0.2.0, which resulted in some removal of existing logic, and API changes.

## Removals

Those are removed because it was logic surrounding two UDP sockets. Now with laminar 0.2.0 we can do it with one. 
- server/receive_handler.rs
- server/send_handler.rs
- server/server_socket_event.rs
- src/server/udp.rs

The following API changes occurred:
- `NetConnection` now accepts one address for the target UDP socket.
- `NetworkBundle`  takes in one address for the host UDP socket.

## PR Checklist

- [X ] Ran `cargo +stable fmt --all` locally if this modified any rs files.


amethyst_network/src/connection.rs
```diff
@@ -62,10 +59,7 @@ impl<E: Send + Sync + 'static> NetConnection<E> {
 
 impl<E> PartialEq for NetConnection<E> {
     fn eq(&self, other: &Self) -> bool {
-        self.target_receiver == other.target_receiver
-            && self.state == other.state
-            && self.target_sender == other.target_sender
-            && self.state == other.state
+        self.target_addr == other.target_addr && self.state == other.state
```
LucioFranco: I want to say that addr should be the only thing here, since there can only be one connection with said address, therefore its unique.

amethyst_network/src/network_socket.rs
```diff
@@ -85,15 +82,15 @@ where
     }
 
     /// Start a thread to send all queued packets.
-    fn start_sending(sender: Arc<SendHandler>) -> Sender<InternalSocketEvent<E>> {
+    fn start_sending(sender: crossbeam_channel::Sender<Packet>) -> Sender<InternalSocketEvent<E>> {
```
LucioFranco: since we are importing this above, can we import and use `Sender<Packet>` and same thing below on 107.

amethyst_network/src/network_socket.rs
```diff
@@ -107,33 +104,20 @@ where
...
-                                    error!(
-                                        "`NetworkSocketSystem` was dropped. Reason: {:?}",
-                                        error
-                                    );
-                                    break;
-                                }
-                            }
-                            ServerSocketEvent::Error(error) => error!("{:?}", error),
-                            _ => error!("Event not supported"),
+        let receiver = receiver.clone();
```
LucioFranco: I wonder if we dont want to clone here and panic if this is called twice?

amethyst_network/src/server/host.rs
```diff
@@ -1,61 +1,50 @@
...
-            udp_receiver,
+            packet_sender,
+            packet_receiver,
         })
     }
 
     /// Get the handle to the internals of the UDP-receiving threat.
-    pub fn udp_receive_handle(&self) -> Arc<Mutex<ReceiveHandler>> {
-        self.udp_receiver.clone()
+    pub fn udp_receive_handle(&self) -> Receiver<SocketEvent> {
```
LucioFranco: just dropping this here to think about in the future but we probably do not want to expose crossbeam directly but instead wrap it in our own handle struct, this way if crossbeam does decide to make a breaking change we are safe.

amethyst_network/src/test.rs
```diff
@@ -94,7 +76,7 @@ mod test {
             }
         }
         cl_dispatch.dispatch(&mut world_cl.res);
-        sleep(Duration::from_millis(100));
+        sleep(Duration::from_millis(10000));
```
jstnlef: 10 secs? Why so long? It's pretty annoying to have a test last so long.

TimonPost: Ah, it was for testing, should be changed to 100.

amethyst_network/src/server/host.rs
TimonPost: True, I am a bit searching for a correct implementation but it isn't public so we can change it without causing any problems.

TimonPost:
bors +r=LucioFranco

TimonPost:
No, I meant `r=` stupid bors :)

bors r=LucioFranco

PR #1518, opened 2019-04-16
Change fixed_update pass in app to accumulate delta time and iterate at fixed step

### LINKED ISSUE #1516: [FEATURE MODIFICATION] Set up fixed update according to industry standard
## Feature

Fixed update

## Change

Make it work as a fixed update is expected to work, as explained in the classic article, [Fix Your Timestep](https://gafferongames.com/post/fix_your_timestep/).

## Reason

The way it works now is not how it is typically done in game engines and is not useful due to being tied to framerate.

## Impact

It will impact any states that use fixed update

## Additional Notes

The interpolation step as explained in the article is important, so the Time struct needs to be updated immediately after the fixed updates run, and before update. Currently Time is updated after update is called, which makes interpolation impossible.

jaynus:
## Description

This changes the iteration of fixed_update in app.rs to accumulate delta time and then release it for multiple fixed_update iterations. The end result of this is iterating the fixed_update step rate regardless of frame rate, iterating multiple times a frame or only once every fixed_update rate. This fixes #1516 which you can read for more details

## Additions

- Two unit tests for both use cases (faster than FPS, slower than FPS)

## Removals

None

## Modifications

This changes the fixed update API for core::timing::Time, implementing:
- fixed_update_accumulator - Accumulates delta time to track fixed_update execution rates
- start_fixed_update - Called at the beginning of the fixed_update segment
- step_fixed_update - this function returns true/false on whether a fixed update to should, and handles the time accumulator
- finish_fixed_update - Does nothing right now. 

## PR Checklist


JayPavlina:
This looks great. To apply interpolation, we need to know what time it is when `fixed_update` ends (or when `update` begins). Maybe an `Instant` could be added to `Time` and could be set in finish_fixed_update? Alternatively, set it in app.rs before update runs.

JayPavlina:
Or if you want it built-in, (which I think it should be), add an `interpolation_factor` to `Time` and update it before update is run. This makes it easy if you want to build interpolation into the engine later. Calculation is at bottom of [Fix Your Timestep](https://gafferongames.com/post/fix_your_timestep/)

amethyst_core/src/timing.rs
```diff
@@ -156,13 +151,34 @@ impl Time {
...
+            }
+            false => false,
+        }
     }
+
+    /// This function currently doesn't do anything, but is called at the end of a fixed time update series.
+    ///
+    /// This should only be called by the engine.  Bad things might happen if you call this in
+    /// your game.
+    pub fn finish_fixed_update(&mut self) {}
```
torkleyy: Given it's internal only, what's the motivation for adding it now?

EDIT: I see this existed before. So, does it make sense to keep?

jaynus: I've added functionality to this function now, where it updates the current interpolation alpha factor for systems to be able to fetch.

jaynus:
@JayPavlina I've added updating an interpolation_alpha series of methods and updating on finish_fixed_update. I think its up to the consuming functions to extrapolate interpolation past this point, but I believe this framework gives enough tracking and building blocks to make this possible now.

amethyst_core/src/timing.rs
```diff
@@ -156,12 +165,36 @@ impl Time {
...
+    }
+
+    /// Updates the interpolation alpha factor given the current delta time and accumulator.
     ///
     /// This should only be called by the engine.  Bad things might happen if you call this in
     /// your game.
     pub fn finish_fixed_update(&mut self) {
-        self.last_fixed_update += self.fixed_time
+        self.interpolation_alpha = self.fixed_time_accumulator / self.delta_seconds;
+        self.interpolation_alpha_time = secs_to_duration(self.interpolation_alpha);
```
Frizi: I believe alpha is unitless, you are dividing seconds by seconds. It's an interpolation factor in 0..1 range. There is no point in converting it to duration and naming it `seconds`.

Also this calculation should probably be `self.fixed_time_accumulator / sellf.fixed_seconds`.

JayPavlina: What is the purpose of interpolation_alpha_time? I don't understand what it would be used for. The alpha is the only thing needed as far as I know (the number between 0 and 1).

jaynus: I've updated it to as suggested.

jaynus:
Fixed merge conflicts on changeme. I believe this is ready.

fhaynes:
bors r=frizi,torkleyy

PR #1519, opened 2019-04-17
nalgebra: 0.17 -> 0.18

Laaas:
This is to ensure that identical versions are used if ncollide and
nphysics are used in addition to amethyst, which also use nalgebra.

This change is obviously a breaking change, since amethyst re-exports
nalgebra.

## PR Checklist

### DIFF
--- a/amethyst_animation/src/util.rs
+++ b/amethyst_animation/src/util.rs
@@ -4,7 +4,7 @@ use serde::{Deserialize, Serialize};
 
 use amethyst_core::{
     ecs::prelude::{Entity, WriteStorage},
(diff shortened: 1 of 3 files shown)


Laaas:
I could not test my change, since `cargo test --all` exhausts my tiny 8 GB of RAM.

fhaynes:
Thanks! We'll take a look at this after 0.11 release (May 1st...ish).

distransient:
Thanks for this! I've added this code to the Transform refactor: https://github.com/jojolepro/amethyst/pull/19. It should be making it into 0.11 from there.

PR #1522, opened 2019-04-18
Add editable text events for input submission and modification.

dkushner:
## Description

This PR extends the existing `UiEvent` definition to include two new events: `Change` and `Commit`.

The `Change` event is written whenever a focused, editable `UiText` component has its internal value modified through user action.

The `Commit` event is written when the system receives either the `Enter` or `NumPadEnter` virtual key codes. It is designed to indicate that the user has concluded their input and may want to move onto the next field, submit the form, etc.

## Modifications
- Created two new definitions in the `UiEventType` enumeration, `Change` and `Commit`.
- Added a new match arm in the `TextEditingInputSystem` to produce the new `Commit` event.
- Updated a number of alternate match arms in the same block to properly produce the new `Change` event. 
- Added the `Entity` resource to the system data for `TextEditingInputSystem` in support of the new targeted events. 

## PR Checklist

### DIFF
--- a/amethyst_ui/src/event.rs
+++ b/amethyst_ui/src/event.rs
@@ -48,6 +48,10 @@ pub enum UiEventType {
         /// The entity on which the dragged object was dropped.
         dropped_on: Entity,
     },
+    /// When the value of an `Editable` Ui element has changed.
+    Change,
+    /// When an `Editable` Ui element has been committed.
+    Commit,
 }
 
 /// A ui event instance.
--- a/amethyst_ui/src/event.rs
+++ b/amethyst_ui/src/event.rs
@@ -48,10 +48,15 @@ pub enum UiEventType {
         /// The entity on which the dragged object was dropped.
         dropped_on: Entity,
     },
-    /// When the value of an `Editable` Ui element has changed.
-    Change,
-    /// When an `Editable` Ui element has been committed.
-    Commit,
+    /// When the value of a UiText element has been changed by user input.
+    ValueChange,
+    /// When the value of a UiText element has been committed either by 
+    /// losing focus on mobile or by pressing "Enter".
+    ValueCommit,
+    /// When an editable UiText element has gained focus.
+    Focus,
+    /// When an editable UiText element has lost focus.
+    Blur,
 }
 
 /// A ui event instance.
--- a/amethyst_ui/src/event.rs
+++ b/amethyst_ui/src/event.rs
@@ -50,8 +50,7 @@ pub enum UiEventType {
     },
     /// When the value of a UiText element has been changed by user input.
     ValueChange,
-    /// When the value of a UiText element has been committed either by 
-    /// losing focus on mobile or by pressing "Enter".
+    /// When the value of a UiText element has been committed by user action.
     ValueCommit,
     /// When an editable UiText element has gained focus.
     Focus,
--- a/amethyst_ui/src/text_editing.rs
+++ b/amethyst_ui/src/text_editing.rs
@@ -6,9 +6,9 @@ use unicode_normalization::{char::is_combining_mark, UnicodeNormalization};
 use unicode_segmentation::UnicodeSegmentation;
 use winit::{ElementState, Event, KeyboardInput, ModifiersState, VirtualKeyCode, WindowEvent};
 
-use crate::{Selected, TextEditing, UiText};
+use crate::{Selected, TextEditing, UiText, UiEvent, UiEventType, LineMode};
 use amethyst_core::{
-    ecs::prelude::{Join, Read, ReadStorage, Resources, System, WriteStorage},
+    ecs::prelude::{Join, Read, Write, ReadStorage, Resources, System, WriteStorage, Entities},
     shrev::{EventChannel, ReaderId},
 };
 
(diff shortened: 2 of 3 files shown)


dkushner:
bors retry

fhaynes:
bors retry

azriel91:
bors try

amethyst_ui/src/event.rs
```diff
@@ -48,6 +48,10 @@ pub enum UiEventType {
         /// The entity on which the dragged object was dropped.
         dropped_on: Entity,
     },
+    /// When the value of an `Editable` Ui element has changed.
+    Change,
+    /// When an `Editable` Ui element has been committed.
```
jojolepro: There's no Editable component. Should this documentation be changed to say Selectable?

amethyst_ui/src/text_editing.rs
```diff
@@ -266,6 +287,40 @@ impl<'a> System<'a> for TextEditingInputSystem {
...
                             }
                         }
+                        VirtualKeyCode::Return | VirtualKeyCode::NumpadEnter => {
+                            match focused_text.line_mode {
+                                LineMode::Single => {
+                                    edit_events
+                                        .single_write(UiEvent::new(UiEventType::Commit, entity));
+                                }
+                                LineMode::Wrap => {
+                                    if modifiers.shift
```
jojolepro: Shift+Enter at max length = commit? Shouldn't this be ignored?

amethyst_ui/src/event.rs
dkushner: @jojolepro Good point. Would `TextEditing` be a more appropriate component reference? Perhaps it's more clear to the end user to just use a phrase like: 

> When the value of an editable `UiText` element has been changed by user input.

Moxinilian:
Do you think that in the context of mobile input, using the OS keyboards return features or having it go out of scope would trigger a commit event? If not, do you think this could be added?

amethyst_ui/src/event.rs
happenslol: I think it should at least be called `ValueChange`. I like your proposed description, it clarifies that the change is due to user input, which I think is the important part.

amethyst_ui/src/event.rs
```diff
@@ -48,6 +48,10 @@ pub enum UiEventType {
...
+    Commit,
```
happenslol: By "commit" you mean something like "confirm" or "submit"? I think this is not very obvious from the name, maybe submit confirm would be better since that's what's generally used in "form speak" (unless there's some other convention here that I'm not aware of).

amethyst_ui/src/text_editing.rs
happenslol: I would generally expect shift + enter to make a newline, and enter alone to commit (If I understand correctly what commit means).  Maybe the behavior here should be controllable through options/flags?

amethyst_ui/src/event.rs
```diff
@@ -48,6 +48,10 @@ pub enum UiEventType {
         /// The entity on which the dragged object was dropped.
         dropped_on: Entity,
     },
+    /// When the value of an `Editable` Ui element has changed.
+    Change,
```
happenslol: Could it be of benefit to provide a delta or the full value with the event? Just a thought, might save the user the trouble of having to read from the field and instead just depend on the events. On the other hand, it's easy enough to implement for the user, and might be unwanted overhead that he then can't turn off.

amethyst_ui/src/event.rs
jojolepro: Value change would not indicate the type of the value changed. ie checkbox vs text

Imo TextEdited would be a good name.

jojolepro: :+1:

amethyst_ui/src/event.rs
jojolepro: Impossible because you can't know the value type.

happenslol: Ok, fair

amethyst_ui/src/event.rs
dkushner: Precisely that, yeah. I avoided "submit" simply because it is generally associated with the larger concept of "forms", which were out of scope for this. I figured a submit generally implies multiple fields, but I'd be happy to change it if the consensus is that "submit" is more clear.

amethyst_ui/src/event.rs
```diff
@@ -48,10 +48,15 @@ pub enum UiEventType {
...
-    Change,
-    /// When an `Editable` Ui element has been committed.
-    Commit,
+    /// When the value of a UiText element has been changed by user input.
+    ValueChange,
+    /// When the value of a UiText element has been committed either by 
+    /// losing focus on mobile or by pressing "Enter".
+    ValueCommit,
+    /// When an editable UiText element has gained focus.
+    Focus,
```
jojolepro: what about GainedFocus + LostFocus? :)

amethyst_ui/src/text_editing.rs
```diff
@@ -62,6 +63,19 @@ impl<'a> System<'a> for TextEditingInputSystem {
                     .join()
                     .next()
             {
+                match self.last_selected {
+                    Some(last) => {
+                        if last != entity {
+                            edit_events.single_write(UiEvent::new(UiEventType::Blur, last));
```
jojolepro: Focus events should be emitted from the SelectionSystem. The TextEditing system isn't responsible for focus event handling emission.

amethyst_ui/src/event.rs
happenslol: I like it a lot better now, thanks!

amethyst_ui/src/event.rs
dkushner: I just lifted the names of the DOM events so they'd be familiar (at least to those familiar with web development).

amethyst_ui/src/text_editing.rs
```diff
@@ -266,10 +294,49 @@ impl<'a> System<'a> for TextEditingInputSystem {
...
+                                    }
+                                }
+                            }
+                        }
                         _ => {}
                     },
                     _ => {}
                 }
+            } else {
+                self.last_selected = None
```
jojolepro: Unused var.

Ready for merge otherwise. :)

PR #1523, opened 2019-04-19
Allow user to control Ordering and reliability.

TimonPost:
## Description

This PR implements the ordering and sequencing functionality from laminar into `amethyst_netework`. This required some API changes.

## Additions

- Added `NetEvent::Packet(NetPacket)` where `NetPacket` is a packet with different reliabilities. 
- OrderingGuarantee, same as the one we have in laminar, is needed so that we can implement `Serialize, Deserialize` for it. 
- DeliveryGuarantee, same as the one we have in laminar, is needed so that we can implement `Serialize, Deserialize` for it. 

## Removals

- Removed an UDP receiving thread
- Removed `NetEvent::Unreliable` and `NetEvent::Reliable`
- Removed `NetEvent::TextMessage` which is a `NetEvent::Packet()` now.

## PR Checklist

- [  ] Ran `cargo test --all` locally if this modified any rs files.

### DIFF
--- a/amethyst_network/src/network_socket.rs
+++ b/amethyst_network/src/network_socket.rs
@@ -44,9 +44,9 @@ where
     /// The list of filters applied on the events received.
     pub filters: Vec<Box<dyn NetFilter<E>>>,
     // sender on which you can queue packets to send to some endpoint.
-    transport_sender: Sender<InternalSocketEvent<E>>,
+    event_sender: Sender<InternalSocketEvent<E>>,
     // receiver from which you can read received packets.
-    transport_receiver: Receiver<Packet>,
+    event_receiver: Receiver<laminar::SocketEvent>,
     config: ServerConfig,
 }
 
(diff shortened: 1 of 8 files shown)


amethyst_network/src/network_socket.rs
```diff
@@ -66,27 +66,31 @@ where
...
+            for control_event in event_receiver.try_iter() {
                 match control_event {
                     InternalSocketEvent::SendEvents { target, events } => {
                         for ev in events {
-                            send_event(ev, target, &sender);
+                            match ev {
+                                NetEvent::Packet(packet) => {
+                                    send_event(packet, target, &sender);
+                                }
+                                _ => { /* TODO, handle connect, disconnect etc. */ }
```
fhaynes: If we aren't handling connect/disconnect, doesn't this mean amethyst_network won't work at all?

TimonPost: exactly

fhaynes:
Also looks like the build is failing

amethyst_network/src/network_socket.rs
fhaynes: Shouldn't we handle that?

TimonPost: Yes, we should be, but that's the task for an other PR.

TimonPost:
bors r=fhaynes

PR #1524, opened 2019-04-20
Adding support for AMETHYST_NUM_THREADS environment variable to control size of the threads pool used by thread_pool_builder.

Puciek:
## Description

This change allows for an environmental variable in Amethyst to control thread pool size, instead of relying on the variable in rayon. This helps alleviate extreme CPU saturation when amethyst is run on a system with many cores.

## Modifications

- `ApplicationBuilder:new`

## PR Checklist

PS. Not sure how to cover this best in unit tests, so any help in that department (and any other really, new-ish to RUST here) is highly appreciated and more than welcome :). Also couldn't run the 2nd cargo test due to error:
`Warning: can't set `merge_imports = true`, unstable features are only available in nightly channel.`

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -33,6 +33,7 @@ it is attached to. ([#1282])
 * `AmethystApplication` takes in application name using `with_app_name(..)`. ([#1499])
 * Add `NetEvent::Reliable` variant. When added to NetConnection, these events will eventually reach the target. ([#1513])
 * "How To" guides for defining state-specific dispatchers. ([#1498])
+* Adding support for AMETHYST_NUM_THREADS environment variable to control size of the threads pool used by thread_pool_builder.
 * Add `Input` variant to `StateEvent`. ([#1478])
 * Support type parameters in `EventReader` derive. ([#1478])
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -134,6 +134,7 @@ extra bounds from `AnimatablePrefab` and `AnimationSetPrefab` ([#1435])
 [#1513]: https://github.com/amethyst/amethyst/pull/1513
 [#1509]: https://github.com/amethyst/amethyst/pull/1509
 [#1510]: https:://github.com/amethyst/amethyst/pull/1523/
+[#1524]: https://github.com/amethyst/amethyst/pull/1524
 
 ## [0.10.0] - 2018-12
 
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,10 +1,10 @@
 //! The core engine framework.
 
-use std::{marker::PhantomData, path::Path, sync::Arc, time::Duration, env};
+use std::{env, marker::PhantomData, path::Path, sync::Arc, time::Duration};
 
 use crate::shred::Resource;
 use derivative::Derivative;
-use log::{info, log_enabled, trace, Level};
+use log::{debug, info, log_enabled, trace, Level};
 use rayon::ThreadPoolBuilder;
 use winit::Event;
 
@@ -487,11 +487,14 @@ where
             info!("Rustc git commit: {}", hash);
         }
 
-        let thread_count :usize = env::var("AMETHYST_NUM_THREADS")
+        let thread_count: Option<usize> = env::var("AMETHYST_NUM_THREADS")
             .as_ref()
-            .map(|s| s.as_str().parse()
-                    .expect("AMETHYST_NUM_THREADS was provided but is not a valid number!"))
-            .unwrap_or(0);
+            .map(|s| {
+                s.as_str()
+                    .parse()
+                    .expect("AMETHYST_NUM_THREADS was provided but is not a valid number!")
+            })
+            .ok();
 
         let mut world = World::new();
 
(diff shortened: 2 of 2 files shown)


docs/CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ it is attached to. ([#1282])
 * `AmethystApplication` takes in application name using `with_app_name(..)`. ([#1499])
 * Add `NetEvent::Reliable` variant. When added to NetConnection, these events will eventually reach the target. ([#1513])
 * "How To" guides for defining state-specific dispatchers. ([#1498])
+* Adding support for AMETHYST_NUM_THREADS environment variable to control size of the threads pool used by thread_pool_builder.
```
jojolepro: Add the url at the end of this line like on the other ones please.

src/app.rs
```diff
@@ -484,14 +485,26 @@ where
             info!("Rustc git commit: {}", hash);
         }
 
+        let _thread_count :usize = env::var("AMETHYST_NUM_THREADS")
```
jojolepro: Remove the underscore from this variable name.

docs/CHANGELOG.md
Puciek: Done!

src/app.rs
Puciek: And done! For some reason rust compiler suggested that.

src/app.rs
```diff
@@ -484,14 +485,26 @@ where
             info!("Rustc git commit: {}", hash);
         }
 
+        let thread_count :usize = env::var("AMETHYST_NUM_THREADS")
+            .as_ref()
+            .map(String::as_str)
```
jojolepro: The .parse.expect calls can be moved inside of the map clause, and in the unwrap_or clause you can directly put 0.

This will avoid a string allocation when the user does not use the env variable.

Puciek: Ah, interesting! Though I am not sure how to exactly move the `.parse.except` inside, could you point me in the right direction? I tried few guesses but end up with error akin to:

```

Comment (jojolepro) [reply src/app.rs]: .map(|s| s.as_str().parse().expect("...")).unwrap_or(0)

Comment (Puciek) [reply src/app.rs]: Ah, awesome! Added in, and done, thank you very much!

**Comment by LucioFranco** [comment]
@Puciek looks like youll need to run a cargo fmt pass and fix the conflicts with the changelog.

File: src/app.rs [hunk 4]
```diff
@@ -484,14 +485,26 @@ where
             info!("Rustc git commit: {}", hash);
         }
 
+        let thread_count :usize = env::var("AMETHYST_NUM_THREADS")
+            .as_ref()
+            .map(|s| s.as_str().parse()
+                    .expect("AMETHYST_NUM_THREADS was provided but is not a valid number!"))
+            .unwrap_or(0);
```
azriel91: instead of `unwrap_or(0)`, how about leaving this as `Option<usize>`, and then use `if let Some(thread_count) = thread_count` below?

That way we don't have `0` as a sentinel value, and if the user really wanted to specify 0, then they can do it

src/app.rs
```diff
@@ -484,14 +485,26 @@ where
...
         #[cfg(feature = "profiler")]
         let thread_pool_builder = thread_pool_builder.start_handler(|_index| {
             register_thread_with_profiler();
         });
-        let pool = thread_pool_builder.build().map(Arc::new)?;
+        let pool :ArcThreadPool;
+        if thread_count == 0 {
+            pool = thread_pool_builder.build().map(Arc::new)?;
+        } else {
+            info!("Running Amethyst with fixed thread pool: {}", thread_count);
```
azriel91: How about `debug` level? 

* the user specified the X thread count, echoing it back is okay at debug (when they want to check that it was read properly)
* hide information that a user doesn't need to see when they're not actively checking

Puciek: Fair enough, adjusting and fixing the merge conflict!

src/app.rs
Puciek: Ah, yes, I didn't wrap my head about if-let yet, but this makes sense so will adjust accordingly. Thank you kindly!

Puciek:
@LucioFranco I am working on the merge conflict, but I have no idea what causes the cargo fmt failure, can you point me in the right direction? I didn't change anything besides the code, and I have very little experience with rust and cargo.

Puciek:
Okay, so I think I've found the offender, namely it is in `rustfmt.toml` and the line causing the errors is `merge_imports = true` which was added by @Xaeroxe. I am not sure why it doesn't affect other builds, maybe they are based of older branches? Anyway, should I remove it to bring the build into passing, or is there a better solution?

Puciek:
Oh my bad, due to the spam about nightly builds I've missed that there is actual content available there, both in the jenkins and when I run it locally, duh. Will get on fixing it.

azriel91:
Thanks! One last thing, could you rebase over `master`?
We try to not have merge commits in branches

Puciek:
@azriel91 took a bit of finesse but managed to clear the tree from all the merge mess.

azriel91:
bors r=jojolepro, azriel91

PR #1525, opened 2019-04-21
Add Unit Tests for Audio Module

yl54:
## Description
- Added a few unit tests to test the audio library.
- Added version `0.5.0` of amethyst_utils to audio directory.

## Checklist


fhaynes:
Thank you! I will have to update the Dockerfile to include a /dev/null equivalent for audio, and then I'll re-run this.

amethyst_audio/src/sink.rs
```diff
@@ -59,3 +59,62 @@ impl AudioSink {
         self.sink.stop();
     }
 }
+
+// q: for AudioSink, do we want to test each of these apis as a sanity check?
```
azriel91: I don't think it's necessary; it may be a better idea to do one of the following in a separate PR:

* implement `Deref` and `DerefMut` with `Sink` as the target type, or
* use the [`delegate`](https://crates.io/crates/delegate) crate -- not really liking this option, but worth listing

amethyst_audio/src/components/audio_emitter.rs
```diff
@@ -53,3 +54,79 @@ impl AudioEmitter {
...
+        f.read_to_end(&mut buffer).unwrap();
+
+        // Create a Source and AudioEmitter from those bytes
+        let src = Source { bytes: buffer };
+        let mut emitter = AudioEmitter::default();
+
+        // Call play
+        match emitter.play(&src) {
+            Ok(_pass) => assert!(should_pass),
+            Err(_fail) => assert!(!should_pass),
```
azriel91: for the it-failed-but-should-not-fail case, since the error type contains information, perhaps do something like this:

```rust
e @ Err(_fail) => assert!(!should_pass, "Expected `play` result to be `Ok(..)`, but was {:?}", e),
```

(syntax might be wrong, but there's the idea)

amethyst_audio/src/components/audio_emitter.rs
```diff
@@ -53,3 +54,79 @@ impl AudioEmitter {
 impl Component for AudioEmitter {
     type Storage = BTreeStorage<Self>;
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::{AudioEmitter, Source};
+    use amethyst_utils::app_root_dir::application_root_dir;
+    use std::{fs::File, io::Read, vec::Vec};
```
azriel91: by the way, could you add empty newlines between import groups like this:

```rust
use std::{..};
// newline
use amethyst_utils::{..};
use another::{..};
// newline
use crate::{..};
```

we adopted the convention to quickly see what's from std / other crates / current crate.

oh ya, are you used to `use`ing `std::{boxed::Box, vec::Vec}`? (guessing that's standard practice for a `#[no_std]` crate?)
normally we leave them out as well

amethyst_audio/src/components/audio_emitter.rs
```diff
@@ -53,3 +54,79 @@ impl AudioEmitter {
...
+    #[test]
+    fn test_picker() {
+        // Create the input variables
+        let mut emitter_main = AudioEmitter::default();
+        let box_picker: Box<dyn FnMut(&mut AudioEmitter) -> bool + Send + Sync> =
+            Box::new(use_audio_emitter);
+
+        // Test set_picker and assert that it is not empty
+        emitter_main.set_picker(box_picker);
+        assert!(!emitter_main.picker.is_none());
```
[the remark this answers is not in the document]
yl54: That sounds better I kinda like that instead of the weird not none.

amethyst_audio/src/components/audio_emitter.rs
yl54: I saw the empty new lines in the other src files. Can change for tests as well.

I'm not used to anything I'm new to the language. Happy to help keep conventions consistent.

fhaynes:
OK, this passes on linux now, but not Windows. I'll see if I can add a loopback audio device to Windows later today.

fhaynes:
Hmm. I think we're going to have to only run these tests on linux for the moment. Windows is going to be a bit more challenging...

amethyst_audio/src/sink.rs
yl54: Oops that was not meant for review. I was gonna ask in Discord or something. Will remove the comment.

amethyst_audio/src/components/audio_emitter.rs
yl54: I added more useful commentary.

yl54:
At least to mitigate the issue of adding a larger file, I can add a smaller wav file.

amethyst_audio/src/components/audio_emitter.rs
azriel91: ah, could you either use `e`, or remove the underscore from `_fail` (underscore means it's unused).

Also, could you use the mp3 and wav from https://github.com/mathiasbynens/small ? (gotta rebase the bigger files out of history)
Thanks!

azriel91:
ah, just saw your last push. Unfortunately replacing a file in git doesn't save us the space unless we squash commits -- which [our bot can't seem to d](https://bors.tech/documentation/getting-started/), so could you please rebase / squash the commits? thanks!

amethyst_audio/src/components/audio_emitter.rs
yl54: Removed all instances of the fail underscores. ty for catching it.

yl54:
Squashed everything.

azriel91:
bors r=torkleyy, fhaynes, azriel91

azriel91:
now we'll have a less embarrassing coverage % :stuck_out_tongue:

yl54:
small steps

PR #1527, opened 2019-04-21
[WIP] Implement UI pass for rendy

happenslol:
This is the tracking PR for porting the UI pass over to the new renderer.


amethyst_ui/src/bundle.rs
```diff
@@ -24,26 +24,31 @@ use crate::{
 ///
 /// Will fail with error 'No resource with the given id' if the InputBundle is not added.
 #[derive(new)]
-pub struct UiBundle<A = String, B = String, C = NoCustomUi, W = u32, G = ()> {
+pub struct UiBundle<R, I, T, A = String, B = String, C = NoCustomUi, W = u32, G = ()> {
```
jojolepro: needs docs lol

amethyst_ui/src/bundle.rs
```diff
@@ -110,8 +115,9 @@ where
             &["ui_sound_system"],
         );
 
+        // TODO(happens): Move this into ui. Why is this even here?
         // Required for text editing. You want the cursor image to blink.
-        builder.add(BlinkSystem, "blink_system", &[]);
+        // builder.add(BlinkSystem, "blink_system", &[]);
```
jojolepro: you are in ui already?

amethyst_ui/src/bundle.rs
happenslol: Yeah, the entire `bundle.rs` and `prefab.rs` need to be updated, there is a metric ton of generics now and I didn't want to update it until I'm through with it

torkleyy: Is there no way we can reduce those?

happenslol: @Frizi just did a major refactor to use boxed trait objects instead of generics for this, since it has practically no performance impact and makes this code readable again =P This means that this will continue once he is done with that.

Frizi:
Superseded by https://github.com/amethyst/amethyst/commit/1e39c522f8d33345f653df1f644b18ab2227e057#diff-2f73b2b8472b2f81a559547ec35d37a7R73

Issue #1531, opened 2019-04-23
[BUG] No longer able to implement UI button actions


azriel91:
## Description

since #1390, it isn't possible to implement a custom event handler and tie it to a `UiButton`.

## Reproduction Steps

1. Use `UiButtonBuilder` to create a `UiButton`.
2. Implement a system that listens for a `UiEvent` and reconcile the event to the button.
3. can't 😢

## What You Expected to Happen

One of:

* There is access to the `Entity`(s) created by the button (#1532).
* The `UiButton`'s widget ID is transmitted within the `UiEvent`.
* The `UiButton`'s has a component that stores its widget ID, so it can be looked up.
* ~~The `Widgets` resource also stores the entity for a given widget ID.~~ same as the first bullet point
* The `UiButtonBuilder` allows you to specify an event handler (not sure if this one is feasible).

## What Actually Happened

The API has changed and no longer allows this.

fhaynes:
@jojolepro and @happenslol can you guys please take a look at this. This is a bad regression.

happenslol:
Hiya! Jojo had remarked on this, but I was not immediately aware of the implications. Totally an oversight on my part, but thankfully a very quick fix.

azriel91:
Got it going in my project, all good 👍!

PR #1532, opened 2019-04-23
Make generated UI entity fields `pub`.

azriel91:
## Description

Expose `Entity` fields created by the `define_widget!` macro.

See #1531 for details.

## PR Checklist

### DIFF
--- a/amethyst_ui/src/widgets.rs
+++ b/amethyst_ui/src/widgets.rs
@@ -187,7 +187,10 @@ macro_rules! define_widget {
         /// A $t widget, containing references to its associated entities.
         #[derive(Debug, Clone)]
         pub struct $t {
-            $($field: $crate::Entity),*
+            $(
+                /// `$field` Entity
+                pub $field: $crate::Entity
+            ),*
         }
 
         impl $crate::Widget for $t {}
--- a/amethyst_ui/src/button/builder.rs
+++ b/amethyst_ui/src/button/builder.rs
@@ -260,7 +260,7 @@ impl<'a, G: PartialEq + Send + Sync + 'static, I: WidgetId> UiButtonBuilder<G, I
     pub fn build(mut self, mut res: UiButtonBuilderResources<'a, G, I>) -> (I, UiButton) {
         let image_entity = res.entities.create();
         let text_entity = res.entities.create();
-        let widget = UiButton::new(image_entity, text_entity);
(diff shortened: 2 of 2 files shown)


happenslol:
Hey, thanks for the quick PR! Since we're at it, I also seem to have swapped the `text_entity` and the `image_entity` around when the button widget is created in the builder [here](https://github.com/amethyst/amethyst/blob/3dde11413a088ebac7751a777dbe5401aa9273c2/amethyst_ui/src/button/builder.rs#L263), which would be the next thing you run into when this is fixed. (Awkward... :sweat_smile:)

If you could include that in this PR, it would save us from having to do an additional one. Otherwise hit me up, and I'll do it tonight.

azriel91:
> Awkward... 😅

😄

amethyst_ui/src/widgets.rs
```diff
@@ -187,7 +187,10 @@ macro_rules! define_widget {
         /// A $t widget, containing references to its associated entities.
         #[derive(Debug, Clone)]
         pub struct $t {
-            $($field: $crate::Entity),*
+            $(
+                /// `$field` Entity
```
torkleyy: I don't think macros replace anything in doc comments. So this might look weird in the API docs

happenslol:
Cool, thanks for the change!

amethyst_ui/src/widgets.rs
happenslol: Ah, I was not aware of that. In that case, the attribute variant could be used here:
```rust
#[doc="A $t widget, containing references to its associated entities."]
pub struct $t { ... }
```

torkleyy: Unfortunately, it doesn't work in strings either :(

torkleyy: I've tried before, `stringify!` also doesn't work here. So the only way I know of is to just put docs inside the macro invocation and to-expand them using the macro.

happenslol: Ah, that sucks... But since all generated method do roughly the same thing, maybe the documentation doesn't need to be too detailed. This probably warrants another look, but should not be a blocker for this PR.

azriel91: one day in the future i may turn that into a proc macro

azriel91:
bors r=happenslol, torkleyy, jojolepro

PR #1535, opened 2019-04-24
[WIP] Headless Server Feature

fhaynes:
## Description

This adds a feature for headless servers, and an example server

## Modifications

- Adds a Cargo feature, `headless`

## PR Checklist


fhaynes:
@torkleyy @LucioFranco @jojolepro can you tell me if I added the feature correctly?

torkleyy:
Hmm, it doesn't seem to do anything at all. Is that intended?

If this allows headless execution, it should also be tested on CI.

fhaynes:
> Hmm, it doesn't seem to do anything at all. Is that intended?
> 
> If this allows headless execution, it should also be tested on CI.

The goal was to remove the crates that are not needed when running in a headless environment; audio, renderer, etc. I want to make sure I did that part right before I elaborate on the example.

torkleyy:
Okay, I don't think it can work like this. Features need to be additive, i.e. when enabling a feature no code which worked before should break.

For that reason, `headless` can never disable a crate.

azriel91:
I think it could work the other way round -- having a `"graphics"` or `"display"` feature that is defaulted to on. Headless apps would turn the feature off in the [dependency section](https://doc.rust-lang.org/stable/cargo/reference/specifying-dependencies.html#choosing-features).

Related, `amethyst_test` currently has this:

```toml
default = []

# Used to tag tests that need a display (e.g. X) to run.
graphics = []
```

It defaults to off (because for CI, usually agents don't have a display)

minecrawler:
Didn't we already have this [discussion on Discourse](https://community.amethyst.rs/t/amethyst-feature-gates/491)?

examples/headless_server/main.rs
```diff
@@ -0,0 +1,73 @@
+use amethyst::{
+    core::frame_limiter::FrameRateLimitStrategy,
+    ecs::{Join, System, WriteStorage},
+    network::*,
```
LucioFranco: nit: I'd rather not use a glob import like this

jojolepro:
The idea is that when you create your game, you define two features: one for the client and one for the server. Those features(in your game) decide which amethyst crates get compiled. 

You can't know ahead of time if your headless feature needs or doesn't need amethyst_network. What if you are making a console game? Do you need localization in this case?

Also features absolutely need to be additive, as torkleyy said.

Also also, its better if those kind of headless vs full clients are implemented in the developer's game directly, because they also need to switch the systems that are added to the dispatcher by using cfg macro statements over those features. (For example: Adding RenderBundle only if the "client" feature is enabled on the game).

fhaynes:
> You can't know ahead of time if your headless feature needs or doesn't need amethyst_network. What if you are making a console game? Do you need localization in this case?

Console games are not in scope for this. This is strictly for servers.

> Also also, its better if those kind of headless vs full clients

Do you mean headless server? A headless client is a different thing. 

> Okay, I don't think it can work like this. Features need to be additive, i.e. when enabling a feature no code which worked before should break.

Additive it shall be! Thanks

examples/headless_server/main.rs
fhaynes: That's how it was in the net_server example. =) I am going to leave it like this for consistency's sake. If we want to go back and remove all glob imports from the examples, we can do it as a different PR.

jojolepro:
By console I meant terminal or text based games.

fhaynes:
So, it seems to work fine if I add a headless feature like so:

```toml
headless = [
    "locale", 
    "network"
]
```

And then build the example with:

```bash
cargo build --example headless_server --no-default-features --features=headless
```

@torkleyy @jojolepro is this acceptable?

fhaynes:
> I have some more questions.
> 
> Reading the comments and the code, `headless` seems to be a set of features that you want for an instance which is exclusively a server, e.g. for hosting a multiplayer game session. Is that correct?

Yeah.

> If so, I have two more questions:
> 
> 1. Wouldn't it make more sense to name the feature `server`? Headless for me is the basic state, which you extend when you add e.g. rendering capabilities. However, when you want a game server, you need special functionality which may be exclusive to the server, but not necessary for the client application.

Server is fine with me.

> 2. Given the definition of headless, why is `amethyst_renderer` still included? It's non-optional currently, but for a headless system there is no point for a renderer. Can we make it optional and add it to `default`?

Sure.

fhaynes:
bors r=torkleyy,jojolepro

Issue #1430, opened 2019-02-19
[BUG] Specular Highlights broken in DrawPbmSeparate / GLTF


robert-wallis:
## Description

# Specular Lighting Artifacts

When loading the gltf files, it appears that the specular lighting is not being calculated correctly along some faces.

For example the file [cube.gltf](https://github.com/robert-wallis/cube_spec_error/raw/master/resources/cube.gltf) created from a default cube mapped in blender.  With metallic of 0.0, and roughness of 0.5.

In Amethyst there are triangles that have what appears to be no specular lighting.  And the roughness appears too low.

However the specular lighting is displayed correctly in Don McCurdy's GLTF viewer https://gltf-viewer.donmccurdy.com

I made a repo for this issue at: https://github.com/robert-wallis/cube_spec_error  The code is mostly a copy of [examples/gltf/main.rs](../tree/master/examples/gltf/main.rs) except the animation and vertex weights have been removed.

## Reproduction Steps

1) `git clone https://github.com/robert-wallis/cube_spec_error.git`
2) `cd cube_spec_error`
3) `cargo run`
4) Turn the cube with the mouse so the light shines on face B1 or B2 

## What You Expected to Happen

* The lighting should be consistent across the faces.
* The lighting should be more subtle for a 0.5 roughness material.

## What Actually Happened

* It appears some normals are not being calculated correctly in the fragment shader.

### Amethyst: [cube.gltf](https://github.com/robert-wallis/cube_spec_error/raw/master/resources/cube.gltf)

![docs/amethys-b1.png](https://github.com/robert-wallis/cube_spec_error/raw/master/docs/amethyst-b1.png)
![docs/amethyst-b2.png](https://github.com/robert-wallis/cube_spec_error/raw/master/docs/amethyst-b2.png)

### https://gltf-viewer.donmccurdy.com
The specular lighting is more subtle, but you can see there are no artifacts if you drop the file in the viewer yourself.

![mccurdy-b1.jpg](https://github.com/robert-wallis/cube_spec_error/raw/master/docs/mccurdy-b1.jpg)
![mccurdy-b2.jpg](https://github.com/robert-wallis/cube_spec_error/raw/master/docs/mccurdy-b2.jpg)

Similar good results are visible in babylon.js's WebGL renderer and Blender 2.8 Eevee.

## Your Environment

* macOS 10.14.3 Mojave / MacBook Pro Retina 13-inch Early 2015 / Intel Iris 6100 (1.5gb vram)
[...]

PR #1536, opened 2019-04-24
[GLTF] Properly index vertices

sfanxiang:
## Description

I'm not sure if this is correct, but by looking at the code we indexed into `faces` again when calculating tangents despite having done this before.

The lighting on `gltf` example now looks better, but still incorrect. However this is likely a separate issue because exporting with tangent included has the same lighting as this patch.

cc issue #1430, PR #1534

## PR Checklist


sfanxiang:
@torkleyy Rebased. Thanks for your review.
> add a test that makes sure it uses the correct indices

The fact that `calculate_tangents` no longer takes indices as its argument makes it impossible to use the "wrong" indices. Do you still want a test?

PR #1538, opened 2019-04-24
Adding events example based off the Events Channel page of the documentation

Puciek:
## Description

A very simple PR that merely adds an example of a cli application that sends and receives events. I felt like a working demonstration helps a lot, especially as it also deals with creating a bundle, adding systems to it etc.

## Additions

- Examples/Events

## PR Checklist

### DIFF
--- a/examples/events/main.rs
+++ b/examples/events/main.rs
@@ -1,0 +1,97 @@
+extern crate amethyst;
+
+use amethyst::core::{
+    bundle::SystemBundle,
+    frame_limiter::FrameRateLimitStrategy,
+    shrev::{EventChannel, ReaderId},
+};
+use amethyst::{
+    ecs::{DispatcherBuilder, Read, Resources, System, SystemData, World, Write},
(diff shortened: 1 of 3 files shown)


examples/events/main.rs
```diff
@@ -0,0 +1,97 @@
...
+
+#[derive(Debug)]
+struct MyBundle;
+
+impl<'a, 'b> SystemBundle<'a, 'b> for MyBundle {
+    fn build(self, builder: &mut DispatcherBuilder<'a, 'b>) -> Result<(), Error> {
+        builder.add(SpammingSystem, "spamming_system", &[]);
+        builder.add(
+            ReceivingSystem {
+                reader: std::prelude::v1::Option::None,
```
jojolepro: Why not just Option::None?

Puciek: Good catch! I was converting it from my scrapbook where I simply copy/pasted suggestion from the compiler. Thanks, updating :).

PR #1539, opened 2019-04-24
Add Window centering and monitor size.

mralve:
## Description

Made so windows will now be centered on the current monitor by default.
Small fixes in ScreenDimensions docs to clarify.

## Additions

DisplayConfig - Added `centered_window` bool to be able to toggle the centering.
ScreenDimensions - Added monitor HEIGHT and WIDTH functions to get the current screens res.

## PR Checklist

### DIFF
--- a/amethyst_renderer/src/config.rs
+++ b/amethyst_renderer/src/config.rs
@@ -15,6 +15,7 @@ use winit::{self, Icon, MonitorId, WindowAttributes, WindowBuilder};
 ///     max_dimensions: None,
 ///     min_dimensions: None,
 ///     fullscreen: false,
+///     centered_window: true,
 ///     multisampling: 0,
 ///     visibility: true,
 ///     vsync: true,
@@ -43,6 +44,9 @@ pub struct DisplayConfig {
 
     /// Maximum window dimensions, measured in pixels (px).
     pub max_dimensions: Option<(u32, u32)>,
+
+    /// Whether the window will be centered on the monitor by deafult.
+    pub centered_window: bool,
 
     /// Path to window icon.
     pub icon: Option<String>,
@@ -88,6 +92,7 @@ impl Default for DisplayConfig {
             decorations: true,
             dimensions: Some((640, 480)),
             fullscreen: false,
+            centered_window: true,
             icon: None,
             loaded_icon: None,
             max_dimensions: None,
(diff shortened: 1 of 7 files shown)


amethyst_renderer/src/config.rs
```diff
@@ -44,6 +45,9 @@ pub struct DisplayConfig {
     /// Maximum window dimensions, measured in pixels (px).
     pub max_dimensions: Option<(u32, u32)>,
 
+    /// Whether the window will be centered on the monitor by deafult.
```
jojolepro: Typo in default

docs/CHANGELOG.md
```diff
@@ -36,6 +36,8 @@ it is attached to. ([#1282])
 * Adding support for AMETHYST_NUM_THREADS environment variable to control size of the threads pool used by thread_pool_builder.
 * Add `Input` variant to `StateEvent`. ([#1478])
 * Support type parameters in `EventReader` derive. ([#1478])
+* Added `monitor_width` and `monitor_height` in `ScreenDimensions`. ([#1539])
+* Added `centered_window` bool to `DisplayConfig` Windows will now be centered by deafult. ([#1539])
```
jojolepro: Please add the link to the pull request. See the lines under those ones to see how.

azriel91:
Wondering if this should be in before or after 0.11 ? (breaking change)
Also, am not an expert but should the monitor dimensions be tracked in a separate type (i.e. what if you move the window to a different monitor at runtime, then the stored values aren't consistent)

mralve:
@azriel91 sorry about that, it now uses the get_current_monitor() and is updated by the window_management() so the values should track the change of the screen if that occurs.

mralve:
@azriel91 about 0.11 I am unsure, it would in the worst case be a minor fix to enter centered_window bool in display config? but that's on by default.

azriel91:
ah, re:0.11, it's probably better answered by @Frizi / @fhaynes (rendy integration is nigh, this might creep the scope); otherwise i'd like it in 🙃.

p/s; could you make the 1280 and 800 public constants in `amethyst_test`? I use the screen constants in my extended version of the test crate

fhaynes:
@azriel91 as long as you handle integrating it with rendy, go ahead.

Frizi:
The whole renderer is going to be replaced, including window management being split into it's own crate.
I'm perfectly OK with the change itself, but i would prefer to merge it once rendy branch is merged into master. 

Other option is to make a PR against `rendy` branch so we can merge it immediately :)

mralve:
I guess we can dismiss this PR then, and move this to Rendy then @Frizi Expect a new PR soon :)

fhaynes:
Closing this PR per @mralve and @Frizi doing it on the rendy branch. =) Thanks!

PR #1543, opened 2019-04-26
Connection Event Handling

TimonPost:
## Description

First, I Removed some unused API code, I did this for the next release, I don't think we should have public types that are not used. 

Secondly, I made use of the connection events sent by laminar; on connecting a `NetConnection` is created and inserted into the world. On disconnect, the user will receive an event and has to remove the connection from the world manually. This feature is of big importance because a server can't know which `NetConnections` to add in before, currently, the user needs to know all client IP addresses in front.

## Additions

- Added `NetEvent::Disconnected` and `NetEvent::Connected`.
- Send `NetEvent::Disconnected` when laminar sends a `SocketEvent::Timeout` event.
- Send `NetEvent::Connected` when laminar sends a `SocketEvent::Connect` event.
    - Automatically add `NetConnection` to world
- Added some unit tests
- `NetworkBundle` contains a method to allow custom configuration.
- `NetworkSocket` does not add new `NetworkConnection` to the world when disabled in the config.

## Removals

- Removed all `NetEvent's` because they were not used,
- Removed filter logic, because it didn't do anything, will be added back in a later version.

## PR Checklist

## Notice

This PR uses the newest laminar version because I discovered a bug within laminar, once laminar 0.2.1 is released this PR can be released also.


amethyst_network/Cargo.toml
```diff
@@ -30,6 +30,6 @@ bincode = "1.0"
 log = "0.4.6"
 uuid = { version = "0.7.1", features = ["v4","serde"] }
 thread_profiler = { version = "0.3" , optional = true }
-laminar = "0.2"
+laminar = { path = "../../laminar" }
```
LucioFranco: adding comment here so we dont forget to merge with this in 😄

TimonPost: I also left that comment in the `Notice`

amethyst_network/src/bundle.rs
```diff
@@ -1,33 +1,34 @@
...
+use crate::{server::ServerConfig, NetSocketSystem};
 
 /// A convenience bundle to create the infrastructure needed to send and receive network messages.
 pub struct NetworkBundle<T> {
     /// the configuration used for the networking crate.
     config: ServerConfig,
-
-    /// The filters applied on received network events.
-    filters: Vec<Box<dyn NetFilter<T>>>,
+    _data: PhantomData<T>,
```
LucioFranco: this can be dropped and the `T` can be a bound on `fn build<T>`

amethyst_network/src/net_event.rs
```diff
@@ -3,38 +3,19 @@
...
-    Disconnect {
-        /// The reason of the disconnection.
-        reason: String,
-    },
-    /// Notify the clients(including the one being disconnected) that a client has been disconnected from the server.
-    Disconnected {
-        /// The reason of the disconnection.
-        reason: String,
-    },
+    Connected(SocketAddr),
```
LucioFranco: We probably don't want to expose these variants but instead use forwards compatible methods. So have a NetEventKind and a NetEvent struct where the kind is the enum that is private and the NetEvent has methods to check if its one of the variants and a way to access that inner type.

amethyst_network/src/network_socket.rs
```diff
@@ -135,14 +132,21 @@ where
...
                                 if connection.target_addr == packet.addr() {
+                                    connection_exists = true;
                                     connection
                                         .receive_buffer
                                         .single_write(NetEvent::Packet(event.clone()));
                                 };
                             }
+
+                            if !connection_exists {
+                                panic!("No connection exitsts");
```
LucioFranco: `error!`?

examples/net_server/main.rs
```diff
@@ -34,12 +33,12 @@ fn main() -> Result<()> {
 pub struct State1;
 impl SimpleState for State1 {
     fn on_start(&mut self, data: StateData<'_, GameData<'_, '_>>) {
-        data.world
-            .create_entity()
-            .with(NetConnection::<String>::new(
-                "127.0.0.1:3457".parse().unwrap(),
-            ))
-            .build();
+        //        data.world
```
LucioFranco: reason why this is commented out?

amethyst_network/src/net_event.rs
```diff
@@ -3,38 +3,19 @@
...
+    Disconnected(SocketAddr),
     /// Send a packet to all connected clients
     Packet(NetPacket<T>),
+    #[doc(hidden)]
+    __Nonexhaustive,
```
jojolepro: Why?

amethyst_network/src/net_event.rs
```diff
@@ -3,38 +3,19 @@
...
-        /// The client uuid.
-        client_uuid: Uuid,
-    },
-    /// Reply to the client that the connection has been accepted.
-    Connected {
-        /// The server uuid.
-        server_uuid: Uuid,
-    },
-    /// Reply to the client that the connection has been refused.
-    ConnectionRefused {
```
jojolepro: Those where mostly meant for the user to handle.

amethyst_network/src/network_socket.rs
jojolepro: Might not want to do that.

docs/CHANGELOG.md
```diff
@@ -74,12 +74,14 @@ extra bounds from `AnimatablePrefab` and `AnimationSetPrefab` ([#1435])
 * Removed `NetEvent::Custom` and added `NetEvent::Packet(NetPacket)` ([#1523])
 * Fixed update is no longer frame rate dependent ([#1516])
 * Display the syntax error when failing to parse sprite sheets  ([#1526])
-
+* `NetConnection` is automatically created when client starts sends data to server. ([#1539])
+* User will receive `NetEvent::Connected` on new connection and `NetEvent::Dissconnected` on disconnect. ([#1539])
 
 ### Removed
+- Removed all `NetEvent's` because they were not used. ([#1539])
+- Removed filter logic, because it didn't do anything, will be added back in a later version (NetFilter, FilterConnected). ([#1539])
```
jojolepro: Both were meant to be defined/used by the user.
Filter was implemented before... not sure where the handling code went.

amethyst_network/src/network_socket.rs
TimonPost: was testing code, thanks for pointing out!

amethyst_network/src/net_event.rs
TimonPost: Those were not implemented and used anywhere within the crate, I see the purpose of them, but I think removing them is better than not using it.

amethyst_network/src/net_event.rs
TimonPost: Because more events will come, this will make us a bit more future proof.

examples/net_server/main.rs
TimonPost: Yes, not needed anymore, since connections are created dynamically, let me remove that.

amethyst_network/src/bundle.rs
TimonPost: I am not sure what you mean I will have a 'unconstrained type parameter' error.

LucioFranco: You can drop the type T from this struct

TimonPost: I can leave it away, but I can't add `T` do `fn build<T>` because it isn't part of the `SystemBundle` trait. 

"method `build` has 1 type parameter but its trait declaration has 0 type parameters"

LucioFranco: ahhh I forgot that its based on a trait you're right 👍

amethyst_network/src/net_event.rs
TimonPost: Although this makes pattern matching a bit weird, we can just as well add the 

```
```

docs/CHANGELOG.md
TimonPost: Alright, but those events weren't interpreted anywhere in the code, so what's the point of having them? 

You are right about the filtering, I am not sure when it was removed, and in fact, I tried to look it up but could not find it, but since the connection events weren't interpreted the filtering was basicly doing nothing. There was also a comment to point that out somewhere. Filtering will come back but can be done at a later time.

amethyst_network/src/lib.rs
```diff
@@ -33,16 +33,29 @@ pub fn send_event<T>(event: NetPacket<T>, addr: SocketAddr, sender: &Sender<Pack
...
-            } else {
-                Packet::unreliable(addr, s)
+        Ok(payload) => {
+            let packet = match event.delivery_guarantee() {
+                net_event::DeliveryGuarantee::Unreliable => match event.ordering_guarantee() {
+                    net_event::OrderingGuarantee::None => Packet::unreliable(addr, payload),
+                    net_event::OrderingGuarantee::Sequenced(s) => {
+                        Packet::unreliable_sequenced(addr, payload, s)
+                    }
+                    _ => panic!("This is in no way possible"),
```
jojolepro: use the unreachable!() macro

amethyst_network/Cargo.toml
TimonPost: done

TimonPost:
bors r=justin

PR #1544, opened 2019-04-26
reworked state specific dispatcher documentation

bamling:
## Description

The current documentation for `State`-specific `Dispatcher`s uses the `State` trait, which requires the explicit execution of the application `Dispatcher` alongside the `State`-specific `Dispatcher`. Most other examples and guides in the Amethyst book use the `SimpleState` trait instead, which automatically handles the execution of the application `Dispatcher`. I initially wanted to rework the documentation so it is based on `SimpleState` as well, but ended up expanding on the whole documentation regarding `State`-specific `Dispatcher`s. I also ended up adding documentation for pausable `System`s.

## PR Checklist


Moxinilian:
I believe this could be ambiguous as by default the State does not update the main dispatcher, but SimpleState does. But I'd like to know what others think.

bamling:
> I believe this could be ambiguous as by default the State does not update the main dispatcher, but SimpleState does. But I'd like to know what others think.

Oh, I see... would you say that `SimpleState` is more commonly used than `State`? I personally wasn't aware that `State` requires manually execution of the *main* dispatcher. I could revert my changes and instead add a block mentioning the usage of `SimpleState`.

Moxinilian:
I think the general use cases only needs SimpleState. Being able to create complex states (as in not SimpleState) really exists just for people that specifically don't want the main dispatcher to run all the time.

Maybe, instead of doing it like this, we could just describe it in a different manner. If you are using a complex state, you probably already know you need to update the main dispatcher yourself as you are doing it. So if we just say "you can add another dispatcher like this", then the ambiguity is lifted.

bamling:
> I think the general use cases only needs SimpleState. Being able to create complex states (as in not SimpleState) really exists just for people that specifically don't want the main dispatcher to run all the time.
> 
> Maybe, instead of doing it like this, we could just describe it in a different manner. If you are using a complex state, you probably already know you need to update the main dispatcher yourself as you are doing it. So if we just say "you can add another dispatcher like this", then the ambiguity is lifted.

I updated my pull request before reading your reply. I can take some time to rework the documentation for the state specific dispatchers and base them off of `SimpleState` instead and close this PR for now though if that's fine!

bamling:
I reworked `how_to_define_state_dispatcher_external.md` and would appreciate some feedback. I'll change `how_to_define_state_dispatcher_internal.md` accordingly.

book/src/state_specific_systems/how_to_define_state_dispatcher_external.md
```diff
@@ -1,156 +1,277 @@
...
+#     /// The `State` specific `DispatcherBuilder`. This will be used to build the actual dispatcher.
+#     dispatcher_builder: Option<DispatcherBuilder<'a, 'b>>,
+# 
+#     /// The `State` specific `Dispatcher`, containing `System`s only relevant for this `State`.
+#     dispatcher: Option<Dispatcher<'a, 'b>>,
+# }
+# 
+impl<'a, 'b> SimpleState for CustomDispatcherState<'a, 'b> {
+    fn on_start(&mut self, data: StateData<GameData>) {
+        if self.dispatcher.is_none() {
```
Moxinilian: Isn't that a bit redundant? If we are calling on_start, I'd say it is fairly certain the dispatcher is not defined yet. It would make the code easier to understand in my opinion.

book/src/state_specific_systems/how_to_define_state_dispatcher_external.md
```diff
@@ -1,156 +1,277 @@
...
+            let mut dispatcher = self.dispatcher_builder
+                .take()
+                .expect("Expected `dispatcher_builder` to exist when `dispatcher` is not yet built.")
+                .build();
```
Moxinilian:
```suggestion
                .build(); // We build the dispatcher itself
```

book/src/state_specific_systems/how_to_define_state_dispatcher_external.md
```diff
@@ -1,156 +1,277 @@
...
+            dispatcher.setup(&mut data.world.res);
```
Moxinilian:
```suggestion
            dispatcher.setup(&mut data.world.res); // We register the resources it uses into the world
```

book/src/state_specific_systems/how_to_define_state_dispatcher_external.md
```diff
@@ -1,156 +1,277 @@
...
+            self.dispatcher = Some(dispatcher);
+        }
+    }
+
+    fn update(&mut self, data: &mut StateData<GameData>) -> SimpleTrans {
+        self.dispatcher.as_mut().unwrap().dispatch(&data.world.res);
```
Moxinilian: Maybe you should prefer an if-let pattern here.

book/src/state_specific_systems/how_to_define_state_dispatcher_external.md
```diff
@@ -1,156 +1,277 @@
...
+#     fn new(dispatcher_builder: DispatcherBuilder<'a, 'b>) -> Self {
+#         Self {
+#             dispatcher_builder: Some(dispatcher_builder),
+#             dispatcher: None,
+#         }
+#     }
+# }
+```
+
+Now, any systems in the state-specific `Dispatcher` will only run in this `State`. The *application* `Dispatcher` is automatically executed for `SimpleState` implementations. For more complex states that implement other `State` specific traits, the *application* `Dispatcher` has to be explicitly executed on top of the `State` specific `Dispatcher`. This is done by slightly modifying your `State`s `update(..)` method:
```
Moxinilian: I think you should just get rid of this part. People should know if they use a custom state that they need to update the main dispatcher as well.

book/src/state_specific_systems/how_to_define_state_dispatcher_internal.md
```diff
@@ -88,9 +88,15 @@ This guide explains how to define a state-specific dispatcher whose systems are
...
 
-            Trans::Pop
+            Trans::None
         }
     }
     ```
 
     Now, any systems in the state-specific dispatcher will only run in this state.
+
+    Depending on the trait implemented by `CustomDispatcherState` it is recommended to omit the execution of the application dispatcher during `update(..)`:
```
Moxinilian: Instead of this, I think it would be easier if you assumed SimpleState in this article as well. Not-simple states aren't much different, SimpleState is only here to not have to explain dispatcher updates to beginners.

book/src/state_specific_systems/how_to_define_state_dispatcher_external.md
bamling: Ahh, that's still some old documentation that I haven't fixed up yet. Ignore that one. :)

book/src/state_specific_systems/how_to_define_state_dispatcher_internal.md
bamling: Ahh, haven't reworked the documentation for internal dispatchers yet. Ignore that bit, it's old!

book/src/state_specific_systems/how_to_define_state_dispatcher_external.md
bamling: I'm not sure how I feel about that. I prefer too much over too little and I ran into an issue when I followed the old guide that was based on custom states. I'd personally keep it in just to be sure but I'll let you guys decide!

book/src/controlling_system_execution/pausable_systems.md
```diff
@@ -0,0 +1,41 @@
...
+
+`pausable(CurrentState::Running)` creates a wrapper around our `System` that controls its execution depending on the `CurrentState` `Resource` registered with the `World`. As long as the value of the `Resource` is set to `CurrentState::Running`, the `System` is executed.
+
+To register the `Resource` or change its value, we can use the following line:
+
+```rust,no_run,noplaypen
+*data.world.write_resource::<CurrentState>() = CurrentState::Paused;
+```
+
+Alternatively we can access the `Resource` in `System`s utilising `amethyst::ecs::prelude::Read` and `amethyst::ecs::prelude::Write` as seen before.
```
jaynus: This is not true for a system using the pausable extension. The Pausable wrapper actually accesses the resource with a 'Read' SystemData, so you cannot have a pausable system pause or unpause itself, as it causes a borrow error. The documentation should be clear on this that you *cannot* pause and unpause a system from within a pausable system, you need to lazy.exec the change. Or better yet, specifically state 'A Pausable system can only access its pause resource with immutable Read and cannot modify it' or something.

bamling: Thanks for the feedback! I'll change that part tomorrow.

book/src/controlling_system_execution/state-specific_dispatcher_external.md
```diff
@@ -0,0 +1,279 @@
+# How to Define State Dispatcher: External
+
+This guide explains how to define a state-specific `Dispatcher` whose `System`s are passed in externally, which means at the moment of state struct creation. This is used when the list of `System`s is determined by user choices at runtime. For simplicity sake we'll be using the `SimpleState` trait for our custom state.
```
marot: I'm having trouble understanding the difference between `External` and `Internal`. Both create the final dispatcher in `on_start`, so I would consider them quite equal.

Imo this is missing a transition to showcase how the dispatchers can be modified from outside.

marot: I tried to modify the `state_dispatcher` example and it seems to be possible to pass a dispatcher for your state on creation, in a different state. See here: https://github.com/marot/amethyst/pull/1/files

bamling: The external/internal examples already existed as separate guides... but I feel like they can probably be merged into one. I think in the end it comes down to the `State` having its own `Dispatcher` that is executing during the `State`s `update(..)` method. How the `Dispatcher` is configured and where it comes from is more a personal taste. I'll fix it up, thank you!

book/src/state_specific_systems/how_to_define_state_dispatcher_external.md
bamling: I ended up removing that part in my current version.

bamling:
bors r=approver1,approver2

bamling:
bors r=jaynus,fletcher

bamling:
bors r=jaynus,fletcher

fhaynes:
bors r=jaynus,fletcher

PR #1547, opened 2019-04-27
Fix using SRGB for default texture data

sfanxiang:
## Description

`impl From<[f32; 4]> for TextureData` interprets texture in srgb and caused very weird normals.

cc #855.

## PR Checklist


Frizi:
The new renderer is already using `palette` crate for color conversion, and all the default material textures are also specified correctly, i.e. Albedo and Emission are Srgb by default, while all the others are linear.

As this is already fixed and we are very soon replacing the whole renderer, i'm going to close this PR.

PR #1548, opened 2019-04-27
renderer: Unify name of fragment output color to "out_color"

sfanxiang:
## Description

Also fixes some shaders that don't match up with the passes (eg. physically based lighting). This effectively enables blending to work.

Please also check if I missed anything.

## PR Checklist


torkleyy:
This needs to be rebased to work with `amethyst_rendy` instead of `amethyst_renderer`.

Frizi:
This PR was created on top of old renderer, and all related files are already removed. All existing rendy shaders do use "out_color" consistently, so there is nothing to be done. Closing.

Issue #1562, opened 2019-05-02
[BUG] Transform serialization is wrong


Awpteamoose:
## Description

Transform serialization doesn't work correctly.

## Reproduction Steps
```rust
let t1 = amethyst::core::Transform::default();
let s: String = ron::ser::to_string_pretty(&t1, ron::ser::PrettyConfig::default()).unwrap();
let t2: amethyst::core::Transform = ron::de::from_str(&s).unwrap();
println!("T1: {:?}", &t1);
println!("T2: {:?}", &t2);
```

## What You Expected to Happen

Outputs are the same.

## What Actually Happened

After deserealization, the rotation quaternion is `[0.0, 0.0, 1.0, 0.0]` instead of `[0.0, 0.0, 0.0, 1.0]`.

## Additional Context

imo, the default deriveable Serialize/Deserialize format is good enough, especially since the rotation is serialized as a quaternion rather than euler angles anyway.

Issue #1563, opened 2019-05-03
[BUG] ComponentEvent reader for Transform constantly receives Modified events on master


bamling:
## Description

Since the recent `Transform` refactoring, it seems like the `ComponentEvent` reader for `Transform` keeps receiving `ComponentEvent::Modified` events, despite none of my `System` accessing `Transform`s in a mutable fashion. After having a quick chat with @azriel91 we narrowed it down to the `TransformSystem` triggering these events. This takes away from the `FlaggedStorage` features and makes it impossible to react to `Transform`s that were modified in the context of my game.

## Reproduction Steps

Create a `ReaderId<ComponentEvent>` for `Transform`s:
```
let transforms_reader = {
    let mut transforms = world.write_storage::<Transform<f32>>();
    transforms.register_reader()
};
```

Access the `ReadStorage<'s, Transform<f32>>` and read from its channel:
```
// `self.component_event_reader` is the `transforms_reader`.
for component_event in transforms.channel().read(&mut self.component_event_reader) {
    match component_event {
        ComponentEvent::Modified(id) => { println!("Got Modified for: {}", id); }
        _ => {}
    }
}
```

You'll end up seeing constant `Got Modified for:` messages for each `Entity`.

## What Actually Happened

`ComponentEvent::Modified` events are only triggered when my own, defined `System`s access `Transform` mutably.

## Your Environment

Amethyst version: 
`master` with Commit ID `331e83c46250464c38936f56a97fa1416bbbbe92`

torkleyy:
This seems to be a problem with `global_matrix` stored inside `Transform`.

This way, writing to `Transform` will inevitably create a `Modified` event for this and the next frame (and right now it just creates that event every frame without any condition).

I think we need to write the global transform to a second component (I think this was the case before the transform refactor).

@jojolepro What was the reason for the change?

bamling:
The referenced issue has been merged 2 days ago but that didn't seem to fix this. I used the current `master` earlier today and it still floods me with `ComponentEvent`s for `Transform`. Not sure if the issue was **supposed** to fix it but I figured I'd let you guys know anyways.

bamling:
As of now (`git+https://github.com/amethyst/amethyst#85365d5543b9be0daa891706c5d46c80ee38f9ca`) this is still an issue. Any update on this?

torkleyy:
This is still an issue. The PR only fixed serialization, so this bug remaons.

There are essentially three problems:

[...]

PR #1564, opened 2019-05-03
Fix Transform serialization and Transform system event loop.

jojolepro:
## Description

title

fixes https://github.com/amethyst/amethyst/issues/1562

## PR Checklist

### DIFF
--- a/amethyst_core/Cargo.toml
+++ b/amethyst_core/Cargo.toml
@@ -36,6 +36,7 @@ thread_profiler = { version = "0.3" , optional = true }
 
 [dev-dependencies]
 amethyst = { path = "..", version = "0.10.0" }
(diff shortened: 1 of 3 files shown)


jojolepro:
that's fine. It was supposed to do that to signal that the global matrix changed.

The issue is that it was also signaling to itself, making an infinite event loop.

azriel91:
The other part of the issue is, what if someone wants to detect if the local transform changes, but not global. For example, an entity with a parent, if the parent's transform changes, but the entity itself does not, the global transform would be modified, but the local wasn't. there isn't a way to detect that now

torkleyy:
If you remove "Fixes #1563", we should be able to merge this.

PR #1568, opened 2019-05-04
Implement builder like functionality for AnimationSet and AnimationCo…

In-line:
…ntrolSet

## Modifications

- Implemented builder like functionality for AnimationSet and AnimationControlSet

## PR Checklist

### DIFF
--- a/amethyst_animation/src/resources.rs
+++ b/amethyst_animation/src/resources.rs
@@ -737,13 +737,14 @@ where
     ///
     /// This should be used with care, as this will leave all linked samplers in place. If in
     /// doubt, use `abort()` instead.
-    pub fn remove(&mut self, id: I) {
+    pub fn remove(&mut self, id: I) -> &mut Self {
         if let Some(index) = self.animations.iter().position(|a| a.0 == id) {
             self.animations.remove(index);
         }
+        self
     }
 
-    fn set_command(&mut self, id: I, command: AnimationCommand<T>) {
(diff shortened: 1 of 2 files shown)


In-line:
Should I write changelog entry?

azriel91:
bors r=azriel91, jojolepro

azriel91:
bors retry

(linker OOM)

azriel91:
bors retry

azriel91:
bors retry

amethyst_animation/src/resources.rs
```diff
@@ -737,13 +737,14 @@ where
     ///
     /// This should be used with care, as this will leave all linked samplers in place. If in
     /// doubt, use `abort()` instead.
-    pub fn remove(&mut self, id: I) {
+    pub fn remove(&mut self, id: I) -> &mut Self {
```
jstnlef: Does it make sense to have these consume and return `self` rather than a reference to it?
`pub fn remove(self, id: I) -> Self`. If we expect it to work like a builder it's a bit more common.

jojolepro: Would not work when pulled from a resource that is inside of `World` already.

jstnlef: ah fair. Should have looked deeper into where it's called from 👍

azriel91:
bors retry

PR #1586, opened 2019-05-09
Sentry integration

jojolepro:
Adds Sentry integration


.sentry_dsn.txt
```diff
@@ -0,0 +1 @@
+https://6bc2b763b60e4651ac237871056eb21b@sentry.io/1445091
```
LucioFranco: probably dont want this?

build.rs
```diff
@@ -1,8 +1,90 @@
...
+            true
+        },
+        false => {
+            let _ = file.write_all(b"false");
+            false
+        }
+    }
+}
+
+fn load_sentry_dsn() {
```
LucioFranco: should just load it from env var?

src/app.rs
```diff
@@ -477,6 +486,10 @@ where
         info!("Version: {}", env!("CARGO_PKG_VERSION"));
         info!("Platform: {}", env!("VERGEN_TARGET_TRIPLE"));
         info!("Amethyst git commit: {}", env!("VERGEN_SHA"));
+        if let Some(sentry) = option_env!("SENTRY_DSN") {
+            info!("Sentry DSN: {}", sentry);
```
LucioFranco: and probably dont want to print this atleast not at an info level?

.sentry_dsn.txt
fhaynes: We do. It is the only way to send events.

build.rs
fhaynes: No, we want people to be able to use it without having to worry about finding the DSN

src/app.rs
fhaynes: No, it is fine for the DSN to be public. The service has rate limiting and IP restriction built in if we need it.

fhaynes:
@jojolepro cargo fmt and good to go

jojolepro:
bors retry

torkleyy:
How will this work for game deployment? Also, what happens for non-interactive builds?

fhaynes:
```
How will this work for game deployment?
```
For now, it shouldn't be deployed into a release version of a game to customers. It is meant as a debugging tool for our direct users, not their customers.

```
Also, what happens for non-interactive builds?
```
It should check the files as normal, and default to not including it if the config files aren't present.

PR #1590, opened 2019-05-10
Fix float book changes

jojolepro:
## PR Checklist

### DIFF
--- a/amethyst_core/Cargo.toml
+++ b/amethyst_core/Cargo.toml
@@ -16,14 +16,15 @@ appveyor = { repository = "amethyst/amethyst" }
 travis-ci = { repository = "amethyst/amethyst" }
 
 [dependencies]
-nalgebra = { git = "https://github.com/jojolepro/nalgebra", branch = "pub", features = ["serde-serialize", "mint"] }
-alga = { git = "https://github.com/jojolepro/alga", branch = "pub",default-features = false }
+nalgebra = { version = "0.18.0", features = ["serde-serialize", "mint"] }
+alga = { version = "0.9.0", default-features = false }
 alga_derive = "0.9.0"
 approx = "0.3"
 amethyst_error = { path = "../amethyst_error", version = "0.1.0" }
 fnv = "1"
 hibitset = { version = "0.5.2", features = ["parallel"] }
 log = "0.4.6"
+num-traits = "0.2.0"
 rayon = "1.0.2"
 serde = { version = "1", features = ["derive"] }
 shred = { version = "0.7" }
--- a/amethyst_core/src/float.rs
+++ b/amethyst_core/src/float.rs
@@ -44,30 +44,6 @@ impl Float {
     }
 }
 
-impl PartialEq<f32> for Float {
-    fn eq(&self, other: &f32) -> bool {
-        (self.0 as f32).eq(other)
-    }
-}
-
-impl PartialEq<f64> for Float {
-    fn eq(&self, other: &f64) -> bool {
-        (self.0 as f64).eq(other)
-    }
-}
-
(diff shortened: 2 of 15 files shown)


jojolepro:
image

amethyst_core/Cargo.toml
```diff
@@ -16,14 +16,15 @@ appveyor = { repository = "amethyst/amethyst" }
...
-nalgebra = { git = "https://github.com/jojolepro/nalgebra", branch = "pub", features = ["serde-serialize", "mint"] }
-alga = { git = "https://github.com/jojolepro/alga", branch = "pub",default-features = false }
+nalgebra = { version = "0.18.0", features = ["serde-serialize", "mint"] }
+alga = { version = "0.9.0", default-features = false }
 alga_derive = "0.9.0"
 approx = "0.3"
 amethyst_error = { path = "../amethyst_error", version = "0.1.0" }
 fnv = "1"
 hibitset = { version = "0.5.2", features = ["parallel"] }
 log = "0.4.6"
```
Frizi: Use alga::num

jojolepro: ?

jojolepro:
bors retry

amethyst_core/src/float.rs
```diff
@@ -42,6 +44,30 @@ impl Float {
     }
 }
 
+impl PartialEq<f32> for Float {
+    fn eq(&self, other: &f32) -> bool {
+        (self.0 as f32).eq(other)
+    }
+}
+
+impl PartialEq<f64> for Float {
```
Frizi: This can have bad consequences. When `Float` is using `f32` internally, it will sill lead to comparisons being made with casts to `f64`s on both sides. This is because float literals default to f64 when multiple variants are available. I think we can have `PartialEq<f32>` though, as you have to cast and can't really do much than dropping a few bits of precission when comparing against low precision value in the first place.

Frizi: Or actually i'm not really sure about that. Whichever we pick ad "default", we are always suspectible to autocasting errors.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8477abe94ff6ac674d0a2d637e592f20
That playground shifts calculations to `f64` unintentionally, but it's equally easy to do this the other way around. The same issue persists with all `impl Into<Float>`
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7b624857eb01782b7ae6f7f6443ffa79

jojolepro: soooo what do we do? ^^

jojolepro: Actually I think this is fine. If it defaults to f64, it will get converted later to f32 internally. Sure, we are causing accidental float casts, but I don't see how to avoid that with the current state of type inference without totally removing the f64 impl. For example the From<f64>. That would prevent the cast, but would force users using f64 to cast to f32 and then back to f64 internally.

jojolepro:
bors try

jojolepro:
bors try

jojolepro:
bors try

jojolepro:
bors try

jojolepro:
bors try

jojolepro:
bors try

jojolepro:
bors try

Frizi:
```
failures:

---- tests::test_backtrace stdout ----
thread 'tests::test_backtrace' panicked at 'a backtrace', src\libcore\option.rs:1038:5

failures:
    tests::test_backtrace

test result: FAILED. 6 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

azriel91:
bors retry

azriel91:
that backtrace issue is flakey

Issue #1588, opened 2019-05-10
[BUG] `VERGEN_TARGET_TRIPLE` undefined


torkleyy:
## Description

The compilation of `amethyst` (987bba3d900d68b5216bf3624765d42a52fa6386) is failing with this error:

```
error: environment variable `VERGEN_TARGET_TRIPLE` not defined
   --> src/app.rs:487:31
    |
487 |         info!("Platform: {}", env!("VERGEN_TARGET_TRIPLE"));
    |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error
```

## Reproduction Steps

1. Clone `master`
2. `cargo +stable build`

## Your Environment

1. OS: Fedora
2. Kernel: 5.0.9-200.fc29.x86_64
3. 987bba3d900d68b5216bf3624765d42a52fa6386

## Additional Context

It seems this is a regression caused by #1586, and happens exclusively on stable Rust.

fhaynes:
Hmm, interesting. vergen is nothing this added in, I don't think

fhaynes:
Hmm, can't repro this:
```
   Compiling amethyst_core v0.5.0 (/Users/fletcher/Projects/amethyst/amethyst/amethyst_core)
   Compiling amethyst_assets v0.6.0 (/Users/fletcher/Projects/amethyst/amethyst/amethyst_assets)
   Compiling amethyst_network v0.3.0 (/Users/fletcher/Projects/amethyst/amethyst/amethyst_network)
   Compiling amethyst_renderer v0.10.1 (/Users/fletcher/Projects/amethyst/amethyst/amethyst_renderer)
   Compiling amethyst_locale v0.4.0 (/Users/fletcher/Projects/amethyst/amethyst/amethyst_locale)
   Compiling amethyst_input v0.6.0 (/Users/fletcher/Projects/amethyst/amethyst/amethyst_input)
   Compiling amethyst_animation v0.5.0 (/Users/fletcher/Projects/amethyst/amethyst/amethyst_animation)
   Compiling amethyst_controls v0.4.0 (/Users/fletcher/Projects/amethyst/amethyst/amethyst_controls)
   Compiling amethyst_utils v0.5.0 (/Users/fletcher/Projects/amethyst/amethyst/amethyst_utils)
   Compiling amethyst_audio v0.5.0 (/Users/fletcher/Projects/amethyst/amethyst/amethyst_audio)
[...]

PR #1592, opened 2019-05-11
Use stderr to print message to user.

azriel91:
This doesn't actually show up in the console, but printing to stdout
pollutes the stream that cargo reads from.

Issue #1588

## Description

Attempt to eliminate a source of error in #1588.

## PR Checklist

- **n/a** Ran `cargo test --all` locally if this modified any rs files.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,7 +1,7 @@
 use dirs::home_dir;
 use std::{
     fs::{create_dir_all, read_to_string, remove_file, File},
-    io::{stdin, stdout, Write},
+    io::{stderr, stdin, Write},
     path::Path,
 };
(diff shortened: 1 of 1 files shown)


Frizi:
That does make a lot of sense :) But it seems like we should remove it from build.rs completely, and instead put it somewhere else. I have brought that issue before, and the idea was that we really need a project structure for that. Unfortunately our amethyst-cli or editor isn't widely used yet. Are there any other options left?

Frizi:
I wonder if running a blocking message box from build.rs would be OK :P Maybe using something like `tkinter` would do the trick https://pythonspot.com/tkinter-askquestion-dialog/

PR #1593, opened 2019-05-11
Add CONTRIBUTING step for mdbook

### LINKED ISSUE #1589: [MAINTENANCE] CONTRIBUTING.md
## Note

https://github.com/amethyst/amethyst/blob/master/docs/CONTRIBUTING.md#writing-documentation should indicate how to test book changes with the mdbook -L ... command.

## Reason

I have bad memory and keep forgetting the exact command :P

distransient:
Fixes #1589

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -66,6 +66,7 @@ done the following things first:
    * `cargo build`
    * `cargo test --all`
    * `cargo run --example {example-name}`
+   * `mdbook test book -L target/debug/deps`
 4. You have granted non-exclusive right to your source code under both the
    [MIT License][lm] and the [Apache License 2.0][la]. Unless you explicitly
    state otherwise, any contribution intentionally submitted for inclusion in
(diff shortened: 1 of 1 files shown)


docs/CONTRIBUTING.md
```diff
@@ -82,6 +83,10 @@ done the following things first:
 [la]: LICENSE-APACHE
 [rustfmt]: https://github.com/rust-lang-nursery/rustfmt
 
+Note that if you do not have mdbook already installed you may do so with `cargo install mdbook`. 
+If you find dependency resolution problems when testing mdbook,
```
azriel91: how about, "If mdbook fails with a multiple crate resolution error"

distransient: Would "If testing mdbook fails with crate resolution errors," work?

azriel91: it wouldn't always work -- there can be snippets that use `extern crate something`, but the `something` is missing from (dev-?) dependencies and it would give a different kind of crate resolution error

azriel91:
bors r=jojolepro, torkleyy, azriel91

PR #1605, opened 2019-05-16
Better sprite origin fix

valkum:
Avoids lines like: `let top = pixel_bottom / image_h;`


valkum:
As we did not change the spritegrids and spritelists in the examples and the visual result is equal, existing applications don't have to change their Sprites/SpritesheetsDefinitions.

We only changed the computation of UVs. In their context top,left, bottom, right is used in the SpritesheetsDefinitions. Not sure what you mean with Y. I think we did not change Y up to y down

In the old renderer the origin for x and y was the top left corner, but opengl has the origin for textures in the bottom left corner, thus the old renderer translated the x and y to fit opengls origin. Vulkan has the origin of textures in the top left corner thus we removed basically the old translation.

jojolepro:
Okay thank you! :+1:

PR #1611, opened 2019-05-17
Update the pong example for rendy

Moxinilian:
Updated the main pong example for rendy.


Moxinilian:
What am I supposed to do for format checks?

PR #1614, opened 2019-05-18
Custom game data example

valkum:
Fix custom game data for rendy.

As `renderable.ron` and the example was not Y-up I changed that to have consistency across examples.


valkum:
If the change to Y-up is not ok I can remove it after the WE.

PR #1616, opened 2019-05-18
Work around false positive warning

torkleyy:
Due to a regression in the Rust nightly compiler, `no_mangle` is incorrectly detected as "unused attribute".

## Description

Adds `allow(warnings)` to the symbol.


torkleyy:
Will revert once fixed on nightly.

bors r+

PR #1619, opened 2019-05-18
fix compilation bug

covercash2:
## Description

fixed a compilation bug with the nightly compiler.

## Modifications

i made it a little more idiomatic and easier to follow, IMO.

## PR Checklist

- [n/a] Updated the content of the book if this PR would make the book outdated.
- [n/a] Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.
- [n/a] Added unit tests for new APIs if any were added in this PR.

### DIFF
--- a/amethyst_assets/src/prefab/mod.rs
+++ b/amethyst_assets/src/prefab/mod.rs
@@ -380,22 +380,20 @@ where
         progress: &mut ProgressCounter,
         system_data: &mut Self::SystemData,
     ) -> Result<bool, Error> {
-        let handle = if let AssetPrefab::File(ref name, ref format, ref options) = *self {
-            Some(system_data.0.load(
-                name.as_ref(),
-                format.clone(),
-                options.clone(),
-                progress,
-                &system_data.2,
-            ))
-        } else {
-            None
-        };
-        if let Some(handle) = handle {
-            *self = AssetPrefab::Handle(handle);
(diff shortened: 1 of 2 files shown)


covercash2:
it should. using `return` is just personal preference. 

the compiler wasn't able to infer the type for `name.as_ref()`, and since the `load` method takes a `Into<String>` I thought just `clone`ing them would be effectively the same resource-wise. although I s'pose `name.as_ref::<&str>()` could work as well, but it doesn't save an allocation to pass a reference and the other resources are being cloned so I thought that would give a clearer picture of what was happening at the call site (not airtight reasoning, admittedly).

azriel91:
ah, sound enough reasoning for me 💯🥇

totorigolo:
This PR fixes the regression for `0.11`, so projects using earlier versions (and which don't want/cannot migrate) are stuck with Rust **1.35.0**. I personally don't really mind, because I consider [my game](https://github.com/totorigolo/fuite-rs) done, but I wanted to warn about this breakage, because older games will now just **not compile**. Maybe you should consider to backport fix this to older versions.

How-to fix for future readers:
* update to Amethyst 0.11+ (if you are actively developing a game)
* `rustup override set 1.35.0` / `cargo +1.35.0 run` (if you just want to run old games)

PR #1620, opened 2019-05-19
Remove Appendix A example

Moxinilian:
Removed the Appendix A example as it had diverged too much to be worth updating.
The only worthwhile feature it showcased was config files, which should be easy to compensate.


Moxinilian:
I do not have time to deal with the LFS issues unfortunately.

PR #1623, opened 2019-05-20
Rendy

Frizi:
## Description

A new rendering engine, among other things

## Modifications

* Brand new way to define rendering pipelines.
* OpenGL support temporarily dropped, Vulkan and Metal support added.
* Normalized texel coordinates are now in Vulkan convention (top-left 0.0, bottom-right 1.0), mirrored vertically compared to old one.
* World space is now Y-up consistently for all projections (2D and 3D).
* `Format` type no longer has associated `Options` and is now object-safe. It is expected to carry required options itself.
* `Format` now supports tag-based deserialization, it is no longer required to provide specific format to prefab type.
* Combined input axis/action generics into single type.
* `Material` is now an asset. Must be turned into handle before putting on an entity.
* Removed `Flipped` component. Use `flip_horizontal` and `flip_vertical` sprite property instead.

## PR Checklist


termhn:
@Frizi is the best 😍

mralve:
Absolutely fantastic work! This will be awesome to work with 💖

amethyst_rendy/src/formats/texture.rs
```diff
@@ -0,0 +1,181 @@
...
+        .with_sampler_info(hal::image::SamplerInfo::new(
+            filter,
+            hal::image::WrapMode::Clamp,
+        ))
+        .with_data(data)
+        .into()
+}
+
+impl TextureGenerator {
+    fn data(&self) -> TextureData {
```
azriel91: could this be made `pub`, so that other code can also use the generator without a prefab?

amethyst_rendy/src/formats/texture.rs
```diff
@@ -0,0 +1,181 @@
...
+    /// Placeholder during loading
+    #[serde(skip)]
+    Placeholder,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub enum TextureGenerator {
+    Srgba(f32, f32, f32, f32),
+    LinearRgba(f32, f32, f32, f32),
+    //LinearRgbaFloat(f32, f32, f32, f32),
```
azriel91: probably want to remove

omni-viral:
Now it's time to switch to `rendy = "0.2"`

jondo2010:
I'm not sure how the CI is building successfully, but when building `cargo build` I get
```
amethyst_rendy/src/system.rs:245:5
    |
245 | /     fn dispose(mut self: Box<Self>, res: &mut Resources) {
246 | |         if let Some(graph) = self.graph.take() {
247 | |             let mut factory = res.fetch_mut::<Factory<B>>();
248 | |             log::debug!("Dispose graph");
...   |
261 | |         drop(self.families);
262 | |     }
    | |_____^ not a member of trait `RunNow`
```
And indeed, `dispose()` is not a method on the `specs::RunNow` trait. The only `dispose()` I could find is in `shred=0.8.0`, which is more recent than the `shred=0.7.1` used transitively in the build.

omni-viral:
Github can't show me list of changed files.
How do I supposed to review this?

minecrawler:
@omni-viral Do you mean [this view](https://github.com/amethyst/amethyst/pull/1623/files)? Also, there is [reviewable.io](https://reviewable.io/reviews/amethyst/amethyst/1623#-), which gives you a better overview 😉

Frizi:
@jondo2010 This is because that change was backported to `shred` and no `specs` version bump have happened. You need to `cargo update`. This issue will not come up in fresh projects as well.

jaynus:
> @omni-viral Do you mean [this view](https://github.com/amethyst/amethyst/pull/1623/files)? Also, there is [reviewable.io](https://reviewable.io/reviews/amethyst/amethyst/1623#-), which gives you a better overview



> Github can't show me list of changed files.
> How do I supposed to review this?

Or https://help.github.com/en/articles/checking-out-pull-requests-locally :P

Frizi:
it's happening

PR #1624, opened 2019-05-21
Feature/amethyst test rendy

azriel91:
## Description

Added support to use `RenderingSystem` in `amethyst_test` based on `rendy`.

## Additions

* `amethyst_test` Rendering features are added under a `RenderAppExt` trait.

## PR Checklist

### DIFF
--- a/amethyst_rendy/src/render_test_bundle.rs
+++ b/amethyst_rendy/src/render_test_bundle.rs
@@ -1,81 +1,59 @@
 use std::{ops::Deref, sync::Arc};
 
-use amethyst::{
-    assets::Processor,
-    core::TransformBundle,
-    ecs::{ReadExpect, Resources, SystemData},
-    renderer::{
-        pass::{DrawFlat2DDesc, DrawFlat2DTransparentDesc},
-        rendy::{
-            factory::Factory,
-            graph::{
-                render::{RenderGroupDesc, SubpassBuilder},
-                GraphBuilder,
-            },
-            hal::{
-                command::{ClearDepthStencil, ClearValue},
-                format::Format,
-                image::Kind,
-            },
+use amethyst_assets::Processor;
+use amethyst_core::{
+    bundle::SystemBundle,
+    ecs::{DispatcherBuilder, ReadExpect, Resources, SystemData},
+};
+use amethyst_error::Error;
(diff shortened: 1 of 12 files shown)


Frizi:
This whole `X11_GL_MUTEX` is probably no longer needed, or at minimum should be renamed. There are now multiple possible backends, and none of them is GL :)

azriel91:
@Frizi Okay, except for some spurious / flakey segfaults in `rendy`, this branch appears to work end-to-end in my project.
`amethyst_renderer` isn't fully removed as a dependency, but I'll do that in a subsequent PR that makes `amethyst_test` only depend on:

* `amethyst_core`
* `amethyst_error` (for `Error`)
* `amethyst_utils` (for `application_root_dir` -- this could be in `core` though, to remove the dependency on `amehtyst_utils`)

torkleyy:
What's the status here?

azriel91:
waiting on @Frizi's review, i've tested it end-to-end in my project's CI, which was the main thing i wanted before calling it "ready for consumption"

amethyst_rendy/src/render_test_bundle.rs
```diff
@@ -0,0 +1,181 @@
...
+            "sprite_sheet_processor",
+            &[],
+        );
+        builder.add(
+            SpriteVisibilitySortingSystem::new(),
+            "sprite_visibility_system",
+            &["transform_system"],
+        );
+
+        builder.add_thread_local(RenderingSystem::<DefaultBackend, _>::new(
```
Frizi: `DefaultBackend` isn't really safe to be used here. This type alias is unavailable when you compile with multiple backends. We don't want to limit testing to one backend only.

azriel91: I've replaced it with a type parameter, is that okay?
(used the "i don't know what i'm doing, but it compiles" strategy)

azriel91:
btw @Frizi I've replaced `DefaultBackend` with a type parameter, is that okay? (comment is hidden by last push, not sure if you saw it. also, no pressure intended)

Frizi:
@azriel91 yeah it looks great :) This parameter is needed when creating the rendering system itself, so it's coherent.

torkleyy:
:shipit:
bors r+

PR #1625, opened 2019-05-21
PrefabData derive support for enums

alec-deason:
## Description

PrefabData derive support for enums.

This is my first ever macro code in rust and I'm still learning about how the prefab system works so I would super, super appreciate it if folks could double check that I haven't missed cases in the code or the test.

This requires an update to the book which I'm still working on (though this already includes an updated example). I'll add that to the PR shortly but I was hoping to get some eyes on the code while I work on the docs.

## Additions

You can now derive PrefabData for enums which is a thing I would like to do.

## Removals

n/a

## Modifications

n/a

## PR Checklist


amethyst_derive/tests/test.rs
```diff
@@ -75,3 +75,15 @@ pub struct Outer {
     #[prefab(Component)]
     external: External,
 }
+
+#[derive(PrefabData, Clone)]
```
jaynus: Since the test prefab data was switched over to only struct variants, will enum tuples and regular enums be supported by this code? Can all the different variant types get added to the tests and prefab example?

alec-deason: The existing code only supported named fields and I didn't change that so this only supports struct variants. My intention was to leave it at that but I think I see how you could support tuple variants (and tuple structs as well), it would add a bit of complexity but probably isn't too bad. I actually hadn't thought about regular enums at all. Let me check that...

alec-deason: Yeah, regular enums work fine. I've added a test for that.

alec-deason: Ok. There's support for tuple structs and tuple enum variants. Wasn't too bad. I think that left the code a bit cluttered so I may come back and tidy up some but it works and there are tests for it on both the struct and enum side.

alec-deason:
Ok. I think that's a reasonable update to the book for this stuff. I'm done changing things unless someone has additional comments.

book/src/prefabs/how_to_define_prefabs_aggregate.md
```diff
@@ -35,95 +30,98 @@ If you intend to include a [`Component`] that has not yet got a corresponding [`
...
-        /// **Note:** If a field is not specified in the prefab, it will default
-        /// to `None`.
-        #[derive(Debug, Default, Deserialize, Serialize, PrefabData)]
-        #[serde(default)]
-        #[serde(deny_unknown_fields)]
-        pub struct CustomPrefabData {
-            player: Option<Named>,
-            weapon: Option<Weapon>,
+    ```rust,edition2018,no_run,noplaypen
+    # extern crate amethyst;
```
torkleyy: I think we can get rid of all those `extern crate` statements.

book/src/prefabs/how_to_define_prefabs_aggregate.md
```diff
@@ -35,95 +30,98 @@ If you intend to include a [`Component`] that has not yet got a corresponding [`
...
+    # extern crate derivative;
+    # extern crate serde;
+    # extern crate specs_derive;
+    #
+    # use amethyst::{
```
torkleyy: Wouldn't it make sense to show these imports (and maybe even the two structs below)?

amethyst_derive/tests/test.rs
```diff
@@ -75,3 +75,20 @@ pub struct Outer {
...
+        number: Stuff<usize>,
+    },
+    Two {
+        #[prefab(Component)]
+        component: External,
+    },
+    Three {},
+    Four,
+    Five(Stuff<usize>, #[prefab(Component)] External),
+}
```
torkleyy: Could you add some actual tests (functions tagged with `#[test]`)? I think `EnumPrefab` covers the cases pretty well, we just need to check that it correctly instantiates the prefab.

book/src/prefabs/how_to_define_prefabs_aggregate.md
alec-deason: Yeah, I was following the style of the existing writeup but I think you're right, showing more is useful. I'll open it up.

amethyst_derive/tests/test.rs
alec-deason: Yeah, that's a good idea.

alec-deason:
As I start to use this in my project I realize that it falls down for a more complex, but common, case. If two variants both use the same resource then that leads to a double borrow so you have to go back to the superset struct with nullable fields model.

That's a bad experience obviously and makes this much less useful in practice. I think I see how to support the shared resource case but I'm going to need fiddle with it to make sure it works. I think I'd like to leave the scope of this PR as is and do the remaining work as a second PR unless folks thing this isn't worth having without support for the shared resource case.

book/src/prefabs/how_to_define_prefabs_aggregate.md
alec-deason: Is there a good way to check syntax on these snippets? `mdbook test` doesn't seem to run them.

torkleyy:
> If two variants both use the same resource then that leads to a double borrow so you have to go back to the superset struct with nullable fields model.

Ah, right, that was the reason it wasn't implemented in the first! I'm not sure how useful enum support is on its own, with the current behaviour.

If we want to merge enum support with the current limitation, it would be great to check duplicated resource access in the derive macro; I'm not sure how easy this is, and I believe we can only do this for collisions caused by one struct (not being able to check nested prefabs).

cc @azriel91 Any opinion on this?

alec-deason:
Assuming SystemData works the way I think it does then I can check for duplicates and emit a useful error even with arbitrarily deep nesting. So there's that at least.

If every SystemData in the tree is a tuple then I think I should be able to automatically construct everything that needs to get passed down while correctly handling duplicates. Since tuple SystemDatas are (I think) overwhelmingly the most common case then I believe that would be enough to be useful.

Both those things assume that my mental model of some stuff I don't really understand well is correct so I'll have to write code to know if I can actually do it. But I really want this feature so I'll poke at it today.

alec-deason:
No, wait. The derive macro only cares about the first level of the tree relative to itself. I think that's true... Makes it simpler.

book/src/prefabs/how_to_define_prefabs_aggregate.md
alec-deason: I included the structs but now that I look at this again I see that the imports are already visible in a block just above. I think that's cleaner that showing them twice so I'm going to leave it that way.

torkleyy: Ah, you're right, sorry. I've missed that.

alec-deason:
No, this is much harder. I'm not thinking clearly about what it means to be inside a macro. Sigh. Seems like it must be possible though...

book/src/prefabs/prefabs_in_amethyst.md
```diff
@@ -165,7 +165,7 @@ cargo run --example prefab_multi
 
 ### Multiple Entities, Different Components
 
-The next level is to instantiate multiple entities, each with their own set of [`Component`]s. The current implementation of [`Prefab`] requires the `data` field to be the same type for *every* [`PrefabEntity`] in the list. This means we would be unable to declare something like the following snippet, because it uses a `Player` prefab data in one entity, and `Weapon` in another:
+The next level is to instantiate multiple entities, each with their own set of [`Component`]s. The current implementation of [`Prefab`] requires the `data` field to be the same type for *every* [`PrefabEntity`] in the list. This means that to have different types of entity in the same prefab they must be variants of an enum. For instance, a prefab like this:
 
 ```rust,ignore
 // Note: Invalid / erroneous example
```
azriel91: no longer invalid 😄

book/src/prefabs/how_to_define_prefabs_aggregate.md
```diff
@@ -29,101 +24,91 @@ If you intend to include a [`Component`] that has not yet got a corresponding [`
...
+    
+    #[derive(Clone, Copy, Component, Debug, Derivative, Deserialize, Serialize, PrefabData)]
+    #[prefab(Component)]
+    #[storage(VecStorage)]
+    pub enum Weapon {
+        Axe,
+        Sword,
+    }
+    
+    /// All fields implement `PrefabData`, and are wrapped in `Option<_>`.
```
azriel91: This comment is somewhat inconsistent with the `enum`

book/src/prefabs/how_to_define_prefabs_aggregate.md
```diff
@@ -29,101 +24,91 @@ If you intend to include a [`Component`] that has not yet got a corresponding [`
         Error,
     };
     use serde::{Deserialize, Serialize};
+    use specs_derive::Component;
     ```
 
 3. Define the aggregate prefab data type.
 
     In these examples, `Named`, `Position`, and `Weapon` all derive [`PrefabData`].
```
azriel91: This line and the code snippet don't match up any more; I think the enum prefab data should come under here now

amethyst_derive/src/prefab_data.rs
```diff
@@ -33,47 +36,216 @@ fn impl_prefab_data_component(ast: &DeriveInput) -> TokenStream {
...
+                        Ident::new(&format!("field_{}", field_number), Span::call_site());
+                    let number = Literal::usize_unsuffixed(field_number);
+                    quote! {
+                        let #var_name = &self.#number;
+                    }
+                }
+            });
+    let extract_fields_sub = data.fields.iter().enumerate().map(|(field_number, field)| {
+        if !have_component_attribute(&field.attrs[..]) {
+            match &field.ident {
```
azriel91: small comment to add:
```rust
// Unnamed fields (tuple structs) do not have an ident, so we name based on their position instead.
```

amethyst_derive/src/prefab_data.rs
```diff
@@ -33,47 +36,216 @@ fn impl_prefab_data_component(ast: &DeriveInput) -> TokenStream {
...
+                None => Ident::new(&format!("field_{}", field_number), Span::call_site()),
+            })
+            .collect();
+        let field_names_sub: Vec<_> = variant
+            .fields
+            .iter()
+            .enumerate()
+            .map(|(field_number, field)| match &field.ident {
+                Some(name) => {
+                    if !have_component_attribute(&field.attrs[..]) {
```
azriel91: this negation appears in a number of places, so this tiny helper function would help my brain a bit better 😛

```rust
#[inline]
fn is_aggregate_prefab(attrs: &[Attribute]) -> bool {
    !have_component_attribute(attrs)
}
```

Maybe rename `have_component_attribute` to `is_component_prefab`

amethyst_derive/src/prefab_data.rs
```diff
@@ -33,47 +36,216 @@ fn impl_prefab_data_component(ast: &DeriveInput) -> TokenStream {
...
+            })
+            .collect();
+        let ident = &variant.ident;
+        add_to_entity.push(match variant.fields {
+            Fields::Named(_) => quote! {
+                #base::#ident {#(#field_names_add,)*} => {
+                    #(#variant_add_to_entity)*
+                }
+            },
+            Fields::Unnamed(_) => quote! {
```
azriel91: does it work if you collapse this with the previous branch? like

```rust
Fields::Named(_) | Fields::Unnamed(_) =>
```

same for `subs`

book/src/prefabs/how_to_define_prefabs_aggregate.md
```diff
@@ -132,48 +117,48 @@ If you intend to include a [`Component`] that has not yet got a corresponding [`
     Finally, the [`#[serde(deny_unknown_fields)]`] ensures that deserialization produces an error if it encounters an unknown field. This will help expose mistakes in the prefab file, such as when there is a typo.
 
 4. Now the type can be used in a prefab.
-
-    * *Superset* prefab data:
+    * *struct* prefab data:
```
azriel91: maybe backticks instead of italics now

alec-deason:
Ok I give up. I've churned through about six designs for how to solve the double borrow problem and all of them fall apart when I get into the details. I'm not sure you can do it without fundamentally changing shred. Or at least I really don't see the solution. I'll respond to the feedback on this PR (thank you everyone for your comments) and be done.

amethyst_derive/src/prefab_data.rs
alec-deason: Legit.

amethyst_derive/src/prefab_data.rs
alec-deason: Sadly, it doesn't. I'm producing tuple matching syntax in this branch and struct matching syntax in the other. I think I could remove some of the duplication from the two `quote!` blocks but I think at the cost of clarity.

azriel91:
Sorry I missed replying to the earlier comment about the double borrow. I think the way @alec-deason implemented it is the cleanest way it can be done (without attempting to change an *insurmountable amount of things* -- shred, specs, prefab system).

book/src/prefabs/how_to_define_prefabs_aggregate.md
alec-deason: Hmm, actually I think I disagree. It shows an example of grouping multiple `PrefabData`s into an aggregate first and then moves to the more complex example of using an enum to support heterogeneous types. That seems like the right flow to me.

alec-deason:
Ok, I believe I've responded to everyone's comments (thanks again for the reviews). The biggest change is a small new section in the book about the limitations of enum `PrefabData`s and how to get around them. Basically it brings back the discussion of superset prefabs from the original version.

book/src/prefabs/how_to_define_prefabs_aggregate.md
azriel91: oh yes, read the whole thing (instead of just the snippet) and it makes sense

azriel91:
bors r=jaynus, torkleyy, azriel91

PR #1633, opened 2019-05-23
[Rendy] Change camera back to global matrix

jaynus:
## Description

This changes the camera gatherer back to using global_matrix and calculating the inverse every frame. This incurs a 30ns/frame cost, but due to the way the transform system currently works with parenting we must fetch and calculate every frame, as view_matrix() on transform is local (so relative) and not global.

## Modifications

- Swaps rendy camera::gather to calculate global inverse view matrix every frame instead of getting local(relative) view_matrix()

## PR Checklist

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,13 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Label to use when marking an issue as stale
+staleLabel: stale
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed in 7 days if no further activity occurs.
+  Feel free to comment if this ticket is still relevant.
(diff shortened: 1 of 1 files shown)


amethyst_rendy/src/submodules/gather.rs
```diff
@@ -54,7 +54,13 @@ impl CameraGatherer {
             convert::<_, Vector3<f32>>(transform.global_matrix().column(3).xyz()).into_pod();
 
         let proj: [[f32; 4]; 4] = (*camera.as_matrix()).into();
-        let view: [[f32; 4]; 4] = convert::<_, Matrix4<f32>>(transform.view_matrix()).into();
+        let view: [[f32; 4]; 4] = convert::<_, Matrix4<f32>>(
+            transform
+                .global_matrix()
+                .try_inverse()
+                .expect("Unable to get inverse of camera transform"),
```
torkleyy: Can we avoid the `expect` here? I'd really prefer falling back to the identity matrix and logging a (one-time) error.

jaynus: This has been updated and removed with a solution from @termhn

PR #1635, opened 2019-05-23
[Rendy] Adds a static assertion during build time for a valid backend feature

### LINKED ISSUE #1632: [BUG] No physical devices found
## Description

When trying to run an example, I get the above panic.

```
cargo run --example material
```

## Screenshots and Logs

```
    Finished dev [unoptimized + debuginfo] target(s) in 9m 39s
     Running `target\debug\examples\material.exe`
[INFO][winit::platform::platform::window] Guessed window DPI factor: 1
[INFO][amethyst::app] Initializing Amethyst...
[INFO][amethyst::app] Version: 0.10.0
[INFO][amethyst::app] Platform: x86_64-pc-windows-gnu
[INFO][amethyst::app] Amethyst git commit: 1e7bd29511ee711c4053f85be82827fb7b7f4920
[INFO][amethyst::app] Rustc version: 1.33.0 Stable
[INFO][amethyst::app] Rustc git commit: 2aa4c46cfdd726e97360c2734835aa3515e8c858
[WARN][rendy_factory::factory] Slow safety checks are enabled! Disable them in production by enabling the 'no-slow-safet
y-checks' feature!
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ErrorMessage { msg: "No physical devices found"
}', src\libcore\result.rs:997:5
error: process didn't exit successfully: `target\debug\examples\material.exe` (exit code: 101)
```

## Your Environment

1. Operating System Windows
2. Operating System Version 10_64
3. Version of Amethyst master (1e7bd29511ee711c4053f85be82827fb7b7f4920)

## Additional Context

Tested on several two different computers with different hardware. Both, however, use Intel/NVIDIA hardware and Win10_64.

jaynus:
## Description

Adds a static assertion during build time for a valid backend feature. I'm not sure why we never included this - this will cause amethyst_rendy to compile time fail if a valid backend is not specified.

Fixes #1632 

## Additions

- static_assertion dependency

## PR Checklist


termhn:
It should just be `metal` and `vulkan` for now, since DX12 does not actually work with rendy currently.

jaynus:
Good point. I didn't realize this would make 'check' CI fail too.

Jenkinsfile
```diff
@@ -38,8 +38,12 @@ pipeline {
                         }
                     }
                     steps {
-                        echo 'Running Cargo check...'
-                        sh 'cargo check --all --all-targets --features sdl_controller,json,saveload'
+                        // Make sure static_assertion fails
+                        echo 'Confirming static assertion...'
+                        sh 'cargo check --all --all-targets || exit 0'
```
azriel91: should we go `&& exit 1` because we want it to fail on success?

jaynus: That command will fail because no backend feature was specified. So we want it to fail to succeed :)

azriel91: ya i mean:

```bash
fail && exit 1 || exit 0 # succeeds on failure, fails on pass
```

amethyst_rendy/src/lib.rs
```diff
@@ -49,3 +49,9 @@ pub use util::{simple_shader_set, ChangeDetection};
 pub mod loaders {
     pub use rendy::texture::palette::{load_from_linear_rgba, load_from_srgb, load_from_srgba};
 }
+
+// feature = "dx12" is removed here because rendy does not currently support it.
+static_assertions::assert_cfg!(
```
torkleyy: Should we warn / error if somebody enables dx12?

Jenkinsfile
torkleyy: @azriel91 is right, we definitely need to fail in case this command succeeds, but we must also not exit if it fails! That would skip the command below.

I'd suggest the following: 

```suggestion
                        sh 'cargo check --all --all-targets && exit 1 || true'
```

**Note:** This is technically no if-then-else; `&&` / `||` have left-associativity, thus in `A && B || C`, `C` will run if either `A` or `B` failed. This is only okay here because `B` will just terminate the whole command. Usually, you'll want to use an actual if-then-else block here.

Jenkinsfile
```diff
@@ -71,7 +80,7 @@ pipeline {
                     }
                     steps {
                         echo 'Beginning tests...'
-                        bat 'C:\\Users\\root\\.cargo\\bin\\cargo test --all'
+                        bat 'C:\\Users\\root\\.cargo\\bin\\cargo test --all --features "vulkan sdl_controller json saveload shader-compiler"'
```
Frizi: why is `shader-compiler` enabled here?

amethyst_rendy/src/lib.rs
Frizi: This technically shoudn't be removed. It's not rendy, but hal that doesn't support command buffers. The support might be added as non-breaking change though, because the api is already defined.

torkleyy: Hmm, but right now people can enable the `dx12` feature which leads to reports like #1643

Frizi:
What about `empty` backend? It is going to be used for testing once it actually has mock implementations, so at minimum it should be possible to explicitly choose it. Another thing is the definition of `type DefaultBackend` that currently is falling back to `empty` when no backend is specified. Removing that and treating `empty` on the same terms as other backends would implicitly fix the issue.

Frizi:
Additionaly, i think instead of erroring out, we should choose default valid backend based on platform when no flag was provided. That way it would be `Metal` on macOS and `Vulkan` everywhere else for now.

jaynus:
> Additionaly, i think instead of erroring out, we should choose default valid backend based on platform when no flag was provided. That way it would be `Metal` on macOS and `Vulkan` everywhere else for now.

As Far as I know, @Frizi this isn't possible with the way we propagate features in rendy right now. Because everything is done via feature flags, we are blocked by https://github.com/rust-lang/cargo/issues/1197 which means you cannot do conditions in targets or dependencies or features, you can only do the cfg(os_name) inclusions.

Because we cannot conditionally set a feature based on platform, the workaround appears to be having intermediate crates (gross). Otherwise, we are stuck forcing the user to do it by default, and then erroring. 

Thoughts?

torkleyy:
I think we should simply merge the compile error, once the review comments are addressed. Having a clear error message is a good improvement already.

jstnlef:
I know this is probably out of scope for this PR but it seemed like a fairly reasonable place to ask this. Is it possible for us to choose a reasonable default feature ("metal" or "vulkan") based on the OS building the project if one isn't specified? Mac -> "metal" else "vulkan"

Jenkinsfile
jaynus: Applied as suggested.

amethyst_rendy/src/lib.rs
jaynus: I've left this removed for now, since dx12 is not supported (and opengl support is coming quickly).  We can always re-add them when they are supported, but I'd rather steer people correctly for now.

Jenkinsfile
jaynus: Because I like to copy and paste things :P

jaynus:
All suggestions have been applied, including allowing empty as a valid backend. Dx12 is left out as it is not supported yet. I believe we can change those appropriately when supported and release if needed.

I'd rather not allow broken backends to be selected for now.

Sooo...this should be ready to merge. @Frizi @torkleyy pending your approval.

amethyst_rendy/src/types.rs
```diff
@@ -14,8 +14,8 @@ macro_rules! impl_backends {
 
         impl_single_default!($([$feature, $backend]),*);
 
-        #[cfg(not(any($(feature = $feature),*)))]
-        pub type DefaultBackend = rendy::empty::Backend;
+        //#[cfg(not(any($(feature = $feature),*)))]
+        //pub type DefaultBackend = rendy::empty::Backend;
```
Frizi: just remove those comments

Frizi: Or actually, we should have the static_assert here instead. Something like that
```rs
 static_assertions::assert_cfg!( 
     any($(feature = $feature),*),
     concat!("You must specify a at least one graphical backend feature:", stringify!($( '$feature'),*))
)

Comment (jaynus) [reply amethyst_rendy/src/types.rs]: Applied in latest commit.

File: amethyst_rendy/src/types.rs [hunk 5]
```diff
@@ -12,10 +12,13 @@ pub trait Backend: rendy::hal::Backend {
...
     ($($variant:ident, $feature:literal, $backend:ty;)*) => {
 
+
         impl_single_default!($([$feature, $backend]),*);
 
-        #[cfg(not(any($(feature = $feature),*)))]
-        pub type DefaultBackend = rendy::empty::Backend;
+        static_assertions::assert_cfg!(
+            any($(feature = $feature),*),
+            concat!("You must specify at least one graphical backend feature.: ", stringify!($($feature),* "See the wiki article https://github.com/amethyst/amethyst/wiki/GraphicalBackendError for more details."))
```
torkleyy:
```suggestion
            concat!("You must specify at least one graphical backend feature: ", stringify!($($feature),* "See the wiki article https://github.com/amethyst/amethyst/wiki/GraphicalBackendError for more details."))
```

torkleyy: Hmm, won't this print them with no whitespace / comma in between?

jaynus: Nope, I tested the output and tweaked it a bit to make sure it appeared clearly.

`jaynus@jaynus-home-desktop:~/dev/amethyst$ cargo build --example rendy --features "animation gltf"
   Compiling amethyst_rendy v0.1.0 (/home/jaynus/dev/amethyst/amethyst_rendy)
error: You must specify at least one graphical backend feature.: "metal" , "vulkan" , "empty"
"See the wiki article https://github.com/amethyst/amethyst/wiki/GraphicalBackendError for more details."
`

azriel91:
bors r+

at least we know it works 😄

jaynus:
bors r+
Try again!

jaynus:
bors retry

Jenkinsfile
```diff
@@ -71,7 +80,7 @@ pipeline {
                     }
                     steps {
                         echo 'Beginning tests...'
-                        bat 'C:\\Users\\root\\.cargo\\bin\\cargo test --all'
+                        bat 'C:\\Users\\root\\.cargo\\bin\\cargo test --all --features "vulkan sdl_controller json saveload"'
```
torkleyy: SDL2 dependency seems to be missing in our Windows setup.

jaynus: @fhaynes is this a possibility to get running, or should've disable the feature on windows?

jaynus: Ah, your watching closely @torkleyy . Yah, I just removed the SDL requirement. CI went swimmingly now.

PR #1637, opened 2019-05-23
Premultiply alpha when loading textures

termhn:
## Description

Makes the TextureData loader premultiply image alpha when loading. This is desirable in most cases 
 as it can prevent many common sorts of image artifacts resulting from texture filtering pixels that are meant to be transparent (see http://www.adriancourreges.com/blog/2017/05/09/beware-of-transparent-pixels/).

This change relies on https://github.com/amethyst/rendy/pull/143, to test it you must change `amethyst_rendy` Cargo.toml to point to that branch for now.

## Additions

- Adds `premultiply_alpha` field to `TexturePrefab`

## PR Checklist

- n/a Updated the content of the book if this PR would make the book outdated.
- n/a Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.
- n/a Added unit tests for new APIs if any were added in this PR.

### DIFF
--- a/amethyst_rendy/src/formats/texture.rs
+++ b/amethyst_rendy/src/formats/texture.rs
@@ -45,6 +45,7 @@ impl Default for ImageFormat {
                 anisotropic: Anisotropic::Off,
             },
             generate_mips: false,
+            premultiply_alpha: true,
         })
     }
 }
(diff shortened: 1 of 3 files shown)


termhn:
Superceeded by #1717

PR #1640, opened 2019-05-24
Add code coverage

torkleyy:
## Description

Make Jenkins collect code coverage reports and upload them to https://codecov.io/

Credit goes to @fhaynes, just opening the PR.

## PR Checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -117,7 +117,7 @@ pipeline {
             steps {
                 withCredentials([string(credentialsId: 'codecov_token', variable: 'CODECOV_TOKEN')]) {
                     echo 'Calculating code coverage...'
-                    sh 'for file in target/debug/amethyst_[a-f0-9]*[^\\.d]; do mkdir -p \"target/cov/$(basename $file)\"; kcov --exclude-pattern=/.cargo,/usr/lib --verify \"target/cov/$(basename $file)\" \"$file\"; done'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -88,6 +88,31 @@ pipeline {
                         echo 'Tests done!'
                     }
                 }
+                stage('Calculate Coverage') {
+                    environment {
+                        CARGO_HOME = '/home/jenkins/.cargo'
```
LucioFranco: you don't need these two iirc on unix.

Jenkinsfile
```diff
@@ -88,6 +88,26 @@ pipeline {
                         echo 'Tests done!'
                     }
                 }
+                stage('Calculate Coverage') {
```
LucioFranco: we should see if we can move this to its own stage? so its

check/rustfmt then tests then coverage

torkleyy: We can, but what's the advantage of that? Wouldn't that just increase the build time?

LucioFranco: only reason I see is if codecov ratelimits us and we prob dont wanna post coverage unless tests pass?

LucioFranco: but I dont mind either way.

torkleyy: I don't think it's gonna be a problem, so let me revert the broken commit I just pushed.

LucioFranco: Minor nit: maybe just `Coverage` here is better? 🤷‍♂

PR #1644, opened 2019-05-24
Version upgrade, added comments, tests, an better encapsulation.

TimonPost:
## Description

There is a problem that the version of `amethyst_network` wasn't upgraded yet, because of that I could not use the features that I added in #1543. This increments the version, adds some unit tests, some encapsulation functions, and an option to select laminar configuration.

## Additions

- `NetConnection`
both `send_buffer` and `receive_buffer` from `NetConnection` where public, I made those private and added some functions to get better encapsulation.

```rust
- conn.receive_buffer.register_id();
+ self.conn.register_id();

- conn.receive_buffer.read(reader_id);
+ self.conn.received_events(reader_id);

- conn.send_buffer.single_write(event);
+ self.conn.queue(event);

- conn.send_buffer.drain_vec_write(Vec<event>);
+self.conn.queue_vec(Vec<event>);

- conn.send_buffer.iter_write(event);
+ self.conn.queue_iter(iterator);
````
- Unit tests
## Modifications

- Added option to specify laminar config for amethyst network.
- I made two separate functions from `lib::send_event()`, `lib::send_event`, `lib::serialize_event()` because of that I could make better tests and it is clearer to see what is happening. 
- I added some comments with better explanations about how things work.
- Increased amethyst_network version and updated Cargo.toml
- Increased laminar version.
## PR Checklist

### CODE REVIEW

File: amethyst_network/src/connection.rs [hunk 1]
```diff
@@ -100,3 +161,59 @@ impl Default for NetIdentity {
...
     type Storage = VecStorage<NetIdentity>;
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::connection::NetConnection;
+    use crate::net_event::NetEvent;
+
+    #[test]
+    fn can_register_reader() {
```
jstnlef: Hrm. Seems to be missing an assert

examples/net_server/main.rs
```diff
@@ -54,19 +54,16 @@ impl<'a> System<'a> for SpamReceiveSystem {
...
-            for ev in connection
-                .receive_buffer
-                .read(self.reader.as_mut().unwrap())
-            {
+            for ev in connection.received_events(self.reader.as_mut().unwrap()) {
                 count += 1;
                 match ev {
                     NetEvent::Packet(packet) => info!("{}", packet.content()),
-                    NetEvent::Connected(addr) => println!("New Client Connection: {}", addr),
+                    NetEvent::Connected(addr) => panic!("New Client Connection: {}", addr),
```
jstnlef: You want to panic on a new connection here?

amethyst_network/src/network_socket.rs
```diff
@@ -24,18 +24,19 @@ enum InternalSocketEvent<E> {
...
-/// To send an event, add it to the NetSendBuffer resource.
+/// The System managing the network state from `NetConnections`.
 ///
-/// If both a connection (Connect or Connected) event is received at the same time as another event from the same connection,
-/// only the connection event will be considered and rest will be filtered out.
-// TODO: add Unchecked Event type list. Those events will be let pass the client connected filter (Example: NetEvent::Connect).
-// Current behaviour: hardcoded passthrough of Connect and Connected events.
+/// This system has a few responsibilities.
+///
+/// - Reading all send packets from `NetConnection` and send those over to some remote endpoint.
```
jstnlef: "Reading all sent packets from `NetConnection` and sending those over to some remote endpoint."

amethyst_network/src/network_socket.rs
```diff
@@ -24,18 +24,19 @@ enum InternalSocketEvent<E> {
...
+/// - Listen for incoming packets and queue the received packets (`NetEvent::Packet(...)`) on the accompanying `NetConnection`.
```
jstnlef: "Listening for incoming..."

amethyst_network/src/server/config.rs
```diff
@@ -37,6 +42,7 @@ impl Default for ServerConfig {
             udp_socket_addr: "0.0.0.0:0".parse().unwrap(),
             max_throughput: 5000,
             create_net_connection_on_connect: true,
+            laminar_config: None,
```
jstnlef: Should the default be a laminar default config `Config::default()` rather than `None`?

TimonPost: I have a reason for that but this might be better.

amethyst_network/src/network_socket.rs
TimonPost: I need to make it 'to send' actually because the packets are queued and need to be sent by the system.

"Reading to send packets from `NetConnection` and sending those over to some remote endpoint."

amethyst_network/src/connection.rs
TimonPost: I am going to remove this test, it can't fail and other tests are also using the register function.

TimonPost:
That PNG is a strange issue from rendy, I asked in #help for this,  they had problems with it as well. I tried to fix that problem by adding the image itself. However, it can be removed now, since I am using an other solution.

torkleyy:
@TimonPost Maybe a problem with git-lfs?

TimonPost:
@torkey yes it was, it seems like I don't have that error anymore.

When I did a git push I got this error: 

https://media.discordapp.net/attachments/476032191567233025/581607984224796702/unknown.png?width=960&height=38

jstnlef:
Seems to be failing a cargo check now.

TimonPost:
I am not sure why it gives the unused error of:  `net_event::NetEvent;` because it is an export at library level that should be fine right?

TimonPost:
Going to merge this one, there is one warning about an unused re-export, but it is meant to be there.

PR #1646, opened 2019-05-25
Fix book/concepts example code

torkleyy:
## Description

Fixes code sampels in the concept book chapters to pass `mdbook test`. Note that there are still chapters to fix (notably Pong), this is just a first pass.

cc #1647 

## PR Checklist


Moxinilian:
Maybe a RenderGraph chapter brief explanation in the concepts chapter could be good too, at least while the API is still a bit rough, what do you think?

torkleyy:
Yes, definitely, but for now I just want the book to build again :)

torkleyy:
Thanks for taking a look!
bors r=Frizi

PR #1661, opened 2019-05-29
Fix bad default in lod_range

alec-deason:
## Description

Creating a default ImageFormat resulted in lod_range having an end of -1536 rather than the expected 8000. This is because when gfx_hal::image::Lod converts an f32 it multiplies it by 8 and converts it to i16 which overflowed.

The behavior of Lod seems strange to me so maybe this is really a gfx bug but I'm not sure.

## Additions
n/a

## Removals
n/a

## Modifications

n/a

## PR Checklist


jaynus:
bors r=jaynus,jojolepro

PR #1662, opened 2019-05-29
[Rendy] Change DynamicVertex to a generic type for index buffers

jaynus:
## Description

The current implementation of `DynamicBuffer` is hard-coded to only create vertex buffers, and has some offsets hard coded. This implementation makes this instead generic `DynamicVertexData` (and `DynamicVertexBuffer` / `DynamicIndexBuffer`), which statically dispatches to the appropriate `bind` function of the rendy encoder based on its type parameter. This lets us dispatch for the correct usage at compile time without any hotpath overhead. 

## Additions

- Adds the generic parameters and other associated data in submodules::vertex

## Modifications

- Changes the struct tuple inner of amethyst_rendy::submodules::texture::TextureSub to public.

## PR Checklist

### DIFF
--- a/amethyst_rendy/src/submodules/texture.rs
+++ b/amethyst_rendy/src/submodules/texture.rs
@@ -29,7 +29,7 @@ enum TextureState<B: Backend> {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-pub struct TextureId(pub u32);
+pub struct TextureId(u32);
 
 #[derive(Debug)]
 pub struct TextureSub<B: Backend> {
--- a/amethyst_rendy/src/submodules/texture.rs
+++ b/amethyst_rendy/src/submodules/texture.rs
@@ -29,7 +29,7 @@ enum TextureState<B: Backend> {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-pub struct TextureId(u32);
+pub struct TextureId(pub u32);
 
 #[derive(Debug)]
 pub struct TextureSub<B: Backend> {
--- a/amethyst_rendy/src/submodules/vertex.rs
+++ b/amethyst_rendy/src/submodules/vertex.rs
@@ -1,3 +1,6 @@
+//! Wrapper and management data structures for providing automatic buffering, resizing and management
+//! of rendy vertex buffer types.
+
 use crate::{
     rendy::{
         command::RenderPassEncoder,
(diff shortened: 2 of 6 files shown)


amethyst_rendy/src/submodules/texture.rs
```diff
@@ -29,7 +29,7 @@ enum TextureState<B: Backend> {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-pub struct TextureId(u32);
+pub struct TextureId(pub u32);
```
Frizi: This wrapper type is done specifically to not allow constructing arbitrary texture ids out of thin air. The inner type should not be public

amethyst_rendy/src/submodules/vertex.rs
```diff
@@ -11,25 +14,121 @@ use crate::{
...
+    }
+}
+
+/// This structure wraps [PerImageDynamicVertexData], managing multiple instances and providing
+/// an easy-to-use interface for having per-image buffers. This is needed because multiple images
+/// (frames) can be in flight at any given time, so multiple buffers are needed for the same data.
 #[derive(Debug)]
-struct PerImageDynamicVertex<B: Backend> {
-    buffer: Option<Escape<Buffer<B>>>,
+pub struct DynamicVertexData<B: Backend, V: VertexDataBufferType<B>, T: 'static> {
```
Frizi: i believe you don't need `T` for anything here.

amethyst_rendy/src/submodules/vertex.rs
```diff
@@ -70,35 +169,53 @@ impl<B: Backend, T: 'static> DynamicVertex<B, T> {
...
         &self,
         index: usize,
         binding_id: u32,
+        offset: u64,
         encoder: &mut RenderPassEncoder<'_, B>,
     ) -> bool {
         self.per_image
             .get(index)
-            .map_or(false, |i| i.bind(binding_id, encoder))
+            .map_or(false, |i| V::bind(binding_id, encoder, &i.buffer, offset))
```
Frizi: it feels wrong to me that `binding_id` is being used only sometimes here. I think the whole `bind` method should be split in half and implemented only for relevant types. One for each impl of
```rs
impl<B: Backend, T: 'static> DynamicVertexData<B, VertexData<T>> { .. }
impl<B: Backend, T: 'static> DynamicVertexData<B, IndexData<T>> { .. }
```

amethyst_rendy/src/submodules/texture.rs
jaynus: Removed.

amethyst_rendy/src/submodules/vertex.rs
jaynus: It's used below at `I::Item: AsRef<[T]>,`

amethyst_rendy/src/submodules/vertex.rs
jaynus: I've changed up the implementation now to specialize for each case and call the appropriate binding function, so there is no spare random value now.

amethyst_rendy/src/submodules/vertex.rs
```diff
@@ -69,36 +157,76 @@ impl<B: Backend, T: 'static> DynamicVertex<B, T> {
...
+        })
     }
 }
 
-impl<B: Backend> PerImageDynamicVertex<B> {
+/// an easy-to-use interface for managing, growing and binding a given vertex buffer type. This
+/// implementation also leverages the [VertexDataBufferType] trait type for statically dispatching
+/// the appropriate binding and allocation functions, preventing hot-path branching.
+#[derive(Debug)]
+pub struct PerImageDynamicVertexData<B: Backend, V: VertexDataBufferType> {
```
Frizi: this should not be `pub`

jaynus:
bors r=jojolepro,Frizi

Issue #1666, opened 2019-05-31
[FEATURE MODIFICATION] The engine struct naming are not clear enough


AndreaCatania:
The standard components in the engine have names that doesn't immediately tells the type of the struct that you are dealing with.
By doing so you don't know if a struct is a component or another thing until you open the relative implementation file.

A solution to this problem could be add "Component" as suffix to all components in the engine in order to have something like: SpriteRenderComponent (or SpriteRenderComp)

In this way the names are way more consistent, and when you read the code is immediately clear what you are dealing with.

Another way to view this is, If you read this code:

>         let sprite_render = SpriteRender {
>             sprite_sheet: sprite_sheet.clone(),
>             sprite_number: 0
>         };

SpriteRender what it is? A System, a component, a state or what?

>         let sprite_render = SpriteRenderComp {
>             sprite_sheet: sprite_sheet.clone(),
>             sprite_number: 0
>         };

Instead in this second case is clear that you are dealing with a component.

What I propose is to add the suffix **comp** to all default components of the engine and thus make it the code way more clear.

Also @piedoom in the discord chat said that currently we use this naming strategy with the Systems structures, and he said that he usually use the namespace to differentiate the components in its code.

BrianMWest:
I would probably bite the bullet and not abbreviate it - `SpriteRenderComponent`. Otherwise, I know that I will at some point forget whether it was abbreviated as C, Com, Comp, or something else. I wouldn't think that reducing verbosity in the type name should be too critical here. If someone wants to reduce the verbosity, they can always type alias on their end.

Only other thing I would say is that all or at least most of the original struct names should be type aliased to the component version with some comment indicating that it is a deprecated type name and the Component suffix version should be used.

AndreaCatania:
This is a good point, instead do you know if is it possible to make the compiler stop when you use the old struct name and tell exactly which type you have to use from now on?

I'll do a PR later today

jojolepro:
Considering the answers I'm seeing in this forum thread https://community.amethyst.rs/t/rename-all-components-to-have-component-suffix/878/27 
I'm closing this issue. Please re-open if it is still valid.

PR #1670, opened 2019-06-02
Improved the naming of the components, by adding "Component" as suffix

AndreaCatania:
## Description
The idea of this PR is to make sure that each component has the suffix "Component" in order to differentiate it to all others structures inside the engine.

Please check this issue, where I open the discussion about this change: https://github.com/amethyst/amethyst/issues/1666

## PR Checklist

About the "Update the book", is done, but I would like if someone check the links because I left them as is because I'm not sure what to do with them.


jojolepro:
Since this is a major change, I opened a discussion on the forum to gets everyone's arguments and reasoning.
https://community.amethyst.rs/t/rename-all-components-to-have-component-suffix/878

AndreaCatania:
Thanks to get on this

PR #1678, opened 2019-06-06
Fix some more clippy warnings

0x6273:
## Description

Fixes 120 clippy lint warnings. The 57 warnings that remain are mostly related to code complexity (very large functions, functions that take 10+ arguments, etc.).

## Additions

- impl Default on some types
- Add is_empty method on types that have a len method.

## Modifications

- Various code style improvements.
- Some function signatures were changed slightly.

## PR Checklist

- [x] Updated the content of the book if this PR would make the book outdated. (it won't)
- [x] Added unit tests for new APIs if any were added in this PR. (none added)

### DIFF
--- a/amethyst_animation/src/systems/control.rs
+++ b/amethyst_animation/src/systems/control.rs
@@ -99,7 +99,7 @@ where
                         .get(&control.animation)
                         .and_then(|animation| {
                             process_animation_control(
-                                &entity,
+                                entity,
                                 animation,
                                 control,
                                 hierarchy,
@@ -126,7 +126,7 @@ where
                 } else {
                     self.state_set.insert(
                         *id,
-                        get_running_duration(&entity, control, hierarchies.get(entity), &samplers),
+                        get_running_duration(entity, control, hierarchies.get(entity), &samplers),
                     );
                 }
             }
@@ -174,7 +174,7 @@ where
                         .get(&def.control.animation)
                         .and_then(|animation| {
                             process_animation_control(
-                                &entity,
+                                entity,
                                 animation,
                                 &mut def.control,
                                 hierarchy,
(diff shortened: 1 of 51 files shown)


amethyst_assets/src/prefab/system.rs
```diff
@@ -129,8 +129,8 @@ where
 
                         children
                             .entry(parent)
-                            .or_insert(vec![])
-                            .push(new_entity.clone());
+                            .or_insert_with(|| vec![])
```
torkleyy: This should probably be `Vec::new` instead.

torkleyy:
```suggestion
                            .or_insert_with(Vec::new)
```

0x6273: Not sure what the difference is so I left it as is. Can I replace all instances of `vec![]` with `Vec::new()`.

torkleyy: Ah, what I meant is not a closure; you can pass the function directly and it will coerce into a closure (I think clippy will even tell you that).

0x6273: Ah I see what you mean. It's fixed now.

amethyst_animation/src/systems/control.rs
```diff
@@ -447,14 +447,15 @@ where
     T: AnimationSampling + Component + Clone,
 {
     // check that hierarchy is valid, and all samplers exist
-    if animation
+    let valid = animation
         .nodes
         .iter()
-        .any(|&(ref node_index, _, ref sampler_handle)| {
```
Dispersia: Wouldn't any be preferred here? All will make it go through all of them even if it's not necessary

PR #1681, opened 2019-06-09
stage -> state typo fix

tpltnt:
## Description

This PR fixes two typos. The section on the concept of state switches to "stage" for no apparent reason and probably meant "state".

## PR Checklist

### DIFF
--- a/book/src/concepts/state.md
+++ b/book/src/concepts/state.md
@@ -2,9 +2,9 @@
 
 ## What is a state?
 The word "state" can mean a lot of different things in computer science.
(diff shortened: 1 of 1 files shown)


jojolepro:
In this context, I'm pretty sure it was meant to be "state".

torkleyy:
Alright ^^
bors r=Moxinilian,jojolepro

Moxinilian:
Yeah in my opinion stage is too narrow of a term. It would have meant something like a platformer level, where here it really does mean any state.

PR #1687, opened 2019-06-10
Implement `const fn new` for `Float`

azriel91:
## Description

~~Change `Float(FloatBase)` to `Float(pub FloatBase)`, so that we can construct constants.~~

Implemented `const fn new(value: <f32/64>)` for `Float` based on feature flags. This allows us to declare `Float` constants.

```rust
const FIELD: Float = Float::new(123.);
```

## Modifications

* Make `Float` inner field public.

## PR Checklist

### DIFF
--- a/amethyst_core/src/float.rs
+++ b/amethyst_core/src/float.rs
@@ -29,14 +29,27 @@ type FloatBase = f32;
 #[derive(Alga, Clone, Copy, PartialOrd, PartialEq, Serialize, Deserialize, Debug)]
 #[alga_traits(Field(Additive, Multiplicative))]
 #[serde(transparent)]
-pub struct Float(pub FloatBase);
+pub struct Float(FloatBase);
 
 impl Float {
+    /// Returns a new `Float`.
+    #[cfg(feature = "float64")]
+    pub const fn new(value: f64) -> Self {
+        Float(value)
+    }
+
+    /// Returns a new `Float`.
+    #[cfg(not(feature = "float64"))]
+    pub const fn new(value: f32) -> Self {
+        Float(value)
+    }
+
     /// Get the internal value as a f32. Will cause a loss in precision if using
     /// the "float64" feature.
     pub fn as_f32(self) -> f32 {
         self.0 as f32
     }
+
     /// Get the internal value as a f64. Guaranteed to be lossless.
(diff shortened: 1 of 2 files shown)


jojolepro:
This leaks the inner type and goes against the encapsulation provided by Float which has for goal to prevent the real type from ever being publicly visible. I would suggest to not do it this way.
Being able to do float.0 + (some_var as f32) would break depending on the feature flag being used or not.

azriel91:
ah, wonder if it works if `Float::new` is made into a `const fn`

azriel91:
How's it looking now?

amethyst_core/src/float.rs
```diff
@@ -32,11 +32,22 @@ type FloatBase = f32;
 pub struct Float(FloatBase);
 
 impl Float {
+    /// Returns a new `Float` from a `f64`.
+    pub const fn from_f64(value: f64) -> Self {
+        Float(value)
+    }
+
+    /// Returns a new `Float` from a `f64`.
```
jojolepro:
```suggestion
    /// Returns a new `Float` from a `f32` using a constant function.
```

amethyst_core/src/float.rs
```diff
@@ -32,11 +32,22 @@ type FloatBase = f32;
 pub struct Float(FloatBase);
 
 impl Float {
+    /// Returns a new `Float` from a `f64`.
```
jojolepro:
```suggestion
    /// Returns a new `Float` from a `f64` using a constant function.
```

amethyst_core/src/float.rs
Frizi: Same as above. Not necessarily const always.

Frizi: Although in this case it should be `f32` indeed :D

Frizi:
We should also make reverse operation `const`, like `into_f32` and `into_f64`. That way you can do whole const expressions in chosen precision without ever leaking the inner type.

amethyst_core/src/float.rs
azriel91: oops, fixed that typo!

azriel91:
cool, that last one -- changing `as_f64` to `const` changed the impl to use `as f64` instead of `f64::from` -- as `from` is not `const`

PR #1689, opened 2019-06-11
Re-export ActiveCamera in the root of amethyst_rendy

jojolepro:


jojolepro:
wip

jojolepro:
done

Frizi:
rustfmt pass required

amethyst_rendy/src/camera.rs
```diff
@@ -468,10 +468,10 @@ impl<'a> PrefabData<'a> for CameraPrefab {
 }
 
 /// Active camera prefab
-pub struct ActiveCameraPrefab(usize);
+pub struct ActiveCameraPrefab(Option<usize>);
```
torkleyy: Is this even used anywhere? It doesn't even implement `Deserialize`, so I don't quite see how this is useful.

jojolepro:
Fixed

@azriel91 the breaking change was that the Option was removed in the first place, and that it was not documented anywhere. This reverts the how it was before the rendy merge.

PR #1701, opened 2019-06-12
[FEEDBACK] Transform Parent Relation

jaynus:
## Description

This adds a new optional relationship to transforms which allows you to dictate how it is updated in relation to its parent.Currently, there is no way to specify this. 

This PR adds 
```rust
pub enum ParentTransformRelation {
    /// A relative transformation, which is the default, will modify this transform based on its
    /// parents
    Relative,
    /// An absolutely transform will ignore parents and be considered a world transform.
    Absolute,
}
```
This specifies whether a Transform should be updated relative to its parent, or stay an absolute world coordinate. 

This allows the use of the built-in parent and Hierarchy system without necessarily forcing relative transforms for everything all the time.  

## Additions

- amethyst::core::transform::ParentTransformRelation
- Transform now has `parent_relation` and appropriate getters/setters
- Added optional deserialization and serialization for the type so its backwards compat.

## PR Checklist

### DIFF
--- a/amethyst_core/src/transform/components/transform.rs
+++ b/amethyst_core/src/transform/components/transform.rs
@@ -698,7 +698,7 @@ impl<'de> Deserialize<'de> for Transform {
 
                 let parent_relation: ParentTransformRelation = seq
                     .next_element()?
-                    .map_or_else(|| ParentTransformRelation::Relative, |v| v);
+                    .unwrap_or(ParentTransformRelation::Relative);
 
                 let isometry = Isometry3::from_parts(
                     Translation3::new(translation[0], translation[1], translation[2]),
--- a/amethyst_core/src/transform/components/transform.rs
+++ b/amethyst_core/src/transform/components/transform.rs
@@ -16,6 +16,21 @@ use serde::{
     ser::{Serialize, Serializer},
 };
 
+/// Defines a transforms relationship to any parents it may have.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
+pub enum ParentTransformRelation {
+    /// A relative transformation, which is the default, will modify this transform based on its
+    /// parents
+    Relative,
+    /// An absolutely transform will ignore parents and be considered a world transform.
+    Absolute,
+}
+impl Default for ParentTransformRelation {
+    fn default() -> Self {
+        ParentTransformRelation::Relative
+    }
+}
+
 /// Local position, rotation, and scale (from parent if it exists).
 ///
 /// Used for rendering position and orientation.
@@ -35,6 +50,11 @@ pub struct Transform {
     /// The global transformation matrix.
     #[get = "pub"]
     pub(crate) global_matrix: Matrix4<Float>,
+
+    #[get = "pub"]
+    #[set = "pub"]
+    #[get_mut = "pub"]
+    pub(crate) parent_relation: ParentTransformRelation,
 }
 
 impl Transform {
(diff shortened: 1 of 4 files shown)


amethyst_core/src/transform/components/transform.rs
```diff
@@ -673,6 +696,10 @@ impl<'de> Deserialize<'de> for Transform {
                     .next_element()?
                     .ok_or_else(|| de::Error::invalid_length(2, &self))?;
 
+                let parent_relation: ParentTransformRelation = seq
+                    .next_element()?
+                    .map_or_else(|| ParentTransformRelation::Relative, |v| v);
```
azriel91: does `|v| v` do anything? (would `unwrap_or_else` be more apt?)
also, there's a `.copied()` in case the mapping copies the value

jaynus: Done and updated! Dunno why I did that, I must be tired.

jaynus:
> If you want to always make a relative change by default, but at rare occasions you want to make an absolute change, then you need to change the inner ParentTransformRelation before and after doing the absolute change?

Are you referring to the edge case of changing a transform from relative to absolute at runtime? I hadn't considered that case, and worked on the assumption that a Transform would stay one or the other.

Perhaps move this to an internal-only variable, and only allow it to be set at construction? That would be a breaking change though to change new, or add a new constructor function.

azriel91:
how about just make the `parent_relation` field `pub`, then people can modify it?

torkleyy:
Maybe it would make sense to make this relation an extra component?

RybekU:
jaynus asked me to add my question about parenting system here, which goes as follow:
> If I wanted to make one entity follow another one, but not outright attached to it on both axises, a bit of freedom on axis Y and strictly attached on axis X setting one as parent is probably not what I'm looking for?

For it to work with parenting system it would have to be able to:
- at minimum allow to choose which axises are relative and which are absolute
- additionaly could allow to force entity to be in a proximity of a parent on given axis (another component?)

kabergstrom:
Unity also has this persistent problem of overloading the semantics of the Transform hierarchy.

- Grouping objects (like folders) in the Scene Hierarchy View
- Local transformation relations (object follow)
- Prefab system sub-objects require usage of Transform parent (even if you don't want the follow behaviour)
- Particle system sub-particles
- UI system uses the Transform hierarchy to determine drawing order (based on child indices)
- UI system traverses the Transform hierarchy to determine Canvas batching and which "UI camera" to draw with
- Scenes clean up whole Transform hierarchies from a set of roots when unloaded

and probably more that I can't think of now.

It's used for so many things, which becomes quite limiting when you want to use multiple of these systems in incompatible ways due to their requirements on the hierarchy. I don't think introducing more fields and complexity in the Transform component is the right way forward for this reason. You are imposing overhead for even default usage.

Can you write down more motivation for this change?

OvermindDL1:
For note, in my engine I had a number of different transforms, which generally mattered based on what it was parented to, and parented things children were *always* relative, the concept of something being 'attached' to something was different from being parented in the scene hierarchy and thus were different components.  In addition GUI components had a GUI-specific transform (useful for attaching anchor points and so forth, and the entire GUI-tree itself could be parented to a 3D component to render it in-scene with all normal transformations and all).

Don't overload transform to mean different things in different contexts.  If anything there shouldn't be a transform just called 'Transform', but rather lots of different ones depending on the context of how it is built and handled (I think I had like 4 different ones on average?).  In addition, there was *always* a parent, everything was relative to some parent, and the base parent was the scene itself that held everything and managed render state (and there could be multiple cameras on a scene each rendering to a different framebuffer for different configurations and overlays and shaders and so forth, in addition there would often by multiple scenes each with their own multiple cameras, for 'distinct' areas that could be viewed into or 'overlaid' between (think portals)).

jojolepro:
Should we close this for the time being and make an rfc discussing in details the changes you want to make as well as the reasoning and their impact?

jaynus:
Sure, let's discuss outside a PR

PR #1704, opened 2019-06-13
Minor fixes

torkleyy:
* Avoid warnings when formatting
* ~~Remove file logging from animation example~~
* Add "Amethyst Foundation" as author

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "amethyst"
 version = "0.10.0"
-authors = ["Eyal Kalderon <ebkalderon@gmail.com>"]
+authors = ["Eyal Kalderon <ebkalderon@gmail.com>", "Amethyst Foundation <contact@amethyst.rs>"]
 edition = "2018"
 description = "Data-oriented game engine written in Rust"
 exclude = ["examples/*", "book/*"]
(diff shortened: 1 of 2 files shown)


azriel91:
ah, `merge_imports` isn't going to be stable any time soon https://github.com/rust-lang/rustfmt/issues/3362

if people still want it, they can add it to `rustfmt.toml` in a parent directory.

Frizi:
Having `merge_imports` in the settings makes it more official that it should be used, even if not strictly required. Also it's used by default for everyone on nightly (and you can use `cargo +nightly fmt` if you use stable compiler). If we will remove it, it's likely that people who use it will have much more noise in their PRs due to random reformats.

I see the point about it being unstable though, the linked issue shows an edge case where it breaks currently. I have never had an issue with that personally though.

Can we somehow remove the warnings without disabling the rule for nightly fmt?

torkleyy:
> Can we somehow remove the warnings without disabling the rule for nightly fmt?

Hmm, not sure. I don't think we can.

azriel91:
There's this:

```bash
$ cargo +stable fmt
Warning: can't set `merge_imports = true`, unstable features are only available in nightly channel.
# ... x100

$ cargo +stable fmt 2> >(grep -vF merge_imports)
# all good
```

torkleyy:
That looks... slightly too complicated :smile:

torkleyy:
Removed the commit.

torkleyy:
bors r=Frizi,azriel91

PR #1705, opened 2019-06-13
Implement the mouse as input axis

minecrawler:
## Description

Added the mouse as an axis for easy rebinding

## Additions

- Added structs for Mouse and MouseAxis to be used for input

## PR Checklist

### DIFF
--- a/amethyst_input/src/axis.rs
+++ b/amethyst_input/src/axis.rs
@@ -1,6 +1,6 @@
 use serde::{Deserialize, Serialize};
 
-use super::{Button, ControllerAxis};
+use super::{Button, ControllerAxis, MouseAxis};
 
 /// Represents any input represented by a float value from -1 to 1.
 /// Retrieve the value of this with [axis_value](struct.InputHandler.html#method.axis_value).
@@ -25,6 +25,17 @@ pub enum Axis {
         /// Treat input values from -dead_zone to dead_zone as 0,
         /// linearly interpolate remaining ranges.
         dead_zone: f64,
+    },
+    /// Represents a mouse as a 2D input device
+    Mouse {
+        /// The axis being bound
+        axis: MouseAxis,
+        /// Should the API be allowed to return values outside [-1..1]?
+        over_extendable: bool,
+        /// Zone to which the movement is relative on the X axis
+        radius_x: f32,
+        /// Zone to which the movement is relative on the Y axis
+        radius_y: f32,
     },
     /// Represents the wheel on a PC mouse.
     MouseWheel {
--- a/amethyst_input/src/input_handler.rs
+++ b/amethyst_input/src/input_handler.rs
@@ -37,6 +37,7 @@ where
     /// First number represents mapped ID visible to the user code,
     /// while second is the ID used by incoming events.
     connected_controllers: SmallVec<[(u32, u32); 8]>,
+    mouse_last_position: Option<(f32, f32)>,
     mouse_position: Option<(f32, f32)>,
     mouse_wheel_vertical: f32,
     mouse_wheel_horizontal: f32,
@@ -404,6 +405,7 @@ where
     pub fn send_frame_begin(&mut self) {
         self.mouse_wheel_vertical = 0.0;
         self.mouse_wheel_horizontal = 0.0;
+        self.mouse_last_position = self.mouse_position.clone();
     }
 
     /// Returns an iterator over all keys that are down.
@@ -552,6 +554,35 @@ where
                     }
                 })
                 .unwrap_or(0.0),
+            Axis::Mouse {
+                axis,
+                over_extendable,
+                radius_x,
+                radius_y,
+            } => {
+                let current_pos = self.mouse_position.unwrap_or((0., 0.));
+                let last_pos = self.mouse_last_position.unwrap_or(current_pos);
+                let delta = match axis {
+                    // These calculations have to be inverses in order to point into the right direction of movement
+                    MouseAxis::X => last_pos.0 - current_pos.0,
+                    MouseAxis::Y => last_pos.1 - current_pos.1,
+                };
+
+                let rel_delta = match axis {
+                    MouseAxis::X => delta / radius_x,
+                    MouseAxis::Y => delta / radius_y,
+                };
+
+                if over_extendable {
+                    rel_delta
+                } else if rel_delta > 1. {
+                    1.
+                } else if rel_delta < -1. {
+                    -1.
+                } else {
+                    rel_delta
+                }
+            }
             Axis::MouseWheel { horizontal } => self.mouse_wheel_value(horizontal),
         })
     }
--- a/amethyst_input/src/mouse.rs
+++ b/amethyst_input/src/mouse.rs
@@ -1,0 +1,10 @@
+use serde::{Deserialize, Serialize};
+
+/// Mouse axis
+#[derive(Eq, PartialEq, Debug, Copy, Clone, Serialize, Deserialize)]
+pub enum MouseAxis {
+    /// The X axis represents moving the mouse left / right
+    X,
+    /// The Y axis represents the mouse moving up / down
+    Y,
+}
--- a/amethyst_input/src/bindings.rs
+++ b/amethyst_input/src/bindings.rs
@@ -125,21 +125,23 @@ pub struct Bindings<T: BindingTypes> {
 #[derive(Clone, Derivative)]
 #[derivative(Debug(bound = ""))]
 pub enum BindingError<T: BindingTypes> {
-    /// You attempted to bind a mousewheel axis twice.
-    MouseWheelAxisAlreadyBound(T::Axis),
-    /// Combo provided for action binding has two (or more) of the same button.
-    ComboContainsDuplicates(T::Action),
-    /// Combo provided was already bound to the contained action.
-    ComboAlreadyBound(T::Action),
-    /// A combo of length one was provided, and it overlaps with an axis binding.
-    ButtonBoundToAxis(T::Axis, Axis),
-    /// Axis buttons provided have overlap with an existing axis.
-    AxisButtonAlreadyBoundToAxis(T::Axis, Axis),
     /// Axis buttons have overlap with an action combo of length one.
     AxisButtonAlreadyBoundToAction(T::Action, Button),
+    /// Axis buttons provided have overlap with an existing axis.
+    AxisButtonAlreadyBoundToAxis(T::Axis, Axis),
+    /// A combo of length one was provided, and it overlaps with an axis binding.
+    ButtonBoundToAxis(T::Axis, Axis),
+    /// Combo provided was already bound to the contained action.
+    ComboAlreadyBound(T::Action),
+    /// Combo provided for action binding has two (or more) of the same button.
+    ComboContainsDuplicates(T::Action),
     /// That specific axis on that specific controller is already in use for an
     /// axis binding.
     ControllerAxisAlreadyBound(T::Axis),
+    /// The given axis was already bound for use
+    MouseAxisAlreadyBound(T::Axis),
+    /// You attempted to bind a mousewheel axis twice.
+    MouseWheelAxisAlreadyBound(T::Axis),
 }
 
 impl<T: BindingTypes> PartialEq for BindingError<T> {
@@ -204,6 +206,9 @@ where
             ),
             BindingError::ControllerAxisAlreadyBound(ref id) => {
                 write!(f, "Controller axis provided is already in use by {}", id)
+            }
+            BindingError::MouseAxisAlreadyBound(ref id) => {
+                write!(f, "Mouse axis provided is already in use by {}", id)
             }
             BindingError::MouseWheelAxisAlreadyBound(ref id) => {
                 write!(f, "Mouse wheel axis provided is already in use by {}", id)
@@ -481,6 +486,18 @@ impl<T: BindingTypes> Bindings<T> {
                     {
                         if controller_id == input_controller_id && axis == input_axis {
                             return Err(BindingError::ControllerAxisAlreadyBound(k.clone()));
+                        }
+                    }
+                }
+            }
+            Axis::Mouse { axis, .. } => {
+                for (k, a) in self.axes.iter().filter(|(k, _a)| *k != id) {
+                    if let Axis::Mouse {
+                        axis: mouse_axis, ..
+                    } = a
+                    {
+                        if axis == mouse_axis {
+                            return Err(BindingError::MouseAxisAlreadyBound(k.clone()));
                         }
                     }
                 }
--- a/amethyst_input/src/lib.rs
+++ b/amethyst_input/src/lib.rs
@@ -19,6 +19,7 @@ pub use self::{
     controller::{ControllerAxis, ControllerButton, ControllerEvent},
     event::InputEvent,
     input_handler::InputHandler,
+    mouse::MouseAxis,
     scroll_direction::ScrollDirection,
     system::{InputSystem, InputSystemDesc},
     util::{
@@ -39,6 +40,7 @@ mod button;
 mod controller;
 mod event;
 mod input_handler;
+mod mouse;
 mod scroll_direction;
 mod system;
 mod util;


Frizi:
I really like the idea :) One worry is that all our axes are normalized to -1..1 range. Mouse "axis" seems to be arbitrary in it's rande, depending on scaling factor. This breaks possible assumptions about axes in general. Should we limit it, add some mapping curcve, or just accept that as a fact of life? :P

minecrawler:
originally, I wanted it to be in range, however that would mean knowing the screen size in the input system. How would you pass it there - especially when there is no need for a mouse to begin with (mobile, consoles,...).

Limiting it would be crass and I feel like it would feel rather weird and mean workarounds (like decreasing sensitivity, only to multiply the mouse axis value with something to get a higher value again... 

One more concern is that the input is jumpy when I leave the window and re-enter it from someplace else, however imho, this should be handled by the user depending on their requirements.

I am open for suggestions, though!

OvermindDL1:
I always mapped the mouse input to an axis in my engine.  Where a gamepad/joystick/otherAbsoluteAxisDevices mapped from -1 to 1, the mouse 'normalized' at -1 to 1, but could exceed that.  Essentially the mouse scaling factor was applied to that normalized value, thus I could still treat the axis as -1 to 1 anywhere in the engine, but if I wanted it specifically limited (rudder, thrusters, etc...) then those already had bounds anyway that it couldn't exceed, but player movement was unbounded, multiplying the mouse movement by it.  In addition mouse events came in as `Immediate`, meaning the entire motion should be applied at once (I.E. it wouldn't exist again next frame if unchanged), and an absolute axis came in as `NonImmediate`, meaning it read the same value every frame if the motion by the player remained unchanged.  This was a simple 0.0 or 1.0 only, which allowed for multiplying it into the calculations with no branching needed to perform an operation or not.

Also remember, I mapped buttons, keys, everything to such 'axis', so even pressing the 'a' button on the keyboard mapped between 0.0 and 1.0 (and you could map the Immediate or NonImmediate versions of it individually), combined with filters on the event bindings you could do things like not have it be sent if, say, 'b' was pressed at the same time as a simple example.

I still hold firm that my style of mapping **every** input to a float normalized _around_ -1.0 to 1.0 with 0.0 being 'off' was a fantastic choice.  The filters and mappings in the path system made building actionmaps trivial, including inverting (just a filter/mapper), sensitivity (also a filter/mapper), etc... etc..., thus allowing every binding to be different in any way needed.

minecrawler:
@OvermindDL1 while I appreciate your insights, I did not plan to redesign the input system with this PR 😉 Maybe you should formulate a new design or your requirements as RFC or issue, so we can work on it in a proper way!

From what I understand though, the relevant parts of your critique boil down to:

- normalize the mouse axis to [-1..1]
- remove sensitivity and inversion, as they should be done in a later step (by a filter; by the user?)
- also provide the raw data

What I don't understand is how I should normalize the input. At the moment, all changes are immediate, and nothing relative to the amount of mouse movement. I could cap the delta and send out the remainder in the next frame (adding new movement to it), but I fear that might lead to sluggish input. Imagine camera movement via mouse in a FPS, and you quickly turn around, but when you stop, the camera moves on... 

or just drop the rest, which would cap the speed, though I am unsure if that would feel good in the same situation as describe before... What if a mouse is set to a high sensitivity and sends out huge values all the time, but we would always cap that at 1? Mouse input would feel binary: all, or nothing. I'd like to reduce that effect, if possible. I guess modern gaming mice are quite a handful...

OvermindDL1:
> remove sensitivity and inversion, as they should be done in a later step (by a filter; by the user?)

Big thing, different inputs will want different sensitivities, even different bindings with different things, like the different between mouse controlling a player or a tank or a ship all in the same game.

> At the moment, all changes are immediate, and nothing relative to the amount of mouse movement. 

That's the usual way, I never set up a NonImmediate form for the mouse.

> I could cap the delta and send out the remainder in the next frame (adding new movement to it), but I fear that might lead to sluggish input.

That's smoothing, a lot of games do it, and yes it makes the inputs feel horribly laggy, hence why I've never been a fan of it and never put it in my engine, but this would be how to do it (which would be done via a filter/mapper in my engine if I ever did).

> or just drop the rest, which would cap the speed, though I am unsure if that would feel good in the same situation as describe before... 

That's turning a relative motion axis (mouse) to be mapping to an absolute motion axis (like a rudder), which for things like player movement makes it feel horribly unnatural, this is what often happens when someone uses a mouse on a game designed for gamepads without the programmers programming well.

This is why I had my system work as I did, it supported all the forms, in whatever way was appropriate for the consumer of the data.

minecrawler:
> This is why I had my system work as I did, it supported all the forms, in whatever way was appropriate for the consumer of the data.

We don't have these concepts (yet), and I am not sure for what kind of input they might be useful for outside of mouse input and when we want to implement them. Maybe a racing wheel or a dj turn-table? Would be cool to get more gadgets supported!

What do you think about making this configurable? So I'd change the mouse struct to look like this:

```rust
    Mouse {
        /// The axis being bound
        axis: MouseAxis,
        /// Can be one of Raw, NormalizeDrop or NormalizeBuffer
        mode: MouseMode,
    },
```

while it is a lot simpler than your system, it still kinda fits with the current input system while making the above three ways of handling the mouse possible.

jojolepro:
Also Raw kind of implies that this is the device event (and not the delta pixel on the window). That would be confusing with winit's actual DeviceEvent.

minecrawler:
There is one issue I still have to investigate, but when I use the input in Pong, it is inverted. Am I missing anything? For example, using the Y axis and BufferDrop mode on the right axis, and move the mouse upward, the paddle moves downward....

amethyst_input/src/input_handler.rs
```diff
@@ -389,6 +393,43 @@ impl<T: BindingTypes> InputHandler<T> {
     pub fn send_frame_begin(&mut self) {
         self.mouse_wheel_vertical = 0.0;
         self.mouse_wheel_horizontal = 0.0;
+
+        // Prepare the mouse axis input
+        {
+            let hits_boundaries = |delta: f64| delta >= 1. || delta <= -1.;
```
skreborn: Why does this check for numbers greater than `1` and less than `-1`? It's only used on normalized values, so it's guaranteed to be in the range of `]-1; 1[`, isn't it?

minecrawler: As the name implies, it looks for values on the boundary, so if the delta is `1.`, it returns `true`

minecrawler: As the name implies, it looks for values on the boundary, so if the delta is `1.`, it returns `true`

OvermindDL1:
> Also Raw kind of implies that this is the device event (and not the delta pixel on the window). That would be confusing with winit's actual DeviceEvent.

Seems fine to me to use some wrapper data, it wasn't just a number on my side but another eventmap that gave information about the raw values, the type of controller, the ID, etc... etc...

I'm unsure if it should be winit's itself, it's possible someone wants to use another library to inject inputs into the actionmap system as well, like say from some bluetooth thing.

amethyst_input/src/input_handler.rs
skreborn: Yes, of course, I understand that. I'm merely asking why it checks for values _outside_ those limits. If the input values range only from `-1` to `1` inclusive, you could use simple equality instead.

```rust
let hits_boundaries = |delta: f64| delta == 1. || delta == -1.;
```

Or even using `abs` if you feel like it, although I suppose it's not worth the overhead to save 10 or so characters.

```rust
let hits_boundaries = |delta: f64| delta.abs() == 1.;
```

minecrawler: > why it checks for values outside those limits

just a tick of mine. All these lines of code are equal, so if there is a performance difference, or a problem from a DX perspective, I am the last person to not do the switch to a better alternative 😉

skreborn: Well, I don't think it matters, really. The CPU will happily do the comparison all the same, so if you're happy with it as it is, feel free to leave it like that.

amethyst_input/src/input_handler.rs
```diff
@@ -389,6 +393,43 @@ impl<T: BindingTypes> InputHandler<T> {
...
+            let normalize = |x: f64| {
```
Xaeroxe: Let's make this a private function rather than an in-line variable since we're using it multiple times.

amethyst_input/src/axis.rs
```diff
@@ -1,6 +1,9 @@
 use serde::{Deserialize, Serialize};
 
-use super::{Button, ControllerAxis};
+use super::{
+    mouse::{MouseAxis, MouseMode},
+    Button, ControllerAxis,
+};
 
 /// Represents any input represented by a float value from -1 to 1.
```
Xaeroxe: If we're going to introduce axes that break this rule, then this documentation needs to be updated.

amethyst_input/src/mouse.rs
```diff
@@ -0,0 +1,21 @@
...
+
+/// Mode of how mouse movement is presented to user code
+#[derive(Eq, PartialEq, Debug, Copy, Clone, Serialize, Deserialize)]
+pub enum MouseMode {
+    /// Pass the pixel distance in a normalized way, but buffer additional movements on top [-1..1]
+    NormalizeBuffer,
+    /// Pass the pixel distance in a normalized way and drop any overflow [-1..1]
+    NormalizeDrop,
+    /// Just pass the pixel distance travelled over to the code directly (??..??)
+    PassThrough,
```
Xaeroxe: Help me understand the use case for this, and why a user shouldn't just use [`mouse_position`](https://docs.amethyst.rs/master/amethyst_input/struct.InputHandler.html#method.mouse_position).

As is this complicates the specification for an axis, and I'd rather not do that if this functionality is redundant.

amethyst_input/src/input_handler.rs
```diff
@@ -389,6 +393,43 @@ impl<T: BindingTypes> InputHandler<T> {
...
+            let last_pos = self.mouse_last_position.unwrap_or((0., 0.));
+            let raw_delta_x = current_pos.0 - last_pos.0;
+            let raw_delta_y = current_pos.1 - last_pos.1;
+            let delta_x = normalize(raw_delta_x);
+            let delta_y = normalize(raw_delta_y);
+
+            // In case the deltas hit the boundaries, not all data could be handled in the last frame,
+            // so the buffer should be updated with the overflow
+            if hits_boundaries(delta_x) {
+                self.mouse_axis_buffer.0 = raw_delta_x - delta_x;
```
Xaeroxe: If we move say, 5 pixels in a frame, wouldn't this code as is only get 2 of those 5 pixels?

Once for the first frame, again for the second frame, and then the buffer gets set to zero.

Xaeroxe: The other problem with this is that even if we correct this code we're reporting several frames of maximum axis values if the user moves, for example,15 pixels in a frame.

So what this would look like is

1. User moves mouse kind of quick
2. at 60 FPS, this axis is maxed out for a full quarter of a second.

minecrawler:
@Xaeroxe Ok, let's start by explaining why I think this feature is useful: Because I can rebind any axes to the mouse. Simple as that. In my fork, I could just go ahead and bind the mouse to the right paddle and use UP/DOWN keys for the left one. I could also use a gamepad or anything with an axis, which imho is really nice. With this PR, that is possible without a user writing lots of non-business logic. A game engine, imho, should take that away.

amethyst_input/src/input_handler.rs
minecrawler: > If we move say, 5 pixels in a frame, wouldn't this code as is only get 2 of those 5 pixels?

It should not, as we only subtract a maximum of one pixel per frame from the buffer.

> we're reporting several frames of maximum axis values if the user moves

exactly, and I was critical of it, too, however Overmind was of the opinion that it might come in handy - if I understood him correctly. I never really had to use buffered mouse input though...

Xaeroxe:
I think this might be reaching into user code a bit too much. Here's an example scenario to make my case. 

I'm making a first person videogame and want to rotate the camera either via a controller axis, or a mouse axis. In this case I should report raw mouse data and tune that according to my own mouse sensitivity value. For this person I'd recommend paying attention to [this input event](https://docs.amethyst.rs/master/amethyst_input/enum.InputEvent.html#variant.MouseMoved). In most games you want the maximum mouse speed to be quite a bit faster than the maximum controller turn speed, as the maximum controller speed is much easier to achieve. So should the controller never be able to achieve maximum axis value? That seems like a strange requirement to enforce on all users just because of this use case, and even if we did want to do that what would the controller maximum be? Should that be configurable?

The other issue with this is, as you asked before, if we're going to be normalizing the mouse data, as an axis expects, what should we be normalizing against? Probably some kind of arbitrary circle, but what should the radius on that circle be? What happens if we exceed that radius? These are some kind of complicated questions whose answers will have a profound impact on how it feels to play the game in question. When I run into these kinds of problems is usually the point at which I decide the input system ends, and the user's gameplay code begins. I merely need to feed them the data to implement what they want.

A non-normalized axis also puts the end user in a kind of weird spot. Should I design my input code around an axis that will only ever be between 1 and -1, or should I design it around axes that might be 15-30x that?

Finally, using a single unit as the maximum axis value, and then just "spilling out" the remainder over several frames will definitely make the input feel laggy, and if the game designer wants that they should implement it themselves.

Xaeroxe:
Oh and @minecrawler another note for your benefit, typically when people talk about buffered input they usually mean "end user receives input in the form of events" as opposed to "end user queries the current input state every frame".

Buffered input:
  Pros: Guarantees that every single button press and release, every axis movement, all of it, will absolutely 100% be received
  Cons: Not as clean to receive in the user code, and requires a bit of input state tracking in the user code.

Unbuffered input:
  Pros: Cleanly receive the current input state every frame, and use it as part of basic simple math calculations that happen every frame.
  Cons: Produces an unreliable stream of input data, making it unsuitable for inputs that may be triggered and released in the span of a single frame.

So the Amethyst input handler provides both. You don't need to worry about buffered input for the mouse, code for that is already written. Axes are specifically for unbuffered input.

minecrawler:
Thank for all the information! I already discussed some of the points you mention above.

- Normalized input is difficult, because what do we normalize against?
- Pass-through is bad, because it does not really fit the axis concept of Amethyst...

However, how would you do the simple re-binding then, which still sounds like a good thing to me? Alternatively, from your experience, is it not required often enough for it to be part of the engine?

Xaeroxe:
> However, how would you do the simple re-binding then

It's not that this isn't a common use case, because it is, however I don't see it as a common use case that we can provide a valuable abstraction for. In implementing such an abstraction we'd need to do one or more of the following.

1. Be strongly opinionated on input sensitivity, possibly in ways that would hamper the ability to tune the input response for each input type.
2. Expose a lot of configuration variables that exist for contrived reasons that may not be applicable to many users, and explain how to use them.
3. Weaken or completely break our guarantees for normalized output, meaning users will just need special case handling in their own code once again anyways.

So, I appreciate the intent but honestly providing this axis is by itself worthy of an RFC as there's a lot of angles we have to account for.

The simple approach is to just not provide it. However, in so doing we need to expose the raw mouse data so that game designers can still implement what they need to. I believe I've done that, in the form of the event stream.

Xaeroxe:
Of course as soon as I finish writing that a solution occurs to me that handles all cases of this scenario pretty well.

1. Eliminate the mouse mode option.
2. Add a radius variable to the mouse axis, and divide all input by that radius.
3. Add a boolean to the mouse axis called "over_extendable" that, if true, will permit the axis to return values outside 1 and -1.
4. Indicate in the axis documentation that being between 1 and -1 is more of a guideline than a guarantee.

This allows the mouse sensitivity to be configured independent of the controller input, by tightening or expanding the circle, and it allows us to choose between the two answers for "what happens when the mouse goes beyond that radius".

minecrawler:
just saw this going stale. I tested Xaeroxe's solution with my project over the past few weeks with different radius values and I am not entirely sure if that works out. Sometimes it feels a bit weird, so I added a `radius_x` and a `radius_y`, which gets rid of the weird feeling.

Surprisingly, there is a lot of optimizing which can be done with these numbers. For Pong, for example, a smaller radius results in faster movement, but also caps the movement at a certain speed. This can be used for game mechanics, and I wondered if these configs should always be available to players, as different values can dramatically change the feeling of this simple game (and allows to cheat by hindering the second player 🤣). For my FP camera movement, however, bigger values felt more natural, as well as using a slightly smaller X than Y value (maybe counter-intuitive). I also found out that if I use smaller values there and remove the cap, I can code my own slow-down easily by buffering the surplus myself, which is also a nice benefit.

All in all, it's a lot of fun to play around with and enables fairly interesting opportunities for mouse control. I will upload my code this weekend, after I rebased it onto amethyst/master.

Should I also add the mouse control to the Pong example or create a new example for this?

amethyst_input/src/input_handler.rs
```diff
@@ -552,6 +554,37 @@ where
...
+                let current_pos = self.mouse_position.unwrap_or((0., 0.));
+                let last_pos = self.mouse_last_position.unwrap_or(current_pos);
+                let delta = match axis {
+                    // These calculations have to be inverses in order to point into the right direction of movement
+                    MouseAxis::X => last_pos.0 - current_pos.0,
+                    MouseAxis::Y => last_pos.1 - current_pos.1,
+                };
+
+                if over_extendable {
+                    return delta;
```
Xaeroxe: I was actually thinking over extendable mouse axes would also be relative, they just wouldn't be clamped between 1 and -1. So after computing rel_delta, I would just return that if over_extendable. What are your thoughts on that approach?

Xaeroxe:
> I wondered if these configs should always be available to players, as different values can dramatically change the feeling of this simple game (and allows to cheat by hindering the second player 🤣). 

This is a good point. Since the developer of the game can control for this by performing some checks on the input config after loading it, I'd actually recommend we go that route. We should update the documentation to alert developers we expect them to validate their input configs are within what they consider an acceptable range.

distransient:
Hey, what's the status of this PR? You've all seemed to evade stalebot :P

minecrawler:
oops, sorry about that.
@distransient I tried to incorporate the changes into the code and rebase onto staging, however I fail to build the glyph_brush dependency. Any tips on working around that?

distransient:
Post your error? You can also chat with me on Discord in #engine-general

minecrawler:
wasn't my day yesterday. So I cargo updated and built it. It's looking good so far. Any more comments?

distransient:
I'd like to ping/give @Xaeroxe a chance to review

distransient:
bors try

distransient:
bors r=distransient,jojolepro

Xaeroxe:
bors r=distransient,jojolepro,xaeroxe

PR #1708, opened 2019-06-13
Allow arc camera example to exit with escape key

piedoom:
## Description

This allows the arc ball camera example to exit when the escape key is pressed. This is useful as the example captures the cursor.


jaynus:
bors r=torkleyy,Moxinilian

PR #1713, opened 2019-06-14
Add more profile statements

Frizi:
For the sake of debugging macos performance, I've added some more profile statements around the code, mainly for frame limiter and all the systems. It seems like a good idea to have that preserved.


Frizi:
bors r=torkleyy,jaynus

jaynus:
bors r=jaynus,torkleyy

PR #1714, opened 2019-06-15
Minor improvement to description of 2d camera position

ohnoimdead:
## Description

Just fixing up a somewhat messy description of the default camera's placement in the pong tut

## Modifications

- Changed one paragraph in the pong tutorial step 2

## PR Checklist

- :white_check_mark: Ran `cargo test --all` locally if this modified any rs files.
- N/A Ran `cargo +stable fmt --all` locally if this modified any rs files.
- :white_check_mark: Updated the content of the book if this PR would make the book outdated.
- N/A Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.
- :white_check_mark: Acknowledged that by making this pull request I release this code under an MIT/Apache 2.0 dual licensing scheme.

### DIFF
--- a/book/src/pong-tutorial/pong-tutorial-02.md
+++ b/book/src/pong-tutorial/pong-tutorial-02.md
@@ -117,15 +117,18 @@ will.
     projection of the size of our arena. We also attach a `Transform` component,
     representing its position in the world.
 
-    A default camera can be created with `standard_2d` which size of our arena.
-    This camera makes the X go right, Y go up and center of the screen being at
-    the position of the camera in space, i.e. whatever we pass to `set_translation_xyz`.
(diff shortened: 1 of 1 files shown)


Frizi:
The new description is not correct. The `standard_2d` method creates camera with center at (0, 0), ranging from (-w/2, -h/2) to (w/2, h/2). You have to translate it with transform entity to make 0 be at the corner of screen.

ohnoimdead:
Hm, I see your point. In this case the tutorial sets the camera entity's transform to w/2, h/2 so the description isn't wrong, but it is a bit misleading. Let me see if I can make that more clear.

ohnoimdead:
@Frizi Let me know if that makes more sense.

PR #1720, opened 2019-06-15
Fix stackoverflow when serializing dyn Format

### LINKED ISSUE #1636: [BUG] Serializing Box<dyn Format<_>> causes a stack overflow
## Description

When serializing the type Box<dyn Format<_>> an infinite loop occurs at https://github.com/amethyst/amethyst/blob/master/amethyst_assets/src/dyn_format.rs#L142

I ran across this while trying to serialize TexturePrefab::File in an asset preparation tool.

## Reproduction Steps
Minimal example:
```
use amethyst::{
    assets::Format,
    renderer::{
        types::TextureData,
        formats::texture::{ImageFormat},
    },
};
fn main() {
    let texture: Box<ImageFormat> = Box::new(ImageFormat::default());
    println!("{}", ron::ser::to_string(&texture).expect("Failed to encode level file"));

    let texture: Box<dyn Format<TextureData>> = Box::new(ImageFormat::default());
    println!("{}", ron::ser::to_string(&texture).expect("Failed to encode level file"));
}
```

I expect it to print the text of two identical ron documents suitable for including in a prefab file.

## What Actually Happened
The first document prints but the second one triggers this:
```
thread 'main' has overflowed its stack
fatal runtime error: stack overflow
Aborted (core dumped)
```

Here is the actual cycle clipped from a gdb backtrace:
```
#27 0x00005555555dc4ff in <&mut ron::ser::Serializer as serde::ser::SerializeTuple>::serialize_element (
    self=0x7fffff800188, value=...)
    at /home/alec/.cargo/registry/src/github.com-1ecc6299db9ec823/ron-0.4.2/src/ser/mod.rs:480
#28 0x00005555555e13cf in <&mut ron::ser::Serializer as serde::ser::SerializeTupleStruct>::serialize_field (self=0x7fffff800188, value=...)
    at /home/alec/.cargo/registry/src/github.com-1ecc6299db9ec823/ron-0.4.2/src/ser/mod.rs:519
#29 0x00005555555da081 in amethyst_assets::dyn_format::<impl serde::ser::Serialize for dyn +amethyst_assets::asset::Format<D>>::serialize (self=..., serializer=0x7fffffffe330)
    at /home/alec/Code/Vendor/amethyst/amethyst_assets/src/dyn_format.rs:142
#30 0x00005555555dc4ff in <&mut ron::ser::Serializer as serde::ser::SerializeTuple>::serialize_element (
    self=0x7fffff800518, value=...)
    at /home/alec/.cargo/registry/src/github.com-1ecc6299db9ec823/ron-0.4.2/src/ser/mod.rs:480
#31 0x00005555555e13cf in <&mut ron::ser::Serializer as serde::ser::SerializeTupleStruct>::serialize_field (self=0x7fffff800518, value=...)
    at /home/alec/.cargo/registry/src/github.com-1ecc6299db9ec823/ron-0.4.2/src/ser/mod.rs:519
#32 0x00005555555da081 in amethyst_assets::dyn_format::<impl serde::ser::Serialize for dyn +amethyst_assets::asset::Format<D>>::serialize (self=..., serializer=0x7fffffffe330)
    at /home/alec/Code/Vendor/amethyst/amethyst_assets/src/dyn_format.rs:142
```

## Your Environment

1. Operating System
Linux
2. Operating System Version
Arch Linux 
3. Version of Amethyst
master (as of 5/23/2019)

## Additional Context

I don't see how to fix this without changing Format to force implementors to handle their own serialization. If that seems like a reasonable solution or someone has another suggestion I'm happy to write up the change. Changing Format feels pretty invasive though.

mvlabat:
## Description

Fixes #1636 

I spent whole day tinkering with it.. It seems like it really fixes the SO. ~~I'm sorry for this totally insane diff. I'll do my best to clean this up.~~ (I've thrown out the redundant typetag copy-paste and rebased the commits, so we don't keep that in the history). Meanwhile you're welcome to review this. Any, even the most obvious remarks, nitpicks are very appreciated, because I'm really a noob in the amethyst codebase, and I don't know how you guys feel about these changes. ~~Especially I'll need your advice on what we need to put instead of those `unimplemented` kludges.~~ (Got rid of them too.)

## Additions

- Introduces new `SerializableFormat` marker trait, which needs to be implemented for all the formats that are supposed to be serializable

## PR Checklist

### DIFF
--- a/amethyst_assets/src/asset.rs
+++ b/amethyst_assets/src/asset.rs
@@ -1,4 +1,4 @@
-use crate::{storage::ProcessingState, Handle, Reload, SingleFile, Source};
+use crate::{storage::ProcessingState, FormatRegisteredData, Handle, Reload, SingleFile, Source};
 use amethyst_core::ecs::storage::UnprotectedStorage;
 use amethyst_error::{Error, ResultExt};
 use std::{fmt::Debug, ops::Deref, sync::Arc};
@@ -51,7 +51,9 @@ impl<T: Asset<Data = T>> ProcessableAsset for T {
 /// The format type itself represents loading options, which are passed to `import`.
 /// E.g. for textures this would be stuff like mipmap levels and
 /// sampler info.
-pub trait Format<D: 'static>: objekt::Clone + Debug + Send + Sync + 'static {
+pub trait Format<D: FormatRegisteredData + 'static>:
+    objekt::Clone + Debug + Send + Sync + erased_serde::Serialize + 'static
+{
     /// A unique identifier for this format.
     fn name(&self) -> &'static str;
 
@@ -105,7 +107,7 @@ pub trait Format<D: 'static>: objekt::Clone + Debug + Send + Sync + 'static {
 objekt::clone_trait_object!(<D> Format<D>);
 
 // Allow using dynamic types on sites that accept format as generic.
-impl<D: 'static> Format<D> for Box<dyn Format<D>> {
+impl<D: FormatRegisteredData + 'static> Format<D> for Box<dyn Format<D>> {
     fn name(&self) -> &'static str {
         self.deref().name()
     }
--- a/amethyst_assets/src/asset.rs
+++ b/amethyst_assets/src/asset.rs
@@ -105,7 +105,10 @@ pub trait Format<D: 'static>: objekt::Clone + Debug + Send + Sync + 'static {
 objekt::clone_trait_object!(<D> Format<D>);
 
 /// SerializableFormat is a marker trait which is required for Format types that are supposed
-/// to be serialized.
+/// to be serialized. This trait implies both `Serialize` and `Deserialize` implementation.
+///
+/// **Note:** For deserialize support you need to register the format.
+/// See [FormatRegisteredData](trait.FormatRegisteredData.html) for full example.
 pub trait SerializableFormat<D: FormatRegisteredData + 'static>:
     Format<D> + erased_serde::Serialize + 'static
 {
--- a/amethyst_assets/src/asset.rs
+++ b/amethyst_assets/src/asset.rs
@@ -107,8 +107,9 @@ objekt::clone_trait_object!(<D> Format<D>);
 /// SerializableFormat is a marker trait which is required for Format types that are supposed
 /// to be serialized. This trait implies both `Serialize` and `Deserialize` implementation.
 ///
-/// **Note:** For deserialize support you need to register the format.
-/// See [FormatRegisteredData](trait.FormatRegisteredData.html) for full example.
+/// **Note:** This trait should never be implemented manually.
+/// Use the `register_format` macro to register it correctly.
+/// See [FormatRegisteredData](trait.FormatRegisteredData.html) for the full example.
 pub trait SerializableFormat<D: FormatRegisteredData + 'static>:
     Format<D> + erased_serde::Serialize + 'static
 {
--- a/amethyst_assets/src/asset.rs
+++ b/amethyst_assets/src/asset.rs
@@ -51,9 +51,7 @@ impl<T: Asset<Data = T>> ProcessableAsset for T {
 /// The format type itself represents loading options, which are passed to `import`.
 /// E.g. for textures this would be stuff like mipmap levels and
 /// sampler info.
-pub trait Format<D: FormatRegisteredData + 'static>:
-    objekt::Clone + Debug + Send + Sync + erased_serde::Serialize + 'static
-{
+pub trait Format<D: 'static>: objekt::Clone + Debug + Send + Sync + 'static {
     /// A unique identifier for this format.
     fn name(&self) -> &'static str;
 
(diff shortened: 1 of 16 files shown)


mvlabat:
Currently it only compiles, but doesn't even serialize it right :)

mvlabat:
@torkleyy, thank you! ❤️ 

1. I dunno.. `Deserialize` trait requires a lifetime, that would spread it to all the code where `SerializeFormat` is used. That'll be painful, I'm afraid :)
2. That's right, from now on those it'll be required to implement `Serialize` for those types. Do we have such cases, when we want to have only `Deserialize` without `Serialize` though?

torkleyy:
> 1. I dunno.. `Deserialize` trait requires a lifetime, that would spread it to all the code where `SerializeFormat` is used. That'll be painful, I'm afraid :)

Ah, you can simply use `DeserializeOwned` here. We don't need zero-copy parsing.

> 2. That's right, from now on it'll be required to implement `Serialize` for those types. Do we have such cases, when we want to have only `Deserialize` without `Serialize` though?

~~Yes, we do, notably all cases where `amethyst::assets::Handle` is used, and the deserialization works by loading the asset.~~

Actually, we don't need to support this. Formats which use `Handle` cannot be deserialized anyway (we use prefabs for those cases), so we can safely ignore this.

amethyst_assets/src/asset.rs
```diff
@@ -104,6 +104,16 @@ pub trait Format<D: 'static>: objekt::Clone + Debug + Send + Sync + 'static {
 
 objekt::clone_trait_object!(<D> Format<D>);
 
+/// SerializableFormat is a marker trait which is required for Format types that are supposed
+/// to be serialized.
```
torkleyy: Please document that this trait indicates both `Deserialize` and `Serialize` support, and that it should be implemented via `register_format!`.
If you want to add one, a doc example would be great here, too.

amethyst_assets/src/asset.rs
```diff
@@ -123,6 +133,26 @@ impl<D: 'static> Format<D> for Box<dyn Format<D>> {
     }
 }
 
+impl<D: 'static> Format<D> for Box<dyn SerializableFormat<D>> {
```
torkleyy: Do we need such an impl for `dyn SerializableFormat<D>`, too?

amethyst_assets/src/formats.rs
```diff
@@ -32,6 +32,11 @@ where
     }
 }
 
+impl<D: FormatRegisteredData> SerializableFormat<D> for RonFormat where
```
torkleyy: This seems to be missing the registration macro, which reminds me of: Please add tests checking that `RonFormat` and `JsonFormat` are properly registered.

mvlabat:
@torkleyy 
> Ah, you can simply use `DeserializeOwned` here
That still implies `Deserialize` implementation:
```
pub trait DeserializeOwned: for<'de> Deserialize<'de> {}
impl<T> DeserializeOwned for T where T: for<'de> Deserialize<'de> {}
```
And as it has generic methods, I then get the following compilation errors:
```
   --> amethyst_assets\src\asset.rs:136:18
    |
136 | impl<D: 'static> Format<D> for Box<dyn SerializableFormat<D>> {
    |                  ^^^^^^^^^ the trait `asset::SerializableFormat` cannot be made into an object
    |
    = note: the trait cannot require that `Self : Sized`
```

amethyst_assets/src/asset.rs
mvlabat: Yeah... I think, we still need to use `Box<dyn SerializableFormat<D>>` as `Format`. Otherwise, without this implementation, we get
```
error[E0277]: the trait bound `std::boxed::Box<(dyn asset::SerializableFormat<D> + 'static)>: asset::Format<D>` is not satisfied
   --> amethyst_assets\src\asset.rs:136:41
    |
136 | impl<D: FormatRegisteredData + 'static> SerializableFormat<D> for Box<dyn SerializableFormat<D>> {}
    |                                         ^^^^^^^^^^^^^^^^^^^^^ the trait `asset::Format<D>` is not implemented for `std::boxed::Box<(dyn asset::SerializableFormat<D> + 'static)>`
    |
    = help: the following implementations were found:
              <std::boxed::Box<(dyn asset::Format<D> + 'static)> as asset::Format<D>>
```
I'm not sure there's any other way to do it. Please correct me if I'm wrong

amethyst_assets/src/formats.rs
mvlabat: @torkleyy 
I'm not sure we can register format for `RonFormat` and `JsonFormat`, as their underlying data parameter (`FormatRegisteredData`) is generic.
I think it'll be possible to register `RonFormat` and `JsonFormat` only if we introduce a new format type for them. (Am I right?..) And with this change we can also introduce a new kind of asset, though maybe it goes out of scope of this PR...

For now I've just dropped `SerializableFormat` implementation for both of them, as far as I notice we don't even use it anywhere for these types. I believe we can bring it back once we decide to introduce something like a `ConfigFile` asset type and the corresponding data type.

Let me know if that sounds reasonable.

Frizi: Generic formats weren't really meant for registration. Every type has it's own storage, and here the type is abstract. Rust prevents generic types from having a static member, so in current form it's impossible.

amethyst_audio/src/formats.rs
```diff
@@ -37,6 +39,8 @@ impl Format<AudioData> for OggFormat {
     }
 }
 
+impl SerializableFormat<AudioData> for OggFormat {}
```
Frizi: I think that all `impl SerializableFormat` should be done as part of `amethyst_assets::register_format!` macro.

mvlabat: @Frizi done!

torkleyy:
Yay! :shipit: 
bors r+

torkleyy:
Thanks for the rebase!
bors r+

PR #1587, opened 2019-05-09
WIP: Adapt Jenkinsfile to run mdbook tests

marot:
## Description

This fixes #1502. I based the changes on https://github.com/amethyst/amethyst/commit/0e5b769553b3d7f67c37900a8450173760852887

I also fixed the book examples so that the CI would be successful.

## PR Checklist

By placing an x in the boxes I certify that I have:

- [ ] Ran `cargo test --all` locally if this modified any rs files.
- [ ] Ran `cargo +stable fmt --all` locally if this modified any rs files.
- [ ] Updated the content of the book if this PR would make the book outdated.
- [ ] Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.
- [ ] Added unit tests for new APIs if any were added in this PR.
- [ ] Acknowledged that by making this pull request I release this code under an MIT/Apache 2.0 dual licensing scheme.

PR #1739, opened 2019-06-19
Ci test book

### LINKED ISSUE #1502: [MAINTENANCE] Update CI to test book
## Description

Concourse CI isn't testing the book ([concourse task](https://github.com/amethyst/amethyst/blob/master/.concourse/tasks/linux-test.yml)). This [used to be done](https://github.com/amethyst/amethyst/blob/master/.travis.yml#L70) in Travis CI.

## Reason

This keeps the book up to date as the code evolves.

## Impact

Would increase build times slightly.

azriel91:
## Description

Fixes #1502 (test book in CI), supersedes #1587.

## PR Checklist

- **n/a** Ran `cargo test --all` locally if this modified any rs files.
- **n/a** Ran `cargo +stable fmt --all` locally if this modified any rs files.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -92,6 +92,8 @@ pipeline {
                         }
                     }
                     steps {
+			echo 'Build all the examples in the book'
+			sh 'mdbook test book -L target/debug/deps'
                         echo 'Beginning tests...'
                         sh 'cargo test --all --features "vulkan sdl_controller json saveload"'
                         echo 'Tests done!'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -92,6 +92,7 @@ pipeline {
                         }
                     }
                     steps {
+		        sh 'cargo build'
 			echo 'Build all the examples in the book'
 			sh 'mdbook test book -L target/debug/deps'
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -92,18 +92,23 @@ pipeline {
                         }
                     }
                     steps {
+                        echo 'Cleaning up old build artifacts'
+                        sh 'rm -rf ./target/debug/deps/libamethyst* || true'
```
torkleyy: This also doesn't work reliably, because as soon as we have a new patch version in any of our dependencies, `mdbook` will fail again.

azriel91: yeap, just realized; I'm crafting a bash script, which isn't great, but I think it's worth testing the book

azriel91:
Crates with multiple lib artifacts in my local machine (`echo`ed):

```bash
$ ./scripts/remove_duplicate_libs.sh  | wc -l
323

# Without `uniq` in the script:
$ ./scripts/remove_duplicate_libs.sh  | wc -l
3143
```

azriel91:
ah, still get `multiple matching crates`, okay I'm convinced.

PR #1753, opened 2019-06-22
Add load_from_data_async to Asset Loader

Friz64:
## Description

This PR enables users to asynchronously compute an asset by passing a loading function that gets called inside the asset loader threadpool.

## Additions

- Add `load_from_data_async` to `AssetLoaderSystemData`
- Add `load_from_data_async` to `Loader`

## PR Checklist

(This took too much ram and crashed for me)


torkleyy:
> (This took too much ram and crashed for me)

Wow, never heard that before. Doesn't rustc track RAM usage?

azriel91:
> > (This took too much ram and crashed for me)
> 
> Wow, never heard that before. Doesn't rustc track RAM usage?

It doesn't track RAM usage sadly -- my laptop with 8 GB RAM sometimes needs `-j 4` when compiling to not fail with out of memory.

Friz64:
I have 8 GB of RAM + 4 GB Swap and 12 threads and i think its because the build invokes too many gcc processes.

Friz64:
I've been wanting to start work on the documentation and tests but am unsure about the process because `load_from_data` (non-async) doesn't have any more documentation or tests than my async function.

PR #1754, opened 2019-06-22
Enable activation of spirv-reflection feature

Friz64:
## Description

This PR allows adding spirv-reflection support using a feature in the amethyst crate

## Additions

- `spirv-reflection` feature in the main and amethyst_rendy crate

## PR Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,7 +94,7 @@ test-support = [
   "amethyst_rendy/test-support",
   "amethyst_window/test-support",
 ]
-spirv-reflection = ["amethyst_rendy/spirv-reflection"]
+experimental-spirv-reflection = ["amethyst_rendy/experimental-spirv-reflection"]
 
 [workspace]
 members = [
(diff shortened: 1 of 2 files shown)


jaynus:
I'm not sure how I feel about this - currently, the spirv-reflection feature set in rendy will pull in spirv-reflect, which pulls is and compiles a bunch of external spirv dependencies.

Some of these dependencies are shared with shader-compiler. Additionally, the spirv-reflect feature is currently mostly useless unless you compile SPIRV images WITH debug symbols on, which requires 'shader-compiler' feature enabled. 

I think this should be wrapped up as an additional-default feature with 'shader-compiler'.

Friz64:
Well, this is still useful when providing precompiled SPIRV images (which in my opinion is superior) instead of using the bundled compiler.

jaynus:
@Friz64 if you can change it to experimental like Frizi requests we will get it merged.

jaynus:
bors r=jaynus,Frizi

PR #1756, opened 2019-06-23
Add Tint support for sprites in 2d

### LINKED ISSUE #1751: [NEW FEATURE] Tint support on 2D passes
`Tint` component should be respected in flat render passes, applied as color multiplication to the sprites.

This feature was unintentionally omitted when porting to new renderer, so it's a regression really.

ohnoimdead:
## Description

Adds Tint component support for sprites.

Fixes #1751 

## Modifications

- Change `SpriteArgs` to include a vec4 `tint` field
- Change the sprite shader to accept a vec4 color argument

## PR Checklist


ohnoimdead:
fixes https://github.com/amethyst/amethyst/issues/1751

ohnoimdead:
sprite_tint

Note about the rendy example: Tint does not implement PrefabData so I manually added some sprite entities to the rendy example to demonstrate tinting, which is why the colorized crates render at the same time as the spheres (and before everything else from the prefab scene).

torkleyy:
It would be great if we had some kind of documentation which components are picked up by which pass.

ohnoimdead:
> It would be great if we had some kind of documentation which components are picked up by which pass.

I was thinking that too. Is that something that should go in the book? I could add doc comments to the Tint struct, but I'm not sure that's the best place. If you point me in the right direction I can add something to this PR.

torkleyy:
The book might be one place, but I was thinking we could add them to the API docs of the specific pass struct (e.g. `DrawFlat2DDesc`).

azriel91:
bors r=torkleyy,azriel91

Shall get this in and let docs come in later in https://github.com/amethyst/amethyst/issues/1759 (don't want to force many rebase/merges).

azriel91:
bors retry

```
No space left on device (os error 28)
```

PR #1763, opened 2019-06-26
Added error checking support to the example prefabs.

vultix:
## Description
Using the [IntelliJ RON Plugin](https://vultix.github.io/intellij-ron-plugin/) IntelliJ now supports type-checking, completion, and navigation for RON files.

This pull request adds this support to all amethyst example prefabs.

## Modifications
Changed how the Transform component is deserialized and serialized.  Previously there was a custom implementation of Deserialize and Serialize, but now we are using serde(from="TransformValues").

This both simplifies the code and allows completion in RON files for the Transform component.

## PR Checklist


amethyst_core/src/transform/components/transform.rs
```diff
@@ -643,144 +639,59 @@ impl From<Vector3<f64>> for Transform {
...
-
-        Serialize::serialize(
-            &TransformValues {
-                translation: [pos[0], pos[1], pos[2]],
-                rotation: [rot[0], rot[1], rot[2], rot[3]],
-                scale: [scale[0], scale[1], scale[2]],
-            },
-            serializer,
-        )
+impl Into<TransformValues> for Transform {
```
azriel91: a little pedantic, but if the following is implemented instead, we get the `Into<TransformValues> for Transform` for free from std's blanket impl.

```rust
impl From<Transform> for TransformValues {
     fn from(transform: Transform) -> TransformValues {
        TransformValues {
            translation: transform.isometry.translation.vector.into(),
            rotation: transform.isometry.rotation.as_ref().coords.into(),
            scale: transform.scale.into(),
        }
    }
}
```

The main advantage is, when `Into` is implemented, then this happens:

```rust
let transform_values = transform.into(); // ok!
let transform_values = TransformValues::from(transform); // fails
```

whereas if you implement `From`, then both work. Normally you'd only implement `Into` if the target type is not under your control.

examples/arc_ball_camera/resources/display_config.ron
```diff
@@ -1,3 +1,8 @@
+/*!
+    @import /amethyst_window/src/config.rs#DisplayConfig
+    DisplayConfig
+*/
+
 (
-  title: "Arc Ball Camera Example",
```
azriel91: btw, is removing trailing commas the default setting for the RON formatter?
I might be caring too much, but `rustfmt` has trailing commas if the type's parameters are over multiple lines
(feel free to ignore this comment, I'll conveniently forget about it next time)

YouGo32: Yes, it is the default setting for now.  Thank you for the heads up about rustfmt, I'll update the plugin to have similar functionality (configurably) shortly.

vultix: Yes, it is the default setting for now. Thank you for the heads up about rustfmt, I'll update the plugin to have similar functionality (in a configurable way) shortly.

azriel91:
bors r=torkleyy,azriel91

azriel91:
ah, fmt:
```patch
Diff in /home/jenkins/workspace/amethyst_staging/amethyst_core/src/transform/components/transform.rs at line 690:
         TransformValues {
             translation: transform.isometry.translation.vector.into(),
             rotation: transform.isometry.rotation.as_ref().coords.into(),
-            scale: transform.scale.into()
+            scale: transform.scale.into(),
         }
     }
 }
```

vultix:
@azriel91 That's my bad, it's fixed now.

azriel91:
bors r=torkleyy,azriel91

all good!

amethyst_ui/src/prefab.rs
```diff
@@ -68,6 +68,7 @@ pub struct UiTransformBuilder<G> {
     /// UI element with the next highest tab order, or if another element with the same tab_order
     /// as this one exists they are ordered according to Entity creation order.  Shift-tab walks
     /// this ordering backwards.
+    pub tab_order: u32,
```
torkleyy: I'm a bit confused by this field. Why was this missing before? And, do we even use it anywhere?

Anyway, the build failed because `selectable` below does not have any API docs.

vultix: This is a change that I didn't mean to commit, my apologies for the extra work.  

The UI prefabs found in `examples/assets/ui` have tab_order fields all over the place.

I was originally going to add the missing fields, as in the change above, but decided to revert that change and keep the scope of this commit to the RON files.

jojolepro: IIRC tab_order was renamed to selectable or something similar.

jojolepro:
There's an unused macro import causing a warning which causes ci to fail.

vultix:
@jojolepro I've removed the unused import.

azriel91:
bors r=torkleyy,azriel91,jojolepro

PR #1769, opened 2019-06-28
Create FUNDING.yml

erlend-sh:
See https://help.github.com/en/articles/displaying-a-sponsor-button-in-your-repository

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,4 @@
+# These are supported funding model platforms
+
+custom: ['https://amethyst.rs/donate']# Replace with up to 4 custom sponsorship URLs e.g., ['link1', 'link2']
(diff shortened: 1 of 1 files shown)


.github/FUNDING.yml
```diff
@@ -0,0 +1,4 @@
+# These are supported funding model platforms
+
+custom: ['https://amethyst.rs/donate']# Replace with up to 4 custom sponsorship URLs e.g., ['link1', 'link2']
```
azriel91: Please could we use either the vertical list format, or at least introduce a space after `]` 👀?
The YAML's valid and does what's intended, but the comment `#` being right next to the `]` made me wonder if it was.

erlend-sh: How about we just remove comments? Updated the PR

azriel91:
bors r=torkleyy,azriel91

PR #1772, opened 2019-06-28
More ergonomic RenderingBundle

### LINKED ISSUE #1746: [FEATURE MODIFICATION] Add ergonomic API layer for rendering system
Currently we expose Rendy's rendergraph api directly, requiring every game to implement it's own rendering pipeline from scratch (only premade passes are provided). This is obviously bad, as it exposes way too much details that most users don't expect to deal with. Additionaly, current API provides challanges in terms of ECS initialization, as passes have no way to provide appropriate system setup.

To solve that, amethyst needs to build it's own API layer encapsulating rendy's graphs.

Problems to solve:
- allow easy "default" graphs setup that allows extension without copying code from the core
- provide a way for render groups to setup the necessary resources
- hide window handling as much as possible, while allowing multiple windows to exist.

Frizi:
Fixes #1746

This adds a `RenderingBundle` with default `GraphCreator` which imlpements a plugin system for render graph creation. ~This is not yet fully complete, as it completely misses the buffers and compute nodes for now. It also needs tests. Still, early feedback very appreciated.~

Should be now complete enough to use :)


distransient:
Ah, I *really* like this! It feels fairly clear (at least to me) how things are being added to the Graph

amethyst_rendy/src/bundle.rs
```diff
@@ -0,0 +1,621 @@
...
+        Ok(())
+    }
+}
+
+/// An action that represents a single transformation to the
+/// render graph, e.g. addition of single render group.
+///
+/// TODO: more actions needed for e.g. splitting pass into subpasses.
+#[derive(Debug)]
+pub enum RenedrableAction<B: Backend> {
```
Friz64: This is probably a typo :D

amethyst_rendy/src/bundle.rs
```diff
@@ -0,0 +1,943 @@
...
+                .collect(),
+            targets: self.targets,
+            passes: Default::default(),
+            outputs: Default::default(),
+            graph_builder: GraphBuilder::new(),
+        };
+
+        for target in self.roots {
+            // prevent evaluation of roots that were accessed recursively or undefined
+            if !ctx.passes.contains_key(&target) {
```
happenslol: This check should be moved inside the `evaluate_target` function, imo. It's not immediately obvious just from this snippet that evaluating the target can add it to the list of passes.

amethyst_rendy/Cargo.toml
```diff
@@ -22,13 +26,13 @@ hibitset = "0.5.4"
 lazy_static = "1.3"
 log = "0.4"
 palette = { version = "0.4", features = ["serde"] }
-rendy = { version = "0.3", default-features = false, features = ["base", "wsi-winit", "empty", "mesh-obj", "texture-image", "texture-palette", "serde-1"] }
+rendy = { git = "https://github.com/amethyst/rendy.git", rev = "0368513", default-features = false, features = ["base", "wsi-winit", "empty", "mesh-obj", "texture-image", "texture-palette", "serde-1"] }
```
azriel91: Should another version of `rendy` be released instead?

amethyst_rendy/src/bundle.rs
```diff
@@ -0,0 +1,943 @@
...
+            subpass: hal::pass::Subpass<'_, B>,
+            buffers: Vec<NodeBuffer>,
+            images: Vec<NodeImage>,
+        ) -> Result<Box<dyn RenderGroup<B, T>>, failure::Error> {
+            unimplemented!()
+        }
+    }
+
+    #[test]
+    #[ignore] // CI can't run tests requring actual backend
```
azriel91: requiring

amethyst_rendy/src/pass/base_3d.rs
```diff
@@ -666,8 +628,14 @@ impl<B: Backend, T: Base3DPassDef<B>> RenderGroup<B, Resources> for DrawBase3DTr
...
                         {
-                            mesh.bind_and_draw(0, &self.vertex_format_base, range.clone(), encoder)
-                                .unwrap();
+                            if let Err(error) = mesh.bind_and_draw(
+                                0,
+                                &self.vertex_format_base,
+                                range.clone(),
+                                encoder,
+                            ) {
+                                log::warn!("Trying to draw a mesh that lacks {:?} vertex attributes. Pass {} requires attributes {:?}.", error.not_found.attributes, T::NAME, T::base_format());
```
azriel91: rustfmt can't handle, haha. I think if you split the format string using `\` it will be able to.
same below

amethyst_rendy/src/plugins.rs
```diff
@@ -0,0 +1,310 @@
...
+    graph::render::RenderGroupDesc,
+    hal::command::{ClearColor, ClearDepthStencil, ClearValue},
+};
+use std::path::Path;
+
+/// A [RenderPlugin] for opening a window and displaying a render target to it.
+///
+/// When you provide [DisplayConfig], it opens a window for you using [WindowBundle].
+#[derive(Default, Debug)]
+pub struct RenderToWindow {
```
azriel91: It can be worth splitting this to a separate crate (or putting behind a feature gate) for server type applications that may want to render to a frame buffer and send it through a video encoder stream, but not a window. Not a hard comment (just resolve if out of scope).

examples/custom_game_data/game_data.rs
```diff
@@ -74,13 +74,6 @@ impl<'a, 'b> CustomGameDataBuilder<'a, 'b> {
         bundle.build(&mut self.base)?;
         Ok(self)
     }
-    pub fn with_thread_local<S>(mut self, system: S) -> Self
```
azriel91: does this change need to be in? it seems like a separate thing

amethyst_controls/src/systems.rs
```diff
@@ -267,16 +267,7 @@ impl<'a> System<'a> for CursorHideSystem {
...
-
-        let win = res.fetch::<Window>();
-
-        if let Err(err) = win.grab_cursor(true) {
-            log::error!("Unable to grab the cursor. Error: {:?}", err);
-        }
-        win.hide_cursor(true);
-
-        self.is_hidden = true;
+        self.is_hidden = false;
```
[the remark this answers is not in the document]
Frizi: Yeah, good catch. I actually did this once but must have ctrl+z out of it by accident :P

amethyst_ui/src/pass.rs
```diff
@@ -43,6 +47,40 @@ use std::cmp::Ordering;
...
+    /// Select render target on which UI should be rendered.
+    pub fn with_target(mut self, target: Target) -> Self {
+        self.target = target;
+        self
+    }
+}
+
+impl<B: Backend> RenderPlugin<B> for RenderUi {
+    fn build<'a, 'b>(&mut self, builder: &mut DispatcherBuilder<'a, 'b>) -> Result<(), Error> {
+        builder.add(crate::UiGlyphsSystem::<B>::new(), "ui_glyphs_system", &[]);
```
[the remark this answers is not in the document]
Frizi: All systems registered by render plugins are added after a barrier. That means the explicit dependencies on all "non-rendering" systems are not needed. This maybe is not ideal, but cuts down a lot of complexity right now until we have better dependency story in specs.

I have considered adding a UiBundle in there, but that would require extra rendering-unrelated type parameter to be held in the RenderUi plugin. It feels wrong to me. I'd prefer to keep only rendering related stuff here. Having to add UiBundle explicitly also means that if we ever add any extra builder methods to it, we don't have to pass that through rendering plugin as well.

amethyst_rendy/src/bundle.rs
```diff
@@ -0,0 +1,943 @@
...
+        }
+    }
+
+    /// Register a [RenderPlugin].
+    pub fn add_plugin(&mut self, plugin: impl RenderPlugin<B> + 'static) {
+        self.plugins.push(Box::new(plugin));
+    }
+
+    /// Register a [RenderPlugin].
+    pub fn with_plugin(mut self, plugin: impl RenderPlugin<B> + 'static) -> Self {
```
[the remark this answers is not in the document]
Frizi: I usually go with both. We might eventually opt-in to cut down some code duplication with builder macros if it ever becomes a bigger inconvinience.

As for the doc comment, I was under the impression that this is quite standard thing. This exact same pattern is all over the place in rendy for example. But there is no harm in adding it i guess.

amethyst_rendy/src/bundle.rs
```diff
@@ -0,0 +1,943 @@
...
+}
+
+/// Basic building block of rendering in [RenderingBundle].
+///
+/// Can be used to register rendering-related systems to the dispatcher,
+/// building render graph by registering render targets, adding [RenderableAction]s to them
+/// and signalling when the graph has to be rebuild.
+pub trait RenderPlugin<B: Backend>: std::fmt::Debug {
+    /// Hook for adding systems and bundles to the dispatcher.
+    fn build<'a, 'b>(&mut self, _builder: &mut DispatcherBuilder<'a, 'b>) -> Result<(), Error> {
```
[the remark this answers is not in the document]
Frizi: Good idea. I'm a bit torn on `rebuild` though. It is a hook, but the return value is used to decide if rebuild is needed. So it's more like `should_rebuild`. Wdyt?

amethyst_rendy/Cargo.toml
Frizi: Yes it will be soon. We will bump this dependency version once it's done.

amethyst_rendy/src/pass/base_3d.rs
Frizi: I think just dropping the macro parameters to separate lines should do the trick as well :)

amethyst_rendy/src/plugins.rs
Frizi: Yeah i was considering such thing. Then i realised we already pull `amethyst_window` as a dependency anyway... :P I just kept it here due to end-user convinience, but definitely will put it under a feature flag now.

examples/custom_game_data/game_data.rs
Frizi: It just isn't used anymore in the example. It was previously because of RenderingSystem, now it was left unused and there was a warning. So I just deleted it.

amethyst_rendy/shaders/vertex/debug_lines.vert
```diff
@@ -14,17 +14,12 @@ layout(location = 1) in vec4 color_a;
 layout(location = 2) in vec3 position_b;
 layout(location = 3) in vec4 color_b;
 
-const mat2 dir_mats[2] = mat2[](
```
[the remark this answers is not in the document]
Frizi: I think the file is still there. It did lost a few bytes in the process, but it wasn't completely deleted. Github UI might be a little bit confusing on that one i guess.

amethyst_rendy/src/bundle.rs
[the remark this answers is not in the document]
Frizi: I went with `on_build`, `on_plan` and `should_rebuild`.

amethyst_rendy/src/pass/base_3d.rs
```diff
@@ -666,8 +666,14 @@ impl<B: Backend, T: Base3DPassDef<B>> RenderGroup<B, Resources> for DrawBase3DTr
...
```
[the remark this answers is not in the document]
Frizi: Yeah it should

amethyst_rendy/src/bundle.rs
```diff
@@ -0,0 +1,621 @@
...
+}
+
+/// An identifier for render target used in render plugins.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Target {
+    /// Default render target for most operations.
+    /// Usually the one that gets presented to the window.
+    Main,
+    /// Custom render target identifier.
+    Custom(&'static str),
```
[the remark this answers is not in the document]
Frizi: It would poison a lot. You would have to surface that generic up to the rendering bundle. I see no point. This is not serializable or anything as well, so no problems with `&'static` here. The reason i wanted strings is that collision is not very likely and at the same time it doesn't require typesystem magic. We use enums for all the intertnal stuff, can as well make a few extra if something is common (like I just did for shadows). Everything user-defined must be hard to collide, easy to use and not pollute the API.

amethyst_rendy/src/pass/base_3d.rs
```diff
@@ -71,12 +71,12 @@ pub trait Base3DPassDef<B: Backend>: 'static + std::fmt::Debug + Send + Sync {
...
-pub struct DrawBase3DDesc<B: Backend, T: Base3DPassDef<B>> {
+pub struct DrawBase3DDesc<B: Backend, T: Base3DPassDef> {
     skinning: bool,
     marker: PhantomData<(B, T)>,
 }
 
-impl<B: Backend, T: Base3DPassDef<B>> DrawBase3DDesc<B, T> {
+impl<B: Backend, T: Base3DPassDef> DrawBase3DDesc<B, T> {
     /// Create pass in default configuration
     pub fn new() -> Self {
```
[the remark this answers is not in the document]
Frizi: It implements Default as well. Although i can use it inside new here.

amethyst_rendy/Cargo.toml
Frizi: Done :)

amethyst_rendy/src/bundle.rs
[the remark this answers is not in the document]
jaynus: sounds good :thumbsup:

amethyst_rendy/src/submodules/gather.rs
```diff
@@ -36,14 +35,13 @@ impl CameraGatherer {
...
-        let (active_camera, cameras, transforms, dimensions) = <(
+        let (active_camera, cameras, transforms) = <(
             Read<'_, ActiveCamera>,
             ReadStorage<'_, Camera>,
             ReadStorage<'_, Transform>,
-            ReadExpect<'_, ScreenDimensions>,
         )>::fetch(res);
 
-        let defcam = Camera::standard_2d(dimensions.width(), dimensions.height());
+        let defcam = Camera::standard_2d(1.0, 1.0);
```
azriel91: I'm not sure that this change means; I'd trust @jaynus ' review on this

Frizi: It's about removing dependency on window size. In pretty much all cases, your defined camera kicks in. The default is instead made to be of size 1.0. I think we could really just not render stuff if there is no camera and display a warning instead. I don't want to introduce that yet though.

README.md
```diff
@@ -132,7 +132,7 @@ $ pacman -Sy grep gcc pkgconfig openssl alsa-lib cmake make python3 freetype2 aw
 ### Debian/Ubuntu
 
 ```
-# apt install gcc pkg-config openssl libasound2-dev cmake build-essential python3 libfreetype6-dev libexpat1-dev libxcb-composite0-dev libssl-dev libx11-dev
+# apt install gcc pkg-config openssl libasound2-dev cmake build-essential python3 libfreetype6-dev libexpat1-dev libxcb-composite0-dev libssl-dev
```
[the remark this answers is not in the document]
Frizi: I have no idea how this happened. I guess this is a merge error. Good catch!

amethyst_rendy/src/submodules/gather.rs
jaynus: Yup!

amethyst_rendy/src/plugins.rs
```diff
@@ -0,0 +1,322 @@
...
+        self.skinning = true;
+        self
+    }
+}
+
+impl<B: Backend, D: Base3DPassDef> RenderPlugin<B> for RenderBase3D<D> {
+    fn on_build<'a, 'b>(&mut self, builder: &mut DispatcherBuilder<'a, 'b>) -> Result<(), Error> {
+        builder.add(VisibilitySortingSystem::new(), "visibility_system", &[]);
+
+        // TODO: We should ideally register `VertexSkinningBundle` here,
```
happenslol: Wouldn't it make sense to then move this renderplugin to the animation bundle? If it can only be used with the `VertexSkinningBundle`, it would be sensible to only allow importing them together.

Might also be a good use case for a glue crate?

amethyst_rendy/src/plugins.rs
```diff
@@ -0,0 +1,322 @@
+//! Set of predefined implementations of `RenderPlugin` for use with `RenderingBundle`.
```
happenslol: I'm torn about whether or not these should be in separate files. Would make it possible to see which bundles are available by just glancing at the files, and not having to scroll through this entire file.

book/src/pong-tutorial/pong-tutorial-01.md
```diff
@@ -192,13 +185,9 @@ game.run();
...
-Here we're creating a new `WindowBundle` that uses the config we prepared above.
-That bundle is being used as a part of `GameDataBuilder`, a central repository
-of all the game logic that runs periodically during the game runtime.
-
-> **Note:** We will cover systems and bundles in more details later, for now, think of the
-> bundle as a group of functionality that together provides a certain feature to the engine.
-> You will surely be writing your own bundles for your own game's features soon.
+Here we're creating a new instance of `GameDataBuilder`, a central repository
+of all the game logic that runs periodically during the game runtime. Right now it's empty,
+but soon we will start adding there all sorts of systems and bundles that run our game code.
```
happenslol: `Right now it's empty, but soon we willst start adding all sorts of systems and bundles to it - which will run our game code.`

book/src/pong-tutorial/pong-tutorial-01.md
```diff
@@ -208,159 +197,54 @@ Then we call `.run()` on `game` which starts the game loop. The game will
 continue to run until our `SimpleState` returns `Trans::Quit`, or when all states
 have been popped off the state machine's stack.
 
-Try compiling the code now. You should be able to see the window already.
-The content of that window right now is undefined and up to the operating system.
-It's time to start drawing on it.
+Try compiling the code now. You should be able to see that application starts but nothing
```
happenslol: `You should be able to see the application start, but nothing will happen and your terminal will hang until you kill the process. This means that the core game loop is running in circles, and is awaiting tasks. Let's give it something to do by adding a renderer!`

book/src/pong-tutorial/pong-tutorial-01.md
```diff
@@ -208,159 +197,54 @@ Then we call `.run()` on `game` which starts the game loop. The game will
...
+is running and awaiting for tasks to run. Let's give it something to do by adding a renderer.
 
 ## Setting up basic rendering
 
-Now, let's define some rendering code so we can keep moving. This part is not strictly
-necessary to show a window, but we need the renderer to display anything inside it.
-
-We'll cover rendering in more depth later in this tutorial, but for now place the
-following code _below_ the `main()` function:
+After preparing the display config and application scaffolding, it's time to actually use it.
```
happenslol: `... now we'll add some systems to it.`

book/src/pong-tutorial/pong-tutorial-01.md
```diff
@@ -208,159 +197,54 @@ Then we call `.run()` on `game` which starts the game loop. The game will
...
+# fn main() -> Result<(), amethyst::Error>{
+let game_data = GameDataBuilder::default()
+    .with_bundle(
+        RenderingBundle::<DefaultBackend>::new()
+            // The RenderToWindow plugin provides all the scaffolding for opening a window and drawing on it
+            .with_plugin(
+                RenderToWindow::from_config_path(display_config_path)
+                    .with_clear([0.0, 0.0, 0.0, 1.0]),
+            )
+            // RenderFlat2D plugin is used to render entities with `SpriteRender` component.
```
happenslol: `... with a SpriteRender component.`

book/src/pong-tutorial/pong-tutorial-01.md
```diff
@@ -208,159 +197,54 @@ Then we call `.run()` on `game` which starts the game loop. The game will
...
+preconfigured to work together, so you don't have to write them all down one by one.
 
-If you want to use a different background color, you can tweak the RGBA
-values inside `ClearValue::Color`. Values range from `0.0` to `1.0`,
-so to get that cool green color you can try `[0.00196, 0.23726, 0.21765, 1.0]`.
-
-Now let's pack everything up and run it back in the `main()` function. We have to
-expand the existing `GameDataBuilder` with `RenderingSystem` that uses our graph:
+> **Note:** We will cover systems and bundles in more details later, for now, think of the
+> bundle as a group of functionality that together provides a certain feature to the engine.
```
happenslol: `We will cover systems and bundles in more detail later. For now, think of a bundle as a collection of systems that, in combination, will provide a certain feature to the engine.`

book/src/pong-tutorial/pong-tutorial-01.md
```diff
@@ -208,159 +197,54 @@ Then we call `.run()` on `game` which starts the game loop. The game will
...
-    // which will always execute on the main thread.
-    .with_thread_local(RenderingSystem::<DefaultBackend, _>::new(
-        ExampleGraph::default(),
-    ));
-
-let assets_dir = app_root.join("assets/");
+The `RenderingBundle` is actually a bit special, as it doesn't really do anything on its own,
+but relies on it's own plugin system to define what should be rendered and how. We have to use
+`with_plugin` method to tell it that we want `RenderToWindow` and `RenderFlat2D` plugins.
+Those plugins will equip our renderer with ability to open a window and draw 2D sprites on it.
```
happenslol: The `RenderingBundle` has a difference to most other bundles: It doesn't really do anything by itself. Instead, it relies on its own plugin system to define what should be rendered and how. We use the `with_plugin` method to tell it that we want to add the `RenderToWindow` and `RenderFlat2D` plugins. Those plugins will equip our renderer with the ability to open a window and draw sprites to it.

book/src/pong-tutorial/pong-tutorial-01.md
```diff
@@ -208,159 +197,54 @@ Then we call `.run()` on `game` which starts the game loop. The game will
...
+In this configuration, our window will have a black background.
+If you want to use a different color, you can tweak the RGBA
+values inside `with_clear`. Values range from `0.0` to `1.0`, to get that cool green color
+you can try `[0.00196, 0.23726, 0.21765, 1.0]`.
 
-Here we're creating a new `RenderingSystem`, adding the `ExampleGraph` we
-created. Additionally we are adding a `Processor::<SpriteSheet>` system,
-which will make sure that all `SpriteSheet` assets are being properly loaded.
-We will learn more about those in the next chapter.
+> **Note:** This setup code is using amethyst's `RenderPlugin` trait based system that
```
happenslol: Shouldn't amethyst generally be capitalized in the docs?

amethyst_rendy/src/plugins.rs
Frizi: Not really. You can still use that plugin without vertex skinning. It's optional feature, you need it only when `self.skinning` is `true`. Also that bundle has it's own set of configuration that you would have to proxy through that call. Having a NOTE in docs is probably enough.

amethyst_rendy/src/plugins.rs
Frizi: You probably mean what plugins. The way to see that in the docs is to navigate to `RenderPlugin` trait and see all implementers.

book/src/pong-tutorial/pong-tutorial-01.md
Frizi: Good catch! It is everywhere else.

amethyst_rendy/src/plugins.rs
Frizi: I'll actually remove that TODO as it's not actionable in any way.

amethyst_rendy/src/plugins.rs
happenslol: Yeah, sorry about that, meant plugins. I'm still not sure about this, I don't really see any benefit to having them in the same file, if there are more or if they expand over time, we would have to break them up anyways...

amethyst_rendy/src/plugins.rs
happenslol: I see, fair enough

Frizi:
bors r=jaynus,happenslol,azriel91

Frizi:
bors retry

amethyst_rendy/src/bundle.rs
```diff
@@ -0,0 +1,950 @@
...
+/// Collection of predefined constants for action ordering in the builtin targets.
+/// Two actions with the same order will be applied in their insertion order.
+/// The list is provided mostly as a comparison point. If you can't find the exact
+/// ordering you need, provide custom `i32` that fits into the right place.
+///
+/// Modules that provide custom render plugins using their own orders should export
+/// similar enum with ordering they have added.
+#[derive(Debug)]
+#[repr(i32)]
+pub enum RenderOrder {
```
omni-viral: You leave 9 between first variants to insert custom ones? What will happen if 10 is not enough?

amethyst_rendy/src/bundle.rs
```diff
@@ -0,0 +1,950 @@
...
+
+    fn depth(&self) -> bool {
+        match self {
+            RenderableAction::RenderGroup(g) => g.depth(),
+        }
+    }
+}
+
+/// Trait for easy conversion of various types into `RenderableAction` shell.
+pub trait IntoAction<B: Backend> {
```
omni-viral: How about `impl<B, G> From<G> for RenderableActrion<B> where G: RenderGroupBuilder<B, Resources>>`?

amethyst_rendy/src/bundle.rs
```diff
@@ -0,0 +1,950 @@
...
+    /// Settings for optional depth output
+    pub depth: Option<ImageOptions>,
+}
+
+#[derive(derivative::Derivative)]
+#[derivative(Debug(bound = ""))]
+struct TargetPlan<B: Backend> {
+    key: Target,
+    #[derivative(Debug = "ignore")]
+    extensions: Vec<Box<dyn FnOnce(&mut TargetPlanContext<'_, B>) -> Result<(), Error> + 'static>>,
```
omni-viral: If plugin would register extensions with names it will be easier debuggable

amethyst_rendy/src/bundle.rs
```diff
@@ -0,0 +1,950 @@
...
+        &mut self,
+        extension: Box<dyn FnOnce(&mut TargetPlanContext<'_, B>) -> Result<(), Error> + 'static>,
+    ) {
+        self.extensions.push(extension);
+    }
+
+    fn evaluate(self, ctx: &mut PlanContext<B>) -> Result<(), Error> {
+        if self.outputs.is_none() {
+            return Err(format_err!(
+                "Trying to evaluate not fully defined pass {:?}. Missing `define_pass` call.",
```
omni-viral: Won't you hit this if plugin registers pipeline to write into `Custom("VeryImportantTarget")` which not get used anywhere?

omni-viral: I can't see any calls to `define_pass` except one that `RenderToWindow` is doing and in tests.

amethyst_rendy/src/plugins.rs
```diff
@@ -0,0 +1,318 @@
...
+            }
+
+            Ok(())
+        }
+
+        #[allow(clippy::map_clone)]
+        fn should_rebuild(&mut self, res: &Resources) -> bool {
+            let new_dimensions = res.try_fetch::<ScreenDimensions>();
+            use std::ops::Deref;
+            if self.dimensions.as_ref() != new_dimensions.as_ref().map(|d| d.deref()) {
```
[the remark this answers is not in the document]
Frizi: I know. It doesn't work well here as things are ambiguous. It's really a deref of a deref or something.

amethyst_rendy/src/plugins.rs
```diff
@@ -0,0 +1,318 @@
...
+        }
+    }
+
+    impl<B: Backend> RenderPlugin<B> for RenderToWindow {
+        fn on_build<'a, 'b>(
+            &mut self,
+            builder: &mut DispatcherBuilder<'a, 'b>,
+        ) -> Result<(), Error> {
+            if let Some(config) = self.config.take() {
+                WindowBundle::from_config(config).build(builder)?;
```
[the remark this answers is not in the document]
Frizi: Yes, this is intentional.

amethyst_rendy/src/plugins.rs
```diff
@@ -0,0 +1,318 @@
...
+
+            let depth_options = ImageOptions {
+                kind: window_kind,
+                levels: 1,
+                format: Format::D32Sfloat,
+                clear: Some(ClearValue::DepthStencil(ClearDepthStencil(1.0, 0))),
+            };
+
+            plan.add_root(Target::Main);
+            plan.define_pass(
```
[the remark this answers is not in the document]
Frizi: This defines the details of renderpass that is behind `Target::Main`. You can add stuff to targets that are never fully defined, as long as those either aren't used in practice or will be defined later. That way plugins can extend potential functionalities without checking for their existance and also all the configuration is order independant.

amethyst_rendy/src/bundle.rs
Frizi: It has to be enough. If the order doesn't matter, you put things under the same enum. I can't really think about a case where you need 9 more distinct values inbetween. In fact, I think you would rarely use anything else than defined on that list already.

amethyst_rendy/src/plugins.rs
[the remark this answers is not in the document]
omni-viral: Isn't this kinda limiting?

amethyst_rendy/src/bundle.rs
Frizi: Orphan rule forbids that

amethyst_rendy/src/bundle.rs
Frizi: Extension types might have names instead. Might be added in the future, I don't see exactly a place to use that right now.

amethyst_rendy/src/bundle.rs
Frizi: There is only one pass in default plugins right now. There will be another once shadow mapping is implemented. You won't hit an error just by extending nonexisting target. It will never be actually evaluated, as the whole system is lazy. If you depend on a target that is never defined, then it is an error and above error would be hit.

Frizi:
cargo retry

Frizi:
bors retry

Frizi:
bors retry

Frizi:
bors retry

Frizi:
bors retry

Frizi:
bors retry

Frizi:
bors retry

Frizi:
bors retry

amethyst_rendy/src/bundle.rs
omni-viral: Ah, it collides with `From<T> for T`

PR #1778, opened 2019-07-03
fix typos

KuSpa:
## Description
I was reading the book and saw some typos. I thought it'ld be nice to fix them :)

## Additions

- fixed typos

## Removals

- typos

### DIFF
--- a/book/src/prefabs/how_to_define_prefabs_aggregate.md
+++ b/book/src/prefabs/how_to_define_prefabs_aggregate.md
@@ -155,7 +155,7 @@ If you intend to include a [`Component`] that has not yet got a corresponding [`
 
     ```
 
-    The problem is that both the `SpriteScenePrefab`s need to write to `Trasform` and several other common `Components`. Because Amythest's underlyng ECS system determins what resources are accessed based on static types it can't determine that only one of the `SpriteScenePrefab`s will be accessed at a time and it attempts a double mutable borrow which fails. The solution is to define the `PrefabData` hierarchically so each component only appears once:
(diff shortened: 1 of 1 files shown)


book/src/prefabs/how_to_define_prefabs_aggregate.md
```diff
@@ -155,7 +155,7 @@ If you intend to include a [`Component`] that has not yet got a corresponding [`
 
     ```
 
-    The problem is that both the `SpriteScenePrefab`s need to write to `Trasform` and several other common `Components`. Because Amythest's underlyng ECS system determins what resources are accessed based on static types it can't determine that only one of the `SpriteScenePrefab`s will be accessed at a time and it attempts a double mutable borrow which fails. The solution is to define the `PrefabData` hierarchically so each component only appears once:
+    The problem is that both the `SpriteScenePrefab`s need to write to `Transform` and several other common `Components`. Because Amythest's underlyng ECS system determines what resources are accessed based on static types, it can't determine that only one of the `SpriteScenePrefab`s will be accessed at a time and it attempts a double mutable borrow which fails. The solution is to define the `PrefabData` hierarchically so each component only appears once:
```
jojolepro:
```suggestion
    The problem is that both the `SpriteScenePrefab`s need to write to `Transform` and several other common `Components`. Because Amythest's underlying ECS system determines what resources are accessed based on static types, it can't determine that only one of the `SpriteScenePrefab`s will be accessed at a time and it attempts a double mutable borrow which fails. The solution is to define the `PrefabData` hierarchically so that each component only appears once:
```

Issue #1493, opened 2019-03-29
[MAINTENANCE] Gradually migrate to `System`s without `setup`


torkleyy:
## Note

* There is an issue to remove / deprecate `System::setup` in the future: https://github.com/slide-rs/specs/issues/525
* Reasoning is provided in that issue

## Description

Right now, `setup` is used for initializing systems. That was not the point of `setup`; it was introduced to create defaults for resources, not to initialize the system.

Using it for initialization leads to a couple of problems (copied from the above issue):

* implicit resource creation -> might be using the wrong resource, especially when generics are involved (one system uses `Foo<i32>`, the other one `Foo<i64>` and both will fail to see the changes of the other one)
* initializing system fields with it means:
    * fields need to be `Option`s
    * resource initialization is decentralized (whether one considers that as a drawback or not)
* resources / system fields can hardly depend on each other when initialized using `setup`, possibly leading to (undocumented) ordering requirements of the setup calls

The alternative I propose for Amethyst is to initialize systems on construction.

One example of this:

```rust
impl MySystem {
	pub fn new(world: &mut World) -> Self {
	    MySystem {
	        foo_reader: world.entry::<EventChannel<Foo>>()
                .or_default().register_reader(),
	    }
	}
}
```

Advantages of this:

* less boilerplate in `System::run` + less `unwrap`s / incentive to use them
* less chance of doing a second insertion of the `EventChannel` as it happened multiple times to beginners (and as a result, registered readers were discarded)
* interaction / control of this process is easier as it is provided in a free standing function; other values than a default value can be provided through alternative constructors instead of complex configuration mechanisms

### Architecture

This will require passing the `World` to systems during graph building; this seems to be an architectural problem since the `GameDataBuilder` creates the `World`, but also takes in all the systems. Creation of the `World` prior to that should, however, not be a problem. The main question that results from this is, as far as I see, how to combine this with the builder pattern.

### Documentation

The book possibly includes examples of system initialization through `setup`; those should be removed, too.

jojolepro:
not stale

jojolepro:
I'm planning to start integrating the getset, new-derive, derivative macros more considering how much they reduce code boilerplate.

torkleyy:
I'm not so sure about `getset` and `new-derive`, considering how they often break auto completion. The boilerplate for getters, setters and `new` isn't that big compared to trait implementations.

How is this related to this issue?

jojolepro:
It doesn't, I got confused :P

PR #1780, opened 2019-07-04
Move away from System::setup, migrate to latest Specs.

jojolepro:
## Description

Move avay from using the System::setup method, as described by https://github.com/amethyst/amethyst/issues/1493

## PR Checklist

- [ ] Ran `cargo clippy --all`


examples/animation/main.rs
```diff
@@ -190,19 +190,21 @@ fn main() -> amethyst::Result<()> {
     let display_config_path = app_root.join("examples/animation/resources/display_config.ron");
     let resources = app_root.join("examples/assets/");
 
+    let mut world = World::new();
+
     let game_data = GameDataBuilder::default()
```
distransient: Could we instead have GameDataBuilder optionally have a constructor where it receives a user's World (constructing one with ::new() by default) and add an argument to SystemBundle::build for receiving the reference to that instance? The costs would be offloaded to changing the API for SystemBundle but I think this solution would be better in the long term since it's a utility interface for building dispatchers anyways.

Then, the calls to ::with_bundle here would not take the &mut world. The problem would arise when users have simple systems which take a reference to world for initialization alongside bundles in their gamedatabuilder, since that would create two simultaneous mutable borrows on world. Such systems might need to be done in a GameDataBuilder method that takes an anonymous function, with the argument being the reference to that World.

azriel91:
Related: https://github.com/slide-rs/specs/issues/525

examples/animation/main.rs
jojolepro: Doesn't work, because the user is also creating systems that require &mut world access for their new() function. :(

distransient: I said that :P 

> The problem would arise when users have simple systems which take a reference to world for initialization alongside bundles in their gamedatabuilder, since that would create two simultaneous mutable borrows on world. Such systems might need to be done in a GameDataBuilder method that takes an anonymous function, with the argument being the reference to that World.

amethyst_animation/src/skinning/systems.rs
```diff
@@ -17,19 +17,27 @@ use super::resources::*;
...
+    updated_id: ReaderId<ComponentEvent>,
 }
 
 impl VertexSkinningSystem {
     /// Creates a new `VertexSkinningSystem`
-    pub fn new() -> Self {
-        Self::default()
+    pub fn new(world: &mut World) -> Self {
+        use amethyst_core::ecs::prelude::SystemData;
+        <Self as System<'_>>::SystemData::setup(&mut world.res);
```
azriel91: Since the `setup` trait method is no longer overridden, this line can be omitted from the `System` (ditto for all other systems that no longer override `setup`).

**edit:** ohh, the system data needs to set up the world for things like `EventChannel` to exist. Wonder if there's a nice way to avoid needing every one to add this line at the start of `new`. Perhaps:

```rust
trait System {
    fn new(world: &mut World) -> Self;
    // ..
}

// Perhaps the Dispatcher would construct the `System`:
dispatcher.add_system::<VertexSkinningSystem>("skinning_system", &[]);

// Then inside `add_system`:
fn add_system<S: System>(&mut self, name: &str, deps: &[&str]) -> Result<_, _> {
    S::setup(&mut self.world);
    let system = S::new(&mut self.world);
    // ..
}
```

It doesn't work so nicely when there are systems that take parameters in constructors like this:

```rust
FreeRotationSystem::new(world, self.sensitivity_x, self.sensitivity_y),
```

 (using an associated type for the "rest of the parameters" is ugly 🙁)

@torkleyy do you have a preference?
I'm happy to live with calling `SystemData::setup` for now, and we can revisit this after.

amethyst_input/src/system.rs
```diff
@@ -17,15 +17,24 @@ use thread_profiler::profile_scope;
...
+        let reader = world
+            .res
+            .fetch_mut::<EventChannel<Event>>()
+            .register_reader();
+        if let Some(ref bindings) = bindings {
+            world.res.fetch_mut::<InputHandler<T>>().bindings = bindings.clone();
+        }
         InputSystem {
-            reader: None,
+            reader: reader,
```
azriel91: sneaky, don't need the `: reader` part 😛

amethyst_network/src/bundle.rs
```diff
@@ -46,7 +46,11 @@ where
     T: Send + Sync + PartialEq + Serialize + Clone + DeserializeOwned + 'static,
 {
     /// Build the networking bundle by adding the networking system to the application.
-    fn build(self, builder: &mut DispatcherBuilder<'_, '_>) -> Result<(), Error> {
+    fn build(
+        self,
+        world: &mut World,
```
azriel91: should be `_world` i think, can check build warnings

examples/animation/main.rs
jojolepro: ah right :P

distransient: Closing this. If I figure out a nice way to do it, I'll open an issue dedicated for it :P

Frizi:
How am I supposed to configure systems at their creation time? As constructor invocation is moved into the dispatcher, i need to be able to skeak additional data into that constructor somehow. This is important for windowing changes that are about to happen. I believe that current form of `new` based system setup in new specs isn't really enough for our purposes.

jojolepro:
I'm not sure I understand. You can simply add more parameters in the new method, or add a builder pattern to create the system? @Friz64

Frizi:
@jojolepro You are right. I'm sorry for some reason i thought that `fn new` was now part of the trait. Should read more before expressing concerns.

Friz64:
(I love getting Notifications because someone accidentally mentioned me instead of Frizi :D)

distransient:
I always assume you two are like hacker twins basically

amethyst_utils/src/fps_counter.rs
```diff
@@ -1,7 +1,7 @@
-//! Util Resources
```
azriel91: should keep

book/src/concepts/event-channel.md
```diff
@@ -173,7 +173,7 @@ and you also need to get read access:
 # }
 ```
 
-Then, in the `System`'s setup method:
+Then, in the `System`'s new method:
```
azriel91: should add backticks around this one -- it's obvious when it has it, but non-obvious when it doesn't

book/src/concepts/system.md
```diff
@@ -540,22 +540,3 @@ impl<'a> System<'a> for MyFirstSystem {
 }
 ```
 
-## The setup method
```
azriel91: I think it's worth keeping the info somehow, for the `new` method. It's difficult for new users to understand why we have the system data setup in `new` otherwise

src/app.rs
```diff
@@ -614,7 +609,7 @@ where
 
     /// Adds the supplied ECS resource which can be accessed from game systems.
     ///
-    /// Resources are common data that is shared with one or more game system.
+    /// World are common data that is shared with one or more game system.
```
azriel91: Resources is correct 💯

jojolepro:
bors try

jojolepro:
bors try

jojolepro:
bors try

jojolepro:
For merge in 0.13.

Friz64:
> For merge in 0.13.

Why not now?

azriel91:
Reduces double migration work:
* current branch needs people to migrate quite a bit of code
* the system-initialization work (in progress) will undo a lot of that shift (it still needs to be shifted to, but less)

Also saves us writing two sets of migration notes.

jojolepro:
Also I had concerns about introducing a big change like that on release day.

book/src/concepts/system.md
distransient: Still relevant.

amethyst_derive/Cargo.toml
```diff
@@ -16,9 +16,11 @@ appveyor = { repository = "amethyst/amethyst" }
 travis-ci = { repository = "amethyst/amethyst" }
 
 [dependencies]
+heck = "0.3.1"
```
distransient: h*ck

amethyst_animation/src/skinning/systems.rs
distransient: I believe this may be resolved for now.

azriel91: oh yes certainly

azriel91:
I have WIP to allow `ReaderId<*Event>` and `ReaderId<ComponentEvent>` to be automatically generated by the macro, but would rather have those reviewed independently in its own PR.

src/game_data.rs
```diff
@@ -168,11 +180,23 @@ impl<'a, 'b> GameDataBuilder<'a, 'b> {
     ///     // It is legal to register a system with an empty name
     ///     .with(NopSystem, "", &[]);
     /// ~~~
-    pub fn with<S>(mut self, system: S, name: &str, dependencies: &[&str]) -> Self
+    pub fn with<SD, S>(
```
jojolepro: I still don't understand how it will be possible to add systems not implementing SystemDesc without having to put them in a bundle. Did I miss something?

Here's how one of my dispatchers looks like, for reference:
image

In there, there's quite a few systems that are defined externally, in crates that have no dependency on amethyst. This PR would mean that I need to put either ALL systems in a single bundle, or make multiple bundles to add the systems that are not implementing SystemDesc. Did I get that right?

azriel91: > in crates that have no dependency on amethyst

ah, this would make it inconvenient, but not impossible. For the crate that *does* depend on `amethyst` and those systems (i.e. the one that references `GameData`), you can implement `SystemDesc` on a separate struct in that crate.

Alternatives:

* Since the systems appear to be constructible without parameters (I may be wrong), it wouldn't be difficult to write a macro to generate a `SomethingSystemDesc` -> `SomethingSystem`. It doesn't need to be a proc macro.
* Crates that provide systems depend on `amethyst`.
* Crates that provide systems depend on `amethyst_core` and optionally `amethyst_derive` for the proc macro.
* We move the trait to `specs`, though it doesn't really appear to be a `specs` concern (lazy instantiation is only required because Amethyst inserts resources into the `World`).

azriel91: Would like to mention, updating the crates that don't depend on `amethyst` (but do depend on `specs`) would not be such a bad thing -- `amethyst` is compiled for the project anyway, and if they do depend on `amethyst`, the `specs` version is always kept consistent.

The con would be, if you build a crate via `cargo build -p <crate>`, then it would compile `amethyst` separately for that crate instead of `specs` downwards.

azriel91: ha, alternatively, we just expose another `with_` method that takes a `System` directly :smile:
why complicate things

azriel91: Okay, implemented this, `with` takes a `S: System` as usual, `with_system_desc` takes the `SD: SystemDesc`

jojolepro:
Allright... everything seems in order, so let's see if this big PR will merge ;)

bors r+

PR #1788, opened 2019-07-06
Add link to starter project to 'Getting Started' book section

happenslol:
## Description

This PR adds a link to the only currently existing starter project to the "Getting Started" section of the book. I initially planned to add an entire chapter for this, which the front page could then also link to - However, I decided against it for now. Once we have more starter projects for different types of games, I think this would be a good idea though.

On a side note, I think we should update the PR template to make sure that the starters don't break. They'll be targeting the same version as the CLI, so maybe just mentioning it in the version release manual would be enough too.

## PR Checklist

### DIFF
--- a/book/src/getting-started.md
+++ b/book/src/getting-started.md
@@ -48,7 +48,7 @@ you should get `Cargo.toml`, `src/main.rs` and `resources/display_config.ron`.
 
 ### Starter Project
 
-If you want to get running as quickly as possibly and start playing around with Amethyst, you can also use a starter project. These are specifically made for certain times of games, and will set you up with the groundwork needed to start right away.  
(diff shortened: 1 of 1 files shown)


book/src/getting-started.md
```diff
@@ -46,6 +46,15 @@ amethyst new <game-name>
 
 you should get `Cargo.toml`, `src/main.rs` and `resources/display_config.ron`.
 
+### Starter Project
+
+If you want to get running as quickly as possibly and start playing around with Amethyst, you can also use a starter project. These are specifically made for certain times of games, and will set you up with the groundwork needed to start right away.
```
0x6273:
```suggestion
If you want to get running as quickly as possibly and start playing around with Amethyst, you can also use a starter project. These are specifically made for certain types of games, and will set you up with the groundwork needed to start right away.  
```

azriel91:
bors r=azriel91,mralve,0x6273

PR #1792, opened 2019-07-08
Fix a compile error in the pong tutorial

KSXGitHub:
### Description

I followed this tutorial until this point where a compile-time error occurs:

```
  (paddle_y + scaled_amount)
            ^ mismatched types: f32 and amethyst::core::Float
```


azriel91:
Ah, it's a bug in the `stable` book. On `master`, the absence of that method call *happens* to be correct as the `Float` type is now removed. I'm not sure how the book gets built/published, but for consistency/accuracy it would be a doc fix. @happenslol do you know how the book gets published?

jojolepro:
As stated by @azriel91 , the Float type does not exist on master anymore. If you want, you can open a PR with this change but targeting the 0.11 release tag to fix the old book. Thanks!

PR #1798, opened 2019-07-09
Add EntityUuid component to core

Xaeroxe:
## Description

Oftentimes while developing a game it's useful to have an entity reference that persists across game sessions, and across networked machines. Uuids serve this purpose quite well, as machines can generate them independently of each other with no fear of ID collisions. So, this PR adds the notion of an Entity Uuid to core, such that it can be used across Amethyst itself, and by downstream users.

## Additions

- amethyst_core::entity_uuid module
  - EntityUuidSystem
  - EntityUuidMap resource.

## Removals

None

## Modifications

None

## PR Checklist

- N/A Updated the content of the book if this PR would make the book outdated.

### DIFF
--- a/amethyst_core/src/entity_uuid.rs
+++ b/amethyst_core/src/entity_uuid.rs
@@ -6,13 +6,14 @@
 //! identification in a networked multiplayer situation.
 
 use fnv::FnvHashMap;
-use hibitset::BitSetLike;
-use shrev::ReaderId;
 use specs::{
-    storage::ComponentEvent, BitSet, Component, Entities, Entity, FlaggedStorage, ReadStorage,
-    Resources, System, SystemData, Write, WriteStorage,
+    Component, Entity, FlaggedStorage,
+    System, Write,
 };
+use serde::{Deserialize, Serialize};
 use uuid::Uuid;
+
+use std::sync::{Arc, atomic::{AtomicBool, Ordering}};
 
 /// An identifier for an entity which can persist across game sessions, and across machines.
 ///
@@ -20,9 +21,11 @@ use uuid::Uuid;
 /// identification in a networked multiplayer situation.
 ///
 /// Once the Uuid is initialized in this structure it should not be altered.
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Debug, Deserialize, Serialize)]
 pub struct EntityUuid {
     uuid: Uuid,
+    #[serde(skip)]
+    dead_signal: Arc<AtomicBool>,
 }
 
 impl Component for EntityUuid {
@@ -30,16 +33,12 @@ impl Component for EntityUuid {
 }
 
 impl EntityUuid {
-    /// Create a new instance with a new randomly generated Uuid.
-    fn new() -> Self {
-        Self {
-            uuid: Uuid::new_v4(),
-        }
-    }
-
     /// Create a new instance with a pre-defined Uuid, typically the Uuid would be deserialized from the disk or network.
     fn new_from_uuid(uuid: Uuid) -> Self {
-        Self { uuid }
+        Self {
+            uuid,
+            dead_signal: Arc::new(AtomicBool::new(false)),
+        }
     }
 
     /// Retrieve the Uuid structure contained in this component.
@@ -48,10 +47,16 @@ impl EntityUuid {
     }
 }
 
+impl Drop for EntityUuid {
+    fn drop(&mut self) {
+        self.dead_signal.store(true, Ordering::Relaxed);
+    }
+}
+
 /// An ECS resource which allows you to create new EntityUuid components, and later retrieve the entities with the contained UUID.
 #[derive(Debug, Default)]
 pub struct EntityUuidMap {
-    map: FnvHashMap<Uuid, Entity>,
+    map: FnvHashMap<Uuid, (Entity, Arc<AtomicBool>)>,
 }
 
 impl EntityUuidMap {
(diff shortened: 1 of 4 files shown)


Xaeroxe:
It just occurred to me there's currently no way to clear the mapping table. I'll need to add some garbage collection to this.

amethyst_core/src/entity_uuid.rs
```diff
@@ -0,0 +1,157 @@
...
+    /// Create a new EntityUuid component with a new randomly generated Uuid.
+    pub fn new_uuid(&mut self, e: Entity) -> EntityUuid {
+        let r = EntityUuid::new();
+        self.map.insert(r.uuid, e);
+        r
+    }
+
+    /// Create a new EntityUuid component with a pre-defined Uuid, typically the Uuid would be deserialized from the disk or network.
+    pub fn with_uuid(&mut self, uuid: Uuid, e: Entity) -> EntityUuid {
+        self.map.insert(uuid, e);
```
jojolepro: Instead of this, I think you should have the System listen to inserted component events. This way it would prevent people from manually creating the component and inserting it on the entity without going through the map.

amethyst_core/src/entity_uuid.rs
```diff
@@ -0,0 +1,157 @@
...
+        for event in storage.channel().read(self_events_id) {
+            match event {
+                ComponentEvent::Removed(id) => {
+                    self.removed.add(*id);
+                }
+                _ => {}
+            }
+        }
+        for entity in (&self.removed).iter().map(|i| entities.entity(i)) {
+            map.map.retain(|_, &mut e| e != entity);
```
jojolepro: Is it possible to use HashMap::remove here?

jojolepro:
Thanks for this PR! <3

amethyst_core/src/entity_uuid.rs
Xaeroxe: The functions to manually create the component are not public, meaning that behavior is already prevented. You can only create the component through the map.

amethyst_core/src/entity_uuid.rs
Xaeroxe: No, unfortunately. Once the Uuid component is removed I have no way to get at the Uuid. So I am only able to remove by the value of the hashmap, rather than the key.

Xaeroxe:
This isn't ready to be merged yet. It's still got a problem. If the component is deserialized into existence then the mapping won't know about it. We'll need to fix this before merging.

kabergstrom:
What is the advantage of having this in `amethyst_core`?

Xaeroxe:
@kabergstrom Well at first I wanted to put it in `amethyst_util` but then it occurred to me that if we do that then we have this new general purpose entity identification system that Amethyst itself can't use in several places. The compilation process roughly goes

Core -> Things that don't depend on renderer + renderer -> Things that do depend on renderer

So the way we can guarantee we'll be able to use this internally as well, is to have it in core. This allows us to use it as part of prefabs, networking, and maybe even built in support for saving the game.

jojolepro:
A agree that this belongs in core. Concerning the deserialization issue, see my previous comment on listening to inserted/modified events.

kabergstrom:
I guess I just don't understand why you want to put this in core before you have fully developed the use-case for it? I could see the type definition being useful as a persistent entity ID, but I am not sure I understand the use-case for the map. I mean, wouldn't it be good to have it in a separate crate first, like with `type-uuid`?

Somewhat relevant is that atelier-assets is using UUIDs as persistent identifiers for assets, and I imagine that individual entity templates in a prefab will be assets too. Once instantiated, they would need the kind of mapping introduced in this PR, but I haven't got to implementing the details yet.

Xaeroxe:
@jojolepro Those events don't contain entity generation data, meaning a solution relying on them will break in the following scenario.

1. Frame start
2. Entity 1 is added with EntityUuid
3. Entity 1 is removed.
4. Entity 2 is added with EntityUuid, and re-uses the id value for entity 1 with a different generation.
5. Frame ends

In this scenario the system can't reliably distinguish between entity 1 and entity 2, resulting in buggy behavior. I've confirmed this happens via unit testing as well.


@kabergstrom What would constitute a sufficiently developed use case to justify the inclusion in core? I'm just hesitant to introduce yet another dependency, and I don't see the drawback to including this in core.

kabergstrom:
> @kabergstrom What would constitute a sufficiently developed use case to justify the inclusion in core? I'm just hesitant to introduce yet another dependency, and I don't see the drawback to including this in core.

Something using it was my thinking. The only drawback is exposing a new public API that may get users, which increases friction for change.

jojolepro:
> Entity 2 is added with EntityUuid, and re-uses the id value for entity 1 with a different generation

Wouldn't that be considered a different entity if the generation is different?

Xaeroxe:
> Wouldn't that be considered a different entity if the generation is different?

Yes, the problem is the system can't tell the difference between

Entity: ID: 3, Generation: 1, Uuid: ec4a61a1-c695-442c-8482-dc77b7f05a96
Entity: ID: 3, Generation: 2, Uuid: 8c8f823d-a4d9-4068-85bb-70c1072fe347

Because the only thing encoded in the events is ID: 3 for both entities. Consequently relying on the events for this will produce a bug if an entity re-uses an existing index before the EntityUuidSystem is given a chance to run, as might happen in a single frame.

Another problem with this approach is that if I add a Uuid component to an Entity and then try and look it up by Uuid without advancing the frame, the lookup may fail.

One approach I'd like to use instead would be to make the Uuid->Entity mapping part of a new wrapper component storage, however this approach has two problems, one is that the `specs::UnprotectedStorage` trait is not generation aware, meaning I can only map Uuid->Index, and the other problem is that you can't expose a new API on a new storage type because the type is erased by the `MaskedStorage` structure. So I can't implement that either without modifying specs.

So honestly I've come to the point where I think maybe I should do away with the component altogether, and instead use a resource with two hash maps, i.e.

```rust
/// An ECS resource that presents a bi-directional mapping between Uuids and Entities.
#[derive(Debug, Default)]
pub struct EntityUuidMap {
    uuid_to_entity: FnvHashMap<Uuid, Entity>,
    entity_to_uuid: FnvHashMap<Entity, Uuid>,
}
```

OvermindDL1:
> Another problem with this approach is that if I add a Uuid component to an Entity and then try and look it up by Uuid without advancing the frame, the lookup may fail.

Isn't this expected considering the async nature of this?  A lot of components can work like this.

Xaeroxe:
I suppose, however we still have the first problem

amethyst_core/src/entity_uuid.rs
```diff
@@ -0,0 +1,140 @@
...
+//!
+//! These properties make them useful for serializing a reference to an entity to disk, such as saving the game, or for
+//! identification in a networked multiplayer situation.
+
+use fnv::FnvHashMap;
+use specs::{Entities, Entity, System, Write};
+use uuid::Uuid;
+
+/// An ECS resource that presents a bi-directional mapping between Uuids and Entities.
+#[derive(Debug, Default)]
```
jojolepro: Should we derive Serialize/Deserialize for this?

amethyst_core/src/entity_uuid.rs
```diff
@@ -0,0 +1,140 @@
...
+    use specs::{world::Builder, System, SystemData, World};
+
+    #[test]
+    fn test_uuid_map() {
+        let mut w = World::new();
+        w.add_resource(EntityUuidMap::new());
+        let e1 = w.create_entity().build();
+        let e2 = w.create_entity().build();
+        let u3;
+        let u4;
```
jojolepro: I understand what is going on with the unasigned value, but its quite confusing written like this.

torkleyy:
Related: https://docs.rs/specs/0.15.0/specs/saveload/struct.UuidMarkerAllocator.html

jojolepro:
For merge in 0.13 once reviewed/fixed/discussed. Not to be merged in 0.12.

Xaeroxe:
Given the lack of an immediate need for this, the difficulty in providing a satisfactory implementation, and the merge conflicts that have come up I've decided to close this PR.

Issue #1748, opened 2019-06-20
No error when transform resource missing


ohnoimdead:
## Description

Adding a transform component to an entity without the Transform resource loaded should cause a panic, but instead silently fails.

## Reproduction Steps

1. Remove line 30 from the pong tutorial step 2: (https://github.com/amethyst/amethyst/blob/master/examples/pong_tutorial_02/main.rs#L30)
2. Run the example

## What You Expected to Happen

A panic as described in the docs at https://book.amethyst.rs/master/pong-tutorial/pong-tutorial-02.html: `thread 'main' panicked at 'Tried to fetch a resource, but the resource does not exist.`

## What Actually Happened

No error and nothing renders.

## Screenshots and Logs

Shouldn't be needed, this is easy to repro.

## Your Environment

1. Operating System: Windows
2. Operating System Version: 10
3. Version of Amethyst: 0.11.0 or master

## Additional Context
Panics as expected in 0.10.0.

Amethyst community: fabulous? or most fabulous? 🤔

Frizi:
I'd say that the problem here is misleading book combined with hard to realise set of set up storages.

The panic doesn't happen, because `RenderingSystem` currently sets up the `Transform` storage as well. If you would remove it too, you would get the panic.

I think the book shoudn't necessarily guide the user through that failure here.

PR #1799, opened 2019-07-10
Remove old error that doesn't occur anymore

jojolepro:
Fixes https://github.com/amethyst/amethyst/issues/1748


jojolepro:
bors try

PR #1802, opened 2019-07-13
Added initial draft of tint docs

michael-hart:
## Description

Add a page to book with details of how Tint works to the sprites chapter.

## Additions

- Modify The Texture page added to book

## Modifications

- New page registered in summary document

## PR Checklist


book/src/sprites/modify_the_texture.md
```diff
@@ -0,0 +1,61 @@
...
+    transform.set_rotation_y_axis(f32::pi());
+
+    // Colour white to show as normal, then update later
+    let colour = Tint(Srgba::new(1.0, 1.0, 1.0, 1.0));
+
+    world
+        .create_entity()
+        .with(transform)
+        .with(colour)
+        .with(texture_handle.clone()) // Use the texture handle as a component
```
0x6273: Why `.clone()` the handle? Also I don't think the comment is needed.

book/src/sprites/modify_the_texture.md
```diff
@@ -0,0 +1,61 @@
+# Modify The `Texture`
+
+The colours of the sprite will show up exactly as in the source file,
```
0x6273: Should probably go with the American spelling "color" instead of "colour" to be consistent with the api.

book/src/sprites/modify_the_texture.md
michael-hart: The code sample was copied from the Display The Texture page, so the comment was left in - I'm happy to remove it, but surely the clone is either necessary in both places and should be kept, or should be removed from both places?

book/src/sprites/modify_the_texture.md
michael-hart: Will change in next commit, but I think it's worth noting that the Appendix A uses the english spelling in several places - is that worth changing as well?

0x6273: The American spelling should be used everywhere. If you find places that use the British spelling you can change it.

book/src/sprites/modify_the_texture.md
0x6273: It doesn't seem necessary in either place. I think you can remove it as well as the `&` in the function signature and where the function is called. (Unless there's a reason for not just using the first `Handle` that I'm overlooking.)

michael-hart: I gave it a quick test and you're right - modified in commit.

book/src/sprites/modify_the_texture.md
michael-hart: Changed everywhere including, slightly concerningly, one .rs file

book/src/sprites/modify_the_texture.md
```diff
@@ -0,0 +1,61 @@
...
+    // Add a transform component to give the image a position
+    let mut transform = Transform::default();
+    transform.set_translation_x(0.0);
+    transform.set_translation_y(0.0);
+    
+    // Flip horizontally
+    transform.set_rotation_y_axis(f32::pi());
+
+    // Color white to show as normal, then update later
+    let color = Tint(Srgba::new(1.0, 1.0, 1.0, 1.0));
```
0x6273: It is not actually updated later. Maybe that's why you call it a draft, but we can't merge this now since it will show up in the book on the website. If you just want this merged now you could remove the link in `SUMMARY.md` and fix it in a future pr.

book/src/sprites/modify_the_texture.md
```diff
@@ -0,0 +1,61 @@
...
+effect on the sprite.
+
+```rust,edition2018,no_run,noplaypen
+# extern crate amethyst;
+use amethyst::assets::{AssetStorage, Handle, Loader};
+use amethyst::core::Transform;
+use amethyst::prelude::*;
+use amethyst::renderer::{ImageFormat, Texture};
+use amethsyt::renderer::resources::Tint;
+use amethyst::renderer::palette::rgb::Srgba;
```
0x6273:
```suggestion
use amethyst::renderer::palette::Srgba;
```

book/src/sprites/modify_the_texture.md
```diff
@@ -0,0 +1,61 @@
...
+
+    // Add a transform component to give the image a position
+    let mut transform = Transform::default();
+    transform.set_translation_x(0.0);
+    transform.set_translation_y(0.0);
+    
+    // Flip horizontally
+    transform.set_rotation_y_axis(f32::pi());
+
+    // Color white to show as normal, then update later
```
jojolepro:
```suggestion
    // Color white to show the sprite as normal. You can change the color at any point to modify the sprite's tint.
```

book/src/sprites/modify_the_texture.md
azriel91: I've updated the comment to match jojo's suggestion, so I think this is resolved now

azriel91:
Rebased over master and committed suggestions

bors r+

PR #1810, opened 2019-07-20
Include `libx11-dev` in debian package dependencies.

azriel91:
## Description

Add `libx11-dev` to debian package dependencies. Without this, you would get:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Failure { command: "\"pkg-config\" \"--libs\" \"--cflags\" \"x11 >= 1.4.99.1\"", output: Output { status: ExitStatus(ExitStatus(256)), stdout: "", stderr: "Package x11 was not found in the pkg-config search path.\nPerhaps you should add the directory containing `x11.pc\'\nto the PKG_CONFIG_PATH environment variable\nNo package \'x11\' found\n" } }', src/libcore/result.rs:999:5
```

## PR Checklist

- **n/a** Ran `cargo +stable fmt --all`
- **n/a** Ran `cargo clippy --all`
- **n/a** Ran `cargo test --all --features "empty"`

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ $ pacman -Sy grep gcc pkgconfig openssl alsa-lib cmake make python3 freetype2 aw
 ### Debian/Ubuntu
 
 ```
-# apt install gcc pkg-config openssl libasound2-dev cmake build-essential python3 libfreetype6-dev libexpat1-dev libxcb-composite0-dev libssl-dev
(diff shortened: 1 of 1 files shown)


jojolepro:
Do you get this even if you use wayland?

JorgeVV:
> Do you get this even if you use wayland?

I do. Installing `libx11-dev` solves it.

PR #1820, opened 2019-07-24
amethyst_tiles

jaynus:
## Description

T his adds the amethyst_tiles subcrate, which provides efficient 2D (and 3D!) tile maps and rendering functionality. The crate provides a `TileMap` struct, which stores underlying `Tile` types in a flat `Vec` storage (could be extended later to more efficient/different storage types). An `Encoder` type is also provided, which performs coordinate encoding into the flat space, allowing for efficient storage and iteration. Region iterator helpers are also included, and can be seen in the render pass. 

Finally, a WIP example is completed at `examples/tiles`. I'd like to include some more complex iteration and access of the tile map before release.

This code has a lot of unit tests.

Features included: 
- `RenderTiles2D` render plugin, which is implemented in `DrawTiles2D` transparency pass.
- `Tile` trait, which is used to implement a tile data type. This is for abstracting away the storage type from the rendering, and allows for efficient storage (a u8 which returns a tile index) to complex storage (an entire struct complex type)
- An `Encoder` trait, which allows for implementing different space filling algorithms for efficient  tile storage and iteration. Default is `MortonEncoder2D`, which implements morton encoding on each Z level. Currently provided are `MortonEncoder`, which is morton encoding across Z-levels, and `HilbertEncoder`, which implements hilbert space filling across Z-levels.

WIP - If something looks very obviously out of place (like a Cargo.toml entry), I probably haven't cleaned it up yet or am experimenting quickly with it.

Created #1858 for further documentation work.

## Modifications

- I had to make a change to camera gatherers in the `gather` submodule of rendy to fetch only the `Entity` I wanted.

## PR Checklist


Cargo.toml
```diff
@@ -309,7 +323,16 @@ path = "examples/events/main.rs"
 [[example]]
 name = "rendy"
 path = "examples/rendy/main.rs"
-required-features = ["animation", "gltf"]
+required-features = ["animation", "tiles_benchmarkgltf"]
```
Dispersia: typo? I don't see tiles_benchmark as a feature, just as a separate example (and if so, separate from gltf or is it just one)

amethyst_tiles/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+thread_profiler = { version = "0.3", optional = true }
+fnv = "1"
+derivative = "1.0"
+hibitset = { version = "0.5.2", features = ["parallel"] }
+smallvec = "0.6"
+failure = "0.1"
+lazy_static = "1.3"
+rayon = "1.0.2"
+bitintr = "0.3"
+space = { path = "../../space" }
```
Dispersia: Space is outside of local repo, change to crates.io or are there changes that need pushed as well?

jaynus: Leftover from testing. Removed!

Cargo.toml
jaynus: Yep, just a typo leftover from when I was testing and benchmarking.

amethyst_tiles/src/morton/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+        let z = morton / d;
+        morton -= z * d;
+
+        Some((morton.pext(0x5555_5555), morton.pext(0xAAAA_AAAA), z))
+    }
+}
+
+/*
+#[inline]
+#[cfg(not(target_feature = "bmi2"))]
```
Frizi: I'm not exatly sure what happens with `bmi2` here. Do you consider it widespread enough to assume? Also, what about non-intel targets (we have to support mobile/wasm)? Or is `Pext` just emulated on non-supporting platforms?

amethyst_tiles/src/pass.rs
```diff
@@ -0,0 +1,503 @@
...
+        if let Some((tile_map, _)) = (&tile_maps, !&hiddens).join().next() {
+            // Render the first tilemap we find
+            let (top, mut bottom) = fetch_world_bounds(tile_map, resources);
+            bottom.z = 0;
+
+            let camera_region = Region::new(top, bottom);
+            camera_region
+                .iter()
+                .filter_map(|coord| {
+                    if let Some(tile) = tile_map.get(&coord) {
```
Frizi: This deep nesting on `Option` inside iterator is a bit disorienting here.

How about some early return?
```rust
let tile = tile_map.get(&coord)?;
let sprite_number = tile.sprite(resources)?;
```

amethyst_tiles/src/pass.rs
```diff
@@ -0,0 +1,503 @@
...
+        let env = FlatEnvironmentSub::new(factory)?;
+        let textures = TextureSub::new(factory)?;
+        let vertex = DynamicVertexBuffer::new();
+
+        let (pipeline, pipeline_layout) = build_tiles_pipeline(
+            factory,
+            subpass,
+            framebuffer_width,
+            framebuffer_height,
+            true,
```
Frizi: As transparency is just assumed here, that `build_tiles_pipeline` might be simplified just a tiny bit.

amethyst_tiles/src/pass.rs
```diff
@@ -0,0 +1,503 @@
...
+
+#[derive(Derivative)]
+#[derivative(Debug(bound = ""))]
+pub struct DrawTiles2D<B: Backend, T: Tile, E: CoordinateEncoder> {
+    pipeline: B::GraphicsPipeline,
+    pipeline_layout: B::PipelineLayout,
+    env: FlatEnvironmentSub<B>,
+    textures: TextureSub<B>,
+    vertex: DynamicVertexBuffer<B, SpriteArgs>,
+    sprites: OneLevelBatch<TextureId, SpriteArgs>,
```
Frizi: For transparent stuff you might want to preserve insertion order, so `OrderedOneLevelBatch` should work better.

amethyst_tiles/src/pass.rs
```diff
@@ -0,0 +1,503 @@
...
+            sprites_ref.prune();
+            self.vertex.write(
+                factory,
+                index,
+                self.sprites.count() as u64,
+                self.sprites.data(),
+            );
+        }
+
+        PrepareResult::DrawRecord
```
Frizi: Once ordered batcher is used, it can be used to detect changes and avoid recording when there weren't any.

amethyst_tiles/src/pass.rs
```diff
@@ -0,0 +1,503 @@
...
+impl<'a, T: Tile, E: CoordinateEncoder> System<'a> for SetupSystem<T, E> {
+    type SystemData = (
+        ReadStorage<'a, Handle<SpriteSheet>>,
+        ReadStorage<'a, Handle<Texture>>,
+        ReadStorage<'a, Hidden>,
+        ReadStorage<'a, TileMap<T, E>>,
+    );
+
+    #[cold]
+    fn run(&mut self, _: Self::SystemData) {}
```
Frizi: This is not `cold`. The function does nothing, but is still called every frame. This is not ideal and we should add an option to setup resources directly in the bundle, but this is how it works now.

examples/assets/texture/cp437_20x20.ron
```diff
@@ -0,0 +1,1542 @@
+(
+    texture_width: 320,
+    texture_height: 320,
+    sprites: [
+        (
+            x: 0,
```
Frizi: Oh crap, this file is long. Not really a PR to fix this i guess, but I think we should start paying more attention to our asset definition files size. This could easily be serialized in a way to keep a sprite per line or something.

examples/tiles/main.rs
```diff
@@ -0,0 +1,252 @@
...
+        .named("player")
+        .build()
+}
+
+fn initialise_camera(world: &mut World, parent: Entity) -> Entity {
+    let (width, height) = {
+        let dim = world.read_resource::<ScreenDimensions>();
+        (dim.width(), dim.height())
+    };
+    //println!("Init camera with dimensions: {}x{}", width, height);
```
Frizi: remove

examples/tiles/main.rs
```diff
@@ -0,0 +1,252 @@
...
+        )?
+        .with(CameraMovementSystem, "movement", &[])
+        .with_bundle(
+            RenderingBundle::<DefaultBackend>::new()
+                .with_plugin(
+                    RenderToWindow::from_config_path(display_config_path)
+                        .with_clear([0.34, 0.36, 0.52, 1.0]),
+                )
+                .with_plugin(RenderUi::default())
+                .with_plugin(RenderFlat2D::default()
```
hyrsky: Missing closing `)`

Cargo.toml
```diff
@@ -25,6 +25,17 @@ vulkan = ["amethyst_rendy/vulkan"]
 metal = ["amethyst_rendy/metal"]
 empty = ["amethyst_rendy/empty"]
 
+tiles = [
+    "amethyst_tiles"
+]
+voxels = [
```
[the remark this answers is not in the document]
jaynus: This was leftover from one of my iterations, deleted.

Makefile
```diff
@@ -14,6 +14,10 @@ COMP_DISASMS = $(patsubst amethyst_rendy/shaders/%,amethyst_rendy/compiled/%.spv
 SHADERS_UI=$(wildcard amethyst_ui/shaders/*)
 COMP_SHADERS_UI = $(patsubst amethyst_ui/shaders/%,amethyst_ui/compiled/%.spv,$(SHADERS_UI))
 COMP_DISASMS_UI = $(patsubst amethyst_ui/shaders/%,amethyst_ui/compiled/%.spvasm,$(SHADERS_UI))
+SHADERS_UI=$(wildcard amethyst_tiles/shaders/*)
+COMP_SHADERS_UI = $(patsubst amethyst_tiles/shaders/%,amethyst_tiles/compiled/%.spv,$(SHADERS_UI))
```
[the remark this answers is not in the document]
jaynus: Fixed.

amethyst_rendy/src/submodules/gather.rs
```diff
@@ -26,6 +26,33 @@ pub struct CameraGatherer {
 }
 
 impl CameraGatherer {
+    /// Collect just the entity which has the current `ActiveCamera`
+    pub fn gather_camera(res: &Resources) -> Option<Entity> {
```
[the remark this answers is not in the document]
jaynus: done

amethyst_tiles/shaders/tiles.vert
```diff
@@ -0,0 +1,38 @@
...
+    uniform mat4 view;
+};
+
+// Quad transform.
+layout(location = 0) in vec2 dir_x;
+layout(location = 1) in vec2 dir_y;
+layout(location = 2) in vec2 pos;
+layout(location = 3) in vec2 u_offset;
+layout(location = 4) in vec2 v_offset;
+layout(location = 5) in float depth;
```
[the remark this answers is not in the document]
jaynus: No, I want them to be different shaders because although they are similar for now, in the future I see this becoming more optimized and specialized to map use cases so i want them to be different to begin with.

amethyst_tiles/shaders/tiles.frag
```diff
@@ -0,0 +1,14 @@
+#version 450
+
+layout(set = 1, binding = 0) uniform sampler2D albedo;
+
+layout(location = 0) in vec2 tex_uv;
```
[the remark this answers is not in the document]
jaynus: See prev

examples/assets/texture/cp437_20x20.ron
jaynus: I mean yah. Just realistically, this is what real projects spritesheets will contain. Not much we can do about it.

amethyst_tiles/src/hilbert/mod.rs
```diff
@@ -0,0 +1,147 @@
+#![allow(
+    non_snake_case,
+    unused_parens,
+    unused_variables,
+    clippy::many_single_char_names
+)]
+/// WIP IMPLEMENTATION, INCOMPLETE
```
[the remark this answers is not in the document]
jaynus: deleted this impl.

amethyst_tiles/src/morton/mod.rs
jaynus: It falls back on a standard implementation in the case that bmi2 doesn't exist. I'd like to still at least use the optimized LUT tables though.

amethyst_tiles/src/pass.rs
```diff
@@ -0,0 +1,468 @@
...
+                p - one
+            } else {
+                p
+            }
+        });
+
+    let mut world_bottom_tile =
+        tile_map
+            .to_tile(&world_bottom)
+            .map_or(Point3::from(*tile_map.dimensions()), |mut p| {
```
hyrsky: I'm also getting [poor performance](https://wetransfer.com/downloads/70d1a823e033c863dc8dd6ac5019ce8320190810005408/46a51cf59600186567b5196d6454818d20190810005409/e55f23) with the example + debug build (I don't know what is causing that. I'll have come back to that some other day, getting too tired). While looking into the performance issue I noticed that it gets worse if the camera comes close to left or top borders of the map.

I'm not sure if this is really a problem at all and I don't understand the 3d maps so this might be wrong:

If I understand this will iterate tiles from ``(0, 0, 0)`` if the camera hits top or left borders (instead of ``(0, tile at camera top y, 0)`` when over the left border). That is potentially feeding render pass with more tiles than necessary. Also if camera bottom right > tilemap top left I think this function will return ``((0,0,0), (dimensions.x, dimensions.y, dimensions.z))`` when nothing should be in view.

amethyst_tiles/src/pass.rs
```diff
@@ -0,0 +1,468 @@
...
+        if let Some((tile_map, _)) = (&tile_maps, !&hiddens).join().next() {
+            // Render the first tilemap we find
+            let (top, mut bottom) = fetch_world_bounds(tile_map, world);
+            bottom.z = 0;
+
+            let camera_region = Region::new(top, bottom);
+            camera_region
+                .iter()
+                .filter_map(|coord| {
+                    let tile = tile_map.get(&coord)?;
```
hyrsky: See my other comment. Should this panic if tile_map.get() return None?

examples/tiles/main.rs
```diff
@@ -0,0 +1,250 @@
...
+        let map_sprite_sheet_handle =
+            load_sprite_sheet(world, "texture/cp437_20x20.png", "texture/cp437_20x20.ron");
+
+        let _reference = init_reference_sprite(world, &circle_sprite_sheet_handle);
+        let player = init_player(world, &circle_sprite_sheet_handle);
+        let _camera = initialise_camera(world, player);
+        let _reference_screen = init_screen_reference_sprite(world, &circle_sprite_sheet_handle);
+
+        let map = TileMap::<ExampleTile, MortonEncoder2D>::new(
+            Vector3::new(64, 64, 64),
```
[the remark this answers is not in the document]
jaynus: I only wanted to perform the tests with a single z-level. `MortonEncoder2D` performs a morton encode on each z-level independently and multiplicatively indexes them;  this allows `MortonEncoder2D` to handle non-cubic spaces.

amethyst_tiles/src/pass.rs
jaynus: The performance issues is due to the large amount of multiplication occuring converting between tile and world space. Because of this, nalgebra just slows everything down a lot. 

I'll look into the other cases.

amethyst_tiles/src/iters.rs
```diff
@@ -0,0 +1,208 @@
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let ret = self.track;
+
+        if self.track.x >= self.region.max.x {
+            self.track.y += 1;
+            self.track.x = self.region.min.x;
+        } else {
+            self.track.x += 1;
+            return Some(ret);
```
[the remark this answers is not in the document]
jaynus: Thanks so much for this! didn't notice, digging in :)

amethyst_tiles/src/pass.rs
jaynus: I'd prefer not to fail hard for cases like this, and fail softly. Maybe a warning?

amethyst_tiles/src/iters.rs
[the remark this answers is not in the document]
jaynus: I fixed this. It was actually not due to the above, but due to a misplacement of the Z-level limit check. Moving it fixed the issue.

amethyst_tiles/src/pass.rs
jaynus: I've added some debug assertions, but wont be failing on out-of-bounds.

distransient:
I propose we delay this until 0.14. Sound fair?

jaynus:
This needs to wait until 0.14, there are a few more changes and cleanup to occur first.

amethyst_tiles/src/pass.rs
jaynus: I've now moved all the math into the shader, and it has resolved this problem.

distransient:
bors try

distransient:
Some clippy errors:
```
error: using `clone` on a `Copy` type

   --> examples/tiles/main.rs:162:37

    |

162 |             let old_camera_entity = old_camera_entity.clone();

    |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^ help: try removing the `clone` call: `old_camera_entity`

    |

    = note: `-D clippy::clone-on-copy` implied by `-D warnings`

    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy


error: using `clone` on a `Copy` type

   --> examples/tiles/main.rs:164:30

    |

164 |             let new_parent = old_parent.entity.clone();

    |                              ^^^^^^^^^^^^^^^^^^^^^^^^^ help: try removing the `clone` call: `old_parent.entity`

    |

    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy


error: you don't need to add `&` to all patterns

   --> examples/tiles/main.rs:165:30

    |

165 |               let new_camera = match old_camera.projection() {

    |  ______________________________^

166 | |                 &Projection::Orthographic(_) => {

167 | |                     Camera::standard_3d(dimensions.width(), dimensions.height())

168 | |                 }

...   |

172 | |                 &Projection::CustomMatrix(_) => unimplemented!(),

173 | |             };

    | |_____________^

    |

    = note: `-D clippy::match-ref-pats` implied by `-D warnings`

    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#match_ref_pats

help: instead of prefixing all patterns with `&`, you can dereference the expression

    |

165 |             let new_camera = match *old_camera.projection() {

166 |                 Projection::Orthographic(_) => {

167 |                     Camera::standard_3d(dimensions.width(), dimensions.height())

168 |                 }

169 |                 Projection::Perspective(_) => {

170 |                     Camera::standard_2d(dimensions.width(), dimensions.height())
```

amethyst_tiles/src/iters.rs
```diff
@@ -0,0 +1,215 @@
...
+use amethyst_core::math::Point3;
+
+/// Cubic region stored and handled as a Morton value.
+#[derive(Copy, Clone, Default, Debug, Eq, PartialEq, Hash)]
+pub struct MortonRegion {
+    min: u32,
+    max: u32,
+}
+impl MortonRegion {
+    /// Create a new `MortonRegion` iterator.
```
Frizi: is this an iterator?

amethyst_tiles/src/iters.rs
```diff
@@ -0,0 +1,215 @@
...
+            morton_region.contains(morton::encode(c.x, c.y, c.z)),
+            region.contains(&c)
+        );
+
+        let c = Point3::new(128, 55, 565);
+        assert_eq!(
+            morton_region.contains(morton::encode(c.x, c.y, c.z)),
+            region.contains(&c)
+        );
+    }
```
Frizi: there should probably be tests that check for not containing a point

amethyst_tiles/src/iters.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// Create a new `MortonRegion` iterator.
+    pub fn new(min: u32, max: u32) -> Self {
+        Self { min, max }
+    }
+
+    /// Check if this `MortonRegion` contains a given morton coordinate.
+    #[inline]
+    pub fn contains(self, morton: u32) -> bool {
+        let m1 = morton::min(self.min, morton) != morton;
+        let m2 = morton::max(self.max, morton) != morton;
```
Frizi: I might misunderstand what's going on there, but for me it seems like the logic checks only the edge conditions, without handling the "inside" of the region.

amethyst_tiles/src/morton/mod.rs
```diff
@@ -0,0 +1,231 @@
...
+pub fn decode(morton: u32) -> (u32, u32, u32) {
+    morton_decode_lut(morton)
+}
+
+#[inline]
+pub fn min(morton1: u32, morton2: u32) -> u32 {
+    let d1 = decode(morton1);
+    let d2 = decode(morton2);
+
+    if d1.0 > d2.0 || d1.1 > d2.1 || d1.2 > d2.2 {
```
Frizi: Is this the logic you want for sure? This logic is constructed in such a way that `min` is not commutative, or even idempotent (think what happens when x is equal). I believe you wanted to chain the comparators instead of running them in parallel.

Really what you want is to implement an comparator and use it in both `min` and `max` to select the value.

```rust
#[inline]
pub fn cmp(morton1: u32, morton2: u32) -> Ordering {
	let (x1, y1, z1) = decode(morton1);
	let (x2, y2, z2) = decode(morton2);
	z1.cmp(z2).then(|| y1.cmp(y2)).then(|| x1.cmp(x2))
}
```

Or, if the (z, y, x) order is not necessary and (x, y, z) is good enough, then we can use the `Ord` on tuples (it chains the comparators from left to right).

```rust
#[inline]
pub fn cmp(morton1: u32, morton2: u32) -> Ordering {
	decode(morton1).cmp(decode(morton2))
}
```

amethyst_tiles/src/morton/mod.rs
```diff
@@ -0,0 +1,231 @@
...
+        morton1
+    }
+}
+
+#[inline]
+pub fn max(morton1: u32, morton2: u32) -> u32 {
+    let d1 = decode(morton1);
+    let d2 = decode(morton2);
+
+    if d1.0 < d2.0 || d1.1 < d2.1 || d1.2 < d2.2 {
```
Frizi: This is wrong in exactly the same way `min` is. See above.

amethyst_tiles/src/morton/mod.rs
```diff
@@ -0,0 +1,231 @@
...
+
+        (0..test_side).into_par_iter().for_each(|x| {
+            for y in 0..test_side {
+                for z in 0..test_side {
+                    let morton = morton_encode_lut(x, y, z);
+                    assert_lt!(morton, max);
+                }
+            }
+        });
+    }
```
Frizi: This file could definitely use some tests for `min`, `max`, and probably `cmp` if it's going to be added.

amethyst_tiles/src/morton/mod.rs
```diff
@@ -0,0 +1,231 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use more_asserts::*;
+    use rayon::prelude::*;
+
+    #[test]
+    fn morton_intr_decode_encode_match() {
```
Frizi: This test should probably be gated with `target_feature = "bmi2"`, but i'm not exactly sure. Is it strictly necessary to run this method, or is it just an compiler generated optimization?

amethyst_tiles/src/pass.rs
```diff
@@ -0,0 +1,444 @@
...
+}
+
+/// Default bounds that returns the entire tilemap
+#[derive(Default, Debug)]
+pub struct DrawTiles2DBoundsDefault;
+impl DrawTiles2DBounds for DrawTiles2DBoundsDefault {
+    fn bounds<T: Tile, E: CoordinateEncoder>(map: &TileMap<T, E>, world: &World) -> Region {
+        Region::new(
+            Point3::new(0, 0, 0),
+            Point3::from(*map.dimensions() - Vector3::new(1, 1, 1)),
```
Frizi: This method should probably just return a Vector, it is a copy type.

amethyst_tiles/src/pass.rs
```diff
@@ -0,0 +1,444 @@
...
+                tile_map.tile_dimensions().x as f32,
+                tile_map.tile_dimensions().y as f32,
+            ]
+            .into();
+
+            let region = Z::bounds(tile_map, world);
+            region
+                .iter()
+                .filter_map(|coord| {
+                    let tile = tile_map.get(&coord).unwrap();
```
Frizi: This unwrap assumes the bounds are always precise. Also, there is no need for so much nesting.

Should just be 
```rust
let tile = tile_map.get(&coord)?;
let sprite_number = tile.sprite(coord, world)?;
```

amethyst_tiles/src/pass.rs
```diff
@@ -0,0 +1,444 @@
...
+                    if let Some(sprite_number) = tile.sprite(coord, world) {
+                        let (batch_data, texture) = {
+                            let sprite_sheet = sprite_sheet_storage
+                                .get(tile_map.sprite_sheet.as_ref().unwrap())?;
```
Frizi: Another unwrap that can probably just be `?` instead.

amethyst_tiles/src/pass.rs
```diff
@@ -0,0 +1,444 @@
...
+    framebuffer_height: u32,
+    layouts: Vec<&B::DescriptorSetLayout>,
+) -> Result<(B::GraphicsPipeline, B::PipelineLayout), failure::Error> {
+    let pipeline_layout = unsafe {
+        factory
+            .device()
+            .create_pipeline_layout(layouts, None as Option<(_, _)>)
+    }?;
+
+    //let shader_vertex = unsafe { VERTEX.module(factory).unwrap() };
```
Frizi: commented code

amethyst_tiles/src/pass.rs
```diff
@@ -0,0 +1,444 @@
...
+            tilemap_args.map_transform = map_transform.into();
+            tilemap_args.sprite_dimensions = [
+                tile_map.tile_dimensions().x as f32,
+                tile_map.tile_dimensions().y as f32,
+            ]
+            .into();
+
+            let region = Z::bounds(tile_map, world);
+            region
+                .iter()
```
Frizi: This iteration order assumes certain depth order and this isn't exactly documented. What if you look at the tilemap from the back side? I believe the semitransparent layers would be rendered incorrectly.

I don't even want to begin thinking about multiple multilayer tilemaps passing through each other :P
Maybe we should just document some cases as "undefined behaviour" (or at least, undefined draw order) for now?

amethyst_tiles/src/pod.rs
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        let sprite = &sprite_sheet.sprites[sprite_number];
+
+        Some((
+            Self {
+                u_offset: [sprite.tex_coords.left, sprite.tex_coords.right].into(),
+                v_offset: [sprite.tex_coords.top, sprite.tex_coords.bottom].into(),
+                tint: tint.map_or([1.0; 4].into(), |t| {
+                    let (r, g, b, a) = t.0.into_components();
```
Frizi: we have `into_pod` for this already

amethyst_tiles/src/pod.rs
```diff
@@ -0,0 +1,114 @@
...
+                    [r, g, b, a].into()
+                }),
+                tile_coordinate: [
+                    tile_coordinate.x as f32,
```
Frizi: maybe those should be kept as integers actually? That would make it a `ivec3` type in glsl

amethyst_tiles/src/pass.rs
```diff
@@ -0,0 +1,444 @@
...
+                .with_shaders(shaders.raw()?)
+                .with_layout(&pipeline_layout)
+                .with_subpass(subpass)
+                .with_framebuffer_size(framebuffer_width, framebuffer_height)
+                .with_blend_targets(vec![pso::ColorBlendDesc(
+                    pso::ColorMask::ALL,
+                    pso::BlendState::PREMULTIPLIED_ALPHA,
+                )])
+                .with_depth_test(pso::DepthTest::On {
+                    fun: pso::Comparison::Less,
```
Frizi: I'm not sure if you want to write depth buffer when rendering sprites, transparency might get in the way. See the comment on region iterator in rendering pass.

amethyst_tiles/src/map.rs
```diff
@@ -0,0 +1,365 @@
...
+    }
+}
+
+/// Trait for providing access to an underlying storage type of a 3-dimensional Tile data. This is abstracted to provide
+/// for allowing more underlying storage types in the future beyond a flat array, such as networking, chunking, etc.
+pub trait Map {
+    /// The world-space (Amethyst) dimensions of a single tile in this map space (1x1x1). This is used to scale our
+    /// sprites to the world coordinate space. This should usually be the tile sprite dimensions. Beware, Z-size is taken
+    /// into consideration as well so you will usually want to se Z-size to 1.
+    fn tile_dimensions(&self) -> &Vector3<u32>;
```
Frizi: those should probably return copies. The nalgebra basic types are `Copy`. I don't have a strong opinion on this though, just worth thinking about maybe.

amethyst_tiles/src/map.rs
```diff
@@ -0,0 +1,365 @@
...
+pub struct TileMap<T: Tile, E: CoordinateEncoder = crate::MortonEncoder2D> {
+    pub(crate) origin: Point3<f32>,
+    pub(crate) tile_dimensions: Vector3<u32>,
+    pub(crate) dimensions: Vector3<u32>,
+    pub(crate) transform: Matrix4<f32>,
+
+    pub(crate) version: u64,
+
+    #[serde(skip)]
+    pub(crate) sprite_sheet: Option<Handle<SpriteSheet>>,
```
Frizi: If it's skipped for serialization, how is it supposed to be bound after?

amethyst_tiles/src/morton/mod.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    #[inline]
+    fn encode(&self, x: u32, y: u32, z: u32) -> Option<u32> {
+        use bitintr::Pdep;
+
+        #[cfg(debug_assertions)]
+        {
+            let check = u32::max_value() / 3;
+            if x > check || y > check || z > check {
+                return None;
```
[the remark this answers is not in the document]
jaynus: I put this under a debug flag to assert that the requested coordinates encode to a valid morton value. In practice (release) all code should be written to prevent this. In debug, this is something you would like to build against. 

I've changed this to a panic in debug, but I am leaving it as a debug check.

The idea here is eventually (tm) this will support 64-bit mortons as well.

amethyst_tiles/src/iters.rs
jaynus: There are, there are tests included which check that both MortonRegion and Region return false together.

```        let c = Point3::new(128, 55, 565);
        assert_eq!(
            morton_region.contains(morton::encode(c.x, c.y, c.z)),
            region.contains(&c)
        );
```

amethyst_tiles/src/iters.rs
jaynus: How so? This checks that the morton values are greater than the min, and less than the max.

amethyst_tiles/src/morton/mod.rs
jaynus: > I believe you wanted to chain the comparators instead of running them in parallel.

rust is lazy-evaluated - that means the comparisons run linearly, left to right, and fail after the first failure and are chained. I'm not sure what you are talking about here needing to use .then comparisons here.

amethyst_tiles/src/morton/mod.rs
jaynus: See above

amethyst_tiles/src/morton/mod.rs
jaynus: Tests added

amethyst_tiles/src/morton/mod.rs
jaynus: I actually dont need to feature gate these, as `bitintr` generates the correct code on unsupported platforms. I just feature gate it for `encode` because I'd like to use the LUT when it is not supported (its faster than non-bmi2 representation)

amethyst_tiles/src/pass.rs
jaynus: I don't underrstand what you mean. I can't represent a 3d bounding box with a Vector?

amethyst_tiles/src/pass.rs
jaynus: I don't want to softly fail for incorrect values, sprites or coordinates - I wanted to fail. Should I wrap those failures in `debug_assertions` gates?

amethyst_tiles/src/pass.rs
jaynus: see above about failures

amethyst_tiles/src/pass.rs
jaynus: Yes. I will document that tilemaps must be viewed from Z-up currently.

amethyst_tiles/src/pass.rs
jaynus: A sprite can be `None` for a tile, making it an empty tile.

amethyst_tiles/src/pod.rs
jaynus: Which are you talking about? There is no `into_pod` for `Tint`, this is copied from the other `Pod` structures that use tint - they map these same way. See SkinnedVertexArgs and others.

amethyst_tiles/src/pod.rs
jaynus: My age was showing.
"In short modern GPUs have equivalent INT and FP performance for 32bit data. So your logical operations will run at the same speed."

I didn't know this was the case now.

amethyst_tiles/src/pass.rs
jaynus: The tile map specifically supports transparent alpha channel tiles. I have been using it without issue, except for the drawing artifacts discussed previously. It seems to work fine, lets discuss offline more.

amethyst_tiles/src/map.rs
jaynus: Yes, performance wise its similar. I determined to stick with references due to their immutable nature - i wanted the return value to be default-immutable and force a user copy if they want to modify it.

amethyst_tiles/src/map.rs
jaynus: I don't know. I didn't design the `Prefab` and `Asset` system, its crap. How is this supposed to happen?

amethyst_tiles/src/morton/mod.rs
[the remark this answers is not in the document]
Frizi: seems good 👍

amethyst_tiles/src/iters.rs
Frizi: Not really (I think). The test is a bit fuzzy on boundaries. Imagine (or write a test for) what happens when only one of the coordinates is actually different. Also, do you want the range to be inclusive on one side, both, or neither? Because currently it's kinda weird about it. I think it's sort of inclusive on some axes, but not really on others. I think all region corners should also be tested for being inclusive/exclusive.

I have a bit hard time thinking about this in combination with the comparators. Writing it as manually decoding values directly in that method might make it much cleaner.

amethyst_tiles/src/iters.rs
Frizi: Oh, i see. This is very not clear, i'd prefer to have the `assert_eq` written directly against `true` and `false` values. That would be just easier to read. Also the tests don't check any boundary values.

amethyst_tiles/src/pass.rs
Frizi: No, i meant that `dimensions` call requires a dereference. But after thinking about it, this is probably preferred as is. So ignore that :)

amethyst_tiles/src/pass.rs
Frizi: Empty tiles are exactly meant to return `None` here and be filtered out from rendering, so that checks out. Requiring that `bounds` call is always precise might be a stretch though. I can easily imagine a situation where i just use camera data to calculate that culling, I'd like to be not bothered about always checking the size of map as well. The renderer should calculate the region intersection before iterating, it's super low cost operation that improves developer experience.

amethyst_tiles/src/iters.rs
jaynus: You're correct actually. Per tests, my method was incorrect. 

I've fallen back on decoding and comparing for ease of reading. It could be optimized but its minimal.

amethyst_tiles/src/pass.rs
Frizi: If the tile can never be `None`, why is that field an option in the first place?

amethyst_tiles/src/pod.rs
Frizi: There is `into_pod` for `palette::Srgba`.

amethyst_tiles/src/map.rs
Frizi: Yeah, let's leave it as is now.

amethyst_tiles/src/pass.rs
jaynus: Done

amethyst_tiles/src/pod.rs
jaynus: Done, fixed in rendy too.

amethyst_tiles/src/map.rs
jaynus: Done.

amethyst_tiles/src/iters.rs
jaynus: Boundaries were added in a new set of tests.

amethyst_tiles/src/pass.rs
```diff
@@ -0,0 +1,451 @@
...
+            .into();
+
+            let region = Z::bounds(tile_map, world);
+            region
+                .iter()
+                .filter_map(|coord| {
+                    let tile = tile_map.get(&coord).unwrap();
+                    if let Some(sprite_number) = tile.sprite(coord, world) {
+                        let (batch_data, texture) = {
+                            if let Some(sprite_sheet_handle) = tile_map.sprite_sheet.as_ref() {
```
Frizi: this whole `sprite_sheet_handle` dance here can be extracted to the outer loop, just below `for (tile_map, _, transform)`.

```rust
let maybe_sheet = tile_map
	.sprite_sheet.map(|handle| sprite_sheet_storage.get(handle))
	.filter(|sheet| tex_storage.contains(&sheet.texture));
	
let sprite_sheet = match maybe_sheet {
  Some(sheet) => sheet,
  None => continue,
}
```

amethyst_tiles/src/pod.rs
```diff
@@ -0,0 +1,107 @@
...
+
+impl TileArgs {
+    #[allow(clippy::cast_precision_loss)]
+    /// Extracts POD vertex data from the provided storages for a sprite.
+    ///
+    /// # Arguments
+    /// * `tex_storage` - `Texture` Storage
+    /// * `sprite_storage` - `SpriteSheet` Storage
+    /// * `sprite_render` - `SpriteRender` component reference
+    /// * `transform` - 'Transform' component reference
```
azriel91: could either correct / remove

docs/CHANGELOG.md
```diff
@@ -28,7 +28,6 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 * `Camera::screen_ray` method added, returning an appropriate `Ray` structure ([#1918]).
 * `amethyst_test`: `InMemorySource` and `WaitForLoad` helpers ([#1933]).
 
-### Changed
```
azriel91: bring me back! (`Ctrl D,S` probably did this)

azriel91:
bors r=Frizi,azriel91

PR #1837, opened 2019-08-01
[WIP] raw Render Graph creation

AndreaCatania:
## Description

Related to this PR #1772.

@Frizi Your work to make the Render Graph creation more easy is really nice and I really like this idea. The only problem that I find is that it doesn't allow (anymore) to create a custom render graph.

This removes completely the possibility to deeply customize the rendering and control the pipeline, decreasing a lot the Flexibility of this engine.

I made this PR to make it possible to also define a custom render graph (as before) in order to take advantage of the two solutions.

~~Actually the code that I written allow to specify a custom render graph, but since you moved part of the code from the RenderingSystem inside the plugins is no more possible to easily create a render graph.
For this reason I'm asking your help, to find a solution and make this engine flexible as before.~~

## Modifications
- Renamed `TextureProcessor` to `TextureProcessorSystem` and `MeshProcessor` to `MeshProcessorSystem` to unify `System`s naming.
- Renamed `RenderingBundle` to `PluggableRenderingBundle` and `RenderingBundleGraphCreator` to `PluggableRenderGraphCreator` to better specify the new functionality.
- Added `RenderingBundle` which accept a raw `RenderGraph` that allow more customization of the `Rendering`.

## PR Checklist

### DIFF
--- a/amethyst_rendy/src/bundle.rs
+++ b/amethyst_rendy/src/bundle.rs
@@ -11,7 +11,7 @@ use crate::{
         hal,
         wsi::Surface,
     },
-    system::{GraphCreator, MeshProcessor, RenderingSystem, TextureProcessor},
+    system::{GraphCreator, MeshProcessorSystem, RenderingSystem, TextureProcessorSystem},
     types::Backend,
     SpriteSheet,
 };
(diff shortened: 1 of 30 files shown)


AndreaCatania:
Closed in favor of #1839.
@Frizi, I was wrong because the plugins have its own implementation so your support is not needed :)

PR #1839, opened 2019-08-01
Added the possibility to use a custom GraphCreator easily

AndreaCatania:
## Description

This PR has the aim to make the already amazing @Frizi work (#1772) even more better by allowing to use a custom `GraphCreator` easily in all games that need to interact directly with the `Pipeline`.

## Modifications
- Renamed `TextureProcessor` to `TextureProcessorSystem` and `MeshProcessor` to `MeshProcessorSystem` to unify it with other `System`s name.
- Exposed `TextureProcessorSystem` and `MeshProcessorSystem`.
- Renamed `RenderingBundle` to `PluggableRenderingBundle` and `RenderingBundleGraphCreator` to `PluggableRenderGraphCreator` to better specify this new functionality.
- Added `RenderingBundle` which accept a custom `RenderGraph` that allow more customization of the `Rendering` pipeline.

## PR Checklist


jondo2010:
New/modified example showing custom graph creation?

amethyst_rendy/src/bundle.rs
```diff
@@ -23,18 +23,69 @@ use amethyst_core::{
 use amethyst_error::{format_err, Error};
 use std::collections::HashMap;
 
+/// A bundle of systems used for rendering using `Rendy` render graph.
+///
+/// Is highly recommended to use `PluggableRenderingBundle` which is way more easy to construct and manage.
```
OvermindDL1: Grammar:
```
It's highly recommended to use `PluggableRenderingBundle` as it is easier to construct and manage at the loss of advanced manual handling.
```
Or something like that.

amethyst_rendy/src/bundle.rs
```diff
@@ -23,18 +23,69 @@ use amethyst_core::{
...
+///
+/// Accepts any `GraphCreator`, This bundle allows to deeply customize the rendering process,
```
OvermindDL1: Grammar:
```
This allows for fully customizable rendering processing by passing in your own `GraphCreator`.
```

amethyst_rendy/src/bundle.rs
```diff
@@ -23,18 +23,69 @@ use amethyst_core::{
...
+    G: GraphCreator<B>,
+{
+    graph_creator: G,
+    phantom_data: std::marker::PhantomData<B>,
+}
+
+impl<B: Backend, G: GraphCreator<B>> RenderingBundle<B, G> {
+    /// Creates a new `RenderingBundle` with a custom `GraphCreator`.
+    ///
+    /// Is highly recommended to use `PluggableRenderingBundle` which is way more easy to construct and manage.
```
OvermindDL1: Etc... as above.  Among others.

docs/CHANGELOG.md
```diff
@@ -14,8 +14,14 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 
 ### Added
 
+* `RenderingBundle` Now allow to easily use a custom `GraphCreator` ([#1839)].
```
OvermindDL1: Grammar:
```
* `RenderingBundle` for full manual control of the rendering pipeline via a custom `GraphCreator` ([#1839]).
```

AndreaCatania:
@jondo2010 Added (pong_tutorial_06)
@OvermindDL1 Fixed

Thanks to both of you

amethyst_rendy/src/bundle.rs
```diff
@@ -23,18 +23,64 @@ use amethyst_core::{
 use amethyst_error::{format_err, Error};
 use std::collections::HashMap;
 
+/// A bundle of systems used for rendering using `Rendy` render graph.
+///
+/// It's highly recommended to use `PluggableRenderingBundle` as it is easier to construct and manage
+/// at the loss of advanced manual handling.
```
jojolepro: Had to re-read this sentence 2-3 times.
Can it be replaced by: "It is highly recommended to use `PluggableRenderingBundle`, as it is easier to construct and manage.

However, at the opposite of `PluggableRenderingBundle`, this (`RenderingBundle`) allows for fully customizable ren......"

amethyst_rendy/src/bundle.rs
```diff
@@ -23,18 +23,64 @@ use amethyst_core::{
...
+        builder.add(MeshProcessorSystem::<B>::default(), "mesh_processor", &[]);
+        builder.add(
+            TextureProcessorSystem::<B>::default(),
+            "texture_processor",
+            &[],
+        );
+        builder.add(Processor::<Material>::new(), "material_processor", &[]);
+
+        builder.add_thread_local(RenderingSystem::<B, G>::new(self.graph_creator));
+        Ok(())
```
jojolepro: Did you forget to add the SpriteSheet processor before the RenderingSystem?
Also, it seems like you forgot to add the barrier.

Would it be possible to put those 4 systems (the 3 you defined + SpriteSheet) in a private/local small bundle called `RenderingProcessorBundle` or something similar?

docs/CHANGELOG.md
```diff
@@ -14,8 +14,14 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
...
 ### Added
 
+* `RenderingBundle` for full manual control of the rendering pipeline via a custom `GraphCreator` ([#1839]).
+
 ### Changed
 
+* `TextureProcessor` renamed to `TextureProcessorSystem` ([#1839)].
+* `MeshProcessor` renamed to `MeshProcessorSystem` ([#1839)].
+* `RenderingBundle` renamed to `PluggableRenderingBundle` ([#1839)].
+
```
jojolepro: Please add the link to this pull request a bit under the `Fixed` title. See the previous release changelog for an example :)

examples/pong_tutorial_06/main.rs
```diff
@@ -0,0 +1,172 @@
...
+    Ok(())
+}
+
+// This graph structure is used for creating a proper `RenderGraph` for rendering.
+// A renderGraph can be thought of as the stages during a render pass. In our case,
+// we are only executing one subpass (DrawFlat2D, or the sprite pass). This graph
+// also needs to be rebuilt whenever the window is resized, so the boilerplate code
+// for that operation is also here.
+#[derive(Default)]
+struct ExampleGraph {
```
jojolepro: In my opinion, it would have been better to have this in a minimal example (of its own) based on the `renderable` example. This way, it would be easier to find the graph code.

jojolepro: Also, usually we reserve pong_tutorial_x entries for when we have a corresponding book tutorial page.

amethyst_rendy/src/bundle.rs
AndreaCatania: > Did you forget to add the SpriteSheet processor before the RenderingSystem?

Before, you had to declare the SpriteSheet processor explicitly; it was good because the 3D games doesn't need this system most of the times.

However on a second thought, following the above reasoning I should remove also the `MeshProcessorSystem`.
What do you think about this?

> Also, it seems like you forgot to add the barrier.

In the PluggableRenderingBundle there is a mechanism that allow the plugins to initialize some `System`s. These `System`s must always be executed after the one defined in the bundle.

Since this mechanism is not used with RenderingBundle, and it's meant to give 100% freedom to the developer is not necessary the barrier.

> Would it be possible to put those 4 systems (the 3 you defined + SpriteSheet) in a private/local small bundle called RenderingProcessorBundle or something similar?

Do you mean move `RenderingBundle` in another module and remove `builder.add_thread_local(RenderingSystem::<B, G>::new(self.graph_creator));` so it must be defined explicitly during the dispatcher creation?

AndreaCatania:
@jojolepro Removed `pong_tutorial_06` in favor of `renderable_custom` which is an example based on `renderable` as you recommended.
Improved the docs.
Let me know how you want to proceed on this: https://github.com/amethyst/amethyst/pull/1839#discussion_r310097021

**Note:**
The check failed, but seems a Jenkins internal error.

Frizi:
I'm not sure where you are really comming from with this @AndreaCatania. The #1772 feature was pretty much additive (except for asset processors being separate). You are totally free to not use RenderingBundle, add the systems manually and specify a custom graph creator. Every usecase that you wanted to keep supporting is indeed already supported.

I don't like the name change as well. `PluggableRenderingBundle` is a mouthful for something that is supposed to be the default. The `RenderingBundle` you are proposing doesn't really do anything except adding three systems, that as you noted should be handled manually in most cases when you want to go custom.

AndreaCatania:
> I'm not sure where you are really comming from with this @AndreaCatania. The #1772 feature was pretty much additive (except for asset processors being separate). You are totally free to not use RenderingBundle, add the systems manually and specify a custom graph creator. Every usecase that you wanted to keep supporting is indeed already supported.

The initial idea was to just expose `TextureProcessor` and `MeshProcessor`. These was not exposed and I was not able to create a render graph.

But then I realized that most part of these systems (like the texture processor) are always used. From this, I got the idea to follow you and add a bundle even to use a "custom" `RenderGraph`.

> I don't like the name change as well. `PluggableRenderingBundle` is a mouthful for something that is supposed to be the default. The `RenderingBundle` you are proposing doesn't really do anything except adding three systems, that as you noted should be handled manually in most cases when you want to go custom.

Revert this name `PluggableRenderingBundle` to `RenderingBundle` is ok to me, the only problem is that if we want to keep the `Bundle` idea; the most correct naming solution to me seems the actual one.

I'm open to discuss it.

Frizi:
If you really want a bundle, i'd opt for a `ManualRenderBundle` or something like that. But really the preferred solution for me would be to just make the processor systems public, also with possible name change for those to keep things consistent. The reason to not have another bundle is to limit the number of possible "supported" configurations, as maintaining all of those takes a lot of time and effort in practice. Being able to use "bare bones" config in the "do it yourself" manner gives you all the possibilities, while also shifting some of that maintenance burden to the user (probably a bit more frequent breaking changes, but shouldn't be very problematic). It may sounds bad at first, but I think it's actually desirable. User is able to maintain their specific config only, while the engine would have to support all weird permutations at the same time.

AndreaCatania:
Considering that most likely is not desirable want the `MeshProcessorSystem` for 2D games, it should never land inside the bundle.
The same reasoning can be also applied to the other `System`s, since when you deal with a custom render graph you never know what you will need or you will not need beforehand.

So I agree with you.
-Removed the `Bundle` and reverted the name of the bundle create by you.
- Changed the example accordingly.
- Added in the doc of the `RenderingBundle` that is also possible to define a custom `RenderGraph`.

amethyst_rendy/src/bundle.rs
AndreaCatania: Solved by removing the `RenderingBundle` as discussed here: https://github.com/amethyst/amethyst/pull/1839#issuecomment-518305303

Frizi:
> Considering that most likely is not desirable want the MeshProcessorSystem for 2D games, it should never land inside the bundle.

I agree, but we do miss some features in specs/shred to auto-enable it only when needed. Having it in bundle is the least surprising option right now. The system is pretty much free when there are no assets to process, so it doesn't really hurt either. I.e. it's fine to keep it in the bundle for nowas the cost of that is negligible. This will change once we have a way to insert a system without duplicates in specs, i.e. `insert_if_not_present` or something like that.

docs/CHANGELOG.md
```diff
@@ -20,9 +20,14 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 
 ### Changed
 
+* `TextureProcessor` renamed to `TextureProcessorSystem` ([#1839]).
+* `MeshProcessor` renamed to `MeshProcessorSystem` ([#1839]).
+* `RenderingBundle` renamed to `PluggableRenderingBundle` ([#1839]).
```
Frizi: this is not true, probably should be about `PluggableRenderGraphCreator`.

AndreaCatania:
@Frizi Removed Changelog entry
Rebased and fixed a problem with the function `with_thread_local` that was asking for a `System` rather then a `RunNow`.

AndreaCatania:
Rebased again;
@Frizi I hope to hear you soon, to avoid rebase it endless.

Frizi:
@AndreaCatania I was offline for a couple of days, sorry for the delay. Yeah, it does seem good to me. Unfortunately one more rebase is needed :)

AndreaCatania:
@Frizi It's fine. However, just rebased.

AndreaCatania:
@Frizi I'm not sure where make sense add some unit test to make the CI pass. What's your input on this?

PR #1843, opened 2019-08-02
Use derive_new on amethyst_controls

iamsauravsharma:
## Description

Use derive_new crates for amethyst_controls crate

## PR Checklist

- [X] Ran `cargo clippy --all --features "empty"`
- [ ] Ran `cargo test --all --features "empty"

### DIFF
--- a/amethyst_controls/src/bundles.rs
+++ b/amethyst_controls/src/bundles.rs
@@ -6,6 +6,8 @@ use amethyst_input::BindingTypes;
 
 use super::*;
 
+use derive_new::new;
+
 /// The bundle that creates a flying movement system.
 ///
 /// Note: Will not actually create a moving entity. It will only register the needed resources and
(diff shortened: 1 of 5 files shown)


amethyst_controls/src/bundles.rs
```diff
@@ -27,33 +29,20 @@ use super::*;
...
 /// * `MouseFocusUpdateSystem`
 /// * `CursorHideSystem`
-#[derive(Debug)]
+#[derive(new, Debug)]
 pub struct FlyControlBundle<T: BindingTypes> {
+    #[new(value = "1.0")]
     sensitivity_x: f32,
+    #[new(value = "1.0")]
     sensitivity_y: f32,
+    #[new(value = "one()")]
```
jojolepro: This should be 1.0 too actually. It seems we forgot to change it following the removal of the `Float` type.

amethyst_controls/src/resources.rs
```diff
@@ -1,20 +1,16 @@
 use serde::{Deserialize, Serialize};
 
+use derive_new::new;
```
jojolepro: Instead of using an `use` statement in each file, you can use the following in lib.rs:

```
#[macro_use]
extern crate derive_new;
```

This way, the macro will be globally exported for the amethyst_control crate :)

iamsauravsharma:
Closing this PR since #1780 have already added half of derive_new macro in crates and change lots of code syntax and I don't have time for analyzing code and changing it anymore soon so anyone can take and work in it.

Issue #1838, opened 2019-08-01
[MAINTENANCE] Use derive-new, derive_builder and getset macros


jojolepro:
## Note

We have a bunch of code that is basically copy pasted, and its a pain to maintain.

First, some facts!
```
grep -r "fn new() " ./  | wc -l
69
grep -r "fn get_" ./  | wc -l
30
grep -r "fn set_" ./  | wc -l
77
egrep -r "fn [a-Z]*\(&self\) ->" ./  | wc -l
134
```

The goal is to get rid of this kind of code
![image](https://user-images.githubusercontent.com/4755185/62293505-c520b700-b436-11e9-98d2-f743bf0f7f90.png)

and this one
![image](https://user-images.githubusercontent.com/4755185/62295171-7b39d000-b43a-11e9-9836-396276d2a969.png)

While I'm already planning on splitting UiTransform in a bunch of components, it would be easier to edit the code if we were using a macro for the new() method, as I would only have to remove the field I'm editing from the struct definition instead of in:

- struct def
- new method signature
- new method body
- accessor ( field_name() method)
- setter ( set_field_name() method)
- sometimes get_mut method

Using both the derive_new macro and the getset macros can help cut down all those methods and instead have

- struct definition
- `#[new(default)] (if needed)`
- `#[get] #[set] #[get_mut]`

So instead of 1+1+1+3+3+(3) = 12 lines per field, in the optimal case we would have 2 lines per field (for public ones).

If we count the mandatory method documentation, we can cut another 3 lines per field and 1 line per struct.

For the derive-builder macro, this is a bit more complex and depends much more on how the struct is defined and used, so it will be looked at on a per-case basis to decide if it is worth replacing a builder by a macro.

## Description

- Add at least the derive-new and getset crates to all amethyst crates
- Make the macros globally accessible in the crates
- Replace ALL `new() -> Self` methods by the `#[derive(new)]` macro
- Use `#[new(default)]` on fields that are currently not in the new() signature but are used by the struct
- Evaluate which Builder structs could be replaced by the derive_builder macro
- Remove ALL get/get_mut/set methods that are not doing anything special and replace them by the corresponding getset macros

## Reason

- Less code, which means faster reviews and easier reading.
- Easier to maintain.
- Lowers modification/editing time significantly.
- Lowers addition time significantly.
- Improves doc consistency (generated docs are always the same for get/get_mut/set methods)

## Impact

[...]

PR #1848, opened 2019-08-03
Use derive_new and getset macros for amethyst_rendy

projecteulerrs:
## Description

As per issue ([#1838]), new methods and most getters and setters were changed to be generated from macros. Getters that did not return a reference were not replaced, since the getset crate doesn't support that yet.

## Modifications

- PlanContext::graph() was renamed to PlanContext::graph_mut(); it returned a mutable reference
- Some private fields were renamed in order to keep the getter and setter methods the same name as before

## PR Checklist


docs/CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 
 ### Changed
 
+* PlanContext::graph() renamed to PlanContext::graph_mut()
```
jojolepro: Missing the link to this pull request. See previous entries for an example. (also don't forget to include the link, and not just the pr number).

projecteulerrs: Ah I missed that thanks. It should be okay now

jojolepro:
bors r+

Thanks!

jojolepro:
Complaining about missing docs in amethyst_rendy.

projecteulerrs:
Missing docs should be fixed. Sorry about that

projecteulerrs:
Not entirely sure why jenkins is getting a linking error :(

amethyst_rendy/src/bundle.rs
```diff
@@ -247,13 +235,14 @@ pub struct TargetMetadata {
...
 
-#[derive(Debug)]
+#[derive(Debug, MutGetters)]
 struct PlanContext<B: Backend> {
     targets: HashMap<Target, TargetPlan<B>>,
     target_metadata: HashMap<Target, TargetMetadata>,
     passes: HashMap<Target, EvaluationState>,
     outputs: HashMap<TargetImage, ImageId>,
-    graph_builder: GraphBuilder<B, World>,
+    #[get_mut = "pub"]
```
Frizi: how about keeping the name `graph` instead of `graph_mut` here? Getting the graph builder for readonly access doesn't make much sense in practice. I'd rather keep the `PlanContext` API consistent with `TargetPlanContext`. It is currently not public, but very likely will be very soon.

amethyst_rendy/src/lib.rs
```diff
@@ -48,6 +48,12 @@
 #[macro_use]
 extern crate amethyst_derive;
 
+#[macro_use]
+extern crate derive_new;
```
Frizi: This should not be required. We use rust 2018, just import the macro where you use it.

Areredify: I thought line "make the macros globally accessible in the crates" was about this, is it not?

projecteulerrs: The reasoning behind this was to make it globally accessible in the crate. I think this was mentioned in PR #1843 where it was suggested to switch from importing the macro in each file to this method. I can change it back though if that's what the amethyst team prefers.

amethyst_rendy/src/bundle.rs
projecteulerrs: Would it make more sense to change `TargetPlanContext::graph()` to `TargetPlanContext::graph_mut()` then? getset doesn't support custom names (aside from changing `graph_mut` to `mut_graph`) so `graph` would have to be written by hand again to keep the name `graph` while returning a mutable reference.

amethyst_rendy/src/lib.rs
Frizi: Well, I guess this is not the right place to complain about it then. I generally prefer to not have `extern crate`, but it already happened and we might need a separate pass to discuss and deal with this globally. So I'd say keep it for now.

amethyst_rendy/src/bundle.rs
distransient: Although getting something immutably may not be useful, I think it'd be best to preserve convention and make all accessors uniform

PR #1855, opened 2019-08-06
Use derive_new and getset macros for amethyst_animation

Areredify:
## Description
As per issue ([#1838]), all new methods (except for one that's not completely trivial) and all getters (there was only one) are replaced with corresponding macros.

## Modifications

## PR Checklist


Areredify:
Oops, I'll fix those conflicts in a minute

PR #1857, opened 2019-08-06
Use derive_new and getset macros for amethyst_utils

Areredify:
## Description

As per issue ([#1838]), new/get/set are replaced with corresponding macros.

## PR Checklist


Areredify:
Is this something I did or?

amethyst_utils/src/auto_fov.rs
```diff
@@ -19,35 +19,38 @@ use serde::{Deserialize, Serialize};
...
 
+    /// Sets `fovx_growth_rate` to the given value
+    #[set = "pub"]
     /// The factor determining how sensitive the FOV change should be
+    #[new(default)]
     fovx_growth_rate: f32,
 
     /// If the FOV grow rate specified in the field `fovx_growth_rate` should be applied as-is
+    #[new(default)]
     fixed_growth_rate: bool,
```
jojolepro: I think you forgot to add the get/set macros for those?

amethyst_utils/src/auto_fov.rs
```diff
@@ -19,35 +19,38 @@ use serde::{Deserialize, Serialize};
...
 
     /// The aspect ratio when the camera's horizontal FOV is identical to `base_fovx`
+    #[new(default)]
     base_aspect_ratio: (usize, usize),
 
     /// The minimum value the horizontal FOV can have
+    #[new(default)]
     min_fovx: f32,
```
jojolepro: If you look at the Default implementation, you'll see that there's default non-zero values for those. :)

amethyst_utils/src/auto_fov.rs
Areredify: They have non-trivial setters and getters that return by value, so I can't add macros for those

amethyst_utils/src/auto_fov.rs
Areredify: Yeah, my bad

PR #1869, opened 2019-08-10
Maintenance/fix rendy example

azriel91:
## Description

Fixes `rendy` example to compile. There is still a runtime issue, but haven't got the time to investigate.

## PR Checklist

- **n/a** Ran `cargo clippy --all --features "empty"`
- **n/a** Ran `cargo test --all --features "empty"`

### DIFF
--- a/examples/rendy/main.rs
+++ b/examples/rendy/main.rs
@@ -13,13 +13,13 @@ use amethyst::{
     core::{
         ecs::{
             Component, DenseVecStorage, DispatcherBuilder, Entities, Entity, Join, Read,
-            ReadExpect, ReadStorage, System, SystemData, World, Write, WriteStorage,
+            ReadStorage, System, SystemData, World, Write, WriteStorage,
         },
         math::{Unit, UnitQuaternion, Vector3},
-        SystemDesc, Time, Transform, TransformBundle,
+        Time, Transform, TransformBundle,
     },
     error::Error,
-    gltf::GltfSceneLoaderSystem,
+    gltf::GltfSceneLoaderSystemDesc,
     input::{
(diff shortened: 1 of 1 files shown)


azriel91:
Hacked Amethyst a bit to see what's wrong in deserialization:

```rust
// amethyst_assets/src/dyn_format.rs
// SeqLookupVisitor::visit_str
eprintln!("key: {}", key);
self.registry
    .map
    .keys()
    .for_each(|key| eprintln!("registry key: {:?}", key));
if self.registry.map.is_empty() {
    eprintln!("No keys in registry.");
}
```

Output:

```bash
Runnin g `target/debug/examples/rendy`
key: IMAGE
No keys in registry.
```

I think the failure is due to the following (same root issue):

* https://github.com/dtolnay/inventory/issues/9
* https://github.com/dtolnay/inventory/issues/7
* https://github.com/rust-lang/rust/issues/47384

cc: @valkum

examples/rendy/main.rs
```diff
@@ -594,16 +594,27 @@ fn main() -> amethyst::Result<()> {
     )?;
 
     let game_data = GameDataBuilder::default()
+        .with_bundle(
+            RenderingBundle::<DefaultBackend>::new()
```
Frizi: This is not right. `RenderingBundle` must always be added at the end, because it adds a barrier at the beginning and expects all game data to be already written. This change adds another frame of delay to the rendering. What should happen instead is that systems should not rely on `Window` being initialized at the setup time. I've made sure of that while introducing RenderingBundle before, in fact it worked like that before specs update. Please make sure that some changes weren't overwritten in that process.

azriel91:
Superseded by #1881.

PR #1870, opened 2019-08-10
Exposed dispatcher operation trait

AndreaCatania:
## Description

I need to construct a `Dispatcher` composed by many parts. So I need, to collect all the `Systems` that compose it, and then I can build the `Dispatcher`.

To create such a thing, I need something similar to the deferred operations that @jojolepro already worked on.

In order to prevent code duplication, my idea is to move it inside the amethyst_core.
Moving this inside the core, allow to use this mechanism even with a custom GameData, and will reduce code duplication even more.

For this reasons, I'm moving it inside the core, in this way these are exposed and more easy to use.

## Modifications

- Moved the `DispatcherOperation` and all related to it, inside the `amethyst_core`.

## PR Checklist

### DIFF
--- a/amethyst_core/src/lib.rs
+++ b/amethyst_core/src/lib.rs
@@ -40,10 +40,11 @@ pub use self::{
     hidden::{Hidden, HiddenPropagate},
     hide_system::HideHierarchySystem,
     named::{Named, WithNamed},
-    system_desc::SystemDesc,
+    system_desc::{RunNowDesc, SystemDesc},
 };
 
 pub mod bundle;
+pub mod deferred_dispatcher_operation;
 pub mod frame_limiter;
 pub mod timing;
 pub mod transform;
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -19,6 +19,8 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 * `SystemDesc` proc macro derive to simplify defining `SystemDesc`s. ([#1780])
 * `UiButtonData` is now exported from `amethyst_ui` and can be used for custom widgets. ([#1859])
 * Add an audio subchapter to the pong chapter. ([#1842])
+* Add `DispatcherOperation` to store dispatcher build logic, which can be executed lazily. ([#1870])
+
 
 ### Changed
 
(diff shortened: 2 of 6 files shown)


src/game_data.rs
```diff
@@ -1,9 +1,8 @@
 use std::marker::PhantomData;
 
-use derivative::Derivative;
-
 use crate::{
     core::{
+        deferred_dispatcher_operation::*,
```
azriel91: Nit: could you list out the imports explicitly?

docs/CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 * `SystemDesc` proc macro derive to simplify defining `SystemDesc`s. ([#1780])
 * `UiButtonData` is now exported from `amethyst_ui` and can be used for custom widgets. ([#1859])
 * Add an audio subchapter to the pong chapter. ([#1842])
+* Add `DispatcherOperation` that can be used to collect `Systems` in a vector, and in case construct a `Dispatcher` later. ([#1870])
```
azriel91: perhaps something more concise like:

```md
* Add `DispatcherOperation` to store dispatcher build logic, which can be executed lazily. ([#1870])
```

amethyst_core/src/lib.rs
```diff
@@ -44,6 +44,7 @@ pub use self::{
 };
 
 pub mod bundle;
+pub mod deferred_dispatcher_operation;
```
azriel91: would prefer the `deferred_` prefix to not be there (am pedantic like that :upside_down_face:)

azriel91:
bors try

AndreaCatania:
bors try

amethyst_core/src/lib.rs
AndreaCatania: I've asked @jojolepro to understand the correct naming, so I suppose that `dispatcher_operation.rs` would make for sense for you.

But I don't like it, it give me the impression that is meaningless. And by the way I don't like the actual one either.

Do you have any other idea?

azriel91: i named the enum so :stuck_out_tongue:
i don't mind that much / can let it go :snowflake:

AndreaCatania:
@azriel91 Changed what you pointed out.
Fixed a problem with `with_thread_local` function that was asking for `System` instead of `RunNow`

AndreaCatania:
bors, your Jenkins friend is making moods... again... please try

jojolepro:
bors try

azriel91:
bors r=jojolepro,azriel91

AndreaCatania:
Thanks guys!

PR #1879, opened 2019-08-14
Small fixes to the pong tutorial

### LINKED ISSUE #1873: [MAINTENANCE] Pong tutorial mismatch with amethyst_cli
## Description

The recommended approach inside of the Getting Started section of the book is to use amethyst_cli to create a project skeleton (https://book.amethyst.rs/stable/getting-started.html).  The later tutorials then build on this skeleton.  However, amethyst_cli now places display_config.ron into resources/display_config.ron instead of config/display_config.ron, so the tutorial code doesn't work without amending.

## Reason

It's pretty trivial to notice and correct yourself, but also a simple change would eliminate the need.  I'm happy to PR the changes if someone more experienced can weigh in and tell me exactly which part of this actually needs changing.

## Impact

I don't see why it would, since it's just documentation?

Adrijaned:
## Description

Fixes #1873

## Additions

Only doc change

## Removals

Only doc change

## Modifications

Only doc change

## PR Checklist

Not applicable


book/src/pong-tutorial/pong-tutorial-01.md
```diff
@@ -125,10 +125,10 @@ from a file. The latter approach is handier, as it allows us to change configura
 (e.g, the window size) without having to recompile our game every time.
 
 Starting the project with `amethyst new` should have automatically generated
-`DisplayConfig` data in `config/display.ron`. If you created the
```
jojolepro: It is stored in the config folder though?

Adrijaned: Ooh, a new commit in the tools repo. In that case, it would be nice if you could push the new version on crates.io. That should also resolve #1873 too.

azriel91:
oh btw, what changed in the pong sprite sheet?

Adrijaned:
Changed background from black to transparent. Especially since the tutorial encourages you to change the background color of the window, you should be able to use the assets even on non black windows.

azriel91:
ah, that's cool

bors r=jojolepro,azriel91

PR #1877, opened 2019-08-12
[WIP] Amethyst Renderer WASM Support (OpenGL)

omni-viral:
## Description

Adds OpenGL/WebGL support

PR #1888, opened 2019-08-17
Upgrade winit to 0.20.0-alpha3

### LINKED ISSUE #1786: [MAINTENANCE] Update to winit 0.20.0
## Description

Since `gfx-hal` is updating -- https://github.com/gfx-rs/gfx/issues/2874 -- we should do this as well.

## Reason

Keep up to date with the ecosystem.

Jaspi10:
Decided to check how complex the changes would be. Turns out it wasn't that complicated.

## Description

This fixes #1786 

## Modifications

As winit is reexported as amethyst::winit all internal API changes in winit are also breaking changes for amethyst. For example in some examples i had to change the from `use winit::` to `use::winit::event::`.  

Also `winit::event::Event` is now generic over a user defined Event type. I just set it to `Event<()>` everywhere, as I don't see a reason why one should inject custom winit events instead of using their own Event channel.

## PR Checklist

### DIFF
--- a/amethyst_window/src/bundle.rs
+++ b/amethyst_window/src/bundle.rs
@@ -1,8 +1,16 @@
-use crate::{DisplayConfig, EventsLoopSystem, WindowSystem};
+use crate::{DisplayConfig, EventLoopSystem, WindowSystem};
 use amethyst_config::Config;
 use amethyst_core::{bundle::SystemBundle, ecs::World, shred::DispatcherBuilder};
 use amethyst_error::Error;
-use winit::EventsLoop;
+use std::{
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        mpsc::{self, Receiver, Sender},
+        Arc, Mutex,
+    },
+    thread,
+};
+use winit::{event::Event, event_loop::EventLoop, window::Window};
 
 /// Screen width used in predefined display configuration.
 #[cfg(feature = "test-support")]
@@ -45,6 +53,43 @@ impl WindowBundle {
 
         WindowBundle::from_config(display_config)
     }
+
+    fn build_event_loop_and_window(self) -> (Receiver<Event<()>>, Window) {
+        let (sender, receiver) = mpsc::channel::<Event<()>>();
+
+        let window_return: Arc<Mutex<Option<Window>>> = Arc::new(Mutex::new(None));
+        let done = Arc::new(AtomicBool::from(false));
+
+        Self::spawn_event_loop(self.config, sender, window_return.clone(), done.clone());
+
+        while !done.load(Ordering::SeqCst) {}
+
+        let window = window_return.lock().unwrap().take().unwrap();
+
+        (receiver, window)
+    }
+
+    fn spawn_event_loop(
+        display_config: DisplayConfig,
+        sender: Sender<Event<()>>,
+        window_return: Arc<Mutex<Option<Window>>>,
+        done: Arc<AtomicBool>,
+    ) {
+        thread::spawn(move || {
+            let event_loop = EventLoop::new();
+            let window = display_config
+                .into_window_builder(&event_loop)
+                .build(&event_loop)
+                .unwrap();
+
+            *window_return.lock().unwrap() = Some(window);
+            done.store(true, Ordering::SeqCst);
+
+            event_loop.run(move |event, _, &mut _control_flow| {
+                sender.send(event).unwrap();
+            });
+        });
+    }
 }
 
 impl<'a, 'b> SystemBundle<'a, 'b> for WindowBundle {
@@ -53,13 +98,9 @@ impl<'a, 'b> SystemBundle<'a, 'b> for WindowBundle {
         world: &mut World,
         builder: &mut DispatcherBuilder<'a, 'b>,
     ) -> Result<(), Error> {
-        let event_loop = EventsLoop::new();
-        builder.add(
-            WindowSystem::from_config(world, &event_loop, self.config),
-            "window",
-            &[],
-        );
-        builder.add_thread_local(EventsLoopSystem::new(event_loop));
+        let (receiver, window) = self.build_event_loop_and_window();
+        builder.add(WindowSystem::new(world, window), "window", &[]);
+        builder.add(EventLoopSystem::new(receiver), "event_loop", &[]);
         Ok(())
     }
 }
--- a/amethyst_window/src/config.rs
+++ b/amethyst_window/src/config.rs
@@ -1,8 +1,8 @@
 use std::path::PathBuf;
 
-use log::error;
+use log::warn;
 use serde::{Deserialize, Serialize};
-use winit::{Icon, WindowAttributes, WindowBuilder};
+use winit::window::{Fullscreen, Icon, WindowAttributes, WindowBuilder};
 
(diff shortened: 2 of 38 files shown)


Jaspi10:
Jenkins seems to be testing with a stable rust prior to 1.36 is that intended?

Jaspi10:
Ok so I thought I was done, but apparently you can upgrade only half of the crates and everything compiles fine and passes all tests. But the moment you try to open a window everything breaks.

I'm gonna try to upgrade the rest of the crates too and see what happens.

Jaspi10:
Updated the rest of the crates.

I'm unsure if the way I implement Event passing now is a good,

The only thing blocking this from compiling on my local clone is that rendy hasn't updated yet.

amethyst_window/src/bundle.rs
```diff
@@ -45,6 +53,43 @@ impl WindowBundle {
...
+
+    fn build_event_loop_and_window(self) -> (Receiver<Event<()>>, Window) {
+        let (sender, receiver) = mpsc::channel::<Event<()>>();
+
+        let window_return: Arc<Mutex<Option<Window>>> = Arc::new(Mutex::new(None));
+        let done = Arc::new(AtomicBool::from(false));
+
+        Self::spawn_event_loop(self.config, sender, window_return.clone(), done.clone());
+
+        while !done.load(Ordering::SeqCst) {}
```
azriel91: probably better to sleep 1 ms instead of having a busy loop?

amethyst_window/src/config.rs
```diff
@@ -124,41 +120,29 @@ impl DisplayConfig {
...
-                        e
-                    );
-
-                    None
-                }
-            };
-
-            builder = builder.with_window_icon(icon);
+        } else if let Some(_) = self.icon {
+            warn!("DisplayConfig specified Icon, but Icon loading is currently not implemented.");
```
azriel91: Is the code from the [icon loading example](https://github.com/rust-windowing/winit/blob/31110be3960bb97caeacdd0a8cbad292722f94f1/examples/window_icon.rs#L43) sufficient to retain this feature?

amethyst_window/src/monitor.rs
```diff
@@ -1,30 +1,30 @@
...
+use winit::{event_loop::EventLoop, monitor::MonitorHandle, window::Window};
 
 /// A struct that can resolve monitors.
 /// Usually either a Window or an EventsLoop.
 pub trait MonitorsAccess {
     /// Returns an iterator over the available monitors
-    fn iter(&self) -> AvailableMonitorsIter;
-    /// Returns the `MonitorId` of the primary display
-    fn primary(&self) -> MonitorId;
+    fn iter(&self) -> Box<dyn Iterator<Item = MonitorHandle>>;
```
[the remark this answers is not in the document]
Jaspi10: No that isn't allowed. I also tried just using `&dyn` but then you can't call `.enumerate()`.

omni-viral:
Have you seen my PR? It has update to winit 0.20 too :stuck_out_tongue:

amethyst_window/src/bundle.rs
omni-viral: Some platforms require event loop on main thread, so this logic won't work.

Consider taking a look for #1877

Jaspi10:
@omni-viral no didn't see it, just looked at the issue. I don't really like letting winit hijack the main thread, but it seems like the only way to go, if we want to support all Platforms. I'll close this in favor of your PR.

omni-viral:
We could split my PR which also adds opengl support and merge new winit sooner.
But IDK how much it will win for us, few days maybe.

Jaspi10:
@omni-viral I don't think we can merge the new winit version before rendy and gfx-hal have made a release with it.

OvermindDL1:
>  don't really like letting winit hijack the main thread

Some platforms absolutely require it as they use callback functions instead of event polling, that's why winit had to change it's API, because it didn't work with those before.

Do you have an idea for how to make an API that would work for all platforms other than winit's full callback style API?

Jaspi10:
@OvermindDL1 
> Some platforms absolutely require it as they use callback functions instead of event polling, that's why winit had to change it's API, because it didn't work with those before.

What platforms are those? I just looked at the winit docs and they only state iOS are there others, because if all other platforms allow winit to run on a different thread it might make sense to optimize for those platforms and add special code for iOS.

> Do you have an idea for how to make an API that would work for all platforms other than winit's full callback style API?

No I don't. I'm not familiar enough with platform APIs and it might not be possible. As you said there is a reason why winits API works like this now.

OvermindDL1:
> What platforms are those? I just looked at the winit docs and they only state iOS are there others, because if all other platforms allow winit to run on a different thread it might make sense to optimize for those platforms and add special code for iOS.

iOS, Android, Web as the big public ones that do that, although some embedded systems do as well (though winit doesn't target those anyway), in addition I'm pretty sure wayland on linux operates via registered event callbacks as well (although it has a polling loop shim that is not as efficient).  The combination of iOS, Android, and Web are significantly large enough (especially the rust community wasm push) that those should work for the 'default' API style.

> No I don't. I'm not familiar enough with platform APIs and it might not be possible. As you said there is a reason why winits API works like this now.

In essence these platforms only allow you to pass in a function pointer into an 'event listing registration function' (on some it's a single function to handle any system event, on others there are different function registrations for different events), which is what winit's `run` API does, at this point the main event handler thread is halted at this point, permanently, and not ever returned from, including when the application is killed (resources should be cleared via the shutdown event).

This is why `run_return` will make your program not compile on iOS/android/wasm (and potentially future others), where `run` will work on all (as you can emulate the function callback style with event polling but you can't emulate event polling via function callback as some events have to be handled 'during' the event and not 'sometime later').

The new `run` style is a very very good thing, but it is what they should have done from the start, this big change later on is a bit painful for projects as they assumed the inherently incompatible event-poller style, though that's why it is not 1.0 yet.

Jaspi10:
@OvermindDL1 
From your explanation I agree that the way `run` is implemented in winit now is a good idea.

But taking over a thread isn't even the problem, it's taking over the **main** thread, because just spawning a thread running the event loop on it and feeding the events back the application is easy. It's what this PR does.

But if platforms insist on using the **main** thread for this. We can either move amethyst main loop onto a different thread. Or use winits `run` function for it.
Both would introduce a lot of coupling between amethyst `CoreApplication` and winit.

OvermindDL1:
> But taking over a thread isn't even the problem, it's taking over the main thread

On some of those systems it is absolutely required to be the main thread, there is no other option and trying to register the hook on another will hard crash.

> because just spawning a thread running the event loop on it and feeding the events back the application is easy. It's what this PR does.

Those systems don't have event loops they have callback handlers owned by the system itself on the main thread.

> But if platforms insist on using the main thread for this. We can either move amethyst main loop onto a different thread. Or use winits run function for it.

Really *really* do need to use winit's `run` function, amethyst is designed to be used on everything from desktops to phones to web and more, so it itself needs to use the sole option that will work on all systems, not just some.

> Both would introduce a lot of coupling between amethyst CoreApplication and winit. As seen in @omni-viral s PR.

Amethyst can always wrap winit's `run`, but regardless it needs to be a single callback called from within the main thread with no code appearing after that callback registration/run.

Issue #1729, opened 2019-06-17
[NEW FEATURE] WebAssembly scripting system


MarkMcCaskey:
Hello! I'm a big fan of the Amethyst project, unfortunately I haven't had the opportunity to contribute yet.

I work at a company called [Wasmer](https://wasmer.io) and we're making a [server-side WebAssembly runtime](https://github.com/wasmerio/wasmer).  One of the neat uses of this is as a plugin system, being able to call WebAssembly functions from Rust and other languages and have the WebAssembly module call back into the host.

## Description

This is less of a feature proposal and more of the start of a discussion around using WebAssembly as a scripting language.  It's our goal to have Wasmer be very easy to use for these cases, but the Wasm ecosystem is still early on.  Whether using Wasmer or not, Wasm plugins are a neat idea that I think will become very popular over the coming years.

## What problem does this solve? What need does it fill?

Reasons why Wasm is useful as a scripting language:
- performance - Wasm JITs currently can run at near native speeds and we expect this to get better over time
- flexibility - use any language that compiles to Wasm; this is great for plugins created by users.  It also opens up tons of libraries and the tooling of all these languages
- security - Wasm can be sandboxed and metered.  It's simple to run untrusted code with limited resources and without worrying about it doing bad things (as of writing, metering works, but is still in development in Wasmer)
- standard - by using Wasm, you benefit from all the tools and work done on Wasm
- good plugin support - Wasm for plugins will likely become very popular and the tooling for connecting the pieces is coming together

## Describe the solution you'd like

Feedback on the idea!  Ideas about how to improve, what we'd need to support game engines well (we have some ideas already).

If you all are interested, we're interested in working with you to help design and implement this.

## Describe alternatives you've considered

- special, endorsed scripting language(s) - probably more work, less flexible, less secure, etc.
- JVM - licensing issues with embedding, less ecosystem support, less flexible (Wasm explicitly supports running without garbage collection)
- no scripting languages - user-created content is awesome + dynamism is nice to have during development
- V8/other JS Wasm VMs - limited by the needs of JS, likely not as focused on plugin use-cases as Wasmer (Wasmer has being easily embeddable as a core design goal)
[...]

PR #1892, opened 2019-08-19
[FEATURE] WebAssembly scripting system

ryzokuken:
Implements #1729 

## Description

Add a PoC WebAssembly scripting system as a subcrate, opening up possibilities of including a full-fledged modding system.

## Additions

- `amethyst_scripting::run_wasm`: Run WebAssembly bytecode.
- `amethyst_scripting::run_wat`: Covert WAT format to WebAssembly bytecode and run it.

## Removals

None

## Modifications

None

## PR Checklist


Friz64:
There's an empty `stdout` file, you may want to remove that.

arlyon:
I agree with @jojolepro in that a WIP would need to hook into the engine in some way. What would a real WIP look like? I think having a script add/overwrite a system (with access to the logger) is a good first step. Providing an FFI for each language we'd like to support would be a benefit. Rust would be a good first step for that as well.

So, in my opinion:
- amethyst provides an api for logging and adding systems
- a small "mod" is written in rust that
  - defines an FFI for this small "modding" api
  - defines an init() function
  - registers a system using the FFI
- at load, wasm "mods" (anything that implements an init function) are loaded, and the init function is called

ryzokuken:
@Friz64 removed! I have no clue how it got in there haha.

azriel91:
CI isn't passing because `wasmer-runtime` relies on `nightly` Rust. If you mask the crate behind a `"nightly"` feature, it should work.

ryzokuken:
@azriel91 are you sure? It's compiling locally with

```
stable-x86_64-apple-darwin (default)
rustc 1.37.0 (eae3437df 2019-08-13)
```

azriel91:
semi :stuck_out_tongue::
```rust
   Checking wasmer-runtime-core v0.6.0
error[E0658]: use of unstable library feature 'maybe_uninit' (see issue #53491)
  --> /home/jenkins/.cargo/registry/src/github.com-1ecc6299db9ec823/wasmer-runtime-core-0.6.0/src/instance.rs:55:22
   |
55 |             Box::new(mem::MaybeUninit::<vm::Ctx>::zeroed());
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

maybe the base image is on an outdated stable rust. have to check

erlend-sh:
*Quote from @arlyon:*

> EDIT some spitballing of problems we need to solve:
> 
> how do we define systems at runtime with specs?
> some infra around logging: each mod should have its own namespace, how do we do that?
> Edit 2: relevant article from mozilla https://hacks.mozilla.org/2019/08/webassembly-interface-types/
> 
> For this to be reasonable in any way we need to wait for wasm interface types.

That article ends with an encouraging note though:

> But if you want to start playing with it, we’ve implemented this across the toolchain, from production to consumption:
> 
> - the Rust toolchain
> - wasm-bindgen
> - the Wasmtime WebAssembly runtime
> 
> And since we maintain all these tools, and since we’re working on the standard itself, we can keep up with the standard as it develops.
> 
> Even though all these parts will continue changing, we’re making sure to synchronize our changes to them. So as long as you use up-to-date versions of all of these, things shouldn’t break too much.
> 
> So here are the many ways you can play with this today. For the most up-to-date version, check out this [repo of demos](https://github.com/CraneStation/wasmtime-demos).

syrusakbary:
We have a lot of plans on WebAssembly Interface Types, expect this to be moved forward in Wasmer very soon :)

azriel91:
bors try

PR #1900, opened 2019-08-25
Refer to `amethyst_rendy` via `amethyst::renderer`

fintelia:
## Description

This PR tweaks a "use" statement in the rendy example. This is necessary for the example to work when run outside of the amethyst repository

## PR Checklist

### DIFF
--- a/examples/rendy/prefab_data.rs
+++ b/examples/rendy/prefab_data.rs
@@ -7,7 +7,7 @@ use amethyst::{
     utils::tag::Tag,
     Error,
 };
-use amethyst_rendy::{
+use amethyst::renderer::{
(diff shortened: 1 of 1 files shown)


mrhatman:
This would need a changelog entry.

examples/rendy/prefab_data.rs
```diff
@@ -7,7 +7,7 @@ use amethyst::{
     utils::tag::Tag,
     Error,
 };
-use amethyst_rendy::{
+use amethyst::renderer::{
```
azriel91: Heya, `cargo fmt` groups it into the previous import, could you also commit that change?
Thanks!

PR #1904, opened 2019-08-26
Add Custom Shader Example

### LINKED ISSUE #1702: [MAINTENANCE] Add rendy basic custom shader example
## Description

We currently have no example for creating a custom RenderPass using your own shader. I think with the flexability we gained from rendy, this would be a good basic example for getting into the internals of the renderer and writing custom passes.

This should consist of something akin to https://vulkan-tutorial.com/Drawing_a_triangle where we simply supply a shader which takes a vertex buffer and index buffer, and draws a triangle. 

## Additional Information
Example pass:
https://github.com/jaynus/amethyst-imgui/blob/master/src/pass.rs

mrhatman:
## Description

Adds an example that uses a custom shader and render.

Any suggestions for code readability and clarity would be greatly appreciated as think is the first renderer I have written. 

Fixes #1702 

## Additions
- Custom Shader Example

## PR Checklist

- [N/A ] Updated the content of the book if this PR would make the book outdated.
- [N/A ] Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.
- [N/A] Added unit tests for new code added in this PR.


examples/custom_shader/custom_pass.rs
```diff
@@ -0,0 +1,358 @@
...
+        &mut self,
+        factory: &Factory<B>,
+        _queue: QueueId,
+        index: usize,
+        _subpass: hal::pass::Subpass<'_, B>,
+        world: &World,
+    ) -> PrepareResult {
+        let (triangles,) = <(ReadStorage<'_, Triangle>,)>::fetch(world);
+
+        //Get our scale value
```
ishitatsuyuki: Space missing after `//`. Also (some of) these comments doesn't say much thing, so they're better removed.

examples/custom_shader/custom_pass.rs
```diff
@@ -0,0 +1,372 @@
...
+        let scale = world.read_resource::<CustomUniformArgs>();
+
+        // Write to our DynamicUniform
+        self.env.write(factory, index, scale.std140());
+
+        // Create and write our vertices to a vector
+        let mut vertex_data_iter: Vec<CustomArgs> = Vec::new();
+
+        for triangle in triangles.join() {
+            vertex_data_iter.extend(triangle.get_args().iter());
```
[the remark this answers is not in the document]
mrhatman: I updated this code to use iterators more. 

Allocation is necessary as the write command requires that it be AsRef<[T]> which requires the data be collected.

examples/custom_shader/custom_pass.rs
```diff
@@ -0,0 +1,372 @@
...
+        types::Backend,
+        util,
+    },
+};
+
+use amethyst_error::Error;
+use derivative::Derivative;
+use glsl_layout::*;
+
+lazy_static::lazy_static! {
```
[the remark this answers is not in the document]
azriel91: I think it's worth including the commented section somehow, even if not in this form directly.
I haven't read the PR properly, but in general am happy with either of:

* doc comment with code snippet (``/// ```rust ``)
* book page with code snippet

Cargo.toml
```diff
@@ -187,6 +189,10 @@ name = "ui"
 path = "examples/ui/main.rs"
 required-features = ["audio"]
 
+[[example]]
```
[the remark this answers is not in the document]
mrhatman: I think I will include an example of how to use a custom shader in a comment as described below.

As for  spirv-reflection. In order to keep true to the original scope. I will keep this example simple.

examples/custom_shader/custom_pass.rs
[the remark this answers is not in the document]
mrhatman: I added a doc comment.

Frizi:
Is this intended to be a custom **render pass**, or indeed just a custom shader used with builtin 3d passes? Because it is possible to define those without copying the whole pass.

I'm going to review it assuming first, but maybe that wasn't the intention.

examples/custom_shader/custom_pass.rs
```diff
@@ -0,0 +1,348 @@
...
+            Err(e)
+        }
+        Ok(mut pipes) => Ok((pipes.remove(0), pipeline_layout)),
+    }
+}
+
+/// A [RenderPlugin] for our custom plugin
+#[derive(Default, Debug)]
+pub struct RenderCustom {
+    target: Target,
```
Frizi: This target is never exposed or changed in any way. There usually is not much point in allowing such configuration in custom render plugins (because those are tailored to your needs anyway), so i would recommend hardcoding `Target::Main` directly to `extend_target` call and get rid of this value. It is of course not really wrong, but here it's incomplete and not really necessary.

examples/custom_shader/custom_pass.rs
```diff
@@ -0,0 +1,348 @@
...
+        // Write the vector to a Vertex buffer
+        self.vertex.write(
+            factory,
+            index,
+            self.vertex_count as u64,
+            Some(vertex_data_iter.collect::<Box<[CustomArgs]>>()),
+        );
+
+        // Return that we request to draw
+        PrepareResult::DrawRecord
```
Frizi: You don't really need to record every frame with this. You need to record only when `self.vertex_count` have changed. Use `util::ChangeDetection` struct as we do in other passes to signal that correctly.

examples/custom_shader/custom_pass.rs
```diff
@@ -0,0 +1,348 @@
...
+/// ''' rust
+/// lazy_static::lazy_static! {
+///     static ref VERTEX_SRC: SpirvShader = PathBufShaderInfo::new(
+///         PathBuf::from(concat!(env!("CARGO_MANIFEST_DIR"), "/examples/assets/shaders/src/vertex/custom.vert")),
+///         ShaderKind::Vertex,
+///         SourceLanguage::GLSL,
+///         "main",
+///     ).precompile().unwrap();
+///
+///     static ref VERTEX: SpirvShader = SpirvShader::new(
```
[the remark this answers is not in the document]
mrhatman: I copied this from https://github.com/amethyst/amethyst-imgui/blob/master/src/pass.rs

I don't fully understand why the duplication is used there.

I removed them and tested. The second was not required.

examples/custom_shader/custom_pass.rs
mrhatman: Removed this.

mrhatman:
> Is this intended to be a custom **render pass**, or indeed just a custom shader used with builtin 3d passes? Because it is possible to define those without copying the whole pass.
> 
> I'm going to review it assuming first, but maybe that wasn't the intention.

The issue linked to https://github.com/amethyst/amethyst-imgui/blob/master/src/pass.rs

I based mine on this. It is a custom render pass mostly to show off all of the steps.

examples/custom_shader/custom_pass.rs
mrhatman: That makes the assumption that the only changes require adding or removing a triangle. If a triangle is changed then it won't cause it to change. correct?

That being said all of this is not need for this example as it is static but i don't want to give an example that doesn't allow for updated entities.

examples/custom_shader/custom_pass.rs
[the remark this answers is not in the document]
jaynus: The master branch currently has some issues like that. That's a legacy reminant. If you check my docking branch that had cleaned up a fair amount.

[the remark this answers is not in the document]
mrhatman: Ok cool

examples/custom_shader/custom_pass.rs
Frizi: No, it will cause it to change. But changing the triangle data doesn't require a draw call. All you need to do is to update the buffer with new data.

You only need to redraw if any inputs to your commands have actually changed. This usually means just the number of objects have changed or a sequence of binds is different. The second never happens in this example, as your bind sequence is static.

Frizi:
> I based mine on this. It is a custom render pass mostly to show off all of the steps.

Okay, but then the example is about custom render pass. Naming it as "custom shader" example can be confusing. You don't need to write a whole render pass if you want to just change a shader, because our builtin render passes are generic over shaders and their parameters.

examples/custom_shader/custom_pass.rs
mrhatman: I will implement that change then.

mrhatman:
I will change the example name to Custom Render Pass

examples/custom_shader/custom_pass.rs
mrhatman: Implemented

examples/custom_render_pass/custom_pass.rs
```diff
@@ -0,0 +1,354 @@
...
+        // Write the vector to a Vertex buffer
+        self.vertex.write(
+            factory,
+            index,
+            self.vertex_count as u64,
+            Some(vertex_data_iter.collect::<Box<[CustomArgs]>>()),
+        );
+
+        // Return with we can reuse the draw buffers using the utility struct ChangeDetection
+        self.change.prepare_result(self.vertex_count, false)
```
Frizi: This usage is incorrect. First argument should also be the frame index, second signifies if the change was detected in this frame.

So, first you need to actually detect this.
```rust
let old_vertex_count = self.vertex_count;
self.vertex_count = triangles.join().count() * 3;
let changed = old_vertex_count != self.vertex_count; 
```

Then, you need to submit that as a change for current frame
```rust
self.change.prepare_result(index, changed)
```

Think about it as a replacement for simple `if changed { redraw } else { reuse }`, except taking "frames in flight" into account.

The whole `ChangeDetection` thing is only necessary because of the details of which command buffer is actually resubmitted, as there is a dedicated one per every frame in flight. That means you can only reuse a buffer if a number of consecutive frames were not changed.

mrhatman: Ok. I have implemented these and I think I get what they do. We might want to clarify the documentation on these as I obviously didn't grok it the first time.

examples/custom_render_pass/custom_pass.rs
```diff
@@ -158,7 +161,7 @@ impl<B: Backend> RenderGroup<B, World> for DrawCustom<B> {
         );
 
         // Return with we can reuse the draw buffers using the utility struct ChangeDetection
-        self.change.prepare_result(self.vertex_count, false)
+        self.change.prepare_result(self.vertex_count, changed)
```
Frizi: This is still wrong, should be `prepare_result(index, changed)`

Frizi:
bors r=frizi,jaynus

Frizi:
bors retry

PR #1907, opened 2019-08-29
Allow disabling parallel support in amethyst_core (for WASM support)

AThilenius:
Hi all. This is both an intro and a tiny patch.

## Background / Intro

I'm using Amethyst (right now only `amethyst_core`) for a browser-based game written in Typescript and Rust. Along with `stdweb` it feels like a match made in heaven, and every day I'm more impressed by both Rust and Amethyst-spawned crates. 😍 Rendering is shelled out to `THREE.js` for now, until `wgpu-rs` or Amethyst are ready for WASM prime-time. Anyway, I figured I would hijack this PR to say hello 👋 and **inquire as to the WASM plan/timeline for Amethyst**. For example, Amethyst (including `core`) pulls in `rayon` directly but treading support isn't there just yet. I'm happy to help anywhere work overlaps! I'll need a few bleeding-edge things like a half decent UDP stack (so `WebRTC`). Would rather my efforts went to something shared than a private code base, if that's ever possible.

You're all awesome, keep up the great work 🍻

## Description
Threading support isn't there yet for browsers, and causes a panic when you try to use a `shred::DispatcherBuilder`:
_`backend.js:1 Panic error message: Invalid configuration: ThreadPoolBuildError { kind: IOError(Custom { kind: Other, error: "operation not supported on wasm yet" }) }`_

This PR continues piping the ability to disable parallel support in `shred` up through `amethyst_core` so that WASM consumers (like me) can disable it.

## PR Checklist

### DIFF
--- a/amethyst_core/Cargo.toml
+++ b/amethyst_core/Cargo.toml
@@ -25,8 +25,8 @@ log = "0.4.6"
 num-traits = "0.2.0"
 rayon = "1.1.0"
 serde = { version = "1", features = ["derive"] }
-specs = { version = "0.15", features = ["shred-derive"] }
-specs-hierarchy = "0.5.1"
+specs = { version = "0.15", default-features = false, features = ["shred-derive"] }
+specs-hierarchy = { version = "0.5.1", default-features = false }
 getset = "0.0.7"
 derive-new = "0.5.6"
 derivative = "1.0"
@@ -38,6 +38,7 @@ amethyst = { path = "..", version = "0.12.0" }
 ron = "0.5.1"
 
 [features]
+default = ["specs/parallel", "specs-hierarchy/parallel"]
 profiler = ["thread_profiler/thread_profiler"]
 nightly = ["specs/nightly"]
 saveload = ["specs/serde"]


AThilenius:
The coverage report looks like a false positive?

jojolepro:
Multithreading is supported by wasm. See this crate: https://github.com/jojolepro/web_worker

For an example on how it is used, see this example: https://github.com/jojolepro/specs_wasm

AThilenius:
Thank you @jojolepro, that's an awesome reference to have, and I'm super excited about the future of WASM and Rust! :) So cool to see all the effort behind this, and the potential of it. After a decade of JS, threading in a browser from a systems level language seems like a dream. It's surreal to see all this crazy powerful tech actually becoming a thing.

Unfortunately in my case with WASM, Firefox-nightly + a user-set flag isn't considered supported, it's hard to require early testers go to that much effort to load a URL. This PR doesn't change any default behaviors, it just allows a consumer of `amethyst_core` to disable parallel support. In my case, it would be disabled until Chrome stable has been shipping with shared buffers enabled for a few months :( But the future is bright!

jaynus:
I have one concern about this. What happens if we begin using par_iter in the engine internals? Does it shim for not-actually-threaded? I believe the build would fail. This means we need to feature gate all future usage of it.

distransient:
I believe this should be only a shim for now, and we shouldn't worry too much about "if this in the future". Jojo's solution should be the actual proper way going forward.

jaynus:
Bors r+

jaynus:
bors retry

PR #1747, opened 2019-06-20
[BREAKING] Removes Float newtype and reverts to f32 everywhere (And clippy while I'm here)

jaynus:
## Description

Discussion: https://community.amethyst.rs/t/improving-float-ergonomics-proposal/889/35

Remove the Float newtype and converts everything back to f32. This leaves the float64 feature flag for now, however it is not used anywhere. This may be used to feature-gate Transform64 or a variant.

This also makes subcrates clippy passing for clippy::all lints. Because this pass already touches so much, I figured it was worth the effort

Items still needing completion
- [x] Update book
- [x] Update docs

Fixes #1744, Closes #1749

## Removals

- Float newtype

## Modifications

- Changes `DisplayConfig::to_window_builder` to `DisplayConfig::into_window_builder`
- `Float` -> `f32` everywhere
- Changes all `f64` in amethyst_input to be `f32`, and downcasts the appropriate f64's from winit. This is uniformity across the engine, at the loss of some precision. 
- Removed as many erroneous `.into()` calls as I could find

## PR Checklist

By placing an x in the boxes I certify that I have:

- [x] Ran `cargo test --all` locally if this modified any rs files.
- [x] Ran `cargo +stable fmt --all` locally if this modified any rs files.
- [x] Updated the content of the book if this PR would make the book outdated.
- [x] Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.
- [x] Acknowledged that by making this pull request I release this code under an MIT/Apache 2.0 dual licensing scheme.

PR #1909, opened 2019-08-30
Remove float64 feature flags

eliaspekkala:
## Description
This PR will remove the float64 feature flags from the Cargo.toml files.

The Float type has been removed from Amethyst a while ago, see: #1747.

## Reason
I think it causes confusion to keep the feature flags if they don't have any affect.

## PR Checklist

- [n/a] Updated the content of the book if this PR would make the book outdated.
- [n/a] Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.
- [n/a] Added unit tests for new code added in this PR.


Frizi:
Good catch, thanks!
bors r+

PR #1912, opened 2019-09-01
Feature/run setup fn before systems

azriel91:
## Description

In `amethyst_test`, changed `AmethystApplication::with_setup` to run before the dispatcher. This allows resources to be computed and inserted from other resources in the `World`.

Previously, if a `System` had `ReadExpect<'_, T>`, and `T ` needed to be computed from the `World`, it was impossible to insert the resource before the system runs.

## Modifications

* `AmethystApplication::with_setup` now takes in `FnOnce(&mut World) + Send + 'static`.
* `AmethystApplication::with_setup` runs the function before the dispatcher.

## PR Checklist


azriel91:
bors r=distransient

Issue #1889, opened 2019-08-18
[NEW FEATURE] Setting custom logger Formatter


mvlabat:
## Description

I would like to use a custom logger formatter in my project. I have several binaries producing output to the same stdout (a client and a server as a child-process in my case), and I'd like to distinguish their output. Like this, for instance:
> [INFO][SERVER][amethyst::app]

## What problem does this solve? What need does it fill?
The only way I've found to accomplish this is to construct `fern::Dispatch` manually and apply it as a global logger, skipping Amethyst logger. I think it's quite easy to provide an API for passing custom formatter, so users don't have to construct the dispatcher manually.

## Describe the solution you'd like
We could provide an alternative constructor that would allow passing both `LoggerConfig` and a formatter function.

nalshihabi:
Hello, I am new here but I believe I can tackle this feature.

mvlabat:
Works perfectly, thanks @nalshihabi for implementing!

PR #1915, opened 2019-09-01
[FEATURE] Custom Logger Formatter

nalshihabi:
Implements #1889 

## Description

Add a custom log formatter option to the Amethyst logger so that instead of setting up their own logger to use their own formatting, a developer using Amethyst can just pass in their own formatter to the Amethyst logger.

## Additions

- `Logger::new_formatter`: a different constructor for Amethyst logger where a custom formatter can be passed as a parameter.
- `Logger::from_config_formatter`: a different function for the Logger for starting a new Logger with a config and a formatter.
- `Logger::new_with_config`: helper function that much of the functionality of `Logger::from_config` was moved into from `Logger::from_config` in order to allow for the functionality to be reused easily by `Logger::from_config_formatter`.

## Removals

None

## Modifications

- `Logger::from_config`: Changed to just call `Logger::from_config_formatter` with valid parameters for the helper function.

## PR Checklist

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -3,7 +3,7 @@ pub use log::LevelFilter;
 use log::debug;
 use serde::{Deserialize, Serialize};
 
-use std::{env, io, path::PathBuf, str::FromStr};
+use std::{env, fmt, io, path::PathBuf, str::FromStr};
 
 /// An enum that contains options for logging to the terminal.
 #[derive(Clone, Copy, PartialEq, Eq, Debug, Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


azriel91:
file system race condition:

```
Could not lock repository. Please try again
```

bors try

nalshihabi:
For the file system race condition, what might need to be done about it?

azriel91:
Not sure yet (maybe run the jobs serially) -- initially I thought it was duplicated work (as in, no value running across `nightly` and `stable`), but then there was once a bug was caught in stable but not on nightly.

Shall fix it when i'm more awake.

bors r=azriel91,distransient

azriel91:
bors r=azriel91,distransient

PR #1917, opened 2019-09-03
Rewriting the network library to provide a nicer baseline

jstnlef:
## Description
Kind of a big one here I know. This is an attempt to both make the networking library play a bit more nicely with Specs and to also decouple it from laminar while still providing the ability to use laminar as a transport backend.

I wanted to get this up to get some feedback from people. Comments, alternative approaches, criticisms are all welcomed and desired!

I'm hoping to use this new implementation as a new baseline with which we can build new functionality off of like component synchronization, matchmaking, etc. This also paves the way for us to implement a websocket implementation backend for wasm games in, what I think, is a fairly painless manner.

I'm also aware that the testing situation is extremely lacking but I'm mostly curious if this is the route we'd like to go down before I shore up the tests.

## Additions
- Complete rewrite of the current networking crate.

## Removals
- The old networking crate

## PR Checklist


amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub fn set_socket(&mut self, socket: S) {
+        self.socket = Some(socket);
+    }
+
+    /// Return whether or not the NetworkResource has a bound socket.
+    pub fn has_socket(&self) -> bool {
+        self.socket.is_some()
+    }
+
+    /// Return the set server address if one was set.
```
TimonPost:
```suggestion
    /// Returns the server address if one was set.
```

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub fn set_server_addr(&mut self, server_addr: Option<SocketAddr>) {
+        self.server_addr = server_addr;
+    }
+
+    /// Set the bound socket to the NetworkResource.
+    pub fn set_socket(&mut self, socket: S) {
+        self.socket = Some(socket);
+    }
+
+    /// Return whether or not the NetworkResource has a bound socket.
```
TimonPost:
```suggestion
    /// Returns whether or not the `NetworkSimulationResource` has a bound socket.
```

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,247 @@
...
+        }
+    }
+
+    /// Update and return the NetworkResource with a bound socket.
+    pub fn with_socket(mut self, socket: S) -> Self {
+        self.socket = Some(socket);
+        self
+    }
+
+    /// Add a number of trusted clients to the NetworkResource for use as a server.
```
TimonPost:
```suggestion
    /// Add a number of trusted clients to the `NetworkSimulationResource` for use as a server
```

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,247 @@
...
+        Self {
+            socket: None,
+            server_addr: None,
+            clients: Vec::new(),
+            messages: VecDeque::new(),
+            is_server: true,
+        }
+    }
+
+    /// Update and return the NetworkResource with a bound socket.
```
TimonPost:
```suggestion
    /// Update and return the `NetworkSimulationResource` with a bound socket
```

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,247 @@
...
+        self
+    }
+
+    /// Add a number of trusted clients to the NetworkResource for use as a server.
+    pub fn with_trusted_clients(mut self, clients: &[SocketAddr]) -> Self {
+        self.clients = clients.iter().map(|addr| Client::new(*addr)).collect();
+        self
+    }
+
+    /// Returns a slice of the tracked clients.
```
TimonPost:
```suggestion
    /// Returns a slice of the tracked clients
```

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub fn has_socket(&self) -> bool {
+        self.socket.is_some()
+    }
+
+    /// Return the set server address if one was set.
+    pub fn server_addr(&self) -> Option<SocketAddr> {
+        self.server_addr
+    }
+
+    /// Returns whether or not this `NetworkSimulationResource` was created as a "server"
```
TimonPost:
```suggestion
    /// Returns whether or not this `NetworkSimulationResource` was created as a 'server'
```

TimonPost: the plural of return should be used

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub fn is_server(&self) -> bool {
+        self.is_server
+    }
+
+    /// Return whether or not there are any outstanding messages to send.
+    pub fn has_messages_to_send(&self) -> bool {
+        !self.messages.is_empty()
+    }
+
+    /// Creates a `Message` with the default guarantees provided by the `Socket` impl and pushes it
```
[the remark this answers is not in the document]
TimonPost: to where is it pushed

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// onto the messages queue.
+    pub fn send(&mut self, payload: &[u8]) {
+        let (delivery, ordering) = S::default_guarantees();
+        self.send_internal(payload, delivery, ordering);
+    }
+
+    /// Creates and queues an unreliable/unordered `Message`.
```
TimonPost:
```suggestion
    /// Create and queue an unreliable/unordered `Message`.
```

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub fn send_unreliable(&mut self, payload: &[u8]) {
+        self.send_internal(
+            payload,
+            DeliveryGuarantee::Unreliable,
+            OrderingGuarantee::None,
+        )
+    }
+
+    /// Creates and queues an unreliable/sequenced `Message`.
```
TimonPost:
```suggestion
    /// Create and queue an unreliable/sequenced `Message`.
```

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub fn send_unreliable_sequenced(&mut self, payload: &[u8]) {
+        self.send_internal(
+            payload,
+            DeliveryGuarantee::Unreliable,
+            OrderingGuarantee::Sequenced(0),
+        )
+    }
+
+    /// Creates and queues an reliable/unordered `Message`.
```
TimonPost:
```suggestion
    /// Create and queue an reliable/unordered `Message`.
```

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub fn send_reliable(&mut self, payload: &[u8]) {
+        self.send_internal(
+            payload,
+            DeliveryGuarantee::Reliable,
+            OrderingGuarantee::None,
+        )
+    }
+
+    /// Creates and queues an reliable/sequenced `Message`.
```
TimonPost:
```suggestion
    /// Create and queue an reliable/sequenced `Message`.
```

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub fn send_reliable_sequenced(&mut self, payload: &[u8]) {
+        self.send_internal(
+            payload,
+            DeliveryGuarantee::Reliable,
+            OrderingGuarantee::Sequenced(0),
+        )
+    }
+
+    /// Creates and queues an reliable/ordered `Message`.
```
TimonPost:
```suggestion
    /// Create and queue an reliable/ordered `Message`.
```

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub fn send_reliable_ordered(&mut self, payload: &[u8]) {
+        self.send_internal(
+            payload,
+            DeliveryGuarantee::Reliable,
+            OrderingGuarantee::Ordered(0),
+        )
+    }
+
+    /// Returns a draining iterator of messages.
```
TimonPost:
```suggestion
    /// Returns a draining iterator of messages.
```

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub fn server_addr(&self) -> Option<SocketAddr> {
+        self.server_addr
+    }
+
+    /// Returns whether or not this `NetworkSimulationResource` was created as a "server"
+    pub fn is_server(&self) -> bool {
+        self.is_server
+    }
+
+    /// Return whether or not there are any outstanding messages to send.
```
TimonPost:
```suggestion
    /// Returns whether or not there are any outstanding messages to send.
```

amethyst_network/src/simulation/guarantees.rs
```diff
@@ -0,0 +1,19 @@
...
+    /// Message will be delivered.
+    Reliable,
+}
+
+#[derive(Copy, Clone, Debug, PartialOrd, PartialEq, Eq)]
+pub enum OrderingGuarantee {
+    /// Messages will be delivered in the order they are received.
+    None,
+    /// Messages will be arranged in order. Older messages will not be sent to client code.
+    /// Type requires a stream_id which will be conditionally used based on the underlying protocol.
```
TimonPost:
```suggestion
    /// Type requires a `stream_id` which will be conditionally used based on the underlying protocol.
```

amethyst_network/src/simulation/guarantees.rs
```diff
@@ -0,0 +1,19 @@
...
+    Sequenced(u8),
+    /// All messages will be arranged in order.
+    /// Type requires a stream_id which will be conditionally used based on the underlying protocol.
```
TimonPost:
```suggestion
    /// Type requires a `stream_id` which will be conditionally used based on the underlying protocol.
```

amethyst_network/src/simulation/events.rs
```diff
@@ -0,0 +1,13 @@
...
+use std::net::SocketAddr;
+
+/// Events which can be received from the network.
+#[derive(Debug)]
+pub enum NetworkSimulationEvent {
+    // A message was received from a remote client
+    Message(SocketAddr, Bytes),
+    // A new host has connected to us
+    Connect(SocketAddr),
+    // a host has disconnected from us
```
TimonPost:
```suggestion
    // A host has disconnected from us
```

amethyst_network/src/simulation/client.rs
```diff
@@ -0,0 +1,16 @@
...
+pub struct Client {
+    addr: SocketAddr,
+}
+
+impl Client {
+    pub fn new(addr: SocketAddr) -> Self {
+        Self { addr }
+    }
+
+    pub fn addr(&self) -> SocketAddr {
```
TimonPost: doc string

amethyst_network/src/simulation/client.rs
```diff
@@ -0,0 +1,16 @@
+use std::net::SocketAddr;
+
+/// Stores information about a known client for the `NetworkSimulationResource`.
+pub struct Client {
+    addr: SocketAddr,
+}
+
+impl Client {
+    pub fn new(addr: SocketAddr) -> Self {
```
TimonPost: doc string

amethyst_network/src/simulation.rs
```diff
@@ -0,0 +1,18 @@
+//! Module containing various utilities to run a client/server based network simulation. Expect
```
TimonPost:
```suggestion
//! Module containing various utilities to run a client/server-based network simulation. Expect
```

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,247 @@
...
+    ) {
+        let message = Message::new(payload, delivery, ordering);
+        self.messages.push_back(message);
+    }
+}
+
+impl<S: Socket> Default for NetworkSimulationResource<S> {
+    fn default() -> Self {
+        panic!(
+            "The `NetworkResource` resource MUST be created and added to the `Application` \
```
TimonPost:
```suggestion
            "The `NetworkSimulationResource` resource MUST be created and added to the `Application` \
```

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,146 @@
...
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct NetworkSimulationTime {
+    /// The current simulation frame
+    frame_number: u32,
+    /// Accumulated duration since last simulation frame
+    elapsed_duration: Duration,
+    /// Duration per frame
+    per_frame_duration: Duration,
+    /// Determines how often we send messages. i.e. "Every N frames" where N is message_send_rate.
```
TimonPost:
```suggestion
    /// Determines how often we send messages. i.e. "Every N frames" where N is message_send_rate
```

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,146 @@
...
+    message_send_rate: u8,
+    /// Number of frames behind the simulation is. Usually 0 or 1 if the ecs system is keeping up.
```
TimonPost:
```suggestion
    /// Number of frames behind the simulation is usually, 0 or 1 if the ECS system is keeping up
```

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+impl NetworkSimulationTime {
+    /// Returns the simulation frame numbers needed to be run this game frame.
+    pub fn sim_frames_to_run(&self) -> RangeInclusive<u32> {
+        (self.frame_number + 1 - self.frame_lag)..=self.frame_number
+    }
+
+    /// Determines whether or not to send a message in the current frame based on the
+    /// message_send_rate.
```
TimonPost:
```suggestion
    /// `message_send_rate`
```

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub fn should_send_message_this_frame(&self) -> bool {
+        self.should_send_message(self.frame_number)
+    }
+
+    /// Determines whether or not to send a message based on the message_send_rate.
```
TimonPost:
```suggestion
    /// Determines whether or not to send a message based on the `message_send_rate`
```

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub fn should_send_message(&self, frame: u32) -> bool {
+        frame % self.message_send_rate as u32 == 0
+    }
+
+    /// Bumps the frame number.
```
TimonPost:
```suggestion
    /// Bumps the frame number
```

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub fn increment_frame_number(&mut self) {
+        self.frame_number += 1;
+        self.elapsed_duration -= self.per_frame_duration;
+        self.frame_lag += 1;
+    }
+
+    /// Resets the frame lag.
```
TimonPost:
```suggestion
    /// Resets the frame lag
```

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub fn reset_frame_lag(&mut self) {
+        self.frame_lag = 0;
+    }
+
+    /// Increase the elapsed_duration by a given duration.
```
TimonPost:
```suggestion
    /// Increase the `elapsed_duration` by the given duration
```

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub fn update_elapsed(&mut self, duration: Duration) {
+        self.elapsed_duration += duration;
+    }
+
+    /// Returns the current simulation frame number.
```
TimonPost:
```suggestion
    /// Returns the current simulation frame number
```

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub fn frame_number(&self) -> u32 {
+        self.frame_number
+    }
+
+    /// Total duration since the last simulation frame.
```
TimonPost:
```suggestion
    /// Returns the total duration since the last simulation frame
```

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub fn elapsed_duration(&self) -> Duration {
+        self.elapsed_duration
+    }
+
+    /// The duration between each simulation frame. This number is calculated when a frame rate
+    /// is set.
```
TimonPost:
```suggestion
    /// is set
```

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub fn per_frame_duration(&self) -> Duration {
+        self.per_frame_duration
+    }
+
+    /// The rate at which messages should be sent over the network.
+    /// i.e. "Every N frames" where N is message_send_rate.
```
TimonPost:
```suggestion
    /// i.e. 'Every N frames' where N is `message_send_rate`.
```

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub fn message_send_rate(&self) -> u8 {
+        self.message_send_rate
+    }
+
+    /// The number of frames behind the simulation is. Usually 0 or 1 if the ECS frame rate is
```
TimonPost:
```suggestion
    /// The number of frames behind the simulation is usually, 0 or 1 if the ECS frame rate is
```

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,146 @@
...
+        self.frame_lag
+    }
+
+    /// Set the rate at which the network simulation progresses. Specified in hertz (frames/second).
+    pub fn with_sim_frame_rate(mut self, new_rate: u32) -> Self {
+        self.per_frame_duration = Duration::from_secs(1) / new_rate;
+        self
+    }
+
+    /// Set the rate which messages are sent. Specified as a "every N frames" where N is new_rate.
```
TimonPost:
```suggestion
    /// Set the rate which messages are sent. Specified as an 'every N frames' where N is `new_rate`.
```

amethyst_network/src/simulation/transport/laminar.rs
```diff
@@ -0,0 +1,154 @@
...
+                    (DeliveryGuarantee::Reliable, OrderingGuarantee::None) => {
+                        Packet::reliable_unordered(addr, message.payload.to_vec())
+                    }
+                    (DeliveryGuarantee::Reliable, OrderingGuarantee::Sequenced(stream_id)) => {
+                        Packet::reliable_sequenced(addr, message.payload.to_vec(), Some(stream_id))
+                    }
+                    (DeliveryGuarantee::Reliable, OrderingGuarantee::Ordered(stream_id)) => {
+                        Packet::reliable_ordered(addr, message.payload.to_vec(), Some(stream_id))
+                    }
+                    _ => unreachable!("Unreliable/Ordered combination is unsupported."),
```
TimonPost:
```suggestion
                    (delivery_guarantee, ordering_guarantee) => unreachable!("{}/{} combination is unsupported.", delivery_guarantee,ordering_guarantee),
```

TimonPost: works for all kind of not supported combinations

amethyst_network/src/simulation/transport/socket.rs
```diff
@@ -0,0 +1,6 @@
+use crate::simulation::guarantees::{DeliveryGuarantee, OrderingGuarantee};
+
+/// Trait defining the interface the generic type of NetworkResource requires.
```
TimonPost:
```suggestion
/// Trait defining the interface for the generic type that `NetworkSimmulationResource` requires.
```

amethyst_network/src/simulation/transport/udp.rs
```diff
@@ -0,0 +1,154 @@
...
+    net::{SocketAddr, UdpSocket},
+};
+
+/// Use this network bundle to add the various underlying UDP network systems to your game.
+pub struct UdpNetworkBundle {
+    recv_buffer_size_bytes: usize,
+}
+
+impl UdpNetworkBundle {
+    pub fn new(recv_buffer_size_bytes: usize) -> Self {
```
TimonPost: just `recv_buffer_size`

amethyst_network/src/simulation/transport/udp.rs
```diff
@@ -0,0 +1,154 @@
...
+        let socket = net.socket.as_mut().unwrap();
+        for message in messages {
+            for addr in addrs.drain(..) {
+                match (message.delivery, message.ordering) {
+                    (DeliveryGuarantee::Unreliable, OrderingGuarantee::None) => {
+                        if let Err(e) = socket.send_to(&message.payload, addr) {
+                            error!("There was an error when attempting to send packet: {:?}", e);
+                        }
+                    }
+                    _ => panic!("Only Unreliable/Unordered guarantees supported by UdpSocket"),
```
TimonPost:
```suggestion
                    (delivery_guarantee, ordering_guarantee) => unreachable!("{}/{} combination is unsupported.", delivery_guarantee,ordering_guarantee),

```

examples/net_server/main.rs
```diff
@@ -1,103 +1,124 @@
...
-    shrev::ReaderId,
+    shrev::{EventChannel, ReaderId},
     utils::application_root_dir,
     Result,
 };
-
 use log::info;
 
-use std::time::Duration;
+// You'll likely want to use a type alias for any place you specify the NetworkResource<T> so
```
TimonPost:
```suggestion
// You'll likely want to use a type alias for any place you specify the `NetworkSimulationResource<T>` so
```

examples/net_client/main.rs
```diff
@@ -54,20 +51,24 @@ impl SpamSystem {
...
-
-                conn.queue(packet);
-            }
+    type SystemData = (
+        Read<'a, NetworkSimulationTime>,
+        Read<'a, Time>,
+        Write<'a, NetworkResourceImpl>,
+    );
+    fn run(&mut self, (sim_time, time, mut net): Self::SystemData) {
+        // Use method "sim_time.sim_frames_to_run()" to determine if the system should send a
```
TimonPost:
```suggestion
        // Use method `sim_time.sim_frames_to_run()` to determine if the system should send a
```

amethyst_network/Cargo.toml
```diff
@@ -22,15 +23,9 @@ profiler = [ "thread_profiler/thread_profiler" ]
...
-serde = { version = "1", features = ["derive"] }
-shrev = "1.0"
-shred = "0.7"
-bincode = "1.0"
-log = "0.4.6"
-uuid = { version = "0.7.1", features = ["v4","serde"] }
+amethyst_core = { path = "../amethyst_core", version = "0.7" }
+amethyst_error = { path = "../amethyst_error", version = "0.2" }
+bytes = "0.4"
+laminar = "0.3"
```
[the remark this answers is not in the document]
jstnlef: Fun fact, cargo treats this exactly the same as with the `^`. https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

amethyst_network/src/simulation/message.rs
```diff
@@ -0,0 +1,21 @@
...
+/// NetworkSystem.
+pub struct Message {
+    pub(crate) payload: Bytes,
+    pub(crate) delivery: DeliveryGuarantee,
+    pub(crate) ordering: OrderingGuarantee,
+}
+
+impl Message {
+    /// Creates and returns a new Message.
+    pub fn new(payload: &[u8], delivery: DeliveryGuarantee, ordering: OrderingGuarantee) -> Self {
```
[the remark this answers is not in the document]
jstnlef: The expectation I have is that the end user will never be making this structure directly. This will only be made through the `NetworkSimulationResource`.

amethyst_network/src/simulation/transport/udp.rs
jstnlef: I really do like putting the units on the names when I can. That way everyone knows that this is the size in bytes

jstnlef:
To address the comments:
1. Made constants for those names.
2. So, what I was going for, though I'm not sure if it's entirely desirable, is that the implementer can use the various methods on `NetworkSimulationResource` (`send, send_*, etc`) to specify the reliability guarantees he/she desires for particular messages. Then, if the backing protocol specified doesn't support the desired guarantees, it will panic. Alternatively we could just log errors but I don't know. I still feel that if someone is trying to use `send_reliable` and using a UDP socket, it should fail fast with a reasonable error message.

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,252 @@
...
+            socket: None,
+            server_addr: None,
+            clients: Vec::new(),
+            messages: VecDeque::new(),
+            is_server: true,
+        }
+    }
+
+    /// Update and return the `NetworkSimulationResource` with a bound socket
+    pub fn with_socket(mut self, socket: S) -> Self {
```
mvlabat: What happens if we call this method when there is a connection with a peer up and running?
Do we want to allow setting new sockets after the resource has been initialized? In case we do not, should we consider a builder there?

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,148 @@
...
+
+impl NetworkSimulationTime {
+    /// Returns the simulation frame numbers needed to be run this game frame.
+    pub fn sim_frames_to_run(&self) -> RangeInclusive<u32> {
+        (self.frame_number + 1 - self.frame_lag)..=self.frame_number
+    }
+
+    /// Determines whether or not to send a message in the current frame based on the
+    /// `message_send_rate`
+    pub fn should_send_message_this_frame(&self) -> bool {
```
mvlabat: Should we consider a separate queue for immediately sent messages? I believe users would want to send some messages as soon as players do any "high priority" actions (for instance movement or shooting actions, so they are processed by server ASAP and get lag compensated properly) and some in batches (rotating a player, chat messages etc). With current design users would need to maintain their custom separate queues for batched messages and set `message_send_rate` to 1, so they can send immediate messages as well

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,148 @@
...
+            sim_time.increment_frame_number();
+        }
+    }
+}
+
+/// Resource to track the state of the network simulation separately from the ECS frame timings
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct NetworkSimulationTime {
+    /// The current simulation frame
+    frame_number: u32,
```
mvlabat: Do we want users to allow setting custom frame number in order to let them replay certain number of frames?
I believe this could be handy when using interpolation techniques for multiplayer. Not sure though if it's the right place to take this into account..

amethyst_network/Cargo.toml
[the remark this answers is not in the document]
TimonPost: oke, I didn't say anything then :)

amethyst_network/src/simulation/resource.rs
jstnlef: That's a good point. We likely will want to be able to arbitrarily set and drop the socket. For example, you may wish to only have the socket connected while connected to a game server instance but not otherwise.

amethyst_network/src/lib.rs
```diff
@@ -1,141 +1,5 @@
...
-
-    #[test]
-    fn can_deserialize_event() {
-        let result =
-            deserialize_event::<NetPacket<String>>(&[3, 0, 0, 0, 0, 0, 0, 0, 97, 98, 99]).unwrap();
-
-        assert_eq!(result.content(), &"abc".to_string());
-    }
-}
+pub mod simulation;
```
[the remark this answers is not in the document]
jstnlef: I actually would prefer that these are accessed through `amethyst::network::simulation`. The idea is that all of the utilities to help build a networked simulation will exist here. Whereas, for example, an http client library or other utilities would live in `clients::http` or something like that.

amethyst_network/src/simulation/events.rs
```diff
@@ -0,0 +1,13 @@
+use bytes::Bytes;
+use std::net::SocketAddr;
+
+/// Events which can be received from the network.
+#[derive(Debug)]
+pub enum NetworkSimulationEvent {
```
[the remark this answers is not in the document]
jstnlef: I expect that these will be used exclusively when we're receiving events when running a network simulation (multiplayer game) with the expectation that there will be other network utilities in this library (like an http client library)

amethyst_network/src/simulation/message.rs
```diff
@@ -0,0 +1,21 @@
+use super::guarantees::{DeliveryGuarantee, OrderingGuarantee};
+use bytes::Bytes;
+
+/// Structure used to hold message payloads before they are consumed and sent by an underlying
+/// NetworkSystem.
+pub struct Message {
```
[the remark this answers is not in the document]
jstnlef: #[derive(Debug, PartialEq, Eq)]

amethyst_network/src/simulation/timing.rs
jstnlef: That's not a bad idea. I like the idea of having an immediate messages queue. 👍

amethyst_network/src/simulation/timing.rs
jstnlef: I would like to say that this is specifically keeping track of the current time (analogous to the Time struct for frame timings). Any sort of storing prior state for rollback/replay would likely exist separate from this. Though I could be convinced otherwise.

amethyst_network/src/simulation/message.rs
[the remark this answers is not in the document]
TimonPost: Yea, good idea, no direct need for the others.

amethyst_network/src/simulation/timing.rs
jstnlef: Added 👍

amethyst_network/src/simulation/transport/udp.rs
TimonPost: Alright, can agree to that, however, buffers are most likely filled with bytes and no other types.

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,192 @@
...
+        self.messages.push_back(message);
+    }
+
+    /// Drains the messages queue and returns the drained messages
+    pub fn drain_messages(&mut self, mut filter: impl FnMut(&mut Message) -> bool) -> Vec<Message> {
+        let mut drained = Vec::with_capacity(self.messages.len());
+        let mut i = 0;
+        while i != self.messages.len() {
+            if filter(&mut self.messages[i]) {
+                let message = self.messages.remove(i);
```
mvlabat: `self.messages.remove(i)`
Ouch, that may be painful performance-wise.
Until `drain_filter` is available for `VecDeque` and stabilized, let's do it at least in two iterations with `.iter().filter(filter)` and `.retain(filter)`? Or what about introducing a real separate queue for immediate messages? That would let us avoid redundant iterations for `OnTick` messages each frame at all.

p.s. but thanks for implementing `UrgencyRequirement` anyway ;)

examples/net_client/main.rs
```diff
@@ -1,50 +1,51 @@
...
 use std::time::Duration;
 
+// You'll likely want to use a type alias for any place you specify the NetworkResource<T> so
+// that, if changed, it will only need to be changed in one place.
+type NetworkResourceImpl = NetworkSimulationResource<LaminarSocket>;
+
 fn main() -> Result<()> {
     amethyst::start_logger(Default::default());
 
+    let socket = LaminarSocket::bind("0.0.0.0:3455").expect("Should bind.");
```
mvlabat: Nitpick: don't use a dot at the end of `expect` messages :)
https://doc.rust-lang.org/src/core/result.rs.html#998

Rust puts a colon right after the message and debug-prints an error

jstnlef: That's a good point. I didn't know that.

amethyst_network/src/simulation/requirements.rs
```diff
@@ -0,0 +1,30 @@
...
+    /// Messages must all be delivered and returned to the client in the order they were sent.
+    /// Client receives 1, 2, 3, 4, 5, 6
+    ReliableOrdered,
+}
+
+/// Specification of urgency of the sending of a message. Typically we'll want to send messages
+/// on simulation tick but the option to send messages immediately is available.
+#[derive(Copy, Clone, Debug, PartialOrd, PartialEq, Eq)]
+pub enum UrgencyRequirement {
+    OnTick,
```
TimonPost: docs

amethyst_network/src/simulation/requirements.rs
```diff
@@ -0,0 +1,30 @@
...
+    Immediate,
```
TimonPost: docs

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,192 @@
...
+    client::Client,
+    message::Message,
+    requirements::{DeliveryRequirement, UrgencyRequirement},
+    transport::socket::Socket,
+};
+use std::{collections::VecDeque, net::SocketAddr};
+
+/// Resource serving as the owner of the underlying socket and the queue of messages to be sent.
+pub struct NetworkSimulationResource<S: Socket> {
+    pub(crate) socket: Option<S>,
```
TimonPost: I prefer to have this since we do use that at different places. 

```rust
pub(crate) fn get_socket_mut() -> &mut S { }
```

But that is up to you since we only call it internally.

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,192 @@
...
+    is_server: bool,
+    server_addr: Option<SocketAddr>,
+    clients: Vec<Client>,
+    pub(crate) messages: VecDeque<Message>,
```
TimonPost: same with this one

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,192 @@
...
+            socket: None,
+            server_addr: None,
+            clients: Vec::new(),
+            messages: VecDeque::new(),
+            is_server: true,
+        }
+    }
+
+    /// Add a number of trusted clients to the `NetworkSimulationResource` for use as a server
+    pub fn with_trusted_clients(mut self, clients: &[SocketAddr]) -> Self {
```
TimonPost: This might be unnecessary in the future. I would like to move this functionality out of our library. Let the user decide what to do with connections. Ignore, accept, deny them. Because in most cases we would never know what our client IP's would be.

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,192 @@
...
+        self.socket.is_some()
+    }
+
+    /// Returns the server address if one was set
+    pub fn server_addr(&self) -> Option<SocketAddr> {
+        self.server_addr
+    }
+
+    /// Returns whether or not this `NetworkSimulationResource` was created as a 'server'
+    pub fn is_server(&self) -> bool {
```
TimonPost: do we want `is_client` for convenience, I guess people could do !server()?

amethyst_network/src/simulation/transport.rs
```diff
@@ -0,0 +1,186 @@
...
+use std::net::SocketAddr;
+
+/// Shared set up code for implementations of `NetworkSendSystem`s
+pub fn run_network_send_system<T: Socket>(
+    net: &mut NetworkSimulationResource<T>,
+    sim_time: &NetworkSimulationTime,
+    mut handle_send: impl FnMut(&mut T, SocketAddr, &Message) -> (),
+) {
+    // If no socket configured, this system should be a no-op.
+    if net.socket.is_none() {
```
TimonPost: `has_socket()`

amethyst_network/src/simulation/transport.rs
```diff
@@ -0,0 +1,186 @@
...
+    }
+}
+
+/// Shared set up code for implementations of `NetworkRecvSystem`s
+pub fn run_network_recv_system<T: Socket>(
+    net: &mut NetworkSimulationResource<T>,
+    mut handle_recv: impl FnMut(&mut T) -> (),
+) {
+    // If no socket configured, this function should be a no-op.
+    if net.socket.is_none() {
```
TimonPost: `has_socket()`

amethyst_network/src/simulation/transport.rs
```diff
@@ -0,0 +1,186 @@
...
+    if messages.is_empty() {
+        return;
+    }
+
+    // If we're a server, we need to broadcast messages to all connected clients. If we're
+    // just a client, we're just going to send to our designated server.
+    let addrs = if net.is_server() {
+        net.clients()
+            .iter()
+            .map(|client| client.addr())
```
TimonPost: `.map(|client| handle_send(socket, *client.addr(), &message);`?

amethyst_network/src/simulation/transport.rs
```diff
@@ -0,0 +1,186 @@
...
+            .collect::<Vec<SocketAddr>>()
+    } else {
+        match net.server_addr() {
+            None => panic!("Expected to send messages but no server destination address!"),
+            Some(server_addr) => vec![server_addr],
```
TimonPost: `handle_send(socket, *server_addr, &message);`

amethyst_network/src/simulation/transport.rs
TimonPost: We can do that, then we don't have to collect into a vector right?

amethyst_network/src/simulation/transport.rs
TimonPost: we now don't have to create the vector

amethyst_network/src/simulation/transport.rs
```diff
@@ -0,0 +1,186 @@
...
+        }
+    };
+
+    let socket = net.socket.as_mut().expect("A socket should be configured.");
+    for message in messages {
```
TimonPost: we can remove if the above comments are handled

amethyst_network/src/simulation/resource.rs
jstnlef: You're probably right. Especially because this will likely have to be made public for people to write their own protocol system implementations

amethyst_network/src/simulation/resource.rs
jstnlef: Could add for convenience. Wouldn't hurt.

amethyst_network/src/simulation/resource.rs
jstnlef: Yeah that's probably fine. Would be pretty straight forward to implement and is likely faster

amethyst_network/src/simulation/resource.rs
jstnlef: Well that's an interesting question. For example, if you had a matchmaking service, before the server code starts running, you would probably know all of the users in the lobby.

TimonPost: Right, but people might as well leave and join during the game after the lobby. So if we would use all IP's from the lobby, we can't have people joining during the game.

amethyst_network/src/simulation/transport/udp.rs
```diff
@@ -0,0 +1,134 @@
+//! Network systems implementation backed by the UDP network protocol.
+use crate::simulation::transport::run_network_recv_system;
```
TimonPost: move this in import below

amethyst_network/src/simulation/transport/udp.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+}
+
+impl<'a, 'b> SystemBundle<'a, 'b> for UdpNetworkBundle {
+    fn build(
+        self,
+        _world: &mut World,
+        builder: &mut DispatcherBuilder<'_, '_>,
+    ) -> Result<(), Error> {
+        builder.add(
```
TimonPost: we could use method chaining here

jstnlef: Method chaining? `DispatcherBuilder::add` doesn't return a value.

jstnlef: Ah there's a `with` method that allows for that. I'll use that.

jstnlef: Ah actually I can't do that because it requires that I own the builder but I only have a mutable reference.

amethyst_network/src/simulation/transport.rs
```diff
@@ -0,0 +1,176 @@
...
+        let clients = net.clients().to_vec();
+        let socket = net.get_socket_mut().expect("A socket should be configured");
+        clients.iter().map(|client| client.addr()).for_each(|addr| {
+            for message in messages.iter() {
+                handle_send(socket, addr, message);
+            }
+        });
+    } else {
+        let server_addr = net.server_addr();
+        let socket = net.get_socket_mut().expect("A socket should be configured");
```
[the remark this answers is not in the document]
jstnlef: I actually couldn't figure out how to do this because it requires that I have a mutable reference to the net. Once I get one of those, I can't call `net.server_addr()` or `net.clients()`

amethyst_network/src/simulation/resource.rs
```diff
@@ -115,8 +117,9 @@ impl<S: Socket> NetworkSimulationResource<S> {
         payload: &[u8],
         delivery: DeliveryRequirement,
         timing: UrgencyRequirement,
+        stream_id: Option<u8>,
```
TimonPost: Instead of the parameter, can we add it to `DeliveryRequirement` like : `DeliveryRequirement::ReliableOrdered(Option<u8>)`

amethyst_network/src/simulation/transport/udp.rs
TimonPost: Ahh, oke!

amethyst_network/src/simulation/resource.rs
jstnlef: Yeah I guess it is only useful on ordered connections.

jstnlef: Actually, to that point, why do we have a stream option on `sequenced` packets in Laminar?

amethyst_network/src/simulation/resource.rs
jstnlef: Yeah that's fair. I'll put more thought into this. It might even be worth lifting this concept up into a higher abstraction level.

amethyst_network/src/simulation/resource.rs
TimonPost: Because ordering and sequencing are done on separate streams. And by default 255 or something like that is used, unless specifying `Some(id)`

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,210 @@
...
+        self.messages.push_back(message);
+    }
+
+    /// Returns true if there are messages enqueued to be sent.
+    pub fn has_messages(&self) -> bool {
+        !self.messages.is_empty()
+    }
+
+    /// Drains the messages queue and returns the drained messages
+    pub fn drain_messages(&mut self, mut filter: impl FnMut(&mut Message) -> bool) -> Vec<Message> {
```
TimonPost: We drain messages based on the given priority. What do you think of the idea to use a [priority queue](https://crates.io/crates/priority-queue) instead?

jstnlef: Would prefer using the std lib rather than bringing in a dependency for a single data structure. That is, unless we profile and find that it is not performing adequately.

amethyst_network/src/simulation/resource.rs
jstnlef: What I mean is, what problem is the stream_id solving for sequenced messages? For ordering it makes sense because we don't want some messages to wait to be delivered if received out of order. However, nothing waits on sequenced messages so as long as they are handled separately from the Ordered messages they shouldn't need a separate logical stream right?

jstnlef: I thought about it and answered my own question. Will asd 👍

amethyst_network/src/simulation/resource.rs
TimonPost: We might as well consider writing our own, by just copying some of their code. Shouldn't be that many lines.

jaynus:
Would it be possible to re-implement or fork the bytes library to utilize smallvec's? I think performance would be hugely increased if we could drop the number of message allocations happening across this.

jstnlef:
@jaynus before I go and fork bytes to reimplement the storage strategy, it would be nice to have hard numbers proving that it's going to be a problem through profiling. It can also be done as a separate (future) optimization PR. I'm mostly trying to get people onboard with the API and general design. Optimization can be done later.

jaynus:
I mean, thousands of heap allocation a frame (on a server) or hundreds (on a client) vs. stack allocations is a pretty no-brainer optimization that is pretty standard and well known; but if we want to wait for numbers for it, thats cool. I can understand it being super early so we just don't want to go down those rabbit holes yet.

amethyst_network/src/simulation/resource.rs
```diff
@@ -0,0 +1,210 @@
...
+    message::Message,
+    requirements::{DeliveryRequirement, UrgencyRequirement},
+    transport::socket::Socket,
+};
+use std::{collections::VecDeque, net::SocketAddr};
+
+/// Resource serving as the owner of the underlying socket and the queue of messages to be sent.
+pub struct NetworkSimulationResource<S: Socket> {
+    socket: Option<S>,
+    is_server: bool,
```
jaynus: On a higher level design note of those with clients and servers and remote connections. As discussed in the other comment, there is the issue of knowing clients and active game/lobby joining. I'd also like the bring up the P2P scenario; currently this is pretty strictly designed to a client-server architecture. It may be worth making sure to keep in consideration a strictly authoritative based approach as well. This would be something akin:
- Any engine can create entities
- Engines can give away certain authorities to other engines (update, etc)
- Engines can take back authority on any entities they own.

This type of mechanism allows for distributing some entity updates and ownership either between multiple servers, or giving the clients entity authority over a single entity; allowing them to just update their 'player' entity.  

This would require stepping this away from teh design of a `client` and `server`, and instead to a `matchmaker` and `engine`. A matchmaker can also be an engine, but strictly handles providing dispatching to an `engine` for requesting synchronization and authority from.

jstnlef: That is a fair point. I'm likely conflating levels of abstraction by baking this information in here. Presumably, we'd want a higher level to make the determination of which `SocketAddr`s to send messages to and have this resource strictly own the socket and queue messages for the underlying systems to churn through.

amethyst_network/src/simulation/transport/laminar.rs
```diff
@@ -0,0 +1,140 @@
...
+struct LaminarNetworkSendSystem;
+
+impl<'s> System<'s> for LaminarNetworkSendSystem {
+    type SystemData = (
+        Write<'s, SimulationTransportResource<LaminarSocket>>,
+        Read<'s, NetworkSimulationTime>,
+    );
+
+    fn run(&mut self, (mut net, sim_time): Self::SystemData) {
+        run_network_send_system(net.deref_mut(), sim_time.deref(), |socket, message| {
```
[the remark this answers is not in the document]
jstnlef: If `Rust` had inheritance, I would have done it that way. I feel like this is a decent way to share code amongst the various systems without being too restrictive to any particular shape. Am open to suggestions.

amethyst_network/src/simulation/transport/laminar.rs
```diff
@@ -0,0 +1,140 @@
...
+            let packet = match message.delivery {
+                DeliveryRequirement::Unreliable => {
+                    Packet::unreliable(message.destination, message.payload.to_vec())
+                }
+                DeliveryRequirement::UnreliableSequenced(stream_id) => {
+                    Packet::unreliable_sequenced(
+                        message.destination,
+                        message.payload.to_vec(),
```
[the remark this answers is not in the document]
jstnlef: Probably worth looking into on the laminar side 👍

amethyst_network/src/simulation/transport.rs
```diff
@@ -0,0 +1,143 @@
+//! This module holds the underlying system implementations for each of the various transport
+//! protocols. One important thing to note if you're implementing your own, the underlying sockets
+//! MUST be non-blocking in order to play nicely with the ECS scheduler.
+mod resource;
```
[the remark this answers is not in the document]
jstnlef: enter? Oh do you mean a new line?

amethyst_network/src/simulation/transport/resource.rs
```diff
@@ -0,0 +1,165 @@
...
+                i += 1;
+            }
+        }
+        drained
+    }
+}
+
+impl<S: Socket> Default for SimulationTransportResource<S> {
+    fn default() -> Self {
+        panic!(
```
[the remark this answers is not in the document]
jstnlef: So I did this because it seems like specs requires that resources implement `Default`. However, I don't want there to be a default implementation of the resource. Basically, I used this to enforce that the end user must provide an implementation with `.with_resource()`

amethyst_network/src/simulation/transport/udp.rs
```diff
@@ -0,0 +1,133 @@
...
+        );
+    }
+}
+
+pub struct UdpNetworkRecvSystem {
+    recv_buffer: Vec<u8>,
+}
+
+impl UdpNetworkRecvSystem {
+    pub fn with_buffer_capacity(size: usize) -> Self {
```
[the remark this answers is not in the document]
jstnlef: I think I might leave it. Specifically because it's not the capacity of the `UdpNetworkRecvSystem`. It's the capacity of the receive buffer

amethyst_network/src/simulation/transport/udp.rs
```diff
@@ -0,0 +1,133 @@
...
+            }
+        });
+    }
+}
+
+impl Socket for UdpSocket {
+    fn default_requirement() -> DeliveryRequirement {
+        DeliveryRequirement::Unreliable
+    }
+}
```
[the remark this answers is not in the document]
jstnlef: 100%. There's a whole slew of tests that still need to be written before this is merged.

jstnlef:
One thing I also did last night, to @jaynus's point is to remove the connection handling from this PR. The idea is that the connection management can easily be built on top of this and slims down the amount of code necessary for this layer.

jstnlef:
I also did a squash because it was getting to be near 50 commits

amethyst_network/src/simulation/transport/udp.rs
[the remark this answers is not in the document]
TimonPost: Do you need help with that?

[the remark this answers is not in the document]
TimonPost: I might be able to free up some time, and will get me more familiar with this system as well. Just ping me and we can spit some work up

amethyst_network/src/simulation/transport/laminar.rs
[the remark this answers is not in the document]
TimonPost: Will have to think about that, understand your ideas behind it

amethyst_network/src/simulation/transport.rs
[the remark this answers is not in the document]
TimonPost: yea :), above the mod and between te comments

amethyst_network/src/simulation/transport/udp.rs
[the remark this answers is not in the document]
jstnlef: I mean if you'd like to write up some tests, by all means 👍. Otherwise, I have no problems doing it.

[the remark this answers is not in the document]
TimonPost: I will see if there is some time to be shifted ;)

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,172 @@
...
+    pub fn frame_number(&self) -> u32 {
+        self.frame_number
+    }
+
+    /// Returns the total duration since the last simulation frame
+    pub fn elapsed_duration(&self) -> Duration {
+        self.elapsed_duration
+    }
+
+    /// The duration between each simulation frame. This number is calculated when a frame rate
```
TimonPost:
```suggestion
    /// Returns the duration between each simulation frame. This number is calculated when a frame rate
```

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,172 @@
...
+        self.per_frame_duration
+    }
+
+    /// The rate at which messages should be sent over the network.
+    /// i.e. 'Every N frames' where N is `message_send_rate`.
+    pub fn message_send_rate(&self) -> u8 {
+        self.message_send_rate
+    }
+
+    /// The number of frames behind the simulation is. This will usually be 0 or 1 if the ECS system
```
TimonPost:
```suggestion
    /// Returns the number of frames which the simulation is behind. This will usually be 0 or 1 if the ECS system
```

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,172 @@
...
+        self.frame_lag
+    }
+
+    /// Set the rate at which the network simulation progresses. Specified in hertz (frames/second).
+    pub fn with_sim_frame_rate(mut self, new_rate: u32) -> Self {
+        self.per_frame_duration = Duration::from_secs(1) / new_rate;
+        self
+    }
+
+    /// Set the rate which messages are sent. Specified as "every N frames" where N is new_rate.
```
TimonPost:
```suggestion
    /// Set the rate which messages are sent. Specified as 'every N frames' where N is new_rate.
```

amethyst_network/src/simulation/transport/laminar.rs
```diff
@@ -0,0 +1,190 @@
...
+    pub fn new(socket: Option<LaminarSocket>) -> Self {
+        Self { socket }
+    }
+
+    /// Return a mutable reference to the socket if there is one configured.
+    pub fn get_mut(&mut self) -> Option<&mut LaminarSocket> {
+        self.socket.as_mut()
+    }
+
+    /// Set the bound socket to the `LaminarSocketResource`
```
TimonPost:
```suggestion
    /// Sets the bound socket to the `LaminarSocketResource`.
```

amethyst_network/src/simulation/transport/laminar.rs
```diff
@@ -0,0 +1,190 @@
...
+    pub fn set_socket(&mut self, socket: LaminarSocket) {
+        self.socket = Some(socket);
+    }
+
+    /// Drops the socket from the `LaminarSocketResource`
```
TimonPost:
```suggestion
    /// Drops the socket from the `LaminarSocketResource`.
```

amethyst_network/src/simulation/transport/resource.rs
```diff
@@ -0,0 +1,234 @@
+use crate::simulation::{
+    message::Message,
+    requirements::{DeliveryRequirement, UrgencyRequirement},
+};
+use std::{collections::VecDeque, net::SocketAddr};
+
+/// Resource serving as the owner of the queue of messages to be sent. This resource also serves
+/// as the interface for other systems to send messages
```
TimonPost:
```suggestion
/// as the interface for other systems to send messages.
```

amethyst_network/src/simulation/transport/resource.rs
```diff
@@ -0,0 +1,234 @@
...
+pub struct TransportResource {
+    messages: VecDeque<Message>,
+}
+
+impl TransportResource {
+    /// Create a new `TransportResource`
```
TimonPost:
```suggestion
    /// Creates a new `TransportResource`.
```

amethyst_network/src/simulation/transport/resource.rs
```diff
@@ -0,0 +1,234 @@
...
+    pub fn send(&mut self, destination: SocketAddr, payload: &[u8]) {
+        self.send_with_requirements(
+            destination,
+            payload,
+            DeliveryRequirement::Default,
+            UrgencyRequirement::OnTick,
+        );
+    }
+
+    /// Create a `Message` with the default guarantees provided by the `Socket` implementation and
```
TimonPost:
```suggestion
    /// Create a `Message` with the default guarantees provided by the `Socket` implementation and
```
```suggestion
    /// Creates a `Message` with the default guarantees provided by the `Socket` implementation and
```

amethyst_network/src/simulation/transport/resource.rs
```diff
@@ -0,0 +1,234 @@
...
+    /// pushes it onto the messages queue to be sent immediately.
```
TimonPost:
```suggestion
    /// Pushes it onto the messages queue to be sent immediately.
```

amethyst_network/src/simulation/transport/resource.rs
```diff
@@ -0,0 +1,234 @@
...
+    pub fn send_immediate(&mut self, destination: SocketAddr, payload: &[u8]) {
+        self.send_with_requirements(
+            destination,
+            payload,
+            DeliveryRequirement::Default,
+            UrgencyRequirement::Immediate,
+        );
+    }
+
+    /// Create and queue a `Message` with the specified guarantee
```
TimonPost:
```suggestion
    /// Creates and queue a `Message` with the specified guarantee.
```

amethyst_network/src/simulation/transport/tcp.rs
```diff
@@ -0,0 +1,276 @@
...
+use std::{
+    collections::HashMap,
+    io::{self, Read as IORead, Write as IOWrite},
+    net::{SocketAddr, TcpListener, TcpStream},
+    ops::{Deref, DerefMut},
+};
+
+const CONNECTION_LISTENER_SYSTEM_NAME: &str = "connection_listener";
+
+/// Use this network bundle to add the various underlying tcp network systems to your game.
```
TimonPost:
```suggestion
/// Use this network bundle to add the TCP transport layer to your game.
```

amethyst_network/src/simulation/transport/tcp.rs
```diff
@@ -0,0 +1,276 @@
...
+impl<'s> System<'s> for TcpNetworkSendSystem {
+    type SystemData = (
+        Write<'s, TransportResource>,
+        Write<'s, TcpNetworkResource>,
+        Read<'s, NetworkSimulationTime>,
+    );
+
+    fn run(&mut self, (mut transport, mut net, sim_time): Self::SystemData) {
+        let messages = transport.messages_to_send(|_| sim_time.should_send_messages());
+        for message in messages.iter() {
```
TimonPost:
```rust
for message in transport.messages_to_send(|_| sim_time.should_send_messages()).iter() {
}
```
I prefer this to be inlined

amethyst_network/src/simulation/transport/tcp.rs
```diff
@@ -0,0 +1,276 @@
...
+        world.insert(TcpNetworkResource::new(
+            self.listener,
+            self.recv_buffer_size_bytes,
+        ));
+        Ok(())
+    }
+}
+
+/// System to listen for incoming connections and cache them to the resource.
+pub struct TcpConnectionListenerSystem;
```
TimonPost: This can be private right?

amethyst_network/src/simulation/transport/tcp.rs
```diff
@@ -0,0 +1,276 @@
...
+    if let Err(e) = stream.write(&message.payload) {
+        error!(
+            "There was an error when attempting to send message: {:?}",
+            e
+        );
+    }
+}
+
+/// System to receive messages from all open `TcpStream`s.
+pub struct TcpNetworkRecvSystem {
```
TimonPost: This can be private right

amethyst_network/src/simulation/transport/tcp.rs
```diff
@@ -0,0 +1,276 @@
...
+            recv_buffer: vec![0; recv_buffer_size_bytes],
+        }
+    }
+
+    /// Return a mutable reference to the listener if there is one configured.
+    pub fn get_mut(&mut self) -> Option<&mut TcpListener> {
+        self.listener.as_mut()
+    }
+
+    /// Set the bound listener to the `TcpNetworkResource`
```
TimonPost:
```suggestion
    /// Sets the bound listener to the `TcpNetworkResource`.
```

amethyst_network/src/simulation/transport/tcp.rs
```diff
@@ -0,0 +1,276 @@
...
+impl TcpNetworkResource {
+    pub fn new(listener: Option<TcpListener>, recv_buffer_size_bytes: usize) -> Self {
+        Self {
+            listener,
+            streams: HashMap::new(),
+            recv_buffer: vec![0; recv_buffer_size_bytes],
+        }
+    }
+
+    /// Return a mutable reference to the listener if there is one configured.
```
TimonPost:
```suggestion
    /// Returns a mutable reference to the listener if there is one configured.
```

amethyst_network/src/simulation/transport/tcp.rs
```diff
@@ -0,0 +1,276 @@
...
+    pub fn get_mut(&mut self) -> Option<&mut TcpListener> {
+        self.listener.as_mut()
+    }
+
+    /// Set the bound listener to the `TcpNetworkResource`
+    pub fn set_listener(&mut self, listener: TcpListener) {
+        self.listener = Some(listener);
+    }
+
+    /// Drops the listener from the `TcpNetworkResource`
```
TimonPost:
```suggestion
    /// Drops the listener from the `TcpNetworkResource`.
```

amethyst_network/src/simulation/transport/tcp.rs
```diff
@@ -0,0 +1,276 @@
...
+    pub fn drop_listener(&mut self) {
+        self.listener = None;
+    }
+
+    /// Returns the stream associated with the `SocketAddr` or creates one if one doesn't exist.
```
TimonPost:
```suggestion
    /// Returns the stream associated with the `SocketAddr` or creates one if it doesn't exist.
```

amethyst_network/src/simulation/transport/udp.rs
```diff
@@ -0,0 +1,166 @@
...
+    bundle::SystemBundle,
+    ecs::{DispatcherBuilder, Read, System, World, Write},
+    shrev::EventChannel,
+};
+use amethyst_error::Error;
+use bytes::Bytes;
+use log::error;
+use std::{io, net::UdpSocket};
+
+/// Use this network bundle to add the various underlying UDP network systems to your game.
```
TimonPost:
```suggestion
/// Use this network bundle to add the UDP transport layer to your game.
```

amethyst_network/src/simulation/transport/udp.rs
```diff
@@ -0,0 +1,166 @@
...
+pub struct UdpNetworkBundle {
+    socket: Option<UdpSocket>,
+    recv_buffer_size_bytes: usize,
+}
+
+impl UdpNetworkBundle {
+    pub fn new(socket: Option<UdpSocket>, recv_buffer_size_bytes: usize) -> Self {
```
TimonPost: doc string

amethyst_network/src/simulation/transport/udp.rs
```diff
@@ -0,0 +1,166 @@
...
+            UdpNetworkRecvSystem::with_buffer_capacity(self.recv_buffer_size_bytes),
+            NETWORK_RECV_SYSTEM_NAME,
+            &[NETWORK_SIM_TIME_SYSTEM_NAME],
+        );
+        world.insert(UdpSocketResource::new(self.socket));
+        Ok(())
+    }
+}
+
+pub struct UdpNetworkSendSystem;
```
TimonPost: private?

amethyst_network/src/simulation/transport/udp.rs
```diff
@@ -0,0 +1,166 @@
...
+                        break;
+                    }
+                }
+            }
+        });
+    }
+}
+
+/// Resource to own the UDP socket.
+pub struct UdpSocketResource {
```
TimonPost: private

amethyst_network/src/simulation/transport/tcp.rs
```diff
@@ -0,0 +1,276 @@
...
+        Write<'s, TransportResource>,
+        Write<'s, TcpNetworkResource>,
+        Read<'s, NetworkSimulationTime>,
+    );
+
+    fn run(&mut self, (mut transport, mut net, sim_time): Self::SystemData) {
+        let messages = transport.messages_to_send(|_| sim_time.should_send_messages());
+        for message in messages.iter() {
+            match message.delivery {
+                DeliveryRequirement::ReliableOrdered(Some(_)) => {
```
TimonPost:
```rust
  DeliveryRequirement::ReliableOrdered(stream)  | DeliveryRequirement::Default => {
                   if (stream.is_some()) {
                          warn!("Streams are not supported by TCP and will be ignored.");
                    }
                    write_message(message, &mut net);
                }
```

Maybe move this in a single match

amethyst_network/src/simulation/transport/tcp.rs
```diff
@@ -0,0 +1,276 @@
...
+    pub fn get_or_create_stream(&mut self, addr: SocketAddr) -> &mut TcpStream {
+        self.streams.entry(addr).or_insert_with(|| {
+            let s = TcpStream::connect(addr).unwrap();
+            s.set_nonblocking(true).expect("Setting non-blocking mode");
+            s.set_nodelay(true).expect("Setting nodelay");
+            s
+        })
+    }
+
+    /// Drop the stream with the given `SocketAddr`. This will be called when a peer seems to have
```
TimonPost:
```suggestion
    /// Drops the stream with the given `SocketAddr`. This will be called when a peer seems to have
```

amethyst_network/src/simulation/transport/udp.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+}
+
+impl UdpSocketResource {
+    /// Create a new instance of the `UdpSocketResource`
+    pub fn new(socket: Option<UdpSocket>) -> Self {
+        Self { socket }
+    }
+
+    /// Return a mutable reference to the socket if there is one configured.
```
TimonPost:
```suggestion
    /// Returns a mutable reference to the socket if there is one configured.
```

amethyst_network/src/simulation/transport/udp.rs
```diff
@@ -0,0 +1,166 @@
...
+    pub fn get_mut(&mut self) -> Option<&mut UdpSocket> {
+        self.socket.as_mut()
+    }
+
+    /// Set the bound socket to the `UdpSocketResource`
```
TimonPost:
```suggestion
    /// Sets the bound socket to the `UdpSocketResource`.
```

amethyst_network/src/simulation/transport/udp.rs
```diff
@@ -0,0 +1,166 @@
...
+    pub fn set_socket(&mut self, socket: UdpSocket) {
+        self.socket = Some(socket);
+    }
+
+    /// Drops the socket from the `UdpSocketResource`
```
TimonPost:
```suggestion
    /// Drops the socket from the `UdpSocketResource`.
```

examples/net_client/main.rs
```diff
@@ -1,26 +1,32 @@
...
     Result,
 };
 use log::info;
+use std::net::TcpListener;
 use std::time::Duration;
 
 fn main() -> Result<()> {
     amethyst::start_logger(Default::default());
 
+    //    let listener = TcpListener::bind("0.0.0.0:3455")?;
```
TimonPost: don't forget to remove this

amethyst_network/src/simulation/transport/tcp.rs
jstnlef: This can't be done in a single match because DeliveryRequirement::Default doesn't bind stream so it doesn't compile

amethyst_network/src/simulation/transport/tcp.rs
jstnlef: Not if we want to allow people to bind and set the tcplistener after initializing the system

kabergstrom:
I have not looked at amethyst_network prior, so this is all from a fresh pair of eyes. Sorry if some of this has already been discussed - I have not kept up with the conversation.

Some issues I've found:
- Hosts and connections are identified with a SockAddr. It would be preferable if there was a way to customize connection identifiers.
- There is no connection buffer, i.e. "intro message" post handshake or similar. These are really useful to avoid state machines for handling connections in the application layer. It's really annoying to build connection identifiers (if you don't want to fix above issue) on top of a connection layer without this.
- The API for sending messages seems inverted in its control flow from what I'd expect in a game state sync layer. There's a send rate, so I would expect that on each send tick, the transport layer would tell me how much data it can send that tick, and I would get to fill up the send budget with messages. In the API in this PR, from what I can tell, I feed the system messages every frame without really knowing the current transport budget?

jstnlef:
@kabergstrom and I talked and came up with the following action items
1. Move away from SocketAddr as the connection identifiers. (Though the current default impls (tcp, udp, laminar) will still need them) I'll figure out something there.
2. Add methods to TransportResource related to send budgets and implement the systems for tcp, udp, laminar to update them.
Will work on these next.

amethyst_network/src/simulation/transport/laminar.rs
```diff
@@ -0,0 +1,186 @@
...
+
+struct LaminarNetworkPollSystem;
+
+impl<'s> System<'s> for LaminarNetworkPollSystem {
+    type SystemData = Write<'s, LaminarSocketResource>;
+
+    fn run(&mut self, mut socket: Self::SystemData) {
+        socket
+            .get_mut()
+            .map(|socket| socket.manual_poll(Instant::now()));
```
TimonPost: So I am wondering whether we should do something with time zones. Like maybe want to poll based on UTC. Because we are having multiple systems running over the globe.

amethyst_network/src/simulation/transport/tcp.rs
TimonPost: okay

TimonPost:
I think that those are about to be my last comments. If there are more changes. We can start rounding this PR up.

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,178 @@
...
+pub struct NetworkSimulationTime {
+    /// The current simulation frame
+    frame_number: u32,
+    /// Accumulated duration since last simulation frame
+    elapsed_duration: Duration,
+    /// Duration per frame
+    per_frame_duration: Duration,
+    /// Determines how often we send messages. i.e. "Every N frames" where N is message_send_rate
+    message_send_rate: u8,
+    /// Number of frames behind the simulation is. This will usually be 0 or 1 if the ECS system
```
TimonPost:
```suggestion
    /// Number of frames behind the simulation is. This will usually be 0 or 1 if the ECS system
```
```suggestion
    /// The number of frames the simulation is behind. This will usually be 0 or 1 if the ECS system
```

amethyst_network/src/simulation/timing.rs
```diff
@@ -0,0 +1,178 @@
...
+        self.message_send_rate
+    }
+
+    /// Returns the number of frames which the simulation is behind. This will usually be 0 or 1 if the ECS system
+    /// is keeping up.
+    pub fn frame_lag(&self) -> u32 {
+        self.frame_lag
+    }
+
+    /// Set the rate at which the network simulation progresses. Specified in hertz (frames/second).
```
TimonPost:
```suggestion
    /// Set the rate at which the network simulation progresses. Specified in hertz (frames/second).
```
```suggestion
    /// Sets the rate at which the network simulation progresses. Specified in hertz (frames/second).
```

amethyst_network/src/simulation/transport/laminar.rs
jstnlef: Eh. That can be future work.

jstnlef:
bors r=jstnlef

jstnlef:
bors r=TimonPost

TimonPost:
bors r=TimonPost

TimonPost:
bors r=TimonPost

PR #1918, opened 2019-09-03
[CHANGED] Fix screen_to_world, rename to screen_to_world_point

jaynus:
## Description

screen_to_world was incorrect for perspective cameras. Additionally, the naming conventionand data provided was insufficient really for proper 3d screen raycasting. 

This adds screen_ray(), screen_to_world_point(), and a Ray struct  for their use.
This allows users to get a proper Ray from the screen, as well as get projected
points along this ray.  These functions are camera agnostic.

## Additions

- `renderer::Ray` structure.
- `Camera::screen_ray()`
- Caches the inverse of the projection matrix as an optimization, accessible via `as_inverse_matrix()`
## Modifications

- `Camera::screen_to_world` renamed `Camera::screen_to_world_point`. Now takes a Point3, where the z-level is the distance from the ray origin (camera)
- `as_matrix_mut` changed to `set_matrix` and no mutable references can be obtained.

## PR Checklist


amethyst_rendy/src/camera.rs
```diff
@@ -32,12 +34,22 @@ impl Orthographic {
     /// * `z_far` - The distance between the viewer (the origin) and the furthest face of the cuboid parallel to the xy-plane. If used for a 3D rendering application, this is the furthest clipping plane.
     ///
     /// The projection matrix is right-handed and has a depth range of 0 to 1
+    ///
+    /// * panics if left equals right, bottom equals top or near equals right
```
Frizi:
```suggestion
    /// * panics if `left` equals `right`, `bottom` equals `top` or `z_near` equals `z_far
````

File: amethyst_rendy/src/camera.rs [hunk 2]
```diff
@@ -396,36 +437,74 @@ impl Projection {
...
+        Ray {
+            origin: near_t,
+            direction: (far_t.coords - near_t.coords).normalize(),
+        }
+    }
 
-        Point3::from_homogeneous(vector).expect("Vector is not homogeneous")
+    /// Transforms the provided (X, Y, Z) screen coordinate into world coordinates.
+    /// This method fires a ray from the camera in its view direction, and returns the Point at `screen_position.z`
+    /// distance from the camera origin.
```
Frizi:
```suggestion
    /// world space distance from the camera origin.
```

amethyst_core/src/geometry.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub fn intersect_plane(&self, plane: &Plane) -> Point3<f32> {
+        let diff = self.origin - plane.position;
+        let prod1 = diff.coords.dot(&plane.normal);
+        let prod2 = self.direction.dot(&plane.normal);
+        let prod3 = prod1 / prod2;
+
+        Point3::from(self.origin.coords - self.direction.scale(prod3))
+    }
+
+    /// Returns the ray `Point` at the given distance
```
Frizi:
```suggestion
    /// Returns a `Point` along the ray at a distance `t` from it's origin.
```

ishitatsuyuki:
You don't need to cache the inverse matrix, the projection matrix has a fast non-matrix invert function I think. So please add support for inverse_transform_point (called unproject_point for Projective3) instead.

amethyst_core/src/geometry.rs
```diff
@@ -0,0 +1,107 @@
...
+    /// Create a plane which is facing along the Z-Axis at the provided coordinate.
+    pub fn with_z(z: f32) -> Self {
+        Self {
+            normal: Vector3::new(0.0, 0.0, 1.0),
+            position: Vector3::new(0.0, 0.0, z),
+        }
+    }
+}
+
+/// A ray structure providing a position and direction.
```
Frizi:
```suggestion
    /// A Ray represents and infinite half-line starting at `origin` and going in specified unit length `direction`.
```

amethyst_rendy/src/camera.rs
```diff
@@ -396,36 +437,74 @@ impl Projection {
...
-            Projection::Orthographic(ref mut s) => s.as_matrix_mut(),
-            Projection::Perspective(ref mut s) => s.as_matrix_mut(),
+            Projection::Orthographic(ref s) => s.far(),
+            Projection::Perspective(ref s) => s.far(),
         }
     }
 
-    /// Transforms position from screen space to camera space
-    pub fn screen_to_world(
+    /// Returns a `Ray` structure for the provided screen position.
```
Frizi:
```suggestion
    /// Returns a `Ray` going out form the camera through provided screen position. The ray origin lies on camera near plane.
```

jaynus:
> You don't need to cache the inverse matrix, the projection matrix has a fast non-matrix invert function I think. So please add support for inverse_transform_point (called unproject_point for Perspective3) instead.

I will be progressing towards that in the coming weeks. This is just a pass to fix some issues and begin moving in a new direction. I have my complaints and reservations about nalgebra, and especially with the design choices of the transform_* function family and their Perspective/Projection limitations. 

That was a really long way of saying I'm not looking to have feature parity/function parity with nalgebra.

amethyst_rendy/src/camera.rs
```diff
@@ -49,7 +61,12 @@ impl Orthographic {
         matrix[(1, 3)] = -(top + bottom) / (top - bottom);
         matrix[(2, 3)] = -z_near / (z_far - z_near);
 
-        Self { matrix }
+        Self {
+            matrix,
+            inverse_matrix: matrix
```
ishitatsuyuki: While singular projection matrix probably never occurs for correct use cases, I'd prefer deferring this error (panic) until the user perform any operation that actually requires the inverse matrix.

amethyst_rendy/src/camera.rs
```diff
@@ -170,9 +187,19 @@ impl Orthographic {
...
+    pub fn set_matrix(&mut self, matrix: Matrix4<f32>) {
+        self.matrix = matrix;
+        self.inverse_matrix = matrix.try_inverse().expect("Matrix not invertible")
+    }
+
+    /// Returns a reference to the inner matrix representation of this projection.
     #[inline]
-    pub fn as_matrix_mut(&mut self) -> &mut Matrix4<f32> {
-        &mut self.matrix
+    pub fn as_inverse_matrix(&self) -> &Matrix4<f32> {
```
ishitatsuyuki: Even if you don't intend to achieve feature parity with nalgebra, I think cached inverse matrix should be kept as an implemention detail for now.

amethyst_rendy/src/camera.rs
```diff
@@ -435,12 +516,8 @@ impl Projection {
...
         camera_transform: &Transform,
     ) -> Point2<f32> {
-        let render_matrix: Matrix4<f32> =
-            amethyst_core::math::convert(*camera_transform.global_matrix());
-
-        let f = render_matrix * self.as_matrix();
-
-        let screen_pos = f.transform_point(&world_position);
+        let screen_pos =
+            (camera_transform.global_matrix() * self.as_matrix()).transform_point(&world_position);
```
ishitatsuyuki: You need to bump the nalgebra requirement to `0.18.1` to ensure that `transform_point` returns correct value for everyone.

amethyst_rendy/src/camera.rs
Frizi: I'd argue this is an unnecessary complication. All matrix based cameras must have an invertible matrix. Having to check that lazily has also additional cost. There are multiple places where camera inverse is necessary, so having it precomputed is very sensible in my opinion.

ishitatsuyuki:
To summarize the discussion from internal Discord: jaynus does not like the nalgebra API because it has oversights from the perspective of a graphic programmer. To list up, there's weird interaction with storage, that doesn't work well with SIMD, and specialized calculation optimization being implemented in a non-intuitive way. Therefore he does not want to replicate the API.

amethyst_rendy/src/camera.rs
Frizi: The matrix inversion is needed to compute the frustum (and likely other places as well). If we want to make it private, those operations (like getting the frustum) needs to be moved into the camera struct as well. I'm not sure if this is a good idea.

amethyst_core/src/geometry.rs
```diff
@@ -0,0 +1,107 @@
...
+//! Geometry helper functionality.
+use nalgebra::{Point3, Vector3};
+
+/// A plane which can be intersected by a ray.
+#[derive(Debug, Copy, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct Plane {
+    /// The forward normal of the plane.
+    pub normal: Vector3<f32>,
+    /// The origin/point of the plane.
+    pub position: Vector3<f32>,
```
Frizi: This plane representation is a bit wasteful and contains redundant data. Planes usually just have a normal vector and a single float value bias (implicit plane).

amethyst_rendy/src/camera.rs
```diff
@@ -524,19 +699,29 @@ impl Camera {
         ))
     }
 
+    /// Creates a `Projection::CustomMatrix` with the matrix provided.
+    pub fn custom_matrix(
+        matrix: Matrix4<f32>,
+        clip_fn: &'static (dyn Fn(&CustomMatrix) -> (f32, f32) + Send + Sync),
```
Frizi: This looks really bad. I'd say that custom matrix projection should simply not have `near` and `far` as part of the API. Other option is just to use the formulas from perspective projection. Notice that in arbitrary matrices the clip planes doesn't have to be perpendicular. You can in fact force the near plane to be in any orientation you want, so just a float doesn't really cut it.

distransient:
@Frizi if you're happy with these changes we can merge this one for 0.13

amethyst_rendy/src/camera.rs
```diff
@@ -388,44 +482,99 @@ impl Projection {
...
+        }
+    }
+
+    /// Returns the near-clip value of the current camera.
+    pub fn near(&self) -> f32 {
+        match *self {
+            Projection::Orthographic(ref s) => s.near(),
+            Projection::Perspective(ref s) => s.near(),
+            Projection::CustomMatrix(ref s) => {
+                panic!("Custom Matrix does not support near and far field projection!")
```
Frizi: This is still not great. The existance of such API might mean that you can't use the `CustomMatrix` variant in practice, because some code unknowingly uses those methods. The `fn near` and `fn far` should either be removed or some general equasion should be applied on the `CustomMatrix` variant. If it's removed, you can always match on your specific camera type you are interested in and get the value out.

Note that `fn far` might soon not even work on `Perspective` (or at least not on every one), as there is such thing as "infinite far plane".

distransient:
IT LIVES

PR #1919, opened 2019-09-06
Clarify UiImage:PartialTexture Coordinates

### LINKED ISSUE #1906: [BUG] PartialTexture Texture Coordinates are not as the variable names suggest
## Description

When using PartialTexture, the TextureCoordinates do not map to the place expected. 
The [TextureCoordinates](https://docs.amethyst.rs/stable/amethyst_rendy/sprite/struct.TextureCoordinates.html) is documented as containing the properties `left`, `right`, `bottom`,  `top`. The constructor for a partial texture is [implemented as parsing](https://github.com/amethyst/amethyst/blob/c454d4bd54b604fcc94395d069f619a96b8f3ec2/amethyst_ui/src/prefab.rs#L404) this as ```PartialTexture(<image>, left, right, bottom top)```

But when constructed, the extracted partial image doesn't respect the "left, right, bottom top" ordering, instead seeming to be "left top right bottom"

## Reproduction Steps

Looking at the UI example, we see the [file example.ron](https://github.com/amethyst/amethyst/blob/c454d4bd54b604fcc94395d069f619a96b8f3ec2/examples/assets/ui/example.ron#L62) contains:
```
image: PartialTexture(File("texture/test_texture.png", ("IMAGE", ())),0.34,0.0,1.0,0.333),
```

This retrieves [this texture](https://github.com/amethyst/amethyst/blob/c454d4bd54b604fcc94395d069f619a96b8f3ec2/examples/assets/texture/test_texture.png):
test_texture

and extracts the blue "Hi" in the top right.

Yet the set of coordinates used: 
```
 L    R   B   T
0.34,0.0,1.0,0.333
```
doesn't map to those coordinates because the given coordinates span 0.3 width and 0.6 height, but the actual sample is 0.6 width and 0.3 height.

A similar issue applies when writing into the UiImage::PartialTexture coordinates at runtime, where the following mapping was needed to make it work.
```
     Right => Top
       Top => Bottom
    Bottom => Right

 L    T   R   B
0.34,0.0,1.0,0.333
```

I'm not quite sure where the problem lies, but there is something rather fishy with how TextureCoordinates are handled.

## Your Environment

1. Arch
2. Reasonably up to date
3. d4b047582c32a250efd43c7c5d8a4f90bef23e24

## Additional Context

Some extra documentation on the coordinate system for PartialTexture would be nice. Ie: the image is indexed from top left.

mrhatman:
Fixes #1906 

## Description

Fixes the texture coordinates issue and makes UiImage:PartialTexture a struct enum so the members are explicit. 

## Modifications

-  UiImage:PartialTexture update
-  Prefab updated to match
## PR Checklist

### DIFF
--- a/amethyst_ui/src/image.rs
+++ b/amethyst_ui/src/image.rs
@@ -1,6 +1,6 @@
 use amethyst_assets::Handle;
 use amethyst_core::ecs::{Component, DenseVecStorage};
-use amethyst_rendy::{sprite::TextureCoordinates, SpriteRender, Texture};
+use amethyst_rendy::{SpriteRender, Texture};
 
 /// Image used UI widgets, often as background.
 #[derive(Debug, Clone, PartialEq)]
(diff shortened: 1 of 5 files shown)


mrhatman:
The bug existed in sprite too. I has been fixed.

distransient:
bors r=distransient,azriel91

PR #1950, opened 2019-09-29
HOTFIX changes to amethyst_tiles 

jaynus:
In a quick nutshell, this fixes the cases of:
- Allocation was incorrect for asymmetric map dimensions causing a panic on render. This is fixed by padding allocation for the morton encoder (2^n  boundary)
- Adds a FlatEncoder to optimize for space due to the padding needed for morton encoding (And I added it for testing)
- The iterator Ord was incorrect, causing hte y-axis to be one short.

This is a hotfix for these items.

Added: 
`FlatEncoder` added to amethyst_tiles for flat linear encoding which is optimized for space.

Fixed: 
- `TileMap` was not allocating enough space for to compensate for morton encoding alignment. This means that
all tilemap allocation must occur on a 2^n boundary on the correct dimensions boundary. This fixes the case of assemetric tilemaps.

- Fixed an iterator oppsy where the Ord was incorrect, causing the y-axis to be one tile short.

## PR Checklist

### DIFF
--- a/amethyst_tiles/src/lib.rs
+++ b/amethyst_tiles/src/lib.rs
@@ -19,6 +19,8 @@ pub use pass::{
     DrawTiles2D, DrawTiles2DBounds, DrawTiles2DBoundsDefault, DrawTiles2DDesc, RenderTiles2D,
 };
 
+use amethyst_core::math::Vector3;
+
 /// Trait to provide generic access to various encoding schemas. All tile storages use this to encode their coordinates
 /// and provide different spatial encoding algorithms for efficiency.
 pub trait CoordinateEncoder: 'static + Clone + Default + Send + Sync {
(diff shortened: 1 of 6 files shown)


amethyst_tiles/src/lib.rs
```diff
@@ -31,4 +33,48 @@ pub trait CoordinateEncoder: 'static + Clone + Default + Send + Sync {
...
+
+    #[inline]
+    fn encode(&self, x: u32, y: u32, z: u32) -> Option<u32> {
+        Some((z * self.dimensions.x * self.dimensions.y) + (y * self.dimensions.x) + x)
+    }
+
+    #[inline]
+    fn decode(&self, idx: u32) -> Option<(u32, u32, u32)> {
+        let z = idx / (self.dimensions.x * self.dimensions.y);
+        let idx = idx - (z * self.dimensions.x * self.dimensions.y);
```
azriel91: (disclaimer: i'm completely unknowledgeable in this area)
am guessing integer division from the previous line means `idx` may not be 0. my brain wants that as a comment because it couldn't figure it out the first time :sweat_smile:

amethyst_tiles/src/map.rs
```diff
@@ -327,7 +329,53 @@ fn to_tile(transform: &Matrix4<f32>, coord: &Vector3<f32>) -> Option<Point3<u32>
...
+            (0..dimensions.y).into_par_iter().for_each(|y| {
+                for z in 0..dimensions.z {
+                    let _ = map.get(&Point3::new(x, y, z)).unwrap();
+                }
+            });
+        });
+    }
+
+    #[test]
+    pub fn assymetric_maps() {
```
azriel91: typo: `asymmetric`

amethyst_tiles/src/morton/mod.rs
```diff
@@ -79,11 +81,23 @@ pub fn morton_decode_intr_3d(morton: u32) -> (u32, u32, u32) {
     )
 }
 
+fn next_power_of_2(mut v: u32) -> u32 {
```
azriel91: whoa what is this magic. Let's see what happens when `v` begins as 32:

```rust
// using base 2
// v: 0010 0000

v -= 1;       // 0001 1111

              // 0001 1111 bitwise OR
              // 0000 1111
v |= v >> 1;  // 0001 1111

              // 0001 1111 bitwise OR
              // 0000 0111
v |= v >> 2;  // 0001 1111

              // 0001 1111 bitwise OR
              // 0000 0001
v |= v >> 4;  // 0001 1111
v |= v >> 8;  // ..
v |= v >> 16; // ..

v + 1         // 0010 0000
```

hm, this doesn't seem right. did I do something wrong? what's it meant to do?

amethyst_tiles/src/lib.rs
jaynus: If idx = 0, z = 0 and that second line will be 0. Hence 0,0,0. This is fine as it's 0 / integer and not integer / 0

amethyst_tiles/src/morton/mod.rs
jaynus: This causes the dimensions to round UP to the next bitwise 2^n. Because Morton encoding is bit interleaving, we need to work within these bounds for indexing. So for example, it would round 20,20 up to 32,32.

Your example is correct, because it is a 2^n already, so stays at 32.

Frizi: There is a builtin for that. https://doc.rust-lang.org/std/primitive.u32.html#method.next_power_of_two

azriel91:
bors r=Frizi,azriel91

Issue #1652, opened 2019-05-27
[FEATURE MODIFICATION] Rendy/gfx backend log verbosity


bamling:
## Feature

Logging for Rendy and the chosen gfx backend.

## Change

The log level for Rendy and the chose gfx backend (e.g. Vulkan) should by default be set to a less verbose level (`amethyst::LogLevelFilter::Warn` or `amethyst::LogLevelFilter::Error`).

## Reason

The current logs are a bit overly verbose and make it harder to find actual errors in your application. It's possible to configure a less verbose level yourself but that requires the following block of code:

```rust
amethyst::Logger::from_config(Default::default())
    .level_for("gfx_backend_vulkan", amethyst::LogLevelFilter::Warn)
    .level_for("rendy_factory::factory", amethyst::LogLevelFilter::Warn)
    .level_for(
        "rendy_memory::allocator::dynamic",
        amethyst::LogLevelFilter::Warn,
    )
    .level_for(
        "rendy_graph::node::render::pass",
        amethyst::LogLevelFilter::Warn,
    )
    .level_for("rendy_graph::node::present", amethyst::LogLevelFilter::Warn)
    .level_for("rendy_graph::graph", amethyst::LogLevelFilter::Warn)
    .level_for(
        "rendy_memory::allocator::linear",
        amethyst::LogLevelFilter::Warn,
    )
    .level_for("rendy_wsi", amethyst::LogLevelFilter::Warn)
    .start();
```

That's a lot of boiler plate just to achieve more readable logs.

silverweed:
I would like to contribute to this issue.

distransient:
Fixed by #2048

PR #1956, opened 2019-10-01
less verbose default logging for rendy/gfx-backend

tdheff:
## Description

Altered the default logger to reduce the verbosity of gfx-backends and various rendy components

Addresses #1652 

The gfx-backends are all  set to the appropriate level. This does create a code location that will need to be modified for additional backends, but I couldn't see a solution for that which wasn't needlessly complex

## Modifications

- Modified Logger::new_with_config to reduce verbosity

## PR Checklist

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -110,6 +110,19 @@ impl Logger {
             StdoutLog::Off => {}
         }
 
+        logger.dispatch = logger
+            .dispatch
+            .level_for("gfx_backend_empty", LevelFilter::Warn)
+            .level_for("gfx_backend_vulkan", LevelFilter::Warn)
+            .level_for("gfx_backend_dx12", LevelFilter::Warn)
+            .level_for("gfx_backend_metal", LevelFilter::Warn)
+            .level_for("rendy_factory::factory", LevelFilter::Warn)
+            .level_for("rendy_memory::allocator::dynamic", LevelFilter::Warn)
+            .level_for("rendy_graph::node::render::pass", LevelFilter::Warn)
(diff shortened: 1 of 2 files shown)


tdheff:
I wasn't sure if the list of crates in the issue to change the verbosity of was complete or not, so i just added those for now

src/logger.rs
```diff
@@ -110,6 +110,19 @@ impl Logger {
...
+            .level_for("gfx_backend_dx12", LevelFilter::Warn)
+            .level_for("gfx_backend_metal", LevelFilter::Warn)
+            .level_for("rendy_factory::factory", LevelFilter::Warn)
+            .level_for("rendy_memory::allocator::dynamic", LevelFilter::Warn)
+            .level_for("rendy_graph::node::render::pass", LevelFilter::Warn)
+            .level_for("rendy_graph::graph", LevelFilter::Warn)
+            .level_for("rendy_memory::allocator::linear", LevelFilter::Warn)
+            .level_for("rendy_wsi", LevelFilter::Warn);
+
         if let Some(log_gfx_device_level) = config.log_gfx_device_level {
```
distransient: I believe this option `log_gfx_device_level` is outdated so it might be able to be removed, but could you put the rendering logging defaults into a config option like this one which defaults to Warn?

bluecheetah001:
It would be nice to have this behind a config like the old log_gfx_level_device was. Possibly split into a backend_level and a rendy_level.

distransient:
Closing as stale. @tdheff or anyone else, if you'd like to reopen with the requested changes included either open a new PR or just ping me in here

PR #1957, opened 2019-10-01
Fix TileArgs pod value tile_coordinate to be Rgb32Uint

jaynus:
This caused a crash on macos but didnt trigger any validation errors for some reason.

This worked fine on linux/windows vulkan layers, but errors on macos metal backend. This changes it to the correct format of `Rgb32Uint`

## PR Checklist


jaynus:
bors r=distransient

PR #1958, opened 2019-10-01
Tiles pre-compiled spv was incorrect

jaynus:
The Tiles spv file was incorrect and compiled wrong previously. This updated one is correct.


jaynus:
bors r=jaynus

PR #1955, opened 2019-10-01
Maintenance/update dependencies

azriel91:
## Description

Updated dependencies:

```
* itertools:   0.7    -> 0.8
* parking-lot: 0.6    -> 0.9
* image:       0.21.0 -> 0.22.2
* getset:      0.0.7  -> 0.0.8
* sdl2:        0.31.0 -> 0.32.2
* shred:       0.7    -> 0.9
* criterion:   0.2.11 -> 0.3
* glyph_brush: 0.5.0  -> 0.6.0
```

## PR Checklist

By placing an x in the boxes I certify that I have:

- **n/a** Updated the content of the book if this PR would make the book outdated.
- **n/a** Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.
- **n/a** Added unit tests for new code added in this PR.
- [x] Acknowledged that by making this pull request I release this code under an MIT/Apache 2.0 dual licensing scheme.

If this modified or created any rs files:

- **n/a** Ran `cargo +stable fmt --all`
- **n/a** Ran `cargo clippy --all --features "empty"`
- [x] Ran `cargo test --all --features "empty"`

PR #1959, opened 2019-10-01
Update unic-langid to 0.5

zbraniecki:
## Description

I noticed that `amethyst_locale` uses `unic-langid` 0.4, but `fluent` 0.7 (which uses `unic-langid` 0.5).

I probably will want to re-export `langid` from fluent in the future to make it less likely to get desync, but for now, let's synchronize them!

## PR Checklist


azriel91:
haha, wanted to update it as part of #1955, but fluent wasn't released yet.

PR #1962, opened 2019-10-02
Upgrade shred to 0.9 in amethyst_network

aeubanks:


TimonPost:
Thanks for the PR, although it is not needed because it is removed by this [PR](https://github.com/amethyst/amethyst/pull/1917/files)

distransient:
This dependency actually isn't used funnily enough, so it could just be removed in its current state (along with shrev, although its usage [here](https://github.com/amethyst/amethyst/blob/master/amethyst_network/src/connection.rs#L4) should instead refer to the shrev provided through the specs re-exposed in amethyst-core).

@TimonPost what's the state of that rewrite? Should we expect it to be ready for 0.14?

TimonPost:
Ask @jstnlef, I believe it is but could be wrong

distransient:
👍 will close this for now.

PR #1965, opened 2019-10-03
Publish v0.13.1

distransient:
## Description

This increments the patch version of all changed crates and updates the Changelog. Crates which depend on patched crates but have not been changed I've left alone, since the only crates which truly need their hotfixes are `amethyst_locale` and `amethyst_tiles`. After this PR is approved and merged I will begin publishing 0.13.1.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "amethyst"
-version = "0.13.0"
+version = "0.13.1"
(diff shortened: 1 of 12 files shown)


azriel91:
bors try
(fs race condition)

distransient:
Alright. Time for the fun part :D

azriel91:
ship ship :package:

PR #1966, opened 2019-10-04
Fix fluent and unic-langid alignment

zbraniecki:
Due to a mistake in https://github.com/zbraniecki/unic-locale/issues/18 I yanked too many revisions and ended up in a state where either amethyst_locale 0.7.0 or 0.7.1 doesn't work.

I updated all crates and this patch fixes amethyst_locale.

We may want to yank 0.7.1 and release 0.7.2 with this patch.


zbraniecki:
@Ratysz - can you help me with this?

Ratysz:
I would if I could, I'm not even a contributor. I'm testing everything again locally, though.

distransient:
bors cancel

distransient:
bors p=999

distransient:
That's new 
image

PR #1968, opened 2019-10-05
Fix silent shader error and pre-compiled shader. 

jaynus:
This was breaking tiles as they were out of sync and incorrect(ish).

## Description

There was a mismatch between the vertex shader compiled code and regular code, causing the float conversion of the shader to fail. This explicitly makes it occur, fixing the bug. This is a hotfix for a 13.3 release for tiles (yay)

## PR Checklist


distransient:
Why don't you like jokes, bors

bors r+

PR #1986, opened 2019-10-11
Add states_ui example

fkarg:
## Description

Added a new ui example, demonstrating the switching between multiple states, and the loading/removal of associated UI, as it was not demonstrated in a concise way somewhere else. Also demonstrates the possible working of a pause menu.

It contains a whole lot of things I was looking for and could not immediately find.

## Additions

- No API additions.

## Removals

- No API removals.

## Modifications

- No changes to existing structures and functions

## PR Checklist


examples/assets/ui/welcome.ron
```diff
@@ -0,0 +1,25 @@
...
+        width: 20.,
+        height: 20.,
+    ),
+
+    children: [
+
+        Image(
+            transform: (
+                id: "background",
+                // y: 540.,            // <- comment this
```
mrhatman: I am confused by this line.

examples/states_ui/welcome.rs
```diff
@@ -0,0 +1,52 @@
...
+        &mut self,
+        _: StateData<'_, GameData<'_, '_>>,
+        event: StateEvent,
+    ) -> SimpleTrans {
+        match &event {
+            StateEvent::Window(event) => {
+                if is_close_requested(&event) || is_key_down(&event, VirtualKeyCode::Escape) {
+                    Trans::Quit
+                } else if is_mouse_button_down(&event, MouseButton::Left) {
+                    info!("Switching to MainMenu!");
```
mrhatman: I would recommend doing this for each state transition

examples/assets/ui/welcome.ron
fkarg: ah, that was there from a previous version. not sure where the other comment came from.

examples/states_ui/welcome.rs
fkarg: will do.

examples/README.md
```diff
@@ -256,6 +257,12 @@ Renders a custom UI.
 
 ---
 
+## States Example
+
+Simple Example on how to use states and UI (e.g. switching between Screens).
+
```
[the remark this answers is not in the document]
fkarg: ah, yes. Will do.

examples/states_ui/util.rs
```diff
@@ -0,0 +1,25 @@
...
+        error::WrongGeneration,
+        prelude::{Entity, World, WorldExt},
+    },
+};
+
+use std::iter;
+
+/// delete the specified root entity and all of its descendents as specified
+/// by the Parent component and maintained by the ParentHierarchy resource
+// from https://github.com/amethyst/evoli src/utils/hierarchy_util.rs
```
azriel91: Missed a slash

examples/states_ui/main.rs
```diff
@@ -0,0 +1,99 @@
...
+        // makes sense to add it early on
+        .with_bundle(TransformBundle::new())?
+        // this bundle allows us to 'find' the Buttons and other UI elements later on
+        .with_bundle(UiBundle::<StringBindings>::new())?
+        // this allows us to reload '*.ron' files during execution
+        .with_bundle(HotReloadBundle::default())?
+        // without this System, our Program will silently (!) fail when trying to load the 'Game'.
+        // (try it!)
+        // This System takes care of Audio (in this case, the Button audio for hovering/clicking/...
+        .with(Processor::<Source>::new(), "source_processor", &[])
```
azriel91: Hm, there's an `amethyst::audio::AudioBundle` which adds this, as well as the system to play the sound.

examples/states_ui/game.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    fn handle_event(
+        &mut self,
+        _: StateData<'_, GameData<'_, '_>>,
+        event: StateEvent,
+    ) -> SimpleTrans {
+        match &event {
+            StateEvent::Window(event) => {
+                if is_close_requested(&event) || is_key_down(&event, VirtualKeyCode::Escape) {
```
azriel91: The `is_close_requested` should likely return a `Trans::Quit` (it's when the :x: is pressed on a window, as well as alt + f4 I think). Relevant in the credits, game, menu, and paused states.

(realize it's an example, but this part could confuse myself if I were a newcomer)

examples/states_ui/pause.rs
```diff
@@ -0,0 +1,107 @@
...
+                } else if Some(target) == self.exit_to_main_menu_button {
+                    let mut state_transition_event_channel = data
+                        .world
+                        .write_resource::<EventChannel<TransEvent<GameData, StateEvent>>>();
+
+                    log::info!("Exiting Game, Switching to MainMenu!");
+
+                    // this allows us to first 'Pop' this state, and then exchange whatever was
+                    // below that with a new MainMenu state.
+                    state_transition_event_channel.single_write(Box::new(|| Trans::Pop));
```
azriel91: clever :v:, makes me think that instead of having a `State` stack, we should have a state graph, and transitions to indicate which state to go to instead. (so we don't have to hack around jumping two (or more) levels up).

examples/states_ui/pause.rs
```diff
@@ -0,0 +1,107 @@
...
+                } else {
+                    Trans::None
+                }
+            }
+            _ => Trans::None,
+        }
+    }
+
+    fn update(&mut self, data: &mut StateData<GameData>) -> SimpleTrans {
+        data.data.update(&data.world);
```
azriel91: Given `PauseMenuState` implements `SimpleState`, this is done already in the [blanket implementation](https://github.com/amethyst/amethyst/blob/03cac16f94bf569f9c99d5c71db25c43951fe7ff/src/state.rs#L336), so this can be removed, otherwise it would be double-updating in the same tick.

examples/states_ui/util.rs
fkarg: that is actually intentional, as it is no link

examples/states_ui/game.rs
fkarg: Ah, I got confused by this. In my case it still behaves as it should (since my wm does not have a :x: -button), and when trying to close it, it closes no matter what the state. But, as it turns out, my wm just kills it in these instances. Yes, I will change that

examples/states_ui/pause.rs
fkarg: I'm pretty happy with the stack, but an additional graph would be nice. I mean with the stack I can put multiple states on top of each other (since it is much more than just a simple state machine which can ever only be in one state); I don't immediately see how I would do this in the Graph.

examples/states_ui/main.rs
fkarg: Okay, so do you recommend using them instead? How would that be different though?

examples/states_ui/pause.rs
fkarg: Ah, I see. Yes, I will change that then.

examples/states_ui/util.rs
azriel91: ah, initially I meant the first slash, then saw there's a middle slash that isn't there, but I'm guessing both are intentional

examples/states_ui/main.rs
azriel91: Yeap I'd recommend using the bundle -- it's the "thing that knows what loads the sound, and what plays the sound". It's not completely necessary if you simply wish to allow sounds to be loaded but never played (e.g. the button that plays a sound on hover / mouse click)

fkarg: okay, did that just now

azriel91:
bors r=azriel91

fkarg:
alright, nice!

azriel91:
There may be a Rust ecosystem issue:

```
thread '<unnamed>' panicked at 'src/librustc_codegen_ssa/back/write.rs:1548: worker thread panicked', src/librustc/util/bug.rs:37:26
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
error: could not write output to /home/jenkins/workspace/amethyst_PR-1986/target/debug/build/libc-8c5bac30304e52c8/build_script_build-8c5bac30304e52c8.build_script_build.3g43h2vc-cgu.10.rcgu.o: No such file or directory

error: Could not compile `libc`.
```

fkarg:
o.O

fkarg:
where did that happen? the current bors/jenkins build appears to be going smooth

azriel91:
The build before the mergy one, seems to be all good now :v:

fkarg:
Ah, I see. Already running tests :crossed_fingers:

PR #1989, opened 2019-10-13
replace Config::load with Config::load_no_fallback

### LINKED ISSUE #1954: [MAINTENANCE] Replace `Config::load` with `Config::load_no_fallback`?
## Description
This is the trait to which I am referring:

https://github.com/amethyst/amethyst/blob/master/amethyst_config/src/lib.rs#L96-L113

## Reason
The `Config::load` function requires that a concrete value be provided in the case of failure. In many cases, a failure can indicate a larger problem and it should be left to the caller of the function to decide what action to take. Mandating that some concrete value be produced can introduce subtle bugs, like default configs being used when they are unexpected (an error message will be logged yes, but a failed `.unwrap()` with a stack trace is easier to track down).

If the `load` function is removed and the `load_no_fallback` function is renamed to `load`, then this trait will be much more idiomatic. In addition, the blanket implementation here ( https://github.com/amethyst/amethyst/blob/master/amethyst_config/src/lib.rs#L115-L168 ) can be improved to no longer require a `Default` trait bound on `T`. (This is actually the main reason I'm advocating for this change; my config has no reasonable default value, it _must_ be user provided. Because of this, I am forced to implement the `Config` trait manually and hard unwrap in the `load` function.)

## Impact
Unfortunately, this would be a breaking change. The manual fix is easy:

- instances of `Config::load(_)` would be changed to `Config::load(_).unwrap_or_default()`. This will have a nearly identical effect to the code as it stands, but no error message is printed out in the case of failure to load (but since a user provided fallback is provided, no error message is expected).
- instances of `Config::load_no_fallback(_)` would be changed to `Config::load(_)`.

## Additional Information
There are a couple other ways to go about this change that may result in fewer breaking changes, but this seems to be the "right" way to do it, so I wanted to start here first. I would be more than happy to do the work involved with implementing this change, but I wanted to gather some feedback first since this would be a breaking change. Thanks! 🙂

0x6273:
## Description

Closes #1954 

## Modifications

- `Config::load` has been modified to behave like `Config::load_no_fallback`, returning an error when the config fails to load, instead of printing a log message and falling back to the default config values. Fallback is still possible for types with a `Default` implementation, by unwrapping the `Result` with `.unwrap_or_default()`.
- `Config::load_no_fallback` is now deprecated, and is an alias to `Config::load`.
- the `Config` trait no longer requires a `Default` implementation.
- `RenderToWindow`, `WindowBundle`, and `WindowSystem` has had their `from_config_path` methods changed similarly to `Config::load`, returning an error on failure instead of a log message and falling back to the default.

## PR Checklist


distransient:
Could you merge master into your branch? Conflict on changelog.md.

PR #1991, opened 2019-10-13
TileMap to_tile and to_world account for the map's transform

mrhatman:
## Description

The TileMap functions now take an Option<&Transform> to expand the functionality in the case where the entire map is translated. Tests are included.

## Modifications

- TileMaps to_tile and to_world now take an Option<&Transform> that allows them to work if the entire map in translated.

## PR Checklist


mrhatman:
Fixes #1991

amethyst_tiles/src/map.rs
```diff
@@ -42,12 +43,16 @@ pub trait Map {
     /// Convert a tile coordinate `Point3<u32>` to an amethyst world-coordinate space coordinate `Point3<f32>`
     /// This performs an inverse matrix transformation of the world coordinate, scaling and translating using this
     /// maps `origin` and `tile_dimensions` respectively.
-    fn to_world(&self, coord: &Point3<u32>) -> Vector3<f32>;
+    fn to_world(&self, coord: &Point3<u32>, map_transform: Option<&Transform>) -> Vector3<f32>;
```
Xaeroxe: We should enhance the documentation to indicate what this new parameter is used for and how it alters the behavior.

mrhatman: I have done that now. Thanks.

amethyst_tiles/src/map.rs
```diff
@@ -290,13 +301,36 @@ fn create_transform(map_dimensions: &Vector3<u32>, tile_dimensions: &Vector3<u32
...
+fn to_world(
+    transform: &Matrix4<f32>,
+    coord: &Point3<u32>,
+    map_transform: Option<&Transform>,
+) -> Vector3<f32> {
     let coord_f = Point3::new(coord.x as f32, -1.0 * coord.y as f32, coord.z as f32);
-    transform.transform_point(&coord_f).coords
+    if let Some(map_trans) = map_transform {
+        map_trans
+            .global_view_matrix()
```
jaynus: This should use `Transform::global_matrix()`instead of `global_view_matrix().inverse()`, as the view_matrix is already the inverse of the global_matrix, so we are doing 2 inversions here for no reason.

amethyst_tiles/src/map.rs
```diff
@@ -290,13 +301,36 @@ fn create_transform(map_dimensions: &Vector3<u32>, tile_dimensions: &Vector3<u32
...
-fn to_tile(transform: &Matrix4<f32>, coord: &Vector3<f32>) -> Option<Point3<u32>> {
-    let point = Point3::from(*coord);
+fn to_tile(
+    transform: &Matrix4<f32>,
+    coord: &Vector3<f32>,
+    map_transform: Option<&Transform>,
+) -> Option<Point3<u32>> {
+    let point = if let Some(map_trans) = map_transform {
+        map_trans
+            .global_view_matrix()
```
jaynus: see previous about `global_matrix`

mrhatman: This one should be inversed though, so it's only doing one inverse

amethyst_tiles/src/map.rs
mrhatman: Done

PR #1994, opened 2019-10-16
Support for tile u/v flipping with TileMap.

StarArawn:
## Description
Added the ability to flip the uv's for a tile map. This is useful for reusing tiles instead of duplicating them in your sprite sheet facing other ways.

## Additions

Tile trait now has a flip function which returns which dimension the tile should be flipped in.

## Modifications

TileArgs.from_data now flips uv's and accepts a new parameter.

## PR Checklist

### DIFF
--- a/amethyst_tiles/src/map.rs
+++ b/amethyst_tiles/src/map.rs
@@ -20,6 +20,11 @@ pub trait Tile: 'static + Clone + Send + Sync + Default {
     fn tint(&self, coordinates: Point3<u32>, world: &World) -> Srgba {
         Srgba::new(1.0, 1.0, 1.0, 1.0)
     }
+
+    /// Takes an immutable reference to world to process this sprite and return its flip.
+    fn flip(&self, coordinates: Point3<u32>, world: &World) -> (bool, bool) {
+        (false, false)
+    }
 }
 
 /// Trait for providing access to an underlying storage type of a 3-dimensional Tile data. This is abstracted to provide
--- a/amethyst_tiles/src/pass.rs
+++ b/amethyst_tiles/src/pass.rs
@@ -274,6 +274,7 @@ impl<B: Backend, T: Tile, E: CoordinateEncoder, Z: DrawTiles2DBounds> RenderGrou
                             sprite_number,
                             Some(&TintComponent(tile.tint(coord, world))),
                             &coord,
+                            tile.flip(coord, world),
                         )?;
 
                         let (tex_id, this_changed) = textures_ref.insert(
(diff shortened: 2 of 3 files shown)


amethyst_tiles/src/map.rs
```diff
@@ -20,6 +20,11 @@ pub trait Tile: 'static + Clone + Send + Sync + Default {
     fn tint(&self, coordinates: Point3<u32>, world: &World) -> Srgba {
         Srgba::new(1.0, 1.0, 1.0, 1.0)
     }
+
+    /// Takes an immutable reference to world to process this sprite and return its flip.
+    fn flip(&self, coordinates: Point3<u32>, world: &World) -> (bool, bool) {
```
StarArawn: I'm not good at naming stuff. I'm hoping a better suggestion can be made here for the function name.

StarArawn:
I'm not sure if a unit test is required for this, but it seems as though the docs should be updated, but I'm not entirely sure how.

amethyst_tiles/src/map.rs
jaynus: The name is fine, but could you add documentation to this function to specify what the tuple return value should be? It should explicit state that it needs to return (horizontal, vertical)

OvermindDL1:
Traditionally spritemaps for each tile they can be flipped in either axis as well as rotated 4 ways, are these all supported as-is?  Is there a way to set all values at the same time?

jaynus:
I mean, this isn't a Sprite map, it's a 3d nap data representation. There are some inherit constraints we have to apply to get any optimization out of it. If every single tile gets so many parameters they may as well be entities because your losing the optimization with all that branching

OvermindDL1:
This is common for tilemaps though, including at times is coloration.  Spritemaps tend to have even more data.  Encoding the flipping (2 bits) and rotation params (2 bits) is very simple, combined with the hotspot and so forth the extra data is very minimal and can be done entirely in shader or baked directly into the tilemap itself, thus taking no additional space.  Coloration can and is often also baked in to tilemaps for the 4 vertices (not necessarily direct color values but sometimes a pallette texture with an index into it is used, though nowadays I've seen 4 full RGBA colors per tile as well), as these are often used for simple baked lighting and shadows.

Branching is not needed in any of the above.  For colors you just set the color (default white), and for the flip and rotation you just index into an array the rotation params needed, pure math in the machine code with no branches required (also very shader friendly if rendered purely in shaders).

mrhatman:
If you include flipping alone the line x=y the all of the 90 degree rotations will be supported as well.

jojolepro:
Status?

jojolepro:
Closing due to inactivity

PR #2003, opened 2019-10-19
s/Point3/Vector3/ in tiles Map documentation.

ssbr:
## Description

Replace (incorrect) claim in amethyst_tiles::Map of using `Point3<f32>` with the actual type, `Vector3<f32>`.

Map::to_tiles etc should probably actually use Point3 (because it is a point, not an offset), but I figure it's better at least for the docs to be accurate.

## Additions

None.

## Removals

None.

## Modifications

None.

## PR Checklist

I did not do those because I made this from the github web UI, but I am gambling that they wouldn't actually change anything. :)

### DIFF
--- a/amethyst_tiles/src/map.rs
+++ b/amethyst_tiles/src/map.rs
@@ -39,12 +39,12 @@ pub trait Map {
     /// Set the sprite sheet handle which the tile render pass should use for rendering this map.
     fn set_sprite_sheet(&mut self, sprite_sheet: Option<Handle<SpriteSheet>>);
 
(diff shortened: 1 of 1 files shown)


jaynus:
Thanks!

You are correct from a accuracy standpoint; it should technically be a `Point3`. I actually had an iteration where that was the case, hence the documentation.

I changed it to `Vector3` to be inline with the transform system, which operates on vectors everywhere, so ergonomically this removes a layer of conversions.

azriel91:
bors r=jaynus, azriel91

PR #2007, opened 2019-10-20
Store sentry_status.txt in $XDG_CONFIG_HOME/amethyst/

ldesgoui:
Hello,

Currently, building amethyst creates a directory in the home directory and to be honest I'm not fond of it
This PR changes that behavior to adhere to XDG specifications

Bests

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,4 +1,4 @@
-use dirs::home_dir;
+use dirs::config_dir;
 use std::{
     fs::{create_dir_all, read_to_string, remove_file, File},
     io::{stderr, stdin, Write},
@@ -8,7 +8,7 @@ use vergen::{self, ConstantsFlags};
 
 fn main() {
     let amethyst_home =
-        Path::new(&home_dir().expect("Failed to find home directory")).join(".amethyst");
+        Path::new(&config_dir().expect("Failed to find home directory")).join("amethyst");
     match amethyst_home.exists() {
         true => match check_sentry_allowed(&amethyst_home) {
             Some(true) => {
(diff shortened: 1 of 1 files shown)


ldesgoui:
I'm not sure if this is the appropriate place to discuss this, but I disagree with the implementation that exists in `build.rs` (if I have multiple amethyst projects, I can't pick which ones send to sentry; I'm not even sure if `ask_write_user_data_collection` works properly on linux, I surely haven't seen any such warning pop up; if a user were to update `.sentry_status.txt`, it wouldn't reflect on the compilation until amethyst was rebuilt, which is unlikely to happen), feature flags would be a much better fit

0x6273:
I also agree with the current implementation having some issues. Requiring user input in `build.rs` for something that's not really related to the build process seems like it would probably just annoy people. You can probably just delete that function, since at the moment it doesn't even work and just returns `false`.

build.rs
```diff
@@ -8,7 +8,7 @@ use vergen::{self, ConstantsFlags};
 
 fn main() {
     let amethyst_home =
-        Path::new(&home_dir().expect("Failed to find home directory")).join(".amethyst");
+        Path::new(&config_dir().expect("Failed to find home directory")).join("amethyst");
```
0x6273:
```suggestion
        Path::new(&config_dir().expect("Failed to find config directory")).join("amethyst");
```

Or even better, have it continue the build normally instead of panicking if the `.config` dir doesn't exist.

0x6273:
Seems like they deleted their repo. Guessing that's why bors didn't work?

ldesgoui:
woops, my bad.

distransient:
No worries

Issue #2002, opened 2019-10-19
[MAINTENANCE] Arch linux dependency installation line


sahsahae:
## Description
This is wrong: https://github.com/amethyst/amethyst#arch-linux

It should be 

```
$ pacman -Syu grep gcc pkgconfig openssl alsa-lib cmake make python3 freetype2 awk libxcb
```

## Reason

https://wiki.archlinux.org/index.php/System_maintenance#Avoid_certain_pacman_commands

Partial upgrades are bad.

## Impact

Arch linux potentially not booting anymore after shutdown, even if technically it's their own fault, that's probably not something that should be given as a thing to copy paste.

distransient:
Could you make a PR for this?

PR #2012, opened 2019-10-22
Fix Arch Linux installation line in documentation

chaosteil:
## Description

As per Issue #2002, it's not recommended to have pacman perform partial upgrades, and this fixes it.

## Additions

- None

## Removals

- None

## Modifications

- Fix to documentation

## PR Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ If you are compiling on Linux, make sure to install the dependencies below.
 ### Arch Linux
 
 ```
-$ pacman -Sy grep gcc pkgconfig openssl alsa-lib cmake make python3 freetype2 awk libxcb
+$ pacman -Syu grep gcc pkgconfig openssl alsa-lib cmake make python3 freetype2 awk libxcb
 ```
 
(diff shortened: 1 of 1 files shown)


ishitatsuyuki:
Don’t `-Syu` but just `-S`, the user should be already aware of the pacman options.

chaosteil:
How about splitting it up: recommending to do a `pacman -Syuond line?` standalone, and then the `-S <packages>` in the second line?

ishitatsuyuki:
That changes nothing, you don’t need to suggest to a user to update at all.

ishitatsuyuki:
I'd like to note that Arch's philosophy expects the user to read documentation and solve problems on their own. [Reference](https://wiki.archlinux.org/index.php/Arch_Linux#Principles).

However it seems true that users commonly mess up with partial upgrades and in that case I think it's a valid point to play safe. So I'm retracting my opposition.

distransient:
This is a trivial change so I'm just shipping it.
bors r+

PR #2015, opened 2019-10-22
Update system desc derive book chapter with note about passing exprs

Josh-Thompson:
## Description

Minor addition to the book chapter on SystemDesc derives, noting the requirement for quotation marks if the resource to add in an `insert` macro is the result of an expression.

## PR Checklist


Josh-Thompson:
Would have added an example, but I'm unsure how to properly create the code that would go in the `Generated Code:` section.

Josh-Thompson:
The checks died because they hit a "no space left on device".

azriel91:
bors retry

(ran the build that frees disk space)

PR #2017, opened 2019-10-24
[FIX] Fix dimensions assumption error with encoding

jaynus:
## Description

With the changes of pre-allocated dimensions, this change did not occur in the actual encoding schemes as well. Because of this, any encoding beyond z-level 0 was actually incorrect and did not work.

This fixes that, and adds the appropriate unit tests for that test case.

## PR Checklist

### DIFF
--- a/amethyst_tiles/src/lib.rs
+++ b/amethyst_tiles/src/lib.rs
@@ -26,7 +26,7 @@ use amethyst_core::math::Vector3;
 pub trait CoordinateEncoder: 'static + Clone + Default + Send + Sync {
     /// Constructor interface for `Self` which consumes the maps dimensions, which is required for some
     /// encoding types to fit within a given coordinate space.
-    fn from_dimensions(x: u32, y: u32, z: u32) -> Self;
+    fn from_dimensions(dimensions: &Vector3<u32>) -> Self;
 
     /// Encode the provided x, y and z 3-dimensional coordinates into a 1-dimensional array index.
     fn encode(&self, x: u32, y: u32, z: u32) -> Option<u32>;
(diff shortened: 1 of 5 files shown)


amethyst_tiles/src/lib.rs
```diff
@@ -26,7 +26,7 @@ use amethyst_core::math::Vector3;
 pub trait CoordinateEncoder: 'static + Clone + Default + Send + Sync {
     /// Constructor interface for `Self` which consumes the maps dimensions, which is required for some
     /// encoding types to fit within a given coordinate space.
-    fn from_dimensions(x: u32, y: u32, z: u32) -> Self;
+    fn from_dimensions(dimensions: &Vector3<u32>) -> Self;
```
azriel91: Should we just pass `Vector3<u32>` by value? It's [`Copy`](https://nalgebra.org/rustdoc/nalgebra/base/struct.Matrix.html#impl-Copy)

amethyst_tiles/src/lib.rs
```diff
@@ -53,9 +53,9 @@ impl Default for FlatEncoder {
     }
 }
 impl CoordinateEncoder for FlatEncoder {
-    fn from_dimensions(x: u32, y: u32, z: u32) -> Self {
+    fn from_dimensions(dimensions: &Vector3<u32>) -> Self {
         Self {
-            dimensions: Vector3::new(x, y, z),
+            dimensions: *dimensions,
```
azriel91: Then this can be collapsed as well to just `dimensions,`

azriel91:
bors r=distransient,azriel91

PR #2023, opened 2019-10-28
Improve SpriteSheetFormat::import_simple to support grids and lists

### LINKED ISSUE #1997: [FEATURE MODIFICATION] Allow loading of SpriteGrid through ron files
## Feature

`amethyst_rendy::SpriteSheetFormat`

## Change

Make it also accept `SpriteGrid`

## Reason

For large Tilemaps adding hundreds of repetitive ron entries is annoying and error-prone

## Impact

The impact is that it is a breaking change to the ron file, however it is easily fixed by making users specify `SpriteList`.

## Additional Notes

The change seems to rather small, only this part would need to be changed: https://github.com/amethyst/amethyst/blob/master/amethyst_rendy/src/sprite/mod.rs#L481

## Alternatives

The alternative would be to do nothing and let users create their own format. However, since the the `SpriteGrid` is builtin, I expected it to be supported as a `ron` file.

valkum:
## Description

Improve the loader of SpriteSheetFormat to allow loading of Grid definitions in ron files.
Currently you can only specify sprites as `SpriteList` in ron files.
Closes #1997 

## Modifications

- `SpriteSheetFormat::import_simple` now uses the `Sprites` enum instead of using `SpriteList` directly. This is a breaking change in current ron files loaded using SpriteSheetFormat.

## PR Checklist

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -30,7 +30,7 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 - ScreenDimensions now consistently reports window size in physical pixels. ([#1988])
 - `Config::load` now returns an error or failure rather than silently falling back to the default config. Same is true for the `from_config_file` methods on `RenderToWindow`, `WindowBundle`, and `WindowSystem` ([#1989])
 - Adds `get` methods to the underlying net::transport resources ([#2005])
-- Changed `SpriteSheetFormat::import_simple` to allow importing grid based `SpriteSheets` ([#0000])
+- Changed `SpriteSheetFormat::import_simple` to allow importing grid based `SpriteSheets` ([#2023])
   Migration Note: Rons need to wrap their content in either Grid() or List()
 
 ### Deprecated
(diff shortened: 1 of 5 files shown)


docs/CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 - ScreenDimensions now consistently reports window size in physical pixels. ([#1988])
 - `Config::load` now returns an error or failure rather than silently falling back to the default config. Same is true for the `from_config_file` methods on `RenderToWindow`, `WindowBundle`, and `WindowSystem` ([#1989])
 - Adds `get` methods to the underlying net::transport resources ([#2005])
+- Changed `SpriteSheetFormat::import_simple` to allow importing grid based `SpriteSheets` ([#2023])
+  Migration Note: Rons need to wrap their content in either Grid() or List()
```
azriel91: I like the change, but am wondering whether we should use `#[serde(untagged)]` instead. I don't have a strong preference, so am happy to merge in a few days if no one else does

azriel91:
bors try

docs/CHANGELOG.md
valkum: sry for my late reply. Do we have any drawbacks by using `#[serde(untagged)]`?

azriel91: It would leave ambiguity if there are future variants that have similar fields (and if they have the same fields, no way to tell serde which one to use). That said, it does boil down to how likely that's going to be. In my mind, `SpriteList` and `SpriteGrid` sufficiently cover all rectangular sprite use cases.

am happy with the tagged version (it helps discoverability), so shall merge.

Issue #2026, opened 2019-10-30
[FEATURE MODIFICATION] Image load should provide sane defaults for SamplerInfo


valkum:
## Feature/Change/Reason/Impact

The loader for `IMAGE` currently needs a full definition of the `SamplerInfo` struct.
Ideally there should be a default and you only need to specify the values you need different from the defaults values.
The problem now is a very verbose definition in ron files with a lot of repetitive values if you load a lot of spritesheets for example. See rendy_example_scene.ron with 4 times the same struct.

PR #2028, opened 2019-10-30
Fix regression caused by #1908

valkum:
Fix regression in the example RONs.

gfx-hal@0.3.0 added the `normalized` field to `SamplerInfo`. This change was not propagated to the prefabs. 

## PR Checklist

### DIFF
--- a/examples/assets/prefab/renderable.ron
+++ b/examples/assets/prefab/renderable.ron
@@ -57,6 +57,7 @@ Prefab (
                                     ),
                                     comparison: None,
                                     border: (0),
+                                    normalized: true,
                                     anisotropic: On(8),
                                 ),
                             )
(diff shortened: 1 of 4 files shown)


valkum:
Should we add something to Changelog even if this is caused by a change in gfx-hal?

ishitatsuyuki:
That's a good catch. tbh this seems to be gfx-hal having poor defaults (not all backends even support normalized: false), and an option is to fix the defaults on the gfx side.

distransient:
bors try

valkum:
> That's a good catch. tbh this seems to be gfx-hal having poor defaults (not all backends even support normalized: false), and an option is to fix the defaults on the gfx side.

I created #2026 to track the progress with this. Wherever we tackle it.

distransient:
bors try

distransient:
bors cancel

azriel91:
Shall merge this first, we can revert it when #2026 is done.

bors r=distransient, azriel91

Issue #2030, opened 2019-11-02
[BUG] Playing audio causes rendering lags


pustaczek:
## Description

Playing audio with amethyst::audio::output::Output may cause lags ranging from unnoticeable to arbitrarily long, depending on the optimization level and the repeat count. When the sound is about to play, the game hangs and resumes rendering after a very short moment(usually after the sound starts playing, but before it finishes).

The issue is harder to reproduce with lower repeat count and optimization flags, but without --release even using play_once may result in slight pauses on non-gaming hardware.

## Reproduction Steps

1. Open main amethyst repo.
2. Apply the following patch.
3. Run the pong example with `--release --features vulkan,no-slow-safety-checks`.
4. Notice the lag happening every time the ball bounces off a paddle.
5. If the lag is too short, try increasing the number of replays from 30 to e.g. 200, removing the --release flag, or running on weaker hardware.
```diff
diff --git a/examples/pong/audio.rs b/examples/pong/audio.rs
index db209bc0..ee0094b9 100644
--- a/examples/pong/audio.rs
+++ b/examples/pong/audio.rs
@@ -56,7 +56,7 @@ pub fn initialise_audio(world: &mut World) {
 pub fn play_bounce(sounds: &Sounds, storage: &AssetStorage<Source>, output: Option<&Output>) {
     if let Some(ref output) = output.as_ref() {
         if let Some(sound) = storage.get(&sounds.bounce_sfx) {
-            output.play_once(sound, 1.0);
+            output.play_n_times(sound, 1.0, 30);
         }
     }
 }
```

## What You Expected to Happen

Playing sound does not cause rendering lags. Delays caused by loading sound should likely be handled in a different thread or asynchronously, possibly with a way to wait for completion before starting the game.

## What Actually Happened

Noticeable pauses in rendering appear.

## Screenshots and Logs

If the bug turns out hard to reproduce, I could try to prepare a demonstration video but I hope it's reproducible.

## Your Environment

1. Ubuntu
2. 19.10
3. master branch, d2971a21

## Additional Context

The hardware used is a recent laptop without a dedicated graphics card.

distransient:
Running your game without optimizations will always be bad, and copying debuginfo introduces a lot of extra compile time. Try running nalgebra benchmarks with and without --release, and everything in amethyst uses nalgebra. I recommend only using debug builds for gdb/detailed stack traces

pustaczek:
Yes, but this also happens with --release.

[...]

PR #2044, opened 2019-11-18
Maintenance/update rodio cpal

azriel91:
## Description

- Updated `rodio` from `0.9` to `0.10`.
- Updated `cpal` from `0.8` to `0.11`.

A nice side effect with this is:

```
    Removing core-foundation-sys v0.5.1
    Removing generic-array v0.12.3
    Removing nalgebra v0.18.1
```

~~May want to manually `impl PartialEq` back on `Output`.~~

Note: I tried the "back onto `AlsaHost` directly approach, it's not great, plus it makes `amethyst_audio` less portable when `Alsa` is not available.

cc: https://github.com/amethyst/amethyst/issues/2030 (this doesn't fix it, and may worsen it)

## PR Checklist

### DIFF
--- a/amethyst_audio/src/bundle.rs
+++ b/amethyst_audio/src/bundle.rs
@@ -8,7 +8,7 @@ use amethyst_core::{
 };
 use amethyst_error::Error;
 
-use crate::{source::*, systems::AudioSystemDesc};
+use crate::{output::Output, source::*, systems::AudioSystemDesc};
 
 /// Audio bundle
 ///
@@ -18,7 +18,7 @@ use crate::{source::*, systems::AudioSystemDesc};
 ///
 /// The generic N type should be the same as the one in `Transform`.
 #[derive(Default, Debug)]
-pub struct AudioBundle;
+pub struct AudioBundle(Output);
 
 impl<'a, 'b> SystemBundle<'a, 'b> for AudioBundle {
     fn build(
@@ -26,7 +26,11 @@ impl<'a, 'b> SystemBundle<'a, 'b> for AudioBundle {
         world: &mut World,
         builder: &mut DispatcherBuilder<'a, 'b>,
     ) -> Result<(), Error> {
-        builder.add(AudioSystemDesc::new().build(world), "audio_system", &[]);
+        builder.add(
+            AudioSystemDesc::new(self.0).build(world),
+            "audio_system",
+            &[],
+        );
         builder.add(Processor::<Source>::new(), "source_processor", &[]);
         Ok(())
     }
--- a/amethyst_audio/src/bundle.rs
+++ b/amethyst_audio/src/bundle.rs
@@ -8,7 +8,7 @@ use amethyst_core::{
 };
 use amethyst_error::Error;
 
-use crate::{output::Output, source::*, systems::AudioSystemDesc};
+use crate::{source::*, systems::AudioSystemDesc};
 
 /// Audio bundle
 ///
@@ -18,7 +18,7 @@ use crate::{output::Output, source::*, systems::AudioSystemDesc};
 ///
 /// The generic N type should be the same as the one in `Transform`.
 #[derive(Default, Debug)]
-pub struct AudioBundle(Output);
+pub struct AudioBundle;
 
 impl<'a, 'b> SystemBundle<'a, 'b> for AudioBundle {
     fn build(
@@ -26,11 +26,7 @@ impl<'a, 'b> SystemBundle<'a, 'b> for AudioBundle {
         world: &mut World,
         builder: &mut DispatcherBuilder<'a, 'b>,
     ) -> Result<(), Error> {
-        builder.add(
-            AudioSystemDesc::new(self.0).build(world),
-            "audio_system",
-            &[],
-        );
+        builder.add(AudioSystemDesc::new().build(world), "audio_system", &[]);
         builder.add(Processor::<Source>::new(), "source_processor", &[]);
         Ok(())
     }
--- a/amethyst_audio/src/output.rs
+++ b/amethyst_audio/src/output.rs
@@ -4,11 +4,15 @@
 use std::{
     fmt::{Debug, Formatter, Result as FmtResult},
     io::Cursor,
+    sync::Arc,
 };
 
-use cpal::OutputDevices;
+use cpal::traits::DeviceTrait;
 use log::error;
-use rodio::{default_output_device, output_devices, Decoder, Device, Sink, Source as RSource};
+use rodio::{
+    default_output_device, output_devices, Decoder, Device, Devices, OutputDevices, Sink,
+    Source as RSource,
+};
 
 use amethyst_core::ecs::World;
 
@@ -17,9 +21,9 @@ use crate::{sink::AudioSink, source::Source, DecoderError};
 /// A speaker(s) through which audio can be played.
 ///
 /// By convention, the default output is stored as a resource in the `World`.
-#[derive(Clone, Eq, PartialEq)]
+#[derive(Clone)]
 pub struct Output {
-    pub(crate) device: Device,
+    pub(crate) device: Arc<Device>,
 }
 
 /// Convenience method for opening the default output device.
@@ -30,7 +34,9 @@ pub struct Output {
 impl Default for Output {
     fn default() -> Self {
         default_output_device()
-            .map(|re| Output { device: re })
+            .map(|device| Output {
+                device: Arc::new(device),
+            })
             .expect("No default output device")
     }
 }
@@ -38,7 +44,10 @@ impl Default for Output {
 impl Output {
     /// Gets the name of the output
     pub fn name(&self) -> String {
-        self.device.name()
+        self.device.name().unwrap_or_else(|e| {
+            error!("Failed to determine output device name: {}", e);
+            String::from("<unnamed_output_device>")
+        })
     }
 
     /// Play a sound once.  A volume of 1.0 is unchanged, while 0.0 is silent.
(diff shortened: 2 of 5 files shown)


azriel91:
<details>

<summary>alsa attempt (still broken, but 70% there?)</summary>

```rust
//! Provides structures and functions used to get audio outputs.

// We have to use types from this to provide an output iterator type.
use std::{
    fmt::{Debug, Formatter, Result as FmtResult},
    io::Cursor,
};

use cpal::{
    platform::{AlsaDevice, AlsaDevices, AlsaHost},
    traits::{DeviceTrait, HostTrait},
};
use log::error;
use rodio::{Decoder, Device, Sink, Source as RSource};

use amethyst_core::ecs::World;

use crate::{sink::AudioSink, source::Source, DecoderError};

/// A speaker(s) through which audio can be played.
///
/// By convention, the default output is stored as a resource in the `World`.
#[derive(Clone, Eq, PartialEq)]
pub struct Output {
    pub(crate) device: AlsaDevice,
}

/// Convenience method for opening the default output device.
///
/// Since most modern hardware features audio output, this implementation fails if a device can't
/// be initialized. Use an alternative initialization scheme if running on hardware without an
/// integrated audio chip.
impl Default for Output {
    fn default() -> Self {
        AlsaHost::new()
            .map_err(|e| {
                error!("Error retrieving Alsa host: `{}`", e);
                Option::<()>::None
            })
            .ok()
            .and_then(|host| host.default_output_device().map(|device| Output { device }))
            .expect("No default output device")
    }
}

impl Output {
    /// Gets the name of the output
    pub fn name(&self) -> String {
        self.device.name().unwrap_or_else(|e| {
            error!("Failed to determine output device name: {}", e);
            String::from("<unnamed_output_device>")
        })
    }

    /// Play a sound once.  A volume of 1.0 is unchanged, while 0.0 is silent.
    ///
    /// This will return an Error if the loaded audio file in source could not be decoded.
    pub fn try_play_once(&self, source: &Source, volume: f32) -> Result<(), DecoderError> {
        self.try_play_n_times(source, volume, 1)
    }

    /// Play a sound once. A volume of 1.0 is unchanged, while 0.0 is silent.
    ///
    /// This may silently fail, in order to get error information use `try_play_once`.
    pub fn play_once(&self, source: &Source, volume: f32) {
        self.play_n_times(source, volume, 1);
    }

    /// Play a sound n times. A volume of 1.0 is unchanged, while 0.0 is silent.
    ///
    /// This may silently fail, in order to get error information use `try_play_n_times`.
    pub fn play_n_times(&self, source: &Source, volume: f32, n: u16) {
        if let Err(err) = self.try_play_n_times(source, volume, n) {
            error!("An error occurred while trying to play a sound: {:?}", err);
        }
    }

    /// Play a sound n times. A volume of 1.0 is unchanged, while 0.0 is silent.
    ///
    /// This will return an Error if the loaded audio file in source could not be decoded.
    pub fn try_play_n_times(
        &self,
        source: &Source,
        volume: f32,
        n: u16,
    ) -> Result<(), DecoderError> {
        let sink = Sink::new(&Device::from(self.device));
        for _ in 0..n {
            sink.append(
                Decoder::new(Cursor::new(source.clone()))
                    .map_err(|_| DecoderError)?
                    .amplify(volume),
            );
        }
        sink.detach();
        Ok(())
    }
}

impl Debug for Output {
    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
        f.debug_struct("Output")
            .field("device", &self.name())
            .finish()
    }
}

/// An iterator over outputs
#[allow(missing_debug_implementations)]
pub struct OutputIterator {
    devices: Option<AlsaDevices>,
}

impl Iterator for OutputIterator {
    type Item = Output;

    fn next(&mut self) -> Option<Output> {
        self.devices
            .as_mut()
            .and_then(|devices| devices.next().map(|device| Output { device }))
    }
}

fn alsa_host() -> Option<AlsaHost> {
    AlsaHost::new()
        .map_err(|e| {
            error!("Error retrieving Alsa host: `{}`", e);
            Option::<()>::None
        })
        .ok()
}

/// Get the default output, returns none if no outputs are available.
pub fn default_output() -> Option<Output> {
    alsa_host().and_then(|host| host.default_output_device().map(|device| Output { device }))
}

/// Get a list of outputs available to the system.
pub fn outputs() -> OutputIterator {
    let devices = alsa_host().and_then(|host| {
        host.output_devices()
            .map_err(|e| {
                error!("Error retrieving output devices: `{}`", e);
                None
            })
            .ok()
    });
    OutputIterator { devices }
}

/// Initialize default output
pub fn init_output(world: &mut World) {
    if let Some(o) = default_output() {
        world
            .entry::<AudioSink>()
            .or_insert_with(|| AudioSink::new(&o));
        world.entry::<Output>().or_insert_with(|| o);
    } else {
        error!("Failed finding a default audio output to hook AudioSink to, audio will not work!")
    }
}

#[cfg(test)]
mod tests {
    #[cfg(target_os = "linux")]
    use {
        crate::{output::Output, source::Source, DecoderError},
        amethyst_utils::app_root_dir::application_root_dir,
        std::{fs::File, io::Read, vec::Vec},
    };

    #[test]
    #[cfg(target_os = "linux")]
    fn test_play_wav() {
        test_play("tests/sound_test.wav", true)
    }

    #[test]
    #[cfg(target_os = "linux")]
    fn test_play_mp3() {
        test_play("tests/sound_test.mp3", true);
    }

    #[test]
    #[cfg(target_os = "linux")]
    fn test_play_flac() {
        test_play("tests/sound_test.flac", true);
    }

    #[test]
    #[cfg(target_os = "linux")]
    fn test_play_ogg() {
        test_play("tests/sound_test.ogg", true);
    }

    #[test]
    #[cfg(target_os = "linux")]
    fn test_play_fake() {
        test_play("tests/sound_test.fake", false);
    }

    // test_play tests the play APIs for Output
    #[cfg(target_os = "linux")]
    fn test_play(file_name: &str, should_pass: bool) {
        // Get the full file path
        let app_root = application_root_dir().unwrap();
        let audio_path = app_root.join(file_name);

        // Convert the file contents into a byte vec
        let mut f = File::open(audio_path).unwrap();
        let mut buffer = Vec::new();
        f.read_to_end(&mut buffer).unwrap();

        // Create a Source from those bytes
        let src = Source { bytes: buffer };

        // Set volume and number of times to play
        let vol: f32 = 4.0;
        let n: u16 = 5;

        // Test each of the play APIs
        let output = Output::default();

        output.play_once(&src, vol);

        output.play_n_times(&src, vol, n);

        let result_try_play_once = output.try_play_once(&src, vol);
        check_result(result_try_play_once, should_pass);

        let result_try_play_n_times = output.try_play_n_times(&src, vol, n);
        check_result(result_try_play_n_times, should_pass);
    }

    #[cfg(target_os = "linux")]
    fn check_result(result: Result<(), DecoderError>, should_pass: bool) {
        match result {
            Ok(_pass) => assert!(
                should_pass,
                "Expected `play` result to be Err(..), but was Ok(..)"
            ),
            Err(fail) => assert!(
                !should_pass,
                "Expected `play` result to be `Ok(..)`, but was {:?}",
                fail
            ),
        };
    }
}
```

</details>

azriel91:
In `cpal 0.11`, `Device` impls `Send + Sync`, so the `Arc` isn't needed anymore. Shall likely finish this next week.

azriel91:
Going to update to `cpal 0.11`, though it will be short lived -- from discord:

@azriel91:

> does it make sense to play sound on the main thread (thread local system)?
> actually nevermind, `Device` now impls `Send + Sync` in `cpal 0.11`
> (wasn't in `0.10`)

@ishitatsuyuki:

> no it wasn't supposed to
> oh wait
> I mean, this is going to change with the EventLoop refactor
> due to portability concerns we are dropping Send + Sync bounds
> but we will manage the thread spawning for you with that change

amethyst_audio/src/bundle.rs
```diff
@@ -18,19 +18,15 @@ use crate::{output::Output, source::*, systems::AudioSystemDesc};
 ///
 /// The generic N type should be the same as the one in `Transform`.
 #[derive(Default, Debug)]
-pub struct AudioBundle(Output);
+pub struct AudioBundle;
```
azriel91: Removed this field because it's never used, and `Device` is `!Clone`.

amethyst_audio/src/output.rs
```diff
@@ -17,7 +19,6 @@ use crate::{sink::AudioSink, source::Source, DecoderError};
 /// A speaker(s) through which audio can be played.
 ///
 /// By convention, the default output is stored as a resource in the `World`.
-#[derive(Clone, Eq, PartialEq)]
```
azriel91: `Device` is `!Clone + !PartialEq`, but it's not used anywhere anyway.

amethyst_audio/src/systems/audio.rs
```diff
@@ -30,24 +30,21 @@ use crate::{
...
-    pub output: Output,
-}
+pub struct AudioSystemDesc;
 
 impl<'a, 'b> SystemDesc<'a, 'b, AudioSystem> for AudioSystemDesc {
     fn build(self, world: &mut World) -> AudioSystem {
         <AudioSystem as System<'_>>::SystemData::setup(world);
 
-        world.insert(self.output.clone());
+        world.insert(Output::default());
```
Xaeroxe: Being able to support a non default output is really important to gamers. This is an unfortunate limitation we should try and work around.

azriel91: nod -- though it was never possible (`Output` was always private in `AudioBundle`), and I think `cpal` or `rodio` will be the layer that exposes different output devices. Would prefer to wait for that instead of handle it at this layer.

amethyst_audio/src/output.rs
distransient: Eq/PartialEq doesn't really seem meaningful here anyways.

azriel91:
Thanks kel!

bors r+

PR #2070, opened 2019-12-15
Changed Panic to return an error type

### LINKED ISSUE #2020: [FEATURE MODIFICATION] amethyst_tiles::Map::to_tile should accept out-of-bounds input.
## Feature

https://docs.amethyst.rs/stable/amethyst_tiles/trait.Map.html#tymethod.to_tile

This returns an `Option<Point3<u32>>`, but does not document when it's `Some(..)` vs `None`. Currently, it's `Some(..)` if the coordinate is within the map or out-of-bounds in a positive direction (on production builds). If it is out of bounds in a positive direction on debug builds, it panics. if it is out of bounds in a negative direction, it returns `None`.

## Change

I think it should return `None` if out of bounds in a positive direction, and the user should choose whether or not to crash.

## Reason

It is difficult to check in advance whether or not the point coordinate is going to be out of bounds -- figuring out where it is on the grid is the purpose of `to_tile` to begin with! So to correctly call this, one must reproduce some of the logic of `to_tile` to do the bounds checks.

## Impact

Improved documentation for to_tile, since it'd be clear and easy to describe when it returns None. Slightly reduced performance on production builds due to the new bounds check.

## Additional Notes

Alternative approaches:

* Always return a `Point3`, but of `i32` and not `u32`, letting the user check the bounds. (Seems more useful in some ways, but inconvenient due to `i32`/`u32` conversions.)
* Remove the debug crash logic so that it never crashes, only returns `None` (out of bounds, negative) or a `Point3<u32>` (that may or may not be out of bounds, positive). (Slightly weird API, but still useful. Same performance as current approach, just removes the debug-only panic logic.)

IMO any of the above would be easier to use and more helpful than the current API.

* Finally, one could embrace crashing: always return a `Point3<u32>`, and panic if out of bounds. (Same performance impact, but seems less useful.)

For now I've solved this by not clicking out-of-bounds in my toy game. ;)  But I'd like a better solution going forward. If you agree with this approach, or one of the alternatives above, feel free to assign this issue to me and I can write the PR.

mrhatman:
## Description

Resolves #2020 

by returning an error type rather than panic!() .

## Addition

- The error type TileOutOfBoundsError.

## Modification

- to_tile now returns Result<Point<u32>,TileOutOfBoundsError> 
## PR Checklist

- [x ] Ran `cargo test --all --features "empty"`

### DIFF
--- a/amethyst_tiles/Cargo.toml
+++ b/amethyst_tiles/Cargo.toml
@@ -33,6 +33,7 @@ lazy_static = "1.3"
 rayon = "1.0.2"
 bitintr = "0.3"
 glsl-layout = "0.3"
+err-derive = "0.1"
 
 [dev-dependencies]
 criterion = "0.3"
--- a/amethyst_tiles/src/error.rs
+++ b/amethyst_tiles/src/error.rs
@@ -1,0 +1,19 @@
+//! Renderer error types.
+
+use amethyst_core::math::{Point3, Vector3};
+
+use err_derive::Error;
+
+/// Tile is out of bounds.
+#[derive(Debug, Error)]
+#[error(
+    display = "Requested coordinate is outside map dimensions: '{:?}', max dimensions=:{:?}",
+    point_dimensions,
+    max_dimensions
+)]
+pub struct TileOutOfBoundsError {
+    /// Calculated Point Dimension.
+    pub point_dimensions: Point3<i32>,
+    /// Map dimensions.
+    pub max_dimensions: Vector3<u32>,
+}
--- a/amethyst_tiles/src/error.rs
+++ b/amethyst_tiles/src/error.rs
@@ -7,7 +7,7 @@ use err_derive::Error;
 /// Tile is out of bounds.
 #[derive(Debug, Error)]
 #[error(
-    display = "Requested coordinate is outside map dimensions: '{:?}', max dimensions=:{:?}",
+    display = "Requested coordinate is outside map dimensions: '{:?}', max dimensions: '{:?}'",
     point_dimensions,
     max_dimensions
 )]
--- a/amethyst_tiles/src/map.rs
+++ b/amethyst_tiles/src/map.rs
@@ -1,6 +1,6 @@
 #![allow(unused_variables)]
 
-use crate::CoordinateEncoder;
+use crate::{CoordinateEncoder, TileOutOfBoundsError};
 use amethyst_assets::{Asset, Handle};
 use amethyst_core::{
     ecs::{Component, HashMapStorage, World},
@@ -54,7 +54,7 @@ pub trait Map {
         &self,
         coord: &Vector3<f32>,
         map_transform: Option<&Transform>,
-    ) -> Option<Point3<u32>>;
+    ) -> Result<Point3<u32>, TileOutOfBoundsError>;
 
     /// Returns the `Matrix4` transform which was created for transforming between world and tile coordinate spaces.
     fn transform(&self) -> &Matrix4<f32>;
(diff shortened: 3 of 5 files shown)


amethyst_tiles/src/error.rs
```diff
@@ -0,0 +1,19 @@
+//! Renderer error types.
+
+use amethyst_core::math::{Point3, Vector3};
+
+use err_derive::Error;
+
+/// Tile is out of bounds.
+#[derive(Debug, Error)]
+#[error(
+    display = "Requested coordinate is outside map dimensions: '{:?}', max dimensions=:{:?}",
```
azriel91: Curious, is the `=:{:?}` a convention somewhere? (because earlier you've got single quotes)
Oh ya, if you're changing it, I tend to go with backticks surrounding the debugged item -- copying how Rust does it

amethyst_tiles/Cargo.toml
```diff
@@ -33,6 +33,7 @@ lazy_static = "1.3"
 rayon = "1.0.2"
 bitintr = "0.3"
 glsl-layout = "0.3"
+err-derive = "0.1"
```
azriel91: Should we use 0.2? (the readme on `err-derive` does point to 0.1, but I think that's just forgotten to be updated)

amethyst_tiles/src/map.rs
```diff
@@ -340,22 +326,26 @@ fn to_tile(
...
-    if inverse.z < 0.0 {
-        return None;
+    if inverse.x < 0.0
+        || inverse.x as u32 > max_dimensions.x
+        || inverse.y < 0.0
+        || inverse.y as u32 > max_dimensions.y
+        || inverse.z < 0.0
+        || inverse.z as u32 > max_dimensions.z
+    {
+        let point_dimensions = Point3::new(inverse.x as i32, inverse.y as i32, inverse.z as i32);
```
azriel91: Think these should be `as u32`

azriel91:
(restarted windows build agent)

azriel91:
bors try

amethyst_tiles/Cargo.toml
mrhatman: I just copied it from another place in the code. I can go through and update all of them if that makes sense for this PR

amethyst_tiles/src/error.rs
mrhatman: I agree this looks weird. I was copying it from the previous panic message.

I will update this to be consistantly formatted.

amethyst_tiles/Cargo.toml
azriel91: ah don't worry about that, normally I'd do version bumps in a separate PR

amethyst_tiles/src/map.rs
mrhatman: The reason they are unsigned is because the tile coordinates you clicked could be negative. This should be passed in the error.

azriel91: oh yes. Does it make sense to pass in whatever type `inverse` is?
Points I'm thinking of:
* is it clear enough for users to understand what the error is, using the raw type?
* it removes the possibility of `i32` overflow

mrhatman: Inverse is of type Point3<f32> I believe.

I believe because of the round and floor commands that making it an int is better 

Considering f32 can only store integers accurately up to about 2^23 I wouldn't worry about overflow. 

Changing this to allow for more precision would require fixes outside of the scope of this PR.

amethyst_tiles/src/error.rs
mrhatman: Fixed

PR #2080, opened 2019-12-22
Add Draggable component + system for Ui widgets that use UiTransform

MegasKomnenos:
## Description
Ui widgets that use UiTransform didn't have an easy to use draggable component that's integrated in its prefab definition and comes bundled with UiBundle. So, I decided to make one.

## Additions
Ui widgets can become draggable by setting 'draggable' field in UiTransformData to be true. You can also manually add or remove Draggable component to any ui widget entity that also has UiTransform component.

## Removals

None

## Modifications

Included dragging SystemDesc to UiBundle.

## PR Checklist

### DIFF
--- a/amethyst_ui/src/drag.rs
+++ b/amethyst_ui/src/drag.rs
@@ -3,8 +3,8 @@ use std::marker::PhantomData;
 
 use amethyst_core::{
     ecs::{
-        Component, DenseVecStorage, Entities, Entity, Join, Read, ReaderId,
-        System, SystemData, Write, WriteStorage,
+        Component, DenseVecStorage, Entities, Entity, Join, Read, ReaderId, System, SystemData,
+        Write, WriteStorage,
     },
     math::Vector2,
     shrev::EventChannel,
@@ -39,7 +39,7 @@ pub struct DragWidgetSystem<T: BindingTypes> {
     phantom: PhantomData<T>,
 }
 
-impl<T> DragWidgetSystem<T> 
+impl<T> DragWidgetSystem<T>
 where
     T: BindingTypes,
 {
@@ -52,7 +52,7 @@ where
     }
 }
 
-impl<'s, T> System<'s> for DragWidgetSystem<T> 
+impl<'s, T> System<'s> for DragWidgetSystem<T>
 where
     T: BindingTypes,
 {
@@ -62,9 +62,12 @@ where
         Write<'s, EventChannel<UiEvent>>,
         WriteStorage<'s, Draggable>,
         WriteStorage<'s, UiTransform>,
-    ); 
+    );
 
-    fn run(&mut self, (entities, input_events, mut ui_events, mut draggables, mut ui_transforms): Self::SystemData) {
+    fn run(
+        &mut self,
+        (entities, input_events, mut ui_events, mut draggables, mut ui_transforms): Self::SystemData,
+    ) {
         let mut click_stopped: Vec<Entity> = Vec::new();
 
         for event in ui_events.read(&mut self.ui_reader_id) {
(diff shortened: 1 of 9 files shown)


amethyst_ui/src/drag.rs
```diff
@@ -0,0 +1,123 @@
...
+            match event.event_type {
+                UiEventType::ClickStart => {
+                    if let Some(draggable) = draggables.get_mut(event.target) {
+                        draggable.being_dragged = true;
+                    }
+                }
+                UiEventType::ClickStop => {
+                    if let Some(draggable) = draggables.get(event.target) {
+                        if draggable.being_dragged {
+                            click_stopped.push(event.target);
```
jojolepro: event.target contains the entity that we stopped dragging, not the entity that we dropped onto.
I think that its this entity (the dragged one) that you are inserting into the dropped_on field here https://github.com/amethyst/amethyst/pull/2080/files#diff-896eb5ae9178dff5949020bc851a076dR117 instead of the actual target we are dropping on.

MegasKomnenos: Interesting. I did not know that dropped_on field was supposed to be like that. But that does raise some issues. What if the draggable entity is stopped being dragged, but there's no entity on which it was dropped onto? What if there are multiple entities a given draggable entity was dropped onto? I'm not entirely sure how I should handle such scenarios.

ab0v3g4me: As far as I'm concerned when one stops dragging the entity and it had not entered any drop zone it should be returned to it's initial position (the position before the dragging started). Basically this means that we need to save the initial position when the drag starts.

Can't really imagine what do you mean by the latter scenario ? I would say you would drop it to the closest drop zone ?

jojolepro:
Ping me when it is ready for a new review

amethyst_ui/src/drag.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+#[derive(Debug, SystemDesc)]
+#[system_desc(name(DragWidgetSystemDesc))]
+pub struct DragWidgetSystem<T: BindingTypes> {
+    #[system_desc(event_channel_reader)]
+    ui_reader_id: ReaderId<UiEvent>,
+
+    #[system_desc(skip)]
+    record: HashMap<Entity, (Vector2<f32>, Vector2<f32>)>,
```
jojolepro: Documentation? :)

docs/CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 - Log warning when `amethyst_test::WaitForLoad` has not completed in 10 seconds. ([#1984])
 - Derive `Copy` and `PartialEq` for `amethyst::renderer::resources::Tint`. ([#2033])
 - Derive `Hash` for `amethyst::input::{Button, ControllerButton, ScrollDirection}`. ([#2041])
+- Add `Draggable` component that can be used with `UiTransform` to make widgets draggable.
```
jojolepro: Please add the PR number and a link to the pull request underneath (see the other links a bit lower in the file for an example)

amethyst_ui/src/drag.rs
MegasKomnenos: oh, sure!

docs/CHANGELOG.md
MegasKomnenos: will do

jojolepro:
For the changelog, please add this line next to the other ones that look similar

```
[#2080]: https://github.com/amethyst/amethyst/pull/2080
```

Thanks!

amethyst_ui/src/drag.rs
```diff
@@ -0,0 +1,157 @@
...
+            mut ui_events,
+            mut ui_transforms,
+        ): Self::SystemData,
+    ) {
+        if let Some((mouse_x, mouse_y)) = input_handler.mouse_position() {
+            let mouse_pos = Vector2::new(mouse_x, screen_dimensions.height() - mouse_y);
+
+            let mut click_stopped: HashSet<Entity> = HashSet::new();
+
+            for event in ui_events.read(&mut self.ui_reader_id) {
```
azriel91: Hm, something tells me the `ui_events.read(&mut self.ui_reader_id)` should be outside the `if let Some((mouse_x, mouse_y))`, so that, if there are events, but no mouse position, the event channel doesn't get clogged up by this reader not reading them.

I'm not sure when it's possible for the mouse position to be `None` though.

That said, the `ClickStart` and `ClickStop` events need to be collected (because it would conflict with the `ui_events.single_write` a bit further down).

examples/assets/ui/example.ron
```diff
@@ -163,7 +163,8 @@ Container(
             transform: (
                 id: "simple_button",
                 x: 250.,
-                y: -80., // -64-16
+                y: -40., // -64-16
```
azriel91: `-24 - 16` now? :upside_down_face:

examples/assets/ui/example.ron
```diff
@@ -178,6 +179,30 @@ Container(
                 normal_image: SolidColor(0.82, 0.83, 0.83, 1.0),
             )
         ),
+
+        // Draggable Button
+        Button(
+            transform: (
+                id: "draggable_button",
+                x: 250.,
+                y: -120., // -64-16
```
azriel91: :upside_down_face:

amethyst_ui/src/drag.rs
MegasKomnenos: That is a fair point. Thanks for pointing that out. Will work on it right away.

examples/assets/ui/example.ron
MegasKomnenos: :D

azriel91:
ah, changelog needs merging

azriel91:
bors r=jojolepro,azriel91

azriel91:
CI wants some more docs:

```rust
error: missing documentation for a trait
  --> amethyst_ui/src/event_retrigger.rs:40:1
   |
40 | pub trait EventRetrigger: Component {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `-D missing-docs` implied by `-D warnings`

error: missing documentation for an associated type
  --> amethyst_ui/src/event_retrigger.rs:41:5
   |
41 |     type In: Clone + Send + Sync + TargetedEvent;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: missing documentation for an associated type
  --> amethyst_ui/src/event_retrigger.rs:42:5
   |
42 |     type Out: Clone + Send + Sync;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: missing documentation for a trait method
  --> amethyst_ui/src/event_retrigger.rs:44:5
   |
44 | /     fn apply<R>(&self, event: &Self::In, out: &mut R)
45 | |     where
46 | |         R: EventReceiver<Self::Out>;
   | |____________________________________^
```

azriel91:
bors r=jojolepro,azriel91

PR #2085, opened 2019-12-29
Update to a newer, Rusty version of mikktspace

MilanVasko:
Hello, this is my first PR, so apologies in case I've done something wrong.

## Description

I was trying to run the GLTF example on Windows and I got told off by Visual Studio while compiling `amethyst_gltf`. The error message was as follows:

```
error: linking with `link.exe` failed: exit code: 1120
  |
  = note: "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Community\\VC\\Tools\\MSVC\\14.24.28314\\bin\\HostX64\\x86\\link.exe" "/NOLOGO" "/NXCOMPAT" ... line too long to paste here
  = note: libmikktspace-ed048d53099c6187.rlib(mikktspace-ed048d53099c6187.mikktspace.24tkbvwf-cgu.0.rcgu.o) : error LNK2019: unresolved external symbol _genTangSpaceDefault@4 referenced in function __ZN10mikktspace8Closures8generate17hfe0598a16c7e4f72E
          C:\Users\mvask\Projects\amethyst\target\debug\examples\gltf.exe : fatal error LNK1120: 1 unresolved externals
```

To be honest I had no idea how to fix it, but given that the problem seemed to be with the `mikktspace` library, I figured I could try and update to a newer version. Doing a little research, I found out that `mikktspace` started out as a wrapper around C code (which is what Amethyst is using), and has been later transformed to Rust.

After the update, I had to adapt the code a bit, since the API of `mikktspace` has changed. Having done that, I was able to compile and run the GLTF example without problems, with the same visual result.

One thing to note is that the library has not released this new, Rusty version yet and the repository looks kind of inactive (last commit was about 7 months ago), so I decided to pin the Amethyst dependency on the latest commit on master.

## Additions

None

## Removals

None

## Modifications

Adapt the code using the `mikktspace` library to a new API.

## PR Checklist

### DIFF
--- a/amethyst_gltf/Cargo.toml
+++ b/amethyst_gltf/Cargo.toml
@@ -27,7 +27,7 @@ gltf = "0.14"
 hibitset = { version = "0.6.2", features = ["parallel"] }
 itertools = "0.8"
 log = "0.4.6"
-mikktspace = { version = "0.1" }
+mikktspace = { git = "https://github.com/gltf-rs/mikktspace", rev = "abbae9b486f280ce6e331821e2454c2cfd746d5b" }
 serde = { version = "1.0", features = ["derive"] }
 
 thread_profiler = { version = "0.3", optional = true }
--- a/amethyst_gltf/Cargo.toml
+++ b/amethyst_gltf/Cargo.toml
@@ -27,7 +27,7 @@ gltf = "0.14"
 hibitset = { version = "0.6.2", features = ["parallel"] }
 itertools = "0.8"
 log = "0.4.6"
-mikktspace = { git = "https://github.com/gltf-rs/mikktspace", rev = "abbae9b486f280ce6e331821e2454c2cfd746d5b" }
+mikktspace = "0.2.0"
 serde = { version = "1.0", features = ["derive"] }
 
 thread_profiler = { version = "0.3", optional = true }
--- a/amethyst_gltf/src/format/mesh.rs
+++ b/amethyst_gltf/src/format/mesh.rs
@@ -7,8 +7,8 @@ use amethyst_rendy::{
     skinning::JointCombined,
 };
 use log::{trace, warn};
-use std::{iter::repeat, ops::Range};
 use mikktspace::{generate_tangents_default, Geometry};
+use std::{iter::repeat, ops::Range};
 
 fn compute_if<T, F: Fn() -> T>(predicate: bool, func: F) -> Option<T> {
     if predicate {
--- a/amethyst_gltf/src/format/mesh.rs
+++ b/amethyst_gltf/src/format/mesh.rs
@@ -6,8 +6,9 @@ use amethyst_rendy::{
     rendy::mesh::{Color, MeshBuilder, Normal, Position, Tangent, TexCoord},
     skinning::JointCombined,
 };
-use log::trace;
+use log::{trace, warn};
 use std::{iter::repeat, ops::Range};
+use mikktspace::{generate_tangents_default, Geometry};
 
 fn compute_if<T, F: Fn() -> T>(predicate: bool, func: F) -> Option<T> {
     if predicate {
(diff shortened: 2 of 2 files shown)


amethyst_gltf/Cargo.toml
```diff
@@ -27,7 +27,7 @@ gltf = "0.14"
 hibitset = { version = "0.6.2", features = ["parallel"] }
 itertools = "0.8"
 log = "0.4.6"
-mikktspace = { version = "0.1" }
+mikktspace = { git = "https://github.com/gltf-rs/mikktspace", rev = "abbae9b486f280ce6e331821e2454c2cfd746d5b" }
```
distransient: I don't think we should be adding git dependencies to the engine, even if pinned to a revision

MilanVasko: I agree that it is not the prettiest solution ever, but my point of view is this:

- it fixes an error (am I really the only one who had it?)
- it gets rid of a C dependency
- the `rev` part ensures there are no moving parts, unless we decide to update explicitly
- it works

Is there a practical reason for not doing it this way?

ishitatsuyuki: - You can't release a crate with git dependency on crates.io
- It's way less flexible than a semver (a fixed revision vs allowed range of versions)

ishitatsuyuki:
> repository looks kind of inactive (last commit was about 7 months ago)

Disagree, all PRs seem to be merged and you should open an issue asking for a release before resorting to git dependencies.

amethyst_gltf/Cargo.toml
MilanVasko: > You can't release a crate with git dependency on crates.io

Oh, I did not know that, sorry! In that case, I agree it would not be a good idea.

MilanVasko:
> Disagree, all PRs seem to be merged and you should open an issue asking for a release before resorting to git dependencies.

Good point, I've just done that (gltf-rs/mikktspace#16). We'll see how it goes.

chemicstry:
Can confirm, I had the exact problem on one of my windows machines. Solved by using latest `mikktspace` as described here. Looking forward for this to be merged 👍

MilanVasko:
Hello, @alteous has kindly agreed to [release](https://github.com/gltf-rs/mikktspace/issues/16#issuecomment-581012416) a new version of mikktspace.  I've already pushed a new commit using it, hopefully everything's OK now.

amethyst_gltf/src/format/mesh.rs
```diff
@@ -202,20 +203,52 @@ fn calculate_tangents(
...
-    let mut tangents = vec![Tangent([0.0, 0.0, 0.0, 0.0]); positions.len()];
-    let num_faces = indices.len().unwrap_or_else(|| positions.len()) / 3;
-    mikktspace::generate_tangents(
-        &|| 3,
-        &|| num_faces,
-        &|face, vert| &positions[indices.map(face, vert)].0,
-        &|face, vert| &normals[indices.map(face, vert)].0,
-        &|face, vert| &tex_coords[indices.map(face, vert)].0,
-        &mut |face, vert, tangent| {
+    struct TangentsGeometry<'a> {
```
azriel91: Heya @MilanVasko, thank you for this! Could you move `TangentsGeometry` and the trait impl outside the function? Shall be good to take the PR then :v:

MilanVasko: Hello, sure thing. Commit is coming in a couple of minutes. 😄

azriel91:
@distransient did you want to review, or shall I go ahead and r+?
(have been following it from the `mikktspace` repo, so am comfortable/confident with the change)

distransient:
You may go ahead 👍

azriel91:
yayness :v:
bors r=azriel91,distransient

azriel91:
bors r=azriel91,distransient

PR #2117, opened 2020-02-02
States ui hierarchy issue

mrhatman:
## Description

Fixed an issue where the states ui example would load incorrectly during fast transitions.

The issue was with delete hierarchy and its call to all_children_iter. https://github.com/rustgd/specs-hierarchy/issues/12

## PR Checklist


azriel91:
Hm, the logic is now changed to delete only the root entity. `specs-hierarchy` doesn't actually delete the child entities when the root entity is deleted -- it only removes them from the `Hierarchy` tracking.

What's the symptom / issue when transitioning quickly?

mrhatman:
The issue was that child entity's weren't deleted when you transitioned sometimes. Especially when transitioning quickly. 

specs-hierarchy doesn't delete children? I appears to remove them in this change.

azriel91:
> specs-hierarchy doesn't delete children? I appears to remove them in this change.

ah! I was looking at the wrong code -- it does delete children in `HierarchySystem` indeed :v:

azriel91:
bors try

azriel91:
ah, if you wait for the dependency bump PR, that has a fix for the current build failure (due to bumping rust version)

azriel91:
@mrhatman could you rebase? then I think we're good to go

mrhatman:
Yes I can,  sorry life got busy.

PR #2125, opened 2020-02-04
Reorganised the dependencies to combat a race condition in amethyst network

Machine-Hum:
## Description

Fixed race condition by reorg'ing dependency order. In addition, networking examples can now sent / receive from the client end, before the client could just send data. This makes them more comprehensive and easier for beginners to get going.

## PR Checklist

- [NA] Updated the content of the book if this PR would make the book outdated.
- [NA] Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.
- [NA] Added unit tests for new code added in this PR.

### DIFF
--- a/amethyst_network/src/simulation/transport/tcp.rs
+++ b/amethyst_network/src/simulation/transport/tcp.rs
@@ -49,26 +49,42 @@ impl<'a, 'b> SystemBundle<'a, 'b> for TcpNetworkBundle {
         world: &mut World,
         builder: &mut DispatcherBuilder<'_, '_>,
     ) -> Result<(), Error> {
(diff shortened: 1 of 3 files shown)


jstnlef:
I'm not sure if we need the changes to the example files, but the changes to tcp.rs look good. Just verified that the deps are indeed backward. 👍

azriel91:
oh, that's gonna fail. @Machine-Hum there's an extra import:

```rust
error: unused import: `std::net::TcpListener`
  --> examples/net_client/main.rs:16:5
   |
16 | use std::net::TcpListener;
   |     ^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `-D unused-imports` implied by `-D warnings`
```

PR #2128, opened 2020-02-05
Export the get_parent_pixel_size function

bonsairobo:
## Description

The function can be used to get the pixel (width, height) of an entity's parent's `UiTransform`. The function is a simplified version of what already existed in the `ui::drag` module.

## Additions

- export a new function from `amethyst_ui` module

## Modifications

- change the `get_scale_for_entity` to only check the parent's pixel size rather than checking the "scaled" size on all ancestors

## PR Checklist

### DIFF
--- a/amethyst_ui/src/drag.rs
+++ b/amethyst_ui/src/drag.rs
@@ -6,8 +6,8 @@ use std::{
 
 use amethyst_core::{
     ecs::{
-        storage::GenericReadStorage, Component, DenseVecStorage, Entities, Entity, Join, Read,
-        ReadExpect, ReadStorage, ReaderId, System, SystemData, Write, WriteStorage,
+        Component, DenseVecStorage, Entities, Entity, Join, Read, ReadExpect, ReadStorage,
+        ReaderId, System, SystemData, Write, WriteStorage,
     },
     math::Vector2,
     shrev::EventChannel,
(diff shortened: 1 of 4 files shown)


bonsairobo:
Forgot to mention this PR also adds public getters for `pixel_width` and `pixel_height` on the `UiTransform`.

amethyst_ui/src/drag.rs
```diff
@@ -128,12 +131,17 @@ where
...
             let ui_transform = ui_transforms.get_mut(*entity).unwrap();
-            ui_transform.local_x += scale_x * change[0];
-            ui_transform.local_y += scale_y * change[1];
+            let (scale_x, scale_y) = match ui_transform.scale_mode {
+                ScaleMode::Pixel => (1.0, 1.0),
+                ScaleMode::Percent => (parent_width, parent_height),
+            };
+
+            ui_transform.local_x += change[0] / scale_x;
+            ui_transform.local_y += change[1] / scale_y;
```
azriel91: hm, shall think aloud:

* `change` is in pixels.
* `local_x/y` are values respective to the scale mode.

Given:

* screen dimensions: `300`
* a parent with scale mode `Percent(0.1)` but on screen pixel size `30`, we get
* current entity local_x: `0.25`

previously, we had:

| scale mode | example values           |
| ---------- | ------------------------ |
| `Pixel`    | `10.0` += `1.0` * `5.0`  |
| `Percent`  | `0.25` += `0.1` * `5.0`  |

now:

| scale mode | example values           |
| ---------- | ------------------------ |
| `Pixel`    | `10.0` += `5.0` / `1.0`  |
| `Percent`  | `0.25` += `5.0` / `30.0` |

hm, something may have changed (is it more correct now?)

bonsairobo: There shouldn't be any difference in the final local_x/y values resulting from this commit.

Previously, the code would compound the scaling factors from the full path up the hierarchy. So if you had this hierarchy with all `Percent` mode transforms:

```
screen: width 300
   ^
   |
parent: width 0.1 (30 pixels)
   ^
   |
leaf
```

Then you'd get:

`scale_x = (1 / 300) / 0.1 = 0.033...`

and `0.033... * 5.0 = 5.0 / 30.0`

bonsairobo: The previous code was overly confusing, but I didn't know about `pixel_width` and `pixel_height` at the time.

azriel91: ah, thank you :bowing_man:

Issue #2146, opened 2020-02-23
[BUG] Cursor position doesn't get updated


mvlabat:
## Description

Cursor position doesn't get updated until text content of any field gets changed (it's not necessarily the focused one).

## Reproduction Steps

1. Comment out `random_text` UI element out of `example.ron` (otherwise text updates will be retriggered every frame)
2. Run `ui` example
3. Focus the editable field, try to move the cursor with mouse clicks or Left/Right buttons on the keyboard

## What You Expected to Happen

I would expect the cursor to move.

## What Actually Happened

The cursor stays on the same position, until you type something.

## Your Environment

1. Windows 10 Pro
2. 1909
3. The latest master version (d647450)

azriel91:
Looks like https://github.com/amethyst/amethyst/pull/2143 fixes this as well -- I just tried it out.

mvlabat:
@azriel91 it doesn't fix it for me on clicks either.. The bug I've described isn't related to text spaces: the cursor position doesn't get updated at all, not only when it reaches whitespace.

If you're testing it on the `ui` example, are you sure that you've commented out `random_text` UI element? Otherwise you won't catch this particular bug

azriel91:
Heya, could you try #2146 (ideally #2146 rebased over #2143). I've commented out `random_text`, and it kinda works (there's a small lag, but the cursor does update its position).

mvlabat:
@azriel91 #2149 works perfectly now, thanks!
Btw, while a bit unrelated to this issue, could it be possible to put the caret right under the cursor, when we click on a field for the first time? Right now it always appears at the beginning of the field and moves under the cursor only on the second hit (after the field gets focus). If it's not a minor fix that could be implemented as a part of that PR, I can create a separate issue for that

azriel91:
Heya, ya, noticed that one too. Shall look into it in a separate PR soon (today).

PR #2151, opened 2020-02-24
Clear selecteds if selectables do not contain an event target

### LINKED ISSUE #2091: [BUG] UiText cursor moves when clicking outside the transform
## Description

A selected editable text field seems to be affected by clicks outside of its transform, and can't be deselected.

## Reproduction Steps

1. `cargo run --example ui --features vulkan`
2. Click inside the "Editable" text transform to select it.
3. Click anywhere else on the screen.

It should be deselected.

## What Actually Happened

The cursor moves to the closest glyph at the X position that clicked, instead of being deselected. This should only happen if the click was inside the bounds of the transform.

## Your Environment

1. Operating System: Void Linux
2. Operating System Version: latest
3. Version of Amethyst: tested on master, 0.13, 0.12, and 0.11.

mvlabat:
## Description

An attempt to fix #2091

I'm not sure what this comment is about:
https://github.com/amethyst/amethyst/blob/72f7d8f913b4a4d8f05733f2d0c27315d6a09530/amethyst_ui/src/selection.rs#L257
So maybe my change breaks something..

## PR Checklist

### DIFF
--- a/amethyst_ui/src/text.rs
+++ b/amethyst_ui/src/text.rs
@@ -194,74 +194,82 @@ impl<'a> System<'a> for TextEditingMouseSystem {
             }
         }
 
+        let mut just_pressed = false;
+        let mut moved_while_pressed = false;
+
         // Process only if an editable text is selected.
         for event in events.read(&mut self.reader) {
-            for (ref mut text, ref mut text_editing, _) in
-                (&mut texts, &mut text_editings, &selecteds).join()
-            {
-                // Process events for the whole UI.
-                match *event {
-                    Event::WindowEvent {
-                        event: WindowEvent::CursorMoved { position, .. },
-                        ..
-                    } => {
-                        let hidpi = screen_dimensions.hidpi_factor() as f32;
-                        self.mouse_position = (
-                            position.x as f32 * hidpi,
-                            (screen_dimensions.height() - position.y as f32) * hidpi,
-                        );
-                        if self.left_mouse_button_pressed {
-                            let (mouse_x, mouse_y) = self.mouse_position;
-                            text_editing.highlight_vector =
-                                closest_glyph_index_to_mouse(mouse_x, mouse_y, &text.cached_glyphs)
-                                    - text_editing.cursor_position;
-                            // The end of the text, while not a glyph, is still something
-                            // you'll likely want to click your cursor to, so if the cursor is
-                            // near the end of the text, check if we should put it at the end
-                            // of the text.
-                            if should_advance_to_end(mouse_x, text_editing, text) {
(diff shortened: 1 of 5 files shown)


mvlabat:
Also the caret doesn't disappear at once, one can notice that it moves right before disappearing. And if we click on the text field again, it highlights (selects) some characters, where the start position was the previous character to which the caret moved right before disappearing.

Maybe that's something related to what is described here:
https://github.com/amethyst/amethyst/issues/2146#issuecomment-590244355

azriel91:
Cool, the code fixes half the issue (deselects the editable widget when clicking outside), but it deselects it after it responds to it the first time.

> And if we click on the text field again, it highlights (selects) some characters, where the start position was the previous character to which the caret moved right before disappearing.

I can't seem to reproduce the highlighting behaviour though.

Shall try and fix the "move to the correct position on click" and the "incorrectly respond to first click outside the box" issues now.

mvlabat:
*(cross-posting from Discord)*

Ok, I think I've fixed it. The thing I don't like is that we use `InputHandler` in `amethyst_ui/src/selection.rs:240`. That makes it dependent on `InputSystem`, which I had to specify as a dependency in `UiBundle`. It's a breaking change (well, I think it's not a big deal on its own), but it also can be quite obscure to users: the application will def crash if `InputBundle` is loaded after `UiBundle`

amethyst_ui/src/text.rs
```diff
@@ -194,74 +194,82 @@ impl<'a> System<'a> for TextEditingMouseSystem {
...
+                    event:
+                        WindowEvent::MouseInput {
+                            button: MouseButton::Left,
+                            state,
+                            ..
+                        },
+                    ..
+                } => match state {
+                    ElementState::Pressed => {
+                        println!("hey");
```
azriel91: heya :wave:

azriel91:
Yay, am glad you managed to fix it :bowing_man: (I became frustrated trying).

> It's a breaking change (well, I think it's not a big deal on its own), but it also can be quite obscure to users: the application will def crash if `InputBundle` is loaded after `UiBundle`

I think it's fine for a breaking change, an entry in `CHANGELOG.md` should be good enough to let people know (and we can call it out in a release blog post).

amethyst_ui/src/text.rs
```diff
@@ -224,7 +224,6 @@ impl<'a> System<'a> for TextEditingMouseSystem {
                     ..
                 } => match state {
                     ElementState::Pressed => {
-                        println!("hey");
```
mvlabat: hey :)

PR #2155, opened 2020-03-01
Don't remove HiddenPropagate components set by users manually

mvlabat:
### Motivation
Let's imagine the following scenario:
A user has a modal window with two child elements: `TextBody`, `ConfirmationButton`. In some cases a user would like to show a modal window without a confirmation button (for instance, it's a loading process that should just block the UI for some time). If a user removes `HiddenPropagate` from the window, this change will be propagated to all the children, and right now there's no way to mark the `ConfirmationButton` entity that its `HiddenPropagate` shouldn't be removed automatically.

This PR adds a flag to `HiddenPropagate` that can prevent `HideHierarchySystem` from removing the component automatically, which should solve the problem.

## Additions
`HiddenPropagate` now contains a `was_propagated` flag indicating whether this entity was created by `HideHierarchySystem` or not. Its storage has been changed to `DenseVecStorage`.

## Removals
`HiddenPropagate` now doesn't have `Default` trait implementation. That's not a big deal, I can easily bring it back, but I thought that might make other `amethyst_core` developer's choice more conscious, if they want to use it.

## Modifications
`HiddenPropagate` doesn't get removed automatically if it was inserted manually by a user.

## PR Checklist

  ^ (the latest clippy has got a lot of warnings for the code I didn't introduce)

### DIFF
--- a/amethyst_core/src/hidden.rs
+++ b/amethyst_core/src/hidden.rs
@@ -14,9 +14,32 @@ impl Component for Hidden {
 
 /// Like [Hidden](struct.Hidden.html), but can propagate through children when the [HideHierarchySystem](struct.HideHierarchySystem.html)
 /// is enabled in the [RenderBundle](struct.RenderBundle.html).
-#[derive(Clone, Debug, Default)]
-pub struct HiddenPropagate;
+#[derive(Clone, Debug)]
+pub struct HiddenPropagate {
+    pub(crate) was_propagated: bool,
+}
 
 impl Component for HiddenPropagate {
-    type Storage = FlaggedStorage<Self, NullStorage<Self>>;
+    type Storage = FlaggedStorage<Self>;
+}
+
+impl HiddenPropagate {
+    /// Creates an instance of HiddenPropagate.
+    pub fn new() -> Self {
+        Self {
+            was_propagated: false,
+        }
+    }
+
+    /// Is meant to be used only by HideHierarchySystem.
+    pub(crate) fn new_propagated() -> Self {
+        Self {
+            was_propagated: true,
+        }
+    }
+
+    /// Returns true if this component was propagated by [HideHierarchySystem](struct.HideHierarchySystem.html) automatically.
+    pub fn was_propagated(&self) -> bool {
+        self.was_propagated
+    }
 }
--- a/amethyst_core/src/hide_system.rs
+++ b/amethyst_core/src/hide_system.rs
@@ -114,13 +114,24 @@ impl<'a> System<'a> for HideHierarchySystem {
                 if parent_dirty {
                     if hidden.contains(parent_entity) {
                         for child in hierarchy.all_children_iter(parent_entity) {
-                            if let Err(e) = hidden.insert(child, HiddenPropagate::default()) {
-                                error!("Failed to automatically add `HiddenPropagate`: {:?}", e);
-                            };
+                            if !hidden.contains(child) {
+                                if let Err(e) =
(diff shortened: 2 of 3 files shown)


mvlabat:
Ok, it won't always work as expected.. `all_children_iter` returns all children recursively, so sometimes it happens that we remove `HiddenPropagate` for an entity whose parent has got `HiddenPropagate` set manually. I need to think about a cleverer solution here

mvlabat:
Upd. Should be good now

amethyst_core/src/hidden.rs
```diff
@@ -14,9 +14,32 @@ impl Component for Hidden {
...
-#[derive(Clone, Debug, Default)]
-pub struct HiddenPropagate;
+#[derive(Clone, Debug)]
+pub struct HiddenPropagate {
+    pub(crate) was_propagated: bool,
+}
 
 impl Component for HiddenPropagate {
-    type Storage = FlaggedStorage<Self, NullStorage<Self>>;
+    type Storage = FlaggedStorage<Self>;
```
caitp: not: I’d prefer explicit backing storage type, it’s more obvious for new users IMHO

amethyst_core/src/hidden.rs
```diff
@@ -14,9 +14,32 @@ impl Component for Hidden {
 
 /// Like [Hidden](struct.Hidden.html), but can propagate through children when the [HideHierarchySystem](struct.HideHierarchySystem.html)
 /// is enabled in the [RenderBundle](struct.RenderBundle.html).
-#[derive(Clone, Debug, Default)]
-pub struct HiddenPropagate;
```
caitp: This is a breaking change, right? I think this needs a mention in the ChangeLo
g, and possibly examples

amethyst_core/src/hide_system.rs
```diff
@@ -111,32 +120,21 @@ impl<'a> System<'a> for HideHierarchySystem {
...
-                } else if self_dirty {
-                    // in case the parent was already dirty, this entity and its children have
-                    // already been hidden, therefore it only needs to be an else-if, instead of a
-                    // stand-alone if.
-                    if hidden.contains(*entity) {
-                        for child in hierarchy.all_children_iter(*entity) {
-                            if let Err(e) = hidden.insert(child, HiddenPropagate::default()) {
-                                error!("Failed to automatically add `HiddenPropagate`: {:?}", e);
-                            };
+                let parent_hidden = hidden.get(parent_entity);
```
caitp: I really like this simplification, nice!

mvlabat:
@caitp thanks for the comments, I'll definitely fix those. First, I'd just like to wait for @Frizi's or @azriel91's review to understand whether this has a chance to get merged :)

amethyst_core/src/hidden.rs
```diff
@@ -14,9 +14,32 @@ impl Component for Hidden {
...
 impl Component for HiddenPropagate {
-    type Storage = FlaggedStorage<Self, NullStorage<Self>>;
+    type Storage = FlaggedStorage<Self>;
+}
+
+impl HiddenPropagate {
+    /// Creates an instance of HiddenPropagate.
+    pub fn new() -> Self {
+        Self {
+            was_propagated: false,
```
azriel91: kind of feel `is_propagated` is more natural to search for, and think the past tense in "propagated" is enough to capture the meaning

mvlabat: It now sounds like present in passive voice, but maybe that's alright.. We should add a comment for that field then

azriel91:
Incorporated by #2186

Issue #2083, opened 2019-12-28
[BUG] FrameLimiter Sleep strategy sleeps too long


AkkaWakka:
I think there is an error in the FrameLimiter implementation logic for the `Sleep` strategy, unless I'm misunderstanding something.

The docs say (emphasis mine):
> Sleep will call `thread::sleep` with a **duration of 0 milliseconds** until the frame duration has passed.

In the code, the `Sleep` strategy is called with a zero duration:
```rust
const ZERO: Duration = Duration::from_millis(0);
....
Sleep => self.do_sleep(ZERO)
```

Working through the `do_sleep` function (comments mine, durations expressed in milis):
```rust
fn do_sleep(&self, stop_on_remaining: Duration) {    // stop_on_remaining = 0 milis
    let frame_duration = self.frame_duration - stop_on_remaining;    // assume 16ms frame, frame_duration = 16 - 0 = 16 milis
    loop {
        let elapsed = Instant::now() - self.last_call;    // assume last_call 4ms, elapsed = 4 milis
        if elapsed >= frame_duration {    // if 4 > = 16 is false
            break;
        } else {    // so we end up here
            sleep(frame_duration - elapsed);    // sleep(16 - 4) | sleep(12)
        }
    }
}
```

I think the error here is that the `stop_on_remaining` parameter is being used for two different things:
1) For the `SleepAndYield` strategy, setting the time before the next frame that we must stop sleeping and start yielding.  This - I think - works.
2) For the `Sleep` strategy, setting the time we want to sleep for, each time we call sleep.

I see two possible solutions:
a) add another parameter to the `do_sleep` method.  Something like `max_sleep_length`.  This might be useful for the `SleepAndYield` strategy too since it would remove the possibility of very long sleeps (and thread unresponsiveness)
b) use a separate method or loop for the `Sleep` strategy.

I don't know this code well, and I'm finding the names a little obtuse, so please tell me if I've got this wrong.

Thanks

zaethan:
I _think_ the code is fine. The documentation seems to be outdated though. If you look at the git blame of the frame limiter file, you can see that the documentation was written two years ago. However, the do_sleep function was changed just 11 months ago.
Before that change, the do_sleep function would indeed call thread::sleep repeatedly with 0 ms until the remaining time was less than the stop_on_remaining parameter.

[...]

PR #2156, opened 2020-03-01
docs: clarify the behaviour of FrameLimitStrategy::Sleep

caitp:
To say that Sleep performs `std::sleep(0)` is misleading, when in fact the sleep duration is calculated based on the elapsed time since the previous frame, per the target frame rate.

Goes along with #2083

## Description

Updated a bit of documentation

## Additions

`(⊙_⊙)？`

## Removals

`¯\(°_o)/¯`

## Modifications

`o((⊙﹏⊙))o.`

## PR Checklist

- [x] Updated the content of the book if this PR would make the book outdated. **(N/A)**
- [x] Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment. **(N/A)**
- [x] Added unit tests for new code added in this PR. **(N/A)**


azriel91:
bors r=erlend-sh, azriel91

PR #2166, opened 2020-03-08
Add example of creating custom state event

RichoDemus:
Also shows how to send that message from a system and receive it from a state

## Description

This is something I struggled a bit with doing, even after having read the documentation, and after getting some stellar help on discord I decided to contribute an example

## Additions

## Removals

## Modifications

## PR Checklist

### DIFF
--- a/examples/events_custom_state_event/event.rs
+++ b/examples/events_custom_state_event/event.rs
@@ -1,0 +1,33 @@
+use amethyst::{
+    core::{
+        shrev::{EventChannel, ReaderId},
+        EventReader,
+    },
+    derive::EventReader,
+    ecs::{Read, SystemData, World},
+};
+use amethyst_input::{BindingTypes, InputEvent, StringBindings};
+use amethyst_ui::UiEvent;
+use winit::Event;
+
+/// Here's a copy of the original StateEvent with our own type added
+#[derive(Clone, Debug, EventReader)]
+#[reader(MyExtendedStateEventReader)]
+pub enum MyExtendedStateEvent<T = StringBindings>
+where
+    T: BindingTypes + Clone,
+{
+    /// Events sent by the winit window.
+    Window(Event),
+    /// Events sent by the ui system.
+    Ui(UiEvent),
+    /// Events sent by the input system.
+    Input(InputEvent<T>),
+    /// Our own events for our own game logic
+    Game(GameEvent),
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum GameEvent {
+    IncreaseDifficulty,
+}
--- a/examples/events_custom_state_event/state.rs
+++ b/examples/events_custom_state_event/state.rs
@@ -1,0 +1,37 @@
+use crate::event::{GameEvent, MyExtendedStateEvent};
+use amethyst::prelude::*;
+
+pub(crate) struct GameplayState {
+    game_difficulty: i32,
+}
+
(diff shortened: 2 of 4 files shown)


examples/events_custom_state_event/event.rs
```diff
@@ -0,0 +1,33 @@
...
+    T: BindingTypes + Clone,
+{
+    /// Events sent by the winit window.
+    Window(Event),
+    /// Events sent by the ui system.
+    Ui(UiEvent),
+    /// Events sent by the input system.
+    Input(InputEvent<T>),
+    /// Our own events for our own game logic
+    MyCustomGameEvent(MyCustomGameEventType),
```
caitp: nit: Can we use realistic variant names? I find placeholder names can make things harder to grasp

examples/events_custom_state_event/event.rs
```diff
@@ -0,0 +1,33 @@
...
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum MyCustomGameEventType {
+    MyEnumVariant,
```
caitp: ditto

examples/events_custom_state_event/system.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#[derive(SystemDesc)]
+pub(crate) struct SpammingSystem;
+
+impl<'a> System<'a> for SpammingSystem {
+    type SystemData = Write<'a, EventChannel<MyCustomGameEventType>>;
+
+    fn run(&mut self, mut my_event_channel: Self::SystemData) {
+        my_event_channel.single_write(MyCustomGameEventType::MyEnumVariant);
+        println!("Sending event");
```
caitp: I suggest we illustrate this a different way, without spamming the tty

examples/events_custom_state_event/state.rs
```diff
@@ -0,0 +1,26 @@
...
+
+impl<'a, 'b> State<GameData<'a, 'b>, MyExtendedStateEvent> for GameplayState {
+    fn handle_event(
+        &mut self,
+        _data: StateData<'_, GameData<'_, '_>>,
+        event: MyExtendedStateEvent,
+    ) -> Trans<GameData<'a, 'b>, MyExtendedStateEvent> {
+        if let MyExtendedStateEvent::MyCustomGameEvent(MyCustomGameEventType::MyEnumVariant) = event
+        {
+            println!("Received event: {:?}", event);
```
caitp: same here, it would be nice to illustrate this without using println!

examples/events_custom_state_event/system.rs
RichoDemus: Ok, I just copied what was being done in the other event-example, do you have an idea of something better? :)

caitp: I think we could do it with Ui interactions for triggering events, and signalling their receipt. But this is just my preference, im sure it’s fine to land without that.

RichoDemus: I made some changes, let me know what you think :)

caitp: It’s ok to me, I think it could be better if the event was triggered by some interactive action, and signaled by something in the window. But I can respect keeping it simple, too. LGTM

azriel91:
bors retry

azriel91:
bors cancel

PR #2167, opened 2020-03-10
Update Fluent to 0.11

zbraniecki:
We reworked how concurrent vs. single-thread FluentBundle works and I'd like to make sure that concurrent uses like amethyst work well.


azriel91:
bors retry

PR #2173, opened 2020-03-17
WIP: Fix #2170 Overly aggressive 3D model culling

### LINKED ISSUE #2170: [BUG] Overly aggressive 3D model culling in rendering
## Description

When rendering 3D while using a FlyCamera, when you look away from the origin of an object the whole object disappears from rendering

## Reproduction Steps

1. Run literally anything that uses a FlyCamera and 3D rendering, like the AutoFov example
2. Back up a bit with the camera
3. Start looking away from an object
4. Watch it disappear to early

You can go back and forth with your camera, the object will appear and disappear, though it should disappear from rendering only when none of its vertexes are on the screen

## Screenshots and Logs
A short gif presenting the issue:
teapot

## Your Environment

1. Linux (Arch Linux x86_64)
2. Kernel 5.5.8-arch1-1
3. Amethyst version 14.0 (current master)
4. PC specs: AMD Ryzen 7 3700X (16) @ 3.8Ghz, NVIDIA GeForce RTX 2060 SUPER, propertiary nvidia driver, 16GB RAM

MrMinimal:
## Description

**Need help, see comments**
This PR fixes the fly_camera example which culls its cube too early.

## Additions

To fix it, a `BoundingSphere` is added to the cube.

## PR Checklist

### DIFF
--- a/examples/fly_camera/main.rs
+++ b/examples/fly_camera/main.rs
@@ -32,7 +32,7 @@ impl SimpleState for ExampleState {
             .create_entity()
             .named("Fly Camera Scene")
             .with(prefab_handle)
-            .with(BoundingSphere::origin(1.0))
+            .with(BoundingSphere::origin(1.732))
             .build();
     }
 
--- a/examples/fly_camera/main.rs
+++ b/examples/fly_camera/main.rs
@@ -11,6 +11,7 @@ use amethyst::{
         plugins::{RenderShaded3D, RenderToWindow},
         rendy::mesh::{Normal, Position, TexCoord},
         types::DefaultBackend,
+        visibility::BoundingSphere,
         RenderingBundle,
     },
     utils::{application_root_dir, scene::BasicScenePrefab},
@@ -31,6 +32,7 @@ impl SimpleState for ExampleState {
             .create_entity()
             .named("Fly Camera Scene")
             .with(prefab_handle)
+            .with(BoundingSphere::origin(1.0))
             .build();
     }
 


MrMinimal:
The `BoundingSphere` is not found by https://docs.amethyst.rs/stable/src/amethyst_rendy/visibility.rs.html#159.
It still uses the default value of 1.0, which leads to the early culling.

How do I fix that?

azriel91:
By the way, is this still WIP?

MrMinimal:
@azriel91 I thought it would work that way but the sphere is not added to the object. It might have to be added to the prefab but I am not experienced enough for that.

Right now, culling is still not using the sphere.

chemicstry:
The problem is that sphere is added to the root prefab entity, not to the cube entity. The easy way would be to add `BoundingSphere` to the cube in `prefab/fly_camera.ron`, but `BoundingSphere` is not supported in prefabs. Either prefab has to be implemented for `BoundingSphere` or you have to do entity lookup after inserting the scene and adding the sphere there. Maybe someone has a better way that I didn't think of.

caitp:
As an easier short-term fix without changes to the library, you could query for the entity with Mesh in on_start and insert the BoundingSphere on it, right?

valkum:
You would need to add a BoundingSpherePrefab for https://github.com/amethyst/amethyst/blob/af3bf68a3b829a386cf4c04068a49eebbdb6282f/amethyst_rendy/src/visibility.rs#L46and add that to 
https://github.com/amethyst/amethyst/blob/6453a83c4dc5b6152d35004d64931f5e09f98e65/amethyst_utils/src/scene.rs#L31

jojolepro:
Status? :)

jojolepro:
Closing due to inactivity

Issue #2177, opened 2020-03-18
Get `web_worker` `ThreadPool` integrated


azriel91:
[`web_worker`](https://github.com/jojolepro/web_worker) uses a JS `WorkerPool` to service tasks, so tasks can be run in parallel.

Integrate this into `amethyst` and create a small WASM application to demonstrate `ThreadPool` usage.

azriel91:
Note, assets still cannot be loaded simply by getting `rayon` to use a `WorkerPool` -- `amethyst_assets` still loads from the file system, and we have to get it to load from perhaps a `HttpSource`.

chemicstry:
While working on #2180 I experimented with `web_worker` and couldn't get it to work on stable rust, but got it working with nightly as per https://github.com/rustwasm/wasm-bindgen/blob/master/examples/raytrace-parallel/build.sh instructions. Strangely I didn't get any compile errors, but browser refused to load workers with `Failed to execute 'postMessage' on 'Worker': #<Memory> could not be cloned.`. I guess the required rust features are still experimental? If that's the case, then the build instructions will have to be updated.

chemicstry:
Partially solved in #2189. Narrowed down to #2191

PR #2182, opened 2020-03-21
Implement HTTP asset source for WASM

chemicstry:
## Description

Solves #2182. Tested on browser with a hacky implementation of #2177 (not included).

Should we add unit tests? Would have to fire up an HTTP server to test.

## Additions

- Added HTTP source

## PR Checklist

### DIFF
--- a/amethyst_assets/src/source/http.rs
+++ b/amethyst_assets/src/source/http.rs
@@ -14,17 +14,17 @@ use crate::{error, source::Source};
 /// Loads assets inside web worker using XmlHttpRequest.
 /// Used as a default source for WASM target.
 #[derive(Debug)]
-pub struct HTTP {
+pub struct HttpSource {
     loc: PathBuf,
 }
 
-impl HTTP {
+impl HttpSource {
     /// Creates a new http source.
     pub fn new<P>(loc: P) -> Self
     where
         P: Into<PathBuf>,
     {
-        HTTP { loc: loc.into() }
+        HttpSource { loc: loc.into() }
     }
 
     fn path(&self, s_path: &str) -> PathBuf {
@@ -35,7 +35,7 @@ impl HTTP {
     }
 }
 
-impl Source for HTTP {
+impl Source for HttpSource {
     fn modified(&self, _path: &str) -> Result<u64, Error> {
         #[cfg(feature = "profiler")]
         profile_scope!("http_modified_asset");
@@ -59,7 +59,9 @@ impl Source for HTTP {
             .with_context(|_| error::Error::Source)?;
 
         // Synchronous GET request. Should only be run in web worker.
-        xhr.open_with_async("GET", path_str, false).unwrap();
+        xhr.open_with_async("GET", path_str, false)
+            .map_err(|_| format_err!("XmlHttpRequest open failed"))
+            .with_context(|_| error::Error::Source)?;
         xhr.set_response_type(XmlHttpRequestResponseType::Arraybuffer);
 
         // We block here and wait for http fetch to complete
@@ -71,7 +73,7 @@ impl Source for HTTP {
         // Returns 0 if request was not completed.
         let status = xhr.status().unwrap();
         if status != 200 {
-            let msg = xhr.status_text().unwrap_or_else(|_| "".to_string());
+            let msg = xhr.status_text().unwrap();
             return Err(format_err!(
                 "XmlHttpRequest failed with code {}. Error: {}",
                 status,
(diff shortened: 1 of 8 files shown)


amethyst_assets/src/source/http.rs
```diff
@@ -0,0 +1,92 @@
...
+            .to_str()
+            .ok_or_else(|| format_err!("Path contains non-unicode characters: {:?}", path))
+            .with_context(|_| error::Error::Source)?;
+
+        let xhr = XmlHttpRequest::new()
+            .map_err(|_| format_err!("Failed to construct XmlHttpRequest"))
+            .with_context(|_| error::Error::Source)?;
+
+        // Synchronous GET request. Should only be run in web worker.
+        xhr.open_with_async("GET", path_str, false).unwrap();
```
azriel91: can the `unwrap()`s be replaced with `?`s? Just a bit neater

amethyst_assets/src/source/http.rs
```diff
@@ -0,0 +1,92 @@
...
+        // We block here and wait for http fetch to complete
+        xhr.send()
+            .map_err(|_| format_err!("XmlHttpRequest send failed"))
+            .with_context(|_| error::Error::Source)?;
+
+        // Status returns a result but according to javascript spec it should never return error.
+        // Returns 0 if request was not completed.
+        let status = xhr.status().unwrap();
+        if status != 200 {
+            let msg = xhr.status_text().unwrap_or_else(|_| "".to_string());
```
azriel91: might save us some debugging time:

```rust
use std::string::ToString; // somewhere on top of the file
let msg = xhr.status_text().unwrap_or_else(ToString::to_string);
```

amethyst_assets/src/source/http.rs
```diff
@@ -0,0 +1,92 @@
...
+
+use crate::{error, source::Source};
+
+/// HTTP source.
+///
+/// Loads assets inside web worker using XmlHttpRequest.
+/// Used as a default source for WASM target.
+#[derive(Debug)]
+pub struct HTTP {
+    loc: PathBuf,
```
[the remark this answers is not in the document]
chemicstry: That's what I initially came up with, but then noticed that native source is called just `Directory`. Should I rename just this one?

amethyst_assets/src/source/http.rs
chemicstry: Everything in `web_sys` returns `JsValue` as error, so they have to be mapped. Some of the javascript functions (such as this one) do not return any error, but the `Result` return type is still used for the rust API. So I don't know a case where this could panic. Bugs in browser/wasm? Should I add error map even though it could be dead code?

chemicstry:
Regarding tests, I think http server on CI is a good option. Or maybe we could have it as a dev dependency and only run it on wasm tests? So tests would not require any additional configuration.

However, another problem is that `web_sys` is just a wrapper around javascript functions and you can't run tests without javascript runtime. `wasm-bindgen` suggests using either node.js or headless browser https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/browsers.html I think we should test wasm on browser (since that's where the games will run) if it's possible to run headless browser on CI.

amethyst_assets/src/source/http.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+impl HTTP {
+    /// Creates a new http source.
+    pub fn new<P>(loc: P) -> Self
+    where
+        P: Into<PathBuf>,
+    {
+        HTTP { loc: loc.into() }
```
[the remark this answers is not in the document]
chemicstry: Yeah, that seems like a good idea for now to make examples compile. In the future we might want bundler to setup this path

amethyst_assets/src/source/http.rs
chemicstry: On the second thought, it's the same story as with `status`. According to javascript specification this function should never fail. Even if we want to handle the error, its type is `JsValue` and we don't know what type to cast it to (`ToString` is not implemented for `JsValue` neither is debug). I suggest leaving it as unwrap unless there is a good reason to throw a generic error which is not really helpful.

amethyst_assets/src/source/http.rs
azriel91: > Bugs in browser/wasm? Should I add error map even though it could be dead code?

haha, my conservative self always maps errors instead of unwraps -- wary of the 1/1000 case where it could save time debugging. Admittedly it isn't always pragmatic. Am happy to take the unreachable unwraps :v:

azriel91:
bors retry

azriel91:
bors retry

Issue #2192, opened 2020-03-24
WASM: fix GL rendering


chemicstry:
Currently pong in wasm looks like this:

![pong](https://i.imgur.com/HuYkC59.png)

Ball sprite is glitchy, paddles only appear on the first frame.
Needs investigation.

azriel91:
On branch [`feature/2192/fix-gl-rendering`](https://github.com/amethyst/amethyst/tree/feature/2192/fix-gl-rendering):

* While running pong, these log messages appear:

    ```
    [WARN][gfx_backend_gl] [API/Performance] ID 14 : Incompatible sampling format (ISL_FORMAT_R8G8B8A8_UNORM_SRGB) for rbc (MESA_FORMAT_R8G8B8A8_SRGB)
    ```

* These [come from the mesa open GL driver](https://gitlab.freedesktop.org/mesa/mesa/-/blob/master/src/mesa/drivers/dri/i965/intel_mipmap_tree.c#L1956).
* The mapping [appears to be intentional](https://gitlab.freedesktop.org/mesa/mesa/-/blob/master/src/mesa/drivers/dri/i965/brw_surface_formats.c#L68) -- note this is the same repository.

azriel91:
When running the `ui` example with `"gl"`, `gfx` panics in `src/backend/gl/src/device.rs:1306`:

```rust
let desc = conv::describe_format(format).unwrap();
```

This is because `describe_format` returns `None` when `format` is `Rgb8Srgb`:

```
[/mnt/data/work/github/gfx-rs/gfx/src/backend/gl/src/device.rs:1306] format = Rgb8Srgb
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /mnt/data/work/github/gfx-rs/gfx/src/backend/gl/src/device.rs:1306:20
```

Rendy uses this in `texture/src/format/palette.rs`: 

```rust
define_load! {
    pub fn load_from_srgb<T: Component>(Srgb) -> Rgb8Srgb NO;
    // ..
}
```

chemicstry:
I'm not sure if it's because I'm running windows, but `amethyst_ui` also crashes due to incorrect `Swizzle` for gl. I didn't dig deep, but it seems that vulkan accepts any swizzle and does the transformation inside the driver. GL doesn't support that and expects swizzle to always be RGBA.
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Swizzle(Zero, Zero, Zero, R)`,
[...]

PR #1698, opened 2019-06-12
Fix glyph rendering on metal backend.

Frizi:
## Description

Now glyph texture is using metal compatible pixel format and swizzle. Black rgb part of texture is biased to white in the shader instead, which required additional attribute. Final solution should be cross-platform.

Fixes #1663

PR #2217, opened 2020-04-04
WASM: Fix glyph transparency on GL

chemicstry:
This PR fixes UI glyph rendering on GL. Previously glyph data was provided as R8Unorm and swizzled into alpha channel, however, GL backend (and dx11 if it's ever used) does not support swizzling. So we build the full texture instead. I'm not sure if this could cause performance problems? This also makes https://github.com/amethyst/amethyst/pull/1698 obsolete, so I removed redudant shader attributes.

Tested on GL and vulkan, but not on metal (would appreciate someone testing UI examples).

Although vulkan complains about image layout, but this already existed before the changes. I was unable to fix it.
```
[WARN][gfx_backend_vulkan] 
PERFORMANCE [UNASSIGNED-CoreValidation-DrawState-InvalidImageLayout (0)] : vkCmdCopyBufferToImage(): For optimal performance VkNonDispatchableHandle 0xde3f1a000000001f[] layout should be VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL instead of GENERAL.
object info: (type: COMMAND_BUFFER, hndl: 219292848)
```

Part of #2192


chemicstry:
Did some profiling on performance. Ran `states_ui` example and measured execution times of 2 glyph rendering batches (one to render menu and other to render demo screen).

Before this PR: `43.37 ms` and `175.43 ms`
After this PR: `47.59` ms and `184.59 ms`

So there is some performance hit. Alternative would be to `cfg` everything into GL and non-GL, but in my opinion that would really affect readability of the code. Any thoughts?

On a side note, I looked into implementing swizzling in `gfx`, but their API abstractions make it really hard for GL backend.

Issue #2195, opened 2020-03-25
WASM: audio


chemicstry:
Current `cpal` only supports `emscripten`, but not `wasm-bindgen`. We probably don't want any external javascript dependencies.

Blocked on:
https://github.com/RustAudio/cpal/issues/212 / https://github.com/RustAudio/cpal/pull/372
https://github.com/RustAudio/rodio/pull/247

azriel91:
Tried using the above branches, and updated amethyst. Chrome has this error:

```js
Unable to write sample data into the audio context buffer: JsValue(TypeError: Failed to execute 'copyToChannel' on 'AudioBuffer': The provided ArrayBufferView value must not be shared.
// ..
at real (http://localhost:8000/pkg/pong_wasm.js:200:20))', /home/azriel/.cargo/git/checkouts/cpal-df9dae205d491af1/61d1a25/src/host/webaudio/mod.rs:273:25
```

The [web audio API spec](https://webaudio.github.io/web-audio-api/#dom-audiobuffer-copytochannel) also says this:

> Note: This means that `copyToChannel()` cannot be used to change the content of an `AudioBuffer` currently in use by an `AudioNode` that has acquired the content of an `AudioBuffer` since the `AudioNode` will continue to use the data previously acquired.

Perhaps this means we cannot use the same `OutputStreamHandle` to play multiple audio files, and should instead create a new output stream each time.

azriel91:
In [`cpal/src/host/webaudio/mod.rs:273`](https://github.com/amethyst/cpal/blob/ca3db4f/src/host/webaudio/mod.rs#L273-L275) this call happens:

```rust
// panic is from this copy_to_channel / expect
ctx_buffer
    .copy_to_channel(&mut temporary_channel_buffer, channel as i32)
    .expect("Unable to write sample data into the audio context buffer");
```

In the error message `The provided ArrayBufferView value must not be shared`, the `ArrayBufferView` likely refers to the `temporary_channel_buffer`, rather than the `ctx_buffer`. Nevertheless, I have tried cloning both `ctx_buffer` and `temporary_channel_buffer` before that method call, and the same error is hit.

I'm not sure what to do next, save for looking into the generated JS bindings / chromium source code. I have asked for help in:

* Rust community discord
* https://github.com/RustAudio/cpal/issues/212#issuecomment-610144281

Perhaps waiting will give us an answer.

azriel91:
[...]

PR #2219, opened 2020-04-06
Feature/2195/wasm audio

azriel91:
## Description

Audio support for WASM applications (#2195).

## PR Checklist

- [x] Ran `cargo +stable clippy --all --features "gl"`
- [x] Ran `cargo test --all --features "gl"`

### DIFF
--- a/amethyst_audio/src/bundle.rs
+++ b/amethyst_audio/src/bundle.rs
@@ -27,7 +27,7 @@ impl<'a, 'b> SystemBundle<'a, 'b> for AudioBundle {
         builder: &mut DispatcherBuilder<'a, 'b>,
     ) -> Result<(), Error> {
         builder.add_thread_local(AudioSystemDesc::new().build(world));
-        builder.add_thread_local(Processor::<Source>::new());
+        builder.add(Processor::<Source>::new(), "source_processor", &[]);
         Ok(())
     }
 }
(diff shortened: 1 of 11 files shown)


amethyst_audio/src/bundle.rs
```diff
@@ -18,19 +18,15 @@ use crate::{output::Output, source::*, systems::AudioSystemDesc};
...
         self,
         world: &mut World,
         builder: &mut DispatcherBuilder<'a, 'b>,
     ) -> Result<(), Error> {
-        builder.add(
-            AudioSystemDesc::new(self.0).build(world),
-            "audio_system",
-            &[],
-        );
+        builder.add_thread_local(AudioSystemDesc::new().build(world));
```
chemicstry, line 29: `#[cfg(feature = "wasm")]` ?

azriel91: Hm, `rodio::OutputStream` is `!Send + !Sync`, so it can't be stored in the `World` as a resource.
I put it in the `AudioSystem` since that's where `Output` used to live (and that was `Send + Sync`).

We could require the user to instantiate `OutputDevice` outside the `Application`, and pass in the `Output` (simply a wrapper for `rodio::OutputStreamHandle` -- I tried to not break existing API too much).

Haven't actually gotten audio working in pong WASM yet, but we should decide by then what to do

amethyst_audio/src/output.rs
```diff
@@ -2,46 +2,46 @@
...
+impl Default for OutputDevice {
     fn default() -> Self {
-        default_output_device()
-            .map(|device| Output {
-                device: Arc::new(device),
-            })
-            .expect("No default output device")
+        let output_device = default_output_device().expect("Failed to get default output device.");
+
+        error!("Using: {}", output_device.device.name().unwrap());
```
azriel91, line 38: apply eraser

PR #2237, opened 2020-04-14
Add input context support

Xaeroxe:
## Description

Production videogames have several different contexts they need to support for input, whether it be due to game features such as vehicles, inventory menus, or alternate protagonists. Up until now Amethyst's answer to this has been to allow the developer to reassign and alter the `bindings` variable as needed on the `InputHandler`. This approach has a few downsides.

1. Can't store all input bindings in a single file
2. It's not immediately clear to users how they should change the `bindings` variable after load, or that they even can.
3. It requires the user to juggle around various `Bindings` structures, with no provided location as to where they should live.

So, this PR introduces `Context`s. The InputHandler now stores multiple `Bindings` configurations, and decides which one to use based on the `Context` variable set on it. A `Context` can be anything, as it is defined in all places by a generic. If users wish to opt out of this feature, they can use the `()` Context value.

## Additions

- amethyst_input::Context
- amethyst_input::InputHandler::new_with_context
- amethyst_input::InputHandler::active_bindings
- amethyst_input::InputHandler::active_bindings_mut
- amethyst_input::InputHandler::set_context
- amethyst_input::InputHandler::context
- amethyst_input::InputHandler::set_bindings_for_context
- amethyst_input::InputHandler::bindings_for_context
- amethyst_input::InputHandler::bindings_for_context_mut
- amethyst_input::InputHandler::remove_bindings_for_context

## Removals

- amethyst_input::InputHandler::bindings is no longer public.

## Modifications

- amethyst_input::InputHandler now accepts a `Context` generic parameter
- amethyst_input::InputBundle now accepts a `Context` generic parameter
- amethyst_input::InputSystem now accepts a `Context` generic parameter
- amethyst_input::InputSystemDesc now accepts a `Context` generic parameter
- amethyst_input::SdlEventsSystem now accepts a `Context` generic parameter
- amethyst_input::SdlEventsSystemDesc now accepts a `Context` generic parameter
- amethyst_controls::FlyControlBundle now accepts a `Context` generic parameter
- amethyst_controls::FlyMovementSystem now accepts a `Context` generic parameter
- amethyst_test::AmethystApplication::ui_base now accepts a `Context` generic parameter
- amethyst_test::AmethystApplication::with_ui_bundle now accepts a `Context` generic parameter
- amethyst_ui::UiBundle now accepts a `Context` generic parameter
- amethyst_ui::UiMouseSystem now accepts a `Context` generic parameter
- amethyst_ui::SelectionMouseSystem now accepts a `Context` generic parameter
- amethyst_ui::SelectionMouseSystemDesc now accepts a `Context` generic parameter
- amethyst_ui::DragWidgetSystem now accepts a `Context` generic parameter

## PR Checklist

### DIFF
--- a/amethyst_controls/src/bundles.rs
+++ b/amethyst_controls/src/bundles.rs
@@ -7,7 +7,7 @@ use amethyst_core::{
     SystemDesc,
 };
 use amethyst_error::Error;
-use amethyst_input::BindingTypes;
+use amethyst_input::{BindingTypes, Context};
 
 use super::*;
 
@@ -33,16 +33,17 @@ use super::*;
 /// * `MouseFocusUpdateSystem`
 /// * `CursorHideSystem`
 #[derive(Debug)]
-pub struct FlyControlBundle<T: BindingTypes> {
+pub struct FlyControlBundle<C: Context, T: BindingTypes> {
     sensitivity_x: f32,
     sensitivity_y: f32,
     speed: f32,
     right_input_axis: Option<T::Axis>,
     up_input_axis: Option<T::Axis>,
     forward_input_axis: Option<T::Axis>,
+    phantom: PhantomData<C>,
 }
 
-impl<T: BindingTypes> FlyControlBundle<T> {
+impl<C: Context, T: BindingTypes> FlyControlBundle<C, T> {
     /// Builds a new fly control bundle using the provided axes as controls.
     pub fn new(
         right_input_axis: Option<T::Axis>,
@@ -56,6 +57,7 @@ impl<T: BindingTypes> FlyControlBundle<T> {
             right_input_axis,
             up_input_axis,
             forward_input_axis,
+            phantom: PhantomData,
         }
     }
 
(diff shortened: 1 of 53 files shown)


chemicstry:
I'm a bit sceptical about the generic type proliferation, making the code more difficult to understand and write. Wouldn't just regular string work as a context identifier? This wouldn't break any existing code and would result in way less rewrites.

Xaeroxe:
The main drawback to that approach is it's not as easy to validate correctness of the API usage. It becomes a run time problem rather than a compile time problem. The proliferation of the generic is unfortunate, however this approach is easier to validate correctness for, and has the added benefit of being easy to opt out of with the unit type.

maroider:
I have only one problem with this implementation, and that is that contexts are __always__ mutually exclusive. This isn't always what you want. In "Elite: Dangerous", contexts seem to act more like a stack, where each layer either lets non-conflicting button combinations "fall through" or catches all button combinations. This allows you to still (partially) maneuver your spaceship while accessing menus in the cockpit, even though some flight controls (Q, W, E, A, S, D, etc) are also used to navigate menus by default.

The simplest implementation I could think of is something like:
```Rust
#[derive(Derivative)]
#[derivative(Debug(bound = "C: std::fmt::Debug"))]
pub struct InputHandler<C, T>
where
    C: Context,
    T: BindingTypes,
{
    bindings: HashMap<C, Bindings<T>>,
    pub context_stack: Vec<FallThrough<C>>,
    ..
}

#[derive(Derivative)]
#[derivative(Debug(bound = "C: std::fmt::Debug"))]
pub enum FallThrough<C>
where
    C: Context,
{
    Allow(C),
    Block(C),
}
```
This would unfortunately require some changes to `InputHandler::send_event`, but shouldn't be too invasive beyond that.

jojolepro:
I think this is worthy of an RFC?

Xaeroxe:
Yeah at this point I think you're right. This PR was written during a spur of motivation to do work on Amethyst, which I find myself lacking at this instant. So I probably won't author the RFC anytime soon. I might later, but if people want to discuss this now someone else should write the RFC.

PR #2248, opened 2020-04-19
Reset y val on score in pong tutorial

szunami:
## Description

Not technically a bug, but when running through the tutorial, I found it very jarring that ball y doesn't reset on score. I updated the pong tutorial code to reset y val to the center on score.

## PR Checklist

### DIFF
--- a/book/src/pong-tutorial/pong-tutorial-05.md
+++ b/book/src/pong-tutorial/pong-tutorial-05.md
@@ -49,7 +49,7 @@ use amethyst::{
     ecs::prelude::{Join, System, SystemData, World, WriteStorage},
 };
 
-use crate::pong::{Ball, ARENA_WIDTH};
+use crate::pong::{Ball, ARENA_WIDTH, ARENA_HEIGHT};
 
 #[derive(SystemDesc)]
 pub struct WinnerSystem;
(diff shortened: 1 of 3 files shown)


szunami:
Without this change:

https://gyazo.com/00526f99b3da8ad152e98f74ba62f3bd

With this change:

https://gyazo.com/fbb392ea5a4fc44976929021f05e11cc

book/src/pong-tutorial/pong-tutorial-05.md
```diff
@@ -39,6 +39,7 @@ Then, we'll create `systems/winner.rs`:
 #     }
 #
 #     pub const ARENA_WIDTH: f32 = 100.0;
+#     pub const ARENA_HEIGHT: f32 = 100.0;
```
szunami, line 42: I'm unsure if this is legit.

azriel91: it's legit :v:

szunami: I see, makes sense. This explains some of the odd behavior I was seeing where using the code block copy button would generate non-compiling code; not sure if this is a known challenge?

szunami:
@azriel91 I believe this is good to go.

PR #2258, opened 2020-04-27
Fix asset handle reuse in rendy subs

chemicstry:
## Description

Asset handle id is used as a unique identifier in renderer, however, handle ids can be reused and this causes renderer to use old assets. For example if `Material` is unloaded and new one is loaded with the same handle id, `MaterialSub` will not update sampler descriptors and old textures will be used.

To solve this, `WeakHandle` is used instead, which allows unloading of assets (`Handle` was previously used in `TextureSub` which prevented unloads) and checking if original asset is still present with `is_dead()`.

## PR Checklist

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -8,6 +8,14 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 [kc]: http://keepachangelog.com/
 [sv]: http://semver.org/
 
+## [Unreleased]
+
+### Fixed
+
+- Fixed asset handle reuse bug in renderer. ([#2258])
+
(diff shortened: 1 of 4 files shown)


docs/CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 [kc]: http://keepachangelog.com/
 [sv]: http://semver.org/
 
+## [Unreleased]
+
+### Fixed
+
+- Fixed asset handle reuse bug in renderer. ([#2258])
+
```
jojolepro, line 16: Missing the link to the PR.

chemicstry:
bors try

PR #2265, opened 2020-05-01
Fixed a typo in README

terberes:
Pacman should be called with root privileges, just like apt, dnf, etc.


jojolepro:
bors r+
Thanks!

PR #2266, opened 2020-05-02
Move footnotes into links at appropriate places in text

rschulman:
## Description

Prior to this PR, this page in the book had a number of links at the bottom that appeared to be footnotes but which were rendering in such a way as to be unreadable. I fixed this by removing the footnotes entirely and adding links in the text where appropriate.

## Additions

- No API changes

## Removals

- No API changes

## Modifications

- Only book changes

## PR Checklist


rschulman:
Yes, it should if that's preferable. I'll change them around.

PR #2267, opened 2020-05-02
Fix renderable example after PR clash

chemicstry:
Just a quick fix after two PRs (#2243 and #2223) fixed the same thing and we ended up with two `InputBundle`s.


chemicstry:
bors retry

chemicstry:
bors retry

chemicstry:
bors retry

Issue #2272, opened 2020-05-07
[BUG] Resource fetch error despite registering in pong tutorial


jembishop:
## Description

In the tutorial I get `Tried to fetch resource of type MaskedStorage<Paddle>[^1] from the World but the resource does not exist.` error despite registering transform bundle.

## Reproduction Steps

The code is copied from https://github.com/amethyst/amethyst/tree/master/examples/pong_tutorial_02 with the sprite parts and `world.register::<Paddle>();` removed, which should be the code you have at this stage of the tutorial. 

pong.rs
```rust 
use amethyst::{
    assets::{AssetStorage, Handle, Loader},
    core::transform::Transform,
    ecs::prelude::{Component, DenseVecStorage},
    prelude::*,
    renderer::{Camera, ImageFormat, SpriteRender, SpriteSheet, SpriteSheetFormat, Texture},
};

const ARENA_HEIGHT: f32 = 100.0;
const ARENA_WIDTH: f32 = 100.0;

const PADDLE_HEIGHT: f32 = 16.0;
const PADDLE_WIDTH: f32 = 4.0;

pub struct Pong;

impl SimpleState for Pong {
    fn on_start(&mut self, data: StateData<'_, GameData<'_, '_>>) {
        let world = data.world;

        // Load the spritesheet necessary to render the graphics.
        // `spritesheet` is the layout of the sprites on the image;
        // `texture` is the pixel data.

        //world.register::<Paddle>(); should not be necessary right?

        initialise_paddles(world);
        initialise_camera(world);
    }
}

#[derive(PartialEq, Eq)]
enum Side {
    Left,
    Right,
}

struct Paddle {
    pub side: Side,
    pub width: f32,
    pub height: f32,
}

impl Paddle {
    fn new(side: Side) -> Paddle {
        Paddle {
            side,
            width: PADDLE_WIDTH,
            height: PADDLE_HEIGHT,
        }
    }
}

impl Component for Paddle {
    type Storage = DenseVecStorage<Self>;
}

/// Initialise the camera.
fn initialise_camera(world: &mut World) {
    // Setup camera in a way that our screen covers whole arena and (0, 0) is in the bottom left.
    let mut transform = Transform::default();
    transform.set_translation_xyz(ARENA_WIDTH * 0.5, ARENA_HEIGHT * 0.5, 1.0);

    world
        .create_entity()
        .with(Camera::standard_2d(ARENA_WIDTH, ARENA_HEIGHT))
        .with(transform)
[...]

PR #2273, opened 2020-05-07
Added subtitle to avoid confusion.

jojolepro:
fixes https://github.com/amethyst/amethyst/issues/2272


jembishop:
Ah thanks. For future reference I thought that the 
> There must be a better way. Hint: there is.
 
Part was followed by the explanation for how to avoid this, but it was in fact a brand new section, and the explanation comes later. Thanks for changing this to be more clear.

PR #2296, opened 2020-06-05
Bind different TileMapArgs for each TileMap being rendered

### LINKED ISSUE #2210: [BUG] Only the last TileMap created is visible
## Description

If you create more than one `TileMap` Entity, only the last one is visible, while any other created previously to that one are not visible.

## Reproduction Steps

1. Create two TileMap entities
2. Place one at some coordinate, the other at a different coordinate.
3. You'll only see one, the one that was created last.

You can easily do this with the current `tiles` example, duplicating the map being rendered, but changing its transform.

Both `TileMap`s visible at their corresponding positions.

## What Actually Happened

Only one `TileMap` is visible.

## Your Environment

1. Operating System: Windows 10
2. Version of Amethyst: Master

## Additional Context

I'm trying to figure out `pass.rs` for tiles, but I have no experience with rendy and negligible experience in graphics programming, having said that, it looks to me that there is only one `TileMapArgs`, containing the dimensions and transform of the map to be rendered, so its values are from the last iteration of the `join` loop,  and so both TileMaps are rendered at the same spot with same sizes. But I have no idea where to go from here.

https://github.com/amethyst/amethyst/blob/6453a83c4dc5b6152d35004d64931f5e09f98e65/amethyst_tiles/src/pass.rs#L214

jgraef:
## Description

Fixes #2210. In `prepare` only one `TileMapArgs` was constructed for all `TileMap`s. Thus the transformations of the last `TileMap` were applied to all `TileMap`s.

This PR changes the `prepare` and `draw_inline` to keep track of a `Vec` of `TileMapArgs` (one for reach `TileMap`). `draw_inline` then binds a different one for each `TileMap`. I'm not completely sure, if this is the best approach. Feel free to improve this - but this works for me :)

I also moved some code into `compute_region` to please clippy (`prepare` had too many lines :unamused: )

This PR doesn't change any APIs.

## PR Checklist

- [X] Ran `cargo clippy --all --features "empty"`
- [X] Ran `cargo test --all --features "empty"`

### DIFF
--- a/amethyst_tiles/src/pass.rs
+++ b/amethyst_tiles/src/pass.rs
@@ -13,7 +13,7 @@ use amethyst_core::{
 
 use amethyst_assets::{AssetStorage, Handle};
 use amethyst_rendy::{
-    batch::{GroupIterator, OneLevelBatch, OrderedOneLevelBatch},
+    batch::{GroupIterator, OneLevelBatch, OrderedTwoLevelBatch},
     bundle::{RenderOrder, RenderPlan, RenderPlugin, Target},
     camera::{ActiveCamera, Camera, Projection},
     pipeline::{PipelineDescBuilder, PipelinesBuilder},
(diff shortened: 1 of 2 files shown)


jgraef:
There are 2 unwrap's that I'm not so happy about:

 - pass.rs:293
 - pass.rs:321: I'm pretty sure `tilemap_args_index` will never be out of bounds here. But maybe double-check this.

AlveLarsson:
thanks! bors r+

PR #2354, opened 2020-07-07
Add option to opt-out of thread-pool

modestmicha:
## Description

This is useful for rare cases where threading is not available and for some cases of debugging. `cfg(no_threading)` was already used in the codebase; just not consistently. It should be noted that in specs and shred threading is behind a feature called `parallel` instead.

## Modifications

`ArcThreadPool` got renamed to `ThreadPool`. It's now a structure that wraps the `Arc<ThreadPool>` and proxies its `spawn` method so no more deref required. It was used inconsistently thus far as the type alias was often omitted. Most users should've never touched this resource before.

## PR Checklist

- [x] Modified unit tests for new code added in this PR.

### DIFF
--- a/amethyst_assets/src/lib.rs
+++ b/amethyst_assets/src/lib.rs
@@ -27,7 +27,7 @@ pub use crate::{
     storage::{AssetStorage, Handle, ProcessingState, Processor, WeakHandle},
 };
 
-pub use rayon::ThreadPool;
+pub use amethyst_core::ThreadPool;
 
 mod asset;
 mod cache;
--- a/amethyst_core/src/lib.rs
+++ b/amethyst_core/src/lib.rs
@@ -23,8 +23,6 @@ pub use num_traits as num;
 pub use specs as ecs;
 pub use specs::{shred, shrev};
 
-use std::sync::Arc;
-
 pub use crate::{
     bundle::SystemBundle,
     event::EventReader,
@@ -55,6 +53,7 @@ mod hide_system;
 mod named;
 mod system_desc;
 mod system_ext;
+mod thread_pool;
 
-/// A rayon thread pool wrapped in an `Arc`. This should be used as resource in `World`.
-pub type ArcThreadPool = Arc<rayon::ThreadPool>;
+/// A rayon thread pool proxy. This should be used as resource in `World`.
+pub type ThreadPool = thread_pool::ThreadPool;
--- a/amethyst_core/src/thread_pool.rs
+++ b/amethyst_core/src/thread_pool.rs
@@ -1,0 +1,53 @@
+use amethyst_error::Error;
+#[cfg(not(no_threading))]
+use std::sync::Arc;
+
+#[cfg(not(no_threading))]
+#[derive(Debug, Clone)]
+pub struct ThreadPool {
+    pool: Arc<rayon::ThreadPool>,
+}
+
+#[cfg(no_threading)]
+#[derive(Debug, Clone)]
+pub struct ThreadPool {}
+
+impl ThreadPool {
+    #[cfg(not(no_threading))]
+    pub fn new(thread_count: Option<usize>) -> Result<Self, Error> {
+        use rayon::ThreadPoolBuilder;
+        let thread_pool_builder = ThreadPoolBuilder::new();
+        #[cfg(feature = "profiler")]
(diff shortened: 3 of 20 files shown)


amethyst_core/src/lib.rs
```diff
@@ -55,6 +53,7 @@ mod hide_system;
 mod named;
 mod system_desc;
 mod system_ext;
+mod thread_pool;
 
-/// A rayon thread pool wrapped in an `Arc`. This should be used as resource in `World`.
-pub type ArcThreadPool = Arc<rayon::ThreadPool>;
+/// A rayon thread pool proxy. This should be used as resource in `World`.
+pub type ThreadPool = thread_pool::ThreadPool;
```
modestmicha, line 59: Could just as well call it `ArcThreadPool` as it just wraps the `Arc<rayon::ThreadPool>` but the user is not expected to clone it so this may fit better.

amethyst_assets/src/lib.rs
```diff
@@ -27,7 +27,7 @@ pub use crate::{
     storage::{AssetStorage, Handle, ProcessingState, Processor, WeakHandle},
 };
 
-pub use rayon::ThreadPool;
+pub use amethyst_core::ThreadPool;
```
modestmicha, line 30: Not sure why we re-export this here at all. Should this be removed?

amethyst_core/src/thread_pool.rs
```diff
@@ -0,0 +1,53 @@
...
+    pool: Arc<rayon::ThreadPool>,
+}
+
+#[cfg(no_threading)]
+#[derive(Debug, Clone)]
+pub struct ThreadPool {}
+
+impl ThreadPool {
+    #[cfg(not(no_threading))]
+    pub fn new(thread_count: Option<usize>) -> Result<Self, Error> {
```
modestmicha, line 17: Taking an `Option` may be awkward. Two constructors would complicate things a little bit for the caller. An enum would be another option to be more explicit.

amethyst_core/src/thread_pool.rs
```diff
@@ -0,0 +1,53 @@
+use amethyst_error::Error;
+#[cfg(not(no_threading))]
+use std::sync::Arc;
+
+#[cfg(not(no_threading))]
+#[derive(Debug, Clone)]
+pub struct ThreadPool {
```
modestmicha, line 7: There's not necessarily a need to wrap `Arc<rayon::ThreadPool>` but I found that otherwise there would be not so nice to read `#[cfg(` in quite a lot of places.

jojolepro:
Threading should actually be *enabled* by a feature gate that is on by default. Using feature gates to remove features is known to be error prone and confusing.
:)

modestmicha:
@jojolepro In that case we'd add a `parallel` feature to all amethyst crates which would have to be passed to all amethyst crates by default. That's quite a lot of passing down. But can be done.

khionu:
There's two issues, atm:
1) Right now, we're not merging any PRs to the code base until Legion is ported
2) This isn't an insignificant addition that the Amethyst Team would then have to maintain, and as far as I've been able to tell this was opened without any previous discussion. It would be best if you could open a discussion on the [forums](https://community.amethyst.rs/) about your motivations and the benefits this change would bring.

modestmicha:
Thanks, I was unaware of the ongoing efforts.
1. I might adjust and re-open once Legion is ported.
2. This PR contains a feature I'd like to have and therefore is my way to open a discussion.
It would be great if there would be a contributing guideline as it's hard to understand why I'd have to use the forum, start a possibly lengthy discussion that might result in nothingness even if approved due to no one willing to implement.
\
With code on the table, I personally find it easier to reason about. That doesn't mean it's expected to be merged. I usually open against upstream so that I don't have to maintain a fork.

khionu:
I can appreciate your reasoning, for sure. That said, this is a large scale project, I would recommend for projects of this size you check out their communication venues.

CleanCut:
>     1. I might adjust and re-open once Legion is ported.

@modestmicha I'd love to be involved in the conversation when you do!  I'm really curious what platform(s) you would want to disable threading.

I also want to make sure that you know that you are welcome to discuss these sorts of things and submit PRs.  There's no guarantee everyone will agree with everything, but we still welcome your participation! 😄

PR #2356, opened 2020-07-08
Change proposal to define_the_sprite_sheet.md

JohnyProkie:
## Description

On @CleanCut [advice](https://community.amethyst.rs/t/how-can-i-know-sprite-really-got-loaded/1556/3?u=sophisticatedspoon), this PR adds clarity to declaring values for ron files, used by [SpriteGrid](https://docs.amethyst.rs/stable/amethyst_rendy/sprite/struct.SpriteGrid.html) and [SpriteList](https://docs.amethyst.rs/stable/amethyst_rendy/sprite/struct.SpriteList.html)

## PR Checklist

### DIFF
--- a/book/src/sprites/define_the_sprite_sheet.md
+++ b/book/src/sprites/define_the_sprite_sheet.md
@@ -38,7 +38,6 @@ List((
     ],
 ))
 ```
-`offsets: Some((0.0, 0.0)),` can be replaced by `offsets: (0.0, 0.0),` if the line `#![enable(implicit_some)]` is added at the top of the definition file.
 
 Or you can use a grid based definition, for example:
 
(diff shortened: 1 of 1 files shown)


CleanCut:
Thanks for making the PR! This is a great start.

One thing we need to make clear is that it isn't a bounds problem, it's an issue with some types being wrapped in `Option<...>`.  This is a general issue with all prefabs, I believe. If the type of an argument is `Option<u32>`, for example, then you need to wrap the value in the `Some` variant.  You can see which SpriteGrid fields are wrapped in `Option` by looking at the [API documentation](https://docs.amethyst.rs/stable/amethyst_rendy/sprite/struct.SpriteGrid.html) (4 of the fields).

Alternatively, you can attack the problem by putting `#![enable(implicit_some)]` at the top of the file, which I would guess makes it so the prefab loading process attempts to wrap anything that hits a type error in `Some` and tries loading it again.

JohnyProkie:
I see now, it seemed to me it was problem of bounds, but I wasn't able to simulate what brought me to this idea so in the end I was too quick in making an assumption.
So I changed text, but did not feel brave enough to mention it can be an issue with all prefabs.

CleanCut:
Awesome, the information is now accurate!

Next, it would be good to combine this explanation with the previous section, while making it as short and concise as possible.  Do you mind if I push a commit with that change? Then you could take a look at it and see if it would have made sense to you.

CleanCut:
I went ahead and pushed a commit with my suggestion for how to combine and shorten the existing and new sections. Please take a look and let me know what you think.

JohnyProkie:
Didn't realize that, much better :+1:

AlveLarsson:
Is this ready?

JohnyProkie:
I removed redundant description. 
Also information about (what I consider) silent fail when value is not wrapped in `Some` or absence of `enable(implicit_some)` can be confusing. Currently we covered how spritesheet should be defined and properties that can be defined are described in documentation, so unexpected behavior caused by wrong definition is extra information, that naturally comes from context.

Another thing is failing Jenkins, but I would be really surprised if this is related to proposed change.
I would say this is ready.

PR #2364, opened 2020-07-12
GitLab CI

CleanCut:
~Testing bors with GitLab!~

Working on getting GitLab CI set up and working nicely.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -15,7 +15,6 @@ test:
       - target/debug/build/
       - .cargo/
 
-
 # stages:
 #   - build-docs
 #   - deploy-docs
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -7,7 +7,7 @@ stages:
 test:
   script:
     - apt-get -y update
-    - apt-get install -y alsa-utils
+    - apt install -y gcc pkg-config openssl libasound2-dev cmake build-essential python3 libfreetype6-dev libexpat1-dev libxcb-composite0-dev libssl-dev libx11-dev
     - cargo test --features empty
 
 # stages:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -6,9 +6,11 @@ stages:
 
 test:
   script:
-    - apt-get -y update
-    - apt install -y gcc pkg-config openssl libasound2-dev cmake build-essential python3 libfreetype6-dev libexpat1-dev libxcb-composite0-dev libssl-dev libx11-dev
     - cargo test --features empty
+  cache:
+    key: debug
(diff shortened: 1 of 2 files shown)


CleanCut:
Dangit. I got a precompiled target directory into the image, I put it in the right place during the CI run, and it _still_ rebuilt everything.  There's obviously some trick to it that I'm missing.

onelson:
Fwiw, we recently setup https://github.com/mozilla/sccache to help reuse build artifacts in gitlab CI and it's been working well for us. We use redis for the storage backend, but you could also use disk in combination with gitlab's caching if redis isn't possible.

CleanCut:
@onelson Thanks! I think that will help. 😄

onelson:
To maybe give you a jump start, the way we went about it was to add sccache to the docker image we use to run all our rust ci jobs, a la

```Dockerfile
ARG SCCACHE_VER=0.2.13

RUN curl -sSL https://github.com/mozilla/sccache/releases/download/$SCCACHE_VER/sccache-$SCCACHE_VER-x86_64-unknown-linux-musl.tar.gz \
	-o /tmp/sccache.tar.gz \
	&& cd /opt \
	&& tar xf /tmp/sccache.tar.gz \
	&& rm /tmp/sccache.tar.gz

ENV PATH=/opt/sccache-$SCCACHE_VER-x86_64-unknown-linux-musl:$PATH
```

Then in the gitlab ci conf, we activate the cache by setting some env vars and starting/stopping the sscache server, bookending any cargo calls for the job. Worth noting that the start/stop have to happen under the `script` key where the actual cargo calls are, and can't be split up into the `before_script` and `after_script` keys.


```yaml
test:
  script:
    - sccache --start-server
    - cargo fmt --all -- --check
    - cargo tarpaulin
    - sccache --stop-server
  variables:
    RUSTC_WRAPPER: sccache
    SCCACHE_REDIS: redis://HOST:PORT
```

The redis instance is run with the following flags, and backed by a volume so it can persist snapshots to disk

```sh
docker-entrypoint.sh redis-server --appendonly yes --maxmemory 3072mb --maxmemory-policy allkeys-lru
```

Naturally, a good number for the memory limit will depend on the projects being built. `sscache` will spit out a little report each time your jobs use the `--stop-server` flag so you can see what the hit/miss ratio is, and the overall size of the cache. This could help you shrink to fit if you start out too big. A high miss-rate might mean you stated too low.

CleanCut:
@onelson Thanks! That info about sccache will likely be useful.
---
I am closing this GitLab CI PR in favor of the GitHub Actions PR. My rationale is explained in https://github.com/amethyst/amethyst/pull/2382#issuecomment-665251877

PR #2376, opened 2020-07-19
Mark slice_as_bytes and write_into_slice as unsafe

MikailBag:
## Description

These functions are not safe: if they are used incorrectly, program's behavior is undefined.
For example [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=73bfe2c72ca24b46a0f212a5791da09c) code fails under MIRI.

## Modifications
This functions are now marked as unsafe.
Please note that PR currently is incomplete: it does not update usages of `write_into_slice` functions.
There are several ways forward:
1) Make those functions safe again, but require that `T: zerocopy::AsBytes`. It will require adding `#[derive(AsBytes)] somewhere.
2) Manually review that all usages are correct, and put `unsafe` blocks where needed.
3) Make `write_into_slice` function take `&mut [MaybeUninit<u8>]` on input. If that slice it later passed to C ffi, it can be sound.

I cannot do the choice, because I know nothing about Amethyst's rendering subsystem.
 

## PR Checklist

### DIFF
--- a/amethyst_rendy/src/util.rs
+++ b/amethyst_rendy/src/util.rs
@@ -242,18 +242,20 @@ impl<I: Hash + Eq> LookupBuilder<I> {
 }
 
 /// Convert any type slice to bytes slice.
-pub fn slice_as_bytes<T>(slice: &[T]) -> &[u8] {
-    unsafe {
-        // Inspecting any value as bytes should be safe.
-        core::slice::from_raw_parts(
-            slice.as_ptr() as *const u8,
-            core::mem::size_of::<T>() * slice.len(),
-        )
-    }
+/// # Safety
+/// `T` must not contain any padding bytes. In particular, `T` must be repr(C).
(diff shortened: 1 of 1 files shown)


0x6273 (approved):

chemicstry (changes requested):
I'm not sure why this was approved as it is clearly incomplete.

The changes are alright, but I'm not sure what's the best way to handle usage. There aren't many places they are used, so maybe ensuring safety at usage site is the best way to go. Also, this is mostly used for glsl_layout pod types, so maybe a safe impl for these could be done.

jojolepro:
I think this will require further discussions on how to go forward. @MikailBag would you be willing to lead a discussion on this topic on the forums (community.amethyst.rs)? Thanks!

jojolepro:
I think we should go with option 2: adding unsafe at the usage point with a comment explaining why. Would you be available to do it? :)

PR #2379, opened 2020-07-20
Experimenting with GitHub Actions for CI

CleanCut:
Experimenting with GitHub Actions for CI.


CleanCut:
Hmm. GitHub Actions aren't showing up.  I'll try this through the walkthrough on the actions page instead. Maybe there's some magic I missed.

PR #2380, opened 2020-07-20
Create rust ci actions workflow

CleanCut:
Experimenting with GitHub Actions as a CI solution, take 2.

This is a work-in-progress (not ready for review), but I'm not going to put this in "Draft" mode in case that's what's broke my last attempt.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,8 +2,7 @@ name: Amethyst CI via GitHub Actions
 
 on:
   push:
-    branches:
-      - *
+    branches: [ $default-branch ]
   pull_request:
     branches: [ $default-branch ]
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,7 +2,8 @@ name: Amethyst CI via GitHub Actions
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


CleanCut:
@fhaynes Does Actions need to be "enabled" on the `amethyst` org or the `amethyst/amethyst` repo or something?

fhaynes:
Actions is enabled for all repos in the organization, it says. There are no
runners configured, though.

fhaynes:
Did you try going through https://github.com/amethyst/amethyst/actions/new ?

fhaynes:
Looking at main.yml, I think this is telling it to only run on the default branch, which in our case right now would be `master`, or PRs against `master`. Let's see if giving it an approving review helps.

fhaynes:
I changed the on push trigger to be `*` and not just $default-branch, and it seems to at least recognize that it should trigger now.

CleanCut:
@fhaynes  What it recognizes is that the file is now syntactically invalid. 🤣

image

CleanCut:
I'm switching the build to a no-op `echo` command and merging this into master to see if we can get actions working at all.

Issue #2359, opened 2020-07-09
[MAINTENANCE] PR Checklist vs Contributing Checklist


MrTanoshii:
## Description

Is there a particular reason why [Pull Request Template](https://github.com/amethyst/amethyst/blob/master/.github/pull_request_template.md) does not have `mdbook test book -L target/debug/deps` as quoted from [Contributing](https://github.com/amethyst/amethyst/blob/master/docs/CONTRIBUTING.md)

## Reason

Jenkins is, at the moment this Issue is written, failing because of the `mdbook test` which is not checked at all by the contributor.

## Impact

The test is an unnecessary extra step for contributors who did not modify the book.

## Additional Information

Jenkins runs
`mdbook test -L ./target/debug/deps book`
while [Contributing](https://github.com/amethyst/amethyst/blob/master/docs/CONTRIBUTING.md) recommends
`mdbook test book -L target/debug/deps`

CleanCut:
The To-Do's that I see in this issue are:
- [ ] Move the `mdbook`-running information from the "Pull Requests" section to the "Writing Documentation" section (it's extremely useful to run the command yourself and get quick feedback rather than waiting for CI).  Also, once CI works, one of the checks should alert you if your examples don't compile.
- [ ] Compare the "Pull Request" section to the template and make sure the rest of the information correlates.

@MrTanoshii Would you like to take a first pass at doing the update? I'd be happy to be the reviewer!

Related:
- Fixing the book is in progress in https://github.com/amethyst/amethyst/issues/2365
- Migrating to a CI that works with forks is in progress in https://github.com/amethyst/amethyst/pull/2364

MrTanoshii:
Added to my PR #2368

PR #2383, opened 2020-07-21
Fixing consistency between PR template and CONTRIBUTING.md

MrTanoshii:
## Description

As per #2359, this PR aims to fix the align both the PR template and CONTRIBUTING.md closer by ensuring that the procedures required are equivalent

Note: I have not added `cargo run --example {example-name} --features YOUR_BACKEND` to the PR template even though it is in CONTRIBUTING.md, should I?

## PR Checklist


docs/CONTRIBUTING.md
```diff
@@ -286,7 +282,13 @@ mdbook serve book
 Examples in the API can be tested with `cargo test`.  Examples in top-level markdown files (like the one we are currently in) are not tested.
 
 When submitting your pull requests, please follow the same procedures described
-in the [Pull Requests](#pull-requests) section above.
+in the [Pull Requests](#pull-requests) section above and the procedures below.
+All of the following commands completed without errors.
```
jojolepro, line 286: Phrasing is a bit weird and doesn't flow from the rest of the paragraph.

MrTanoshii:
Yup, I see what you mean, is this now better?

.github/pull_request_template.md
```diff
@@ -19,13 +19,14 @@ Add a brief summary of your PR here, between one sentence and two paragraphs
...
 - [ ] Acknowledged that by making this pull request I release this code under an MIT/Apache 2.0 dual licensing scheme.
+- [ ] Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.
+- [ ] Updated the content of the book if this PR would make the book outdated.
 
 If this modified or created any rs files:
 
 - [ ] Ran `cargo +stable fmt --all`
-- [ ] Ran `cargo clippy --all --features "empty"`
+- [ ] Ran `cargo clippy --all --features "empty"` (may require `cargo clean` before)
+- [ ] Ran `cargo build --features "empty"`
```
chemicstry, line 31: Do we really need to `build` since `test` would do that anyway?

MrTanoshii: I think it should be alright since `build` in normal mode while `test` is compiled in test mode.
That being said, I did not think that deeply when making the PR, it is purely to make the docs and template match.

If the more experienced guys say that `build` is unnecessary, I'll make the changes.

PR #2385, opened 2020-07-21
Fix some broken links

CastilloDel:
## Description

Just fixing two broken links in amethyst_core

## PR Checklist


CastilloDel:
I didn't change it because Application still exists. The difference is that now it is a type alias to provide defaults for Event and EventReader.

0x6273:
Your link does not lead to the type alias, but the actual struct. And the name needs to match the type it's linking to, else it's confusing. I think linking to https://docs.amethyst.rs/master/amethyst/type.Application.html would make more sense.

CastilloDel:
Yes, that makes more sense. I didn't know cargo doc generated a file for the type alias. I'll change it

Issue #1846, opened 2019-08-02
[Bug] Image count not supported with Vulkan backend on Wayland


wackbyte:
# Tasks
- Update amethyst's version of rendy

## Description

I use an Intel Ivybridge Desktop GPU, which has a `minImageCount` of `2` ([see here][gpu-surface]). Ever since updating to 0.12, I've been getting the same error every time I try to run my project.

```text
[WARN][rendy_wsi] Image count not supported. Supported: 4..4294967295, requested: 3
```

Is this why Amethyst warns about Ivybridge Vulkan support?

```text
INTEL-MESA: warning: Ivy Bridge Vulkan support is incomplete
```

EDIT: This also doesn't work on `master` and/or with nightly.
EDIT 2: See amethyst/rendy#182

## Reproduction Steps

1. Run the sphere example with Vulkan (the same thing happens)
```sh
cargo run --example sphere --features vulkan
```

## What You Expected to Happen

I expected the example to run and display a sphere.

## What Actually Happened

I got [an error][logs]

## Screenshots and Logs

* [Discord #help conversation][help]
* [Logs][logs]

## Your Environment

1. Operating System: Arch Linux
2. Operating System Version: rolling
3. Version of Amethyst: 0.12

## Additional Context

* GNOME on Wayland
* Mesa 19.1.6
* [GPU & Vulkan info][gpu]

[gpu]: https://vulkan.gpuinfo.org/displayreport.php?id=6579
[gpu-surface]: https://vulkan.gpuinfo.org/displayreport.php?id=6579#surface
[logs]: https://pastebin.com/EpLvybct
[help]: https://imgur.com/a/eggBhQY

ishitatsuyuki:
AFAIK Mesa is known to have a bug where it has a large minImageCount than what is necessary.

valkum:
But I think we checked with the Vulkan Hardware Capability Viewer as well. Why would Mesa return something different to the Capability Viewer than to us?

jakbyte:
Correct me if I'm wrong because I don't really know much about this, but I was thinking that the error was happening because:

* My GPU's `minImageCount` is 2
* Rendy's (specifically `rendy-wsi`'s) minimum is 4
* The supplied image count is 3

So I'm thinking my GPU would accept the image count, but Rendy's minimum causes a panic.

valkum:
Its I think:
- GPU's minImageCount really is 2
[...]

PR #2389, opened 2020-07-24
Correct Arch Linux dependencies installation

RsPlayer:
## Description

This is the correct pacman command to install needed dependencies, this installs only missing dependencies without reinstalling already existing ones.

## Modifications

- correct pacman installation

## PR Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,7 @@ If you are compiling on Linux, make sure to install the dependencies below.
 ```
 # pacman -S --needed grep gcc pkgconf openssl alsa-lib cmake make python3 freetype2 awk libxcb
 ```
+Also check https://wiki.archlinux.org/index.php/Vulkan for Vulkan installation
 
 ### Debian/Ubuntu
 
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ If you are compiling on Linux, make sure to install the dependencies below.
 ### Arch Linux
 
 ```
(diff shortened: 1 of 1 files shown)


RsPlayer:
I have also added Vulkan wiki link since its important to compile and run the Amethyst. 

The packages I need are `vulkan-icd-loader` `vulkan-intel` since I have Intel GPU `vulkan-headers` `vulkan-validation-layers` and `vulkan-tools`

RsPlayer:
Also Amethyst doesn't work on Wayland, I need to run it with  `WINIT_UNIX_BACKEND=x11` to make it work (issue #1846) 
Should I add it to README  and the book? the command is `WINIT_UNIX_BACKEND=x11 cargo run`

0x6273:
Not sure if removing `-yu` is a good idea, since many people will just copy-paste and it will fail due to the repo being out of sync.

RsPlayer:
> Not sure if removing `-yu` is a good idea, since many people will just copy-paste and it will fail due to the repo being out of sync.

`-y` is enough to sync without updating the whole system. So I think `-Sy` would be better.

0x6273:
That's even worse, since it can cause a partial upgrade. https://wiki.archlinux.org/index.php/System_maintenance#Avoid_certain_pacman_commands

RsPlayer:
I still believe `-S -needed` is the perfect one, but as you said some systems are out of sync which it won't fit for all users.

guess I'm closing this pull then.

PR #2395, opened 2020-07-31
Add `SpriteRender::new`

fosskers:
## Description

This PR adds a `new` associated function to `SpriteRender`, so that instantiating the type can be a little cleaner for users.

## Additions

- `SpriteRender::new`

## PR Checklist

- ~Added unit tests for new code added in this PR.~

### DIFF
--- a/amethyst_rendy/src/sprite/mod.rs
+++ b/amethyst_rendy/src/sprite/mod.rs
@@ -183,14 +183,24 @@ impl From<&TextureCoordinates> for [f32; 4] {
 
 /// Information for rendering a sprite.
 ///
-/// Instead of using a `Mesh` on a `DrawFlat` render pass, we can use a simpler set of shaders to
-/// render textures to quads. This struct carries the information necessary for the draw2dflat pass.
+/// Instead of using a `Mesh` on a `DrawFlat` render pass, we can use a simpler
+/// set of shaders to render textures to quads. This struct carries the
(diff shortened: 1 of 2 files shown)


fosskers:
Should I also update all instantiations of `SpriteRender` in the book? One example has:
```rust
         let sprite_render = SpriteRender {
             sprite_sheet: sprite_sheet_handle,
             sprite_number: 0, // First sprite
         };
```
Which has an extra comment for clarity. Should example like these be left in their verbose form?

salpalvv:
If it's all `pub`, and not encapsulated, does this even need a `new` or is that part of the plan later too? I'm in favor of a cleaner API without `pub` fields, to be clear.

fosskers:
My motivation was one of cleanliness. For instance, the 4 lines above become 1 with `new`:
```rust
let sprite_render = SpriteRender::new(sprite_sheet_handle, 0);
```
The effect of this really compounds when you're initializing many `SpriteRender`s at once.

There is also precedent for other types with `pub` fields but having a `new` too, `UiTransform` for instance.

fosskers:
The book hasn't been updated. Should I put up another PR?

CleanCut:
> The book hasn't been updated. Should I put up another PR?

Yes, please!

fosskers:
In that case, I'll change sections of the book where using `new` would make the code sample cleaner.

PR #2400, opened 2020-08-02
Update to latest legion and implement new dispatcher

chemicstry:
Sorry for a big one, but this due to changes to the way dispatcher and system bundles work.

There was a big discussion on discord about system data initialization/disposal (resources and/or entities). It was decided that systems should not contain any initialization logic and it should be solely left to system bundles. Systems are now simple `fn build() -> impl Runnable` functions instead of the horrible `fn build() -> Box<dyn FnOnce(&mut World, &mut Resources) -> Box<dyn Schedulable>>`. New legion codegen macros can be used as well. This is how `SystemBundle` looks now:

```rust
/// A SystemBundle is a structure that adds multiple systems to the [Dispatcher] and loads/unloads all required resources.
pub trait SystemBundle {
    /// This method is lazily evaluated when [Dispatcher] is built with [DispatcherBuilder::build].
    /// It is used to add systems or bundles (recursive) into [Dispatcher]. [World] and [Resources] are
    /// also provided to initialize any entities or resources used by the system.
    fn load(
        &mut self,
        world: &mut World,
        resources: &mut Resources,
        builder: &mut DispatcherBuilder,
    ) -> Result<(), Error>;

    /// This method is called once [Dispatcher] is disposed. It can be used to cleanup entities or resources from ECS.
    fn unload(&mut self, _world: &mut World, _resources: &mut Resources) -> Result<(), Error> {
        Ok(())
    }
}
```

This solves most of the problems that were present previously and also reduces code size considerably. This introduces some inconveniece for simple systems that use `shrev::EventChannel`, because a `SystemBundle` has to be implemented for it to register `ReaderId`. But I believe these ergonomic issues can later be solved with a macro how it was done in specs.

Dispatcher currently does not use `ArcThreadPool` as this is blocked on a fix in legion. It will be resolved soon.

Material example builds & works. Other already ported crates and examples will follow shortly.


chemicstry:
One thing I forgot to mention is the removal of stages and instead using the legion's way of scheduling. Systems are scheduled arbitrarily, but they respect data dependency order. For example if your system writes `Position` and rendering system reads `Position`, and you do `dispatcher_builder.add_system(my_system()).add_system(rendering_system())` they will never violate their order because rendering system has data dependencies on `Position`.

I think this greatly improves code readability as you can observe the full execution order by just looking at dispatcher builder. With stages there was no way to see system order without going into bundle definitions and looking at what stage they are executed.

Another thing is that you can easily reorder systems by just swapping `add_system` (or `add_bundle`) lines in dispatcher builder. With stages this was done with `RelativeStage` and you had to count priorities. Adding system in between meant rescaling all priority numbers to make space, which imo was tedious.

There is one regression though. System bundles now have no way to specify system insertion order. For example if there was `RenderingAndInputBundle` it would have no way to insert input system at the beginning and rendering system at the end. However, I didn't observe any such bundles in amethyst and worst case they can be split up into two.

Any objections?

chemicstry:
I believe this is ready for review now. Updated all ported crates and `material`, `fly_camera`, `arc_ball_camera` examples.

Everything (that is ported so far) compiles without warnings

amethyst_core/Cargo.toml
```diff
@@ -36,9 +34,15 @@ derivative = "2.1.1"
 
 thread_profiler = { version = "0.3", optional = true }
 
+[dependencies.legion]
+git = "https://github.com/TomGillen/legion"
```
khionu, line 38: We should pin a commit, for now, to prevent surprise breakages

khionu:
I don't have time to review this until the weekend, but because of the size, it should have at least two sets of eyes give the entire thing a review. These reviews don't need to guarantee correctness, but they should make an attempt to find any obvious issues.

amethyst_rendy/src/pass/base_3d.rs
```diff
@@ -217,10 +216,26 @@ impl<B: Backend, T: Base3DPassDef> RenderGroup<B, GraphAuxData> for DrawBase3D<B
                     Some((
                         entity,
                         (
-                            world.get_component::<Handle<Material>>(*entity)?,
-                            world.get_component::<Handle<Mesh>>(*entity)?,
-                            world.get_component::<LocalToWorld>(*entity)?,
-                            world.get_component::<Tint>(*entity),
+                            world
+                                .entry_ref(*entity)
+                                .unwrap()
```
jojolepro, line 221: are those unwraps safe?

amethyst_rendy/src/pass/flat2d.rs
```diff
@@ -322,9 +337,21 @@ impl<B: Backend> RenderGroup<B, GraphAuxData> for DrawFlat2DTransparent<B> {
...
+                            world
+                                .entry_ref(*entity)
+                                .unwrap()
+                                .into_component::<SpriteRender>()
+                                .ok()?,
+                            world
+                                .entry_ref(*entity)
+                                .unwrap()
+                                .into_component::<LocalToWorld>()
+                                .ok()?,
```
jojolepro, line 349: I think we should make this a function, seeing how much duplication there is. 
Maybe `get_component::<Type>(world, entity)` ?

amethyst_rendy/src/visibility.rs
```diff
@@ -152,7 +150,11 @@ pub fn build_visibility_sorting_system(
                         .filter(|(_, centroid, radius)| frustum.check_sphere(centroid, *radius))
                         .map(|(entity, centroid, _)| Internals {
                             entity,
-                            transparent: world.get_component::<Transparent>(entity).is_some(),
+                            transparent: world
+                                .entry_ref(entity)
+                                .unwrap()
```
jojolepro, line 155: the number of .entry_ref(entity).unwrap() is very impressive

amethyst_utils/src/removal.rs
```diff
@@ -68,18 +58,16 @@ where
...
 
 /// Removes all entities that have the `Removal<I>` component with the specified removal_id.
-pub fn exec_removal<I, D>(
-    commands: &mut CommandBuffer,
-    subworld: &mut SubWorld<'_>, 
-    removal_id: I,
-) where
+pub fn exec_removal<I, D>(commands: &mut CommandBuffer, subworld: &mut SubWorld<'_>, removal_id: I)
+where
     I: Debug + Clone + PartialEq + Send + Sync + 'static,
```
jojolepro, line 63: I know you didn't change this, but why are we forcing the use of Debug and Clone here? :thinking: 
(its usage in add_removal is fine however)

amethyst_window/src/bundle.rs
```diff
@@ -53,19 +50,34 @@ impl WindowBundle {
...
-            Stage::Begin,
-            build_window_system_from_config(&event_loop, self.config),
-        );
-        builder.add_thread_local_system(build_events_loop_system(event_loop));
+        let window = self
+            .config
+            .clone()
+            .into_window_builder(&events_loop)
+            .build(&events_loop)
+            .unwrap();
```
jojolepro, line 66: Should we use an expect() here instead?

amethyst_rendy/src/pass/base_3d.rs
chemicstry: Yes, because entities come from visibility system inside renderer which just ran before so they must exist

amethyst_rendy/src/pass/flat2d.rs
chemicstry: This code would be really simplified with the query API `<(&SpriteRender, &LocalToWorld, Option<&Tint>)>::query().get(world, entity)`, but the performance is currently quite poor comapred to `entry_ref`. Ayfid mentioned that it will be improved in the future so I'm not sure whether to use it or not.

Query is currently about 2 times slower so maybe not a big deal, since it's linear time.

amethyst_audio/src/systems/dj.rs
```diff
@@ -3,41 +3,79 @@ use log::error;
...
+    ) -> Result<(), Error> {
         init_output(resources);
-        SystemBuilder::<()>::new("DjSystem")
-            .read_resource::<AssetStorage<Source>>()
-            .read_resource::<Option<AudioSink>>()
-            .write_resource::<R>()
-            .build(move |_commands, _world, (storage, sink, res), _queries| {
-                #[cfg(feature = "profiler")]
-                profile_scope!("dj_system");
+        builder.add_system(build_dj_system(self.f.take().unwrap()));
```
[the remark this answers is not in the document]
chemicstry: > Wow, very impressive work!
> I have a few remarks, but nothing major.
> Also, I think all the dispatcher stuff should be moved into a separate crate. legion_dispatch maybe?
> Thanks a lot! ♥️

Thanks!

Dispatcher is just a single file so I don't know if it's worth moving it into a crate?

All the `EventChannel<Event>` unwraps should not fail if you use `Application`, because it is inserted before anything else. It is possible to change it to `get_mut_or_default`, but then it would silently wont work if you forgot to insert your custom input system. Maybe I could just add `expect` everywhere.

chemicstry:
> Wow, very impressive work!
> I have a few remarks, but nothing major.
> Also, I think all the dispatcher stuff should be moved into a separate crate. legion_dispatch maybe?
> Thanks a lot! ♥️

Thanks!

Dispatcher is just a single file so I don't know if it's worth moving it into a crate?

All the `EventChannel<Event>` unwraps should not fail if you use `Application`, because it is inserted before anything else. It is possible to change it to `get_mut_or_default`, but then it would silently wont work if you forgot to insert your custom input system. Maybe I could just add `expect` everywhere.

jojolepro:
> Dispatcher is just a single file so I don't know if it's worth moving it into a crate?

It is very convenient, so I'm sure other legion users would love it.
SystemBundle could even be placed in that new crate. :3

> All the EventChannel<Event> unwraps should not fail if you use Application [...]

Exactly, we can't know if the user will use application or not. We should handle the insertion for them in the case where they don't, instead of crash.

amethyst_utils/src/removal.rs
chemicstry: I guess it's due to trait bounds for `Removal<I>`, which is `Debug + Clone`. So by removing those you won't be able to construct `Removal<I>` for query.

chemicstry:
> Exactly, we can't know if the user will use application or not. We should handle the insertion for them in the case where they don't, instead of crash.

We cannot handle all the custom use cases. If user does not use `Application` then it's up to him to do much more than just inserting window event channel. I think that inserting default is much worse, because instead of crashing and telling you what is missing it would just silently continue and you would start pulling your hair out why no events are coming.

chemicstry:
> It is very convenient, so I'm sure other legion users would love it.
> SystemBundle could even be placed in that new crate. :3

What about moving it into `amethyst_dispatcher`? There is still the dependency of `amethyst_error` though

jojolepro:
Good points, let's go with how it is currently :+1:

chemicstry:
Ping @khionu

amethyst_core/src/lib.rs
```diff
@@ -8,23 +8,23 @@ compile_error!("the cfg flag \"no_threading\" is required when building for emsc
 
 pub use alga;
 pub use approx;
-pub use legion as ecs;
+pub use legion;
```
khionu, line 11: We re-export here (`amethyst_core::legion`) and below (`amethyst_core::ecs::legion`), this could cause confusion at best. We should only re-export once.

amethyst_core/src/system_ext.rs
```diff
@@ -1,140 +0,0 @@
-//! Extension system utilities.
```
khionu, line 1: Why was this file deleted? It appears to be related to functionality that we provided on top of Specs, does Legion provide it natively?

amethyst_core/src/deferred_dispatcher_operation.rs
```diff
@@ -1,178 +0,0 @@
-//! Provides the ability to store `Systems`, `Bundles`, `Barriers`, in a normal vector for deferred dispatcher construction.
```
khionu, line 1: Why is this functionality being deleted?

amethyst_rendy/src/lib.rs
```diff
@@ -45,8 +45,8 @@
 #![allow(clippy::new_without_default)]
 #![allow(unused_variables, dead_code)]
 
-#[macro_use]
-extern crate amethyst_derive;
+// #[macro_use]
```
khionu, line 48: Why is this commented out? If it compiles, it should just be removed.

amethyst_rendy/src/plugins.rs
```diff
@@ -3,14 +3,11 @@
 use crate::{
     bundle::{RenderOrder, RenderPlan, RenderPlugin, Target},
     pass::*,
-    sprite_visibility::build_sprite_visibility_sorting_system,
-    visibility::build_visibility_sorting_system,
+    sprite_visibility::{build_sprite_visibility_sorting_system, SpriteVisibility},
```
khionu, line 6: These two new imports needed? Doesn't look like usage for them was added

amethyst_rendy/src/submodules/vertex.rs
```diff
@@ -159,7 +159,7 @@ impl<B: Backend, V: VertexDataBufferType, T: 'static> DynamicVertexData<B, V, T>
                 let tmp = std::mem::replace(&mut slice, &mut []);
                 let (dst_slice, rest) = tmp.split_at_mut(data_slice.len());
                 dst_slice.copy_from_slice(data_slice);
-                std::mem::replace(&mut slice, rest);
+                let _ = std::mem::replace(&mut slice, rest);
```
khionu, line 162: From the must_use attribute: "if you don't need the old value, you can just assign the new value directly". `slice = rest` should work here.

amethyst_rendy/src/system.rs
```diff
@@ -22,17 +17,20 @@ use rendy::{
...
 /// Auxiliary data for render graph.
+#[allow(missing_debug_implementations)]
 pub struct InternalGraphAuxData<'a> {
+    /// World
     pub world: &'a World,
+    /// Resources
     pub resources: &'a Resources,
 }
 
-// FIXME: It is currently impossible to pass types with lifetimes (except for a single reference)
```
khionu, line 30: This isn't documentation, this is a note of something to fix.

amethyst_window/src/bundle.rs
```diff
@@ -53,19 +50,34 @@ impl WindowBundle {
...
+            .expect("Unable to create window");
+
+        let hidpi = window.get_hidpi_factor();
+        let (width, height) = window
+            .get_inner_size()
+            .expect("Window closed during initialization!")
+            .to_physical(hidpi)
+            .into();
+
+        resources.insert(ScreenDimensions::new(width, height, hidpi));
```
khionu, line 75: A window can be dragged onto another monitor with different dimensions. Is this handled elsewhere?

amethyst_core/src/deferred_dispatcher_operation.rs
chemicstry: This was only needed for specs to lazily build systems. Legion system builds don't have any dependencies on `World` or `Resources` and can be initialized right away. Similar functionality is now provided by system bundles.

amethyst_rendy/src/lib.rs
chemicstry: This was used for prefabs which are scattered throughout the code and are commented out for the time being. Although `#[macro_use]` was deprecated since rust 1.30 so I will just remove this.

amethyst_rendy/src/plugins.rs
chemicstry: Yes, because now these resources are inserted by the rendering system bundles instead of the system itself.

amethyst_window/src/bundle.rs
chemicstry: Nothing has changed here functionality wise from specs, only resource initialization was shuffled around. `ScreenDimensions` are updated from the `build_window_system`.

chemicstry:
CI wouldn't work on legion branch yet because there is still a lot of thinngs to do: update all remaining crates, examples, inline examples, book.

amethyst_core/src/system_ext.rs
chemicstry: Ah yes, I forgot about this. I'm not sure how hard would it be to implement pausable systems in legion, but definitely not as simple. I will add a todo for this to the legion project.

jojolepro:
Yeah definitely don't merge master into legion before most of the legion port is completed.

chemicstry:
Merging master is not necessarily bad and is probably easier now since not everything is ported yet.

I just don't want to do that in this PR as it is already big.

amethyst_window/src/bundle.rs
khionu: Then this might be a bug, but not in the scope of this PR

amethyst_core/src/system_ext.rs
khionu: Once that ToDo is added, this can be resolved.

khionu: Eh, actually, not going to block the PR on it

khionu:
Even if it's not going to work, CI on the Legion branch should happen soon, so we can have some indicator of progress. CI on the Legion branch needs to pass before it's merged.

amethyst_core/src/system_ext.rs
chemicstry: Added it already 👍

PR #2406, opened 2020-08-07
Tcp net fix

Machine-Hum:
## Description

Patched the TCP lib. Previously all the TCP packets would clump together. This places header in the pack, which is then extracted on the other side. The packets are the delivered one by one.

## Additions

## Removals

## Modifications

## PR Checklist

- [x ] Updated the content of the book if this PR would make the book outdated.

- [x ] Ran `cargo +stable fmt --all`
- [x ] Ran `cargo clippy --all --features "empty"` (may require `cargo clean` before)
- [x ] Ran `cargo build --features "empty"`
- [x ] Ran `cargo test --all --features "empty"`

### DIFF
--- a/amethyst_network/src/simulation/transport/tcp.rs
+++ b/amethyst_network/src/simulation/transport/tcp.rs
@@ -25,6 +25,10 @@ use std::{
     ops::DerefMut,
 };
 
+use log::{info};
+
+use std::convert::TryInto;
+
 const CONNECTION_LISTENER_SYSTEM_NAME: &str = "connection_listener";
 const STREAM_MANAGEMENT_SYSTEM_NAME: &str = "stream_management";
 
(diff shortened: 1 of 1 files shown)


amethyst_network/src/simulation/transport/tcp.rs
```diff
@@ -242,32 +253,51 @@ impl<'s> System<'s> for TcpNetworkRecvSystem {
...
-                            _ => {
-                                event_channel.single_write(NetworkSimulationEvent::RecvError(e));
-                            }
+                    break;
+                }
+            }
+
+            // Then grab this amount of data.
+            let mut data = vec![0; pack_size];
+            match stream.read(&mut data) {
```
chemicstry, line 283: What if less than `pack_size` data is actually read? `tcp_pack_size` is not checked. You would have to keep some kind of recv buffer to reassemble incomplete packets between frames

amethyst_network/src/simulation/transport/tcp.rs
```diff
@@ -213,7 +215,10 @@ fn write_message(
     channel: &mut EventChannel<NetworkSimulationEvent>,
 ) {
     if let Some((_, stream)) = net.get_stream(message.destination) {
-        if let Err(e) = stream.write(&message.payload) {
+        let len: Bytes = Bytes::copy_from_slice(&(message.payload.len() as u32).to_le_bytes());
```
chemicstry, line 218: Doesn't `u32` introduce too much overhead for small messages? Maybe not. But to solve this and the other issue I mentioned I think it would be better to use some existing encoding libraries instead of inventing stuff. Although, I don't have any experience with this to suggest something.

chemicstry:
You could use [bincode](https://github.com/servo/bincode), but since `amethyst_network` can only send/recv byte slices it is kinda overkill. And that doesn't solve receive buffer fragmentation either.

So if you could just fix the recv buffer issue I mentioned I think this would be good to go. Thanks for the PR btw!

amethyst_network/src/simulation/transport/tcp.rs
Machine-Hum: Did not think of this, would it be bad to block on the stream.read() until all the data is there?

If so I can implement a buffer.

chemicstry: Yes blocking would be bad because any network lag would freeze the game loop so a buffer is needed.

jojolepro:
Closing due to lack of activity. Feel free to reopen it!

PR #2411, opened 2020-08-10
Remove Workflow Name

CleanCut:
No working bors examples have workflow names, so remove it in hopes of fixing bors.


CleanCut:
Actually, I _think_ this is not needed.

PR #2412, opened 2020-08-11
Enable CI on Forked PRs

CleanCut:
Does it work? Let's see...


CleanCut:
Woot! 🎉 Don't need to do anything. Closing.

PR #2414, opened 2020-08-11
Update stalebot

CleanCut:
The most important change here is making it so that the config uses modern setting names so stalebot stops closing stale issues.  The rest is from the [current default config](https://github.com/probot/stale), with very slight adjustments.


.github/stale.yml
```diff
@@ -1,13 +1,56 @@
...
+# pulls:
+#   daysUntilStale: 30
+#   markComment: >
+#     This pull request has been automatically marked as stale because it has not had
+#     recent activity. It will be closed if no further activity occurs. Thank you
+#     for your contributions.
+
+# issues:
+#   exemptLabels:
+#     - confirmed
```
jojolepro, line 56: you could add a newline at the end of the file :3

CleanCut: ✅ I can indeed!

PR #2417, opened 2020-08-11
Merge master into legion

chemicstry:
Merged master branch into legion to prevent it going too far out of sync.

As a bonus CI should be working now! Except that there is still a lot to do to pass it...

There are no changes here, just minor merge conflict fixups.


khionu:
My only concern is why this has so many additions over so few deletions?

chemicstry:
Added remaining commits to bring it up to `0.15.1`

PR #2429, opened 2020-08-15
Update pong-tutorial-02.md

wanderrful:
## Description

Just a minor touch-up to an import to make it a little bit easier to remember. 

Later on in the tutorial, we reference System with just `amethyst::ecs` (without the prelude). So if we're consistent about it here as well then new people who don't know any better won't get caught up on why one needs prelude and the other doesn't.

## Additions

None

## Removals

Removed `prelude` from an import

## Modifications

Just got rid of the `prelude`, that's all.

## PR Checklist

- [x] Ran `cargo clippy --all --features "empty"` (may require `cargo clean` before)

### DIFF
--- a/book/src/pong-tutorial/pong-tutorial-02.md
+++ b/book/src/pong-tutorial/pong-tutorial-02.md
@@ -423,7 +423,7 @@ First, let's declare the function and load the sprite sheet's image data.
 # use amethyst::{
 #     assets::{AssetStorage, Loader, Handle},
 #     core::transform::Transform,
-#     ecs::prelude::{Component, DenseVecStorage},
+#     ecs::{Component, DenseVecStorage},
 #     prelude::*,
 #     renderer::{
 #         camera::Camera,
@@ -511,7 +511,7 @@ Finally, we load the file containing the position of each sprite on the sheet.
 # use amethyst::{
 #     assets::{AssetStorage, Handle, Loader},
 #     core::transform::Transform,
-#     ecs::prelude::{Component, DenseVecStorage},
+#     ecs::{Component, DenseVecStorage},
 #     prelude::*,
 #     renderer::{
 #         camera::Camera,
--- a/book/src/pong-tutorial/pong-tutorial-02.md
+++ b/book/src/pong-tutorial/pong-tutorial-02.md
@@ -31,7 +31,7 @@ initialization code from the Pong code.
     use amethyst::{
         assets::{AssetStorage, Loader, Handle},
         core::transform::Transform,
-        ecs::prelude::{Component, DenseVecStorage},
+        ecs::{Component, DenseVecStorage},
         prelude::*,
         renderer::{Camera, ImageFormat, SpriteRender, SpriteSheet, SpriteSheetFormat, Texture},
     };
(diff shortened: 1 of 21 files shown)


Jaden-Giordano:
Seems like this is also in https://github.com/amethyst/amethyst/blob/master/book/src/pong-tutorial/pong-tutorial-04.md you wanna change that in this PR too?

wanderrful:
Thanks for noticing - sure, will do.

wanderrful:
I went through the tutorial and did it a bit more methodically -- I removed the prelude from `ecs::prelude` for everything but the star imports.

CleanCut:
Don't forget to make sure to make the same changes to `examples/pong_tutorial_*/...`!

wanderrful:
In most other programming languages it's a bad practice to use star imports.

While adjusting the `pong_tutorial_06/pong.rs` file I noticed this redundancy, for example:
image

World comes from `amethyst::prelude::*` but we're also grabbing it again from `amethyst::ecs::prelude::World` in the same file, which is redundant.

I'll just leave the star imports because it looks like you guys prefer that.

wanderrful:
While I'm in here, I also noticed that your documentation says to create SpriteRender with `{ }` like so:
image

... but in the actual files themselves you guys are using `SpriteRender::new()`, which isn't defined (at least in the latest version, 0.15.0). Maybe that's just not released yet?

CleanCut:
> World comes from `amethyst::prelude::*` but we're also grabbing it again from `amethyst::ecs::prelude::World` in the same file, which is redundant.

> I'll just leave the star imports because it looks like you guys prefer that.

I would normally state that according [docs/CODING_CONVENTIONS.md](https://github.com/amethyst/amethyst/blob/master/docs/CODING_CONVENTIONS.md), we follow the [Rust API Guidelines](https://rust-lang-nursery.github.io/api-guidelines/about.html) -- but wildcard imports are not addressed there!

Happily, we also conform to `clippy`'s lints, and clippy has a lint called [`wildcard_imports`](https://rust-lang.github.io/rust-clippy/master/index.html#wildcard_imports). TL;DR - Never use wildcard imports, _except_ on modules named `prelude`.

Further, I would suggest that in our examples that we avoid using prelude _at all_, whether with wildcard or not.  Using prelude is _convenient_ for quick-n-dirty projects, but I think the priority in examples and tutorials should be _clarity_, not _convenience_.

As an aside, I also think that IDE plugins (ahem, intellij-rust) ought to avoid auto-importing from `prelude` when it can find the item through another path.

> While I'm in here, I also noticed that your documentation says to create SpriteRender with `{ }` like so:
> [snip]
> ... but in the actual files themselves you guys are using `SpriteRender::new()`, which isn't defined (at least in the latest version, 0.15.0). Maybe that's just not released yet?

Yep, just not released yet (I'm working on it this weekend!). Go ahead and change it to `SpriteRender::new()` everywhere.

wanderrful:
Went through and double checked, and it looks like you guys already updated the `tutorial.md` files to use `SpriteRender::new`, it just wasn't deployed to the live version of the gitbook yet.

CleanCut:
No changelog entry required, though this could be included as a link to any generic pong, example, or tutorial update entry.

bors r+

PR #2435, opened 2020-08-16
replace cargo --all with cargo --workspace

Blisto91:
## Description

``--all`` is deprecated to avoid confusion with ``--all-targets`` or ``--all-features``.
We should use ``--workspace`` instead

## PR Checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -27,6 +27,6 @@ By placing an x in the boxes I certify that I have:
 If this modified or created any rs files:
 
 - [ ] Ran `cargo +stable fmt --all`
-- [ ] Ran `cargo clippy --all --features "empty"` (may require `cargo clean` before)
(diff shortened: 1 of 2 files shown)


Blisto91:
Btw i noticed in CONTRIBUTING.md under Submission Checklist that cargo run is an extra step that isn't in pull_request_template.md. Dunno if they should be in sync.

Also isn't it redundant to run cargo build before cargo test since the latter also compiles the project?

CleanCut:
> Btw i noticed in CONTRIBUTING.md under Submission Checklist that cargo run is an extra step that isn't in pull_request_template.md. Dunno if they should be in sync.

Yes, they should be in sync. Though I don't think we necessarily need any of the steps that are also enforced by CI, now.  As I [stated here](https://github.com/amethyst/amethyst/pull/2443#issuecomment-677051908) I think we should be moving as much CI functionality as we can into independent scripts and document how people can run them locally if they want a quicker dev-cycle.

PR #2436, opened 2020-08-16
Bring back the old transform system in place of legion_transform

chemicstry:
As per discord discussion, this brings back the old transform system in place of `legion_transform`. The `legion_transform` did not suit our needs as discussed [here](https://community.amethyst.rs/t/legion-transform-design-discussion/1582/26).

Keeping the old transform system also reduces the amount of porting needed for other crates (especially animation). This was already reviewed by @jojolepro [in my fork](https://github.com/chemicstry/amethyst/pull/1), but this resolves the latest conflicts with the master merge.

Parent system borrowed from https://github.com/amethyst/legion_transform

The old `Transform` component is intact, except for added `parent_matrix` field to simplify legion system implementation. Transform system itself is completely different, because `specs` implementation used `specs_hierarchy` with events. The lack of events inevitably introduces delay for deep hierarchies, but I believe we can live with this until a new transform system from forum discussions is implemented.

Again sorry for a large PR, but transform affects the whole engine. I went through amethyst master code and ensured that all `Transform` usage is the same and is not affected by the previous `LocalToWorld` implementation.


amethyst_core/src/transform/components/transform.rs
```diff
@@ -0,0 +1,788 @@
+//! Local transform component.
```
chemicstry, line 1: This file is exactly the same as from master, but has an added `parent_matrix` field.

PR #2443, opened 2020-08-19
Simplify ci config a bit

Veetaha:
## Description

Simplified the config by removing actions-rs/cargo@v1
See the [uses cases](https://github.com/actions-rs/cargo#use-cases) described in its repo.
- We don't use cross, do we?
- Github annotations from ci failures are generally not that useful, most of the time we just open the job logs with red crosses near them and that's all.
Instead, we get less dependencies for our CI and less code

## Additions

## Removals

## Modifications

## PR Checklist

- [ ] Ran `cargo clippy --all --features "empty"` (may require `cargo clean` before)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,33 +51,23 @@ jobs:
           # Don't use a 'components:' entry--we don't need them with beta/nightly, plus nightly often doesn't have them
           override: true
 
-      - name: install rustfmt+clippy on stable
-        run: |
-          rustup component add rustfmt
-          rustup component add clippy
+      - run: rustup component add rustfmt clippy
         if: contains(matrix.toolchain, 'stable')
 
-      - name: run rustfmt on stable
-        uses: actions-rs/cargo@v1
-        with:
-          command: fmt
-          args: --all -- --check
+      - run: cargo fmt
         if: contains(matrix.toolchain, 'stable')
 
-      - name: run clippy on stable
-        uses: actions-rs/cargo@v1
-        with:
-          command: clippy
-          args: --features=${{matrix.FEATURES}}
+      - run: cargo clippy --features=${{matrix.FEATURES}}
         if: contains(matrix.toolchain, 'stable')
 
       - name: clean clippy-generated amethyst libs
+        # Remove the clippy-generated amethyst files.
+        # They mess up `mdbook test` later on for some reason
         run: rm -rf ./target/debug/deps/libamethyst*
         if: contains(matrix.os, 'macos') || contains(matrix.os, 'linux')
 
       - name: run tests
         run: |
-          # Remove the clippy-generated amethyst files, they mess up `mdbook test` later on for some reason
           cargo test --workspace --features=${{matrix.FEATURES}}
         env:
           MACOS: ${{ matrix.MACOS }}  # Used by some tests


Veetaha:
@CleanCut I'd recommend scaling down the PR template, at least the part about `Ran X`.
The sentences about `clippy, build, test` are not that necessary, because CI will catch these nevertheless.
For `cargo fmt` I would suggest creating a git-pre-commit hook, for example in our repos we generally have a `meta` binary crate called `xtask` for dev things like pre-commit hooks that run `cargo fmt` automatically:
https://github.com/elastio/devx/blob/master/xtask/src/main.rs

.github/workflows/ci.yml
```diff
@@ -51,33 +51,23 @@ jobs:
...
-          rustup component add clippy
+      - run: rustup component add rustfmt clippy
         if: contains(matrix.toolchain, 'stable')
 
-      - name: run rustfmt on stable
-        uses: actions-rs/cargo@v1
-        with:
-          command: fmt
-          args: --all -- --check
+      - run: cargo fmt
```
CleanCut, line 57: `--all` is a now-deprecated synonym for `--workspace`.  Without one of these this command will only run on only the `amethyst` crate and none of the other crates in the workspace.  

`-- --check` makes the command fail if it has to format something.

```suggestion
      - run: cargo fmt --workspace -- --check
```

.github/workflows/ci.yml
```diff
@@ -51,33 +51,23 @@ jobs:
...
         if: contains(matrix.toolchain, 'stable')
 
-      - name: run clippy on stable
-        uses: actions-rs/cargo@v1
-        with:
-          command: clippy
-          args: --features=${{matrix.FEATURES}}
+      - run: cargo clippy --features=${{matrix.FEATURES}}
```
CleanCut, line 60: Aha! You've discovered another bug. We were only running clippy on the root crate.  We can't make it fail yet, though, because we haven't taken care of the warnings.
```suggestion
      - run: cargo clippy --features=${{matrix.FEATURES}} --workspace
```

CleanCut:
> @CleanCut I'd recommend scaling down the PR template, at least the part about `Ran X`.

@Veetaha I agree. I added it to the list.  In addition, in the future I would like as much of each CI step as possible to be encapsulated in standalone scripts, and then link from the PR template to instructions about how to use them locally to have a faster feedback loop (that could include scripts for installing pre-commit hooks, etc.).  So rather than having `cargo fmt --workspace -- --check` in CI, we'd have something like `script/ci-fmt` in CI, and you could also run that locally to get the same effect.

.github/workflows/ci.yml
Blisto91: Cargo fmt actually never made the switch to --workspace and has to use --all.
Can't remember the exact reasoning except they don't have --all-targets etc. to confuse it with. 

There is an old issue talking about it on their repo.

Veetaha: Yup, `--workspace` argument isn't valid with `cargo fmt`, added `--all -- --check`. 
6fcda3a

.github/workflows/ci.yml
Veetaha: Yup, added `--workspace` in 6fcda3a

Veetaha: Surprisingly, with clippy hasn't detected any issues for the rest of the codebase. I guess that `Ran cargo clippy` chebox in PR template did the thing?

Veetaha:
Btw, @CleanCut why not running `cargo fmt --check` only once on `stable` toochain and `ubuntu-latest` os.
I see that we already have `'macos'` selected for doing some additional work which is platform-independent (e.g. mdbook checks), but from my experience 'ubuntu-latest' runners are the fastest ones among other Github-hosted runners

.github/workflows/ci.yml
CleanCut: Yikes! Sorry for the misinformation.

.github/workflows/ci.yml
CleanCut: I won't complain if the clippy issues are all fixed! If that is the case, let's make clippy fail if it finds anything.

CleanCut:
@Veetaha I would be fine running `fmt` and `clippy` only on a single stable OS.  I'm also fine if we want to experiment with whichever OS's are the fastest to reduce overall time.  Since the time I was experimenting with things, GitHub appears to have switched it's macOS runners to some setup that runs significantly slower (it was faster than Linux at the time I chose macOS for the extra stuff).  For clarity, let's do that in a separate PR from this one.

Veetaha:
Sure, though clipply is quite caveat-y. I guess if there are some `cfg(windows)` or other stuff going on it won't check the windows-specific code if we run it on Linux only?

CleanCut:
> Sure, though clipply is quite caveat-y. I guess if there are some `cfg(windows)` or other stuff going on it won't check the windows-specific code if we run it on Linux only?

🤷‍♂️  If clippy gives different results on different platforms, then I withdraw it from consideration. 😄

.github/workflows/ci.yml
Veetaha: @CleanCut 
Well actually clippy does detect issues, but only at warning level and it doesn't fail ci, see [this line in checks log](https://github.com/amethyst/amethyst/pull/2443/checks?check_run_id=1006732344#step:7:197)

Would you merge this if I add these to to global ci env vars we have in our project?
```
  RUSTFLAGS: "--deny warnings -Cdebuginfo=0" // Disabling debug info should speed up ci?
  RUSTDOCFLAGS: "--deny warnings"
```

Blisto91:
I think @Veetaha is correct here. Clippy has to compile the code it is working with so platform (or feature) dependent code might not get lints unless run on that platform.

Fmt works on the raw source files and should be able to check everything no matter the OS. As far as I'm aware.

Issue #2440, opened 2020-08-18
[BUG] `screen_ray` is broken in 0.15.1


bonsairobo:
## Description

After upgrading to 0.15.1, all parts of my app that use the `Camera::screen_ray` function are broken. For example, 3D picking no longer lines up with the cursor.

## Reproduction Steps

I don't think we have any examples to test this, but you can run my project: https://github.com/amethyst/voxel-mapper

## What You Expected to Happen

A pink cube could appear on whichever voxel the cursor is hovering over.

## What Actually Happened

The pink cube does not align with the cursor.

## Your Environment

1. Windows 10
2. version 2004
3. Amethyst v0.15.1

## Additional Context

I am most suspicious of this PR: https://github.com/amethyst/amethyst/pull/2118

CleanCut:
I confirmed the same behavior on macOS after making the following patch:

```diff
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ name = "voxel_mapper"
 path = "src/lib.rs"

 [dependencies]
-amethyst = "0.15"
+amethyst = { version = "0.15", features = ["metal"] }
 bincode = "1.3.1"
 indexmap = "1.5"
 itertools = "0.9.0"
```

/cc @azriel91 @joonazan would #2118 be involved in this regression?

bonsairobo:
I ran my app on both versions and recorded some debug values for nearly identical test cases.

On 0.15.0:

```
Ray = Ray {
    origin: Point { coords: Matrix { data: [0.022067614, 20.89971, 0.07834254] } },
    direction: Matrix { data: [0.2112339, -0.95520276, -0.20728685] }
}

Cursor = Point { coords: Matrix { data: [972.0, 289.0] } }

Diag = Matrix { data: [1600.0, 900.0] }

Projection = Perspective(Perspective {
    matrix: Matrix { data: [
        0.9742786, 0.0, 0.0, 0.0,
        0.0, -1.7320509, 0.0, 0.0,
        0.0, 0.0, -1.00005, -1.0,
        0.0, 0.0, -0.100005, 0.0
    ] },
    inverse_matrix: Matrix { data: [
        1.0264004, 0.0, 0.0, 0.0,
        0.0, -0.57735026, 0.0, 0.0,
        0.0, 0.0, 0.0, -9.9995,
        0.0, 0.0, -1.0, 9.999999
    ] }
})

Transform = Transform {
    isometry: Isometry {
[...]

PR #2447, opened 2020-08-20
Fix the Camera::screen_ray method [#2440]

bonsairobo:
## Description

Add a regression test of Camera::screen_ray, as well as the fix to make it pass. This test was passing on 0.15.0 and broke on 0.15.1.

## Modifications

- Use z coordinate 0.5 instead of 0.0 for the "far" point of the ray before it gets transformed by the inverse perspective projection

## PR Checklist

- [x] Ran `cargo clippy --all --features "empty"` (may require `cargo clean` before)

### DIFF
--- a/amethyst_rendy/src/camera.rs
+++ b/amethyst_rendy/src/camera.rs
@@ -181,7 +181,7 @@ impl Camera {
         let matrix = *camera_transform.global_matrix() * self.inverse;
 
         let near = Point3::new(screen_x, screen_y, 1.0);
-        let far = Point3::new(screen_x, screen_y, 0.0);
+        let far = Point3::new(screen_x, screen_y, 0.5);
 
         let near_t = matrix.transform_point(&near);
         let far_t = matrix.transform_point(&far);
@@ -670,5 +670,37 @@ mod tests {
         let far = Point3::new(0.0, 0.0, -97.0);
         let projected_point = mvp.transform_point(&far);
         assert_abs_diff_eq!(projected_point[2], 0.0);
+    }
+
+    #[test]
+    fn screen_ray_3d() {
+        let width = 1280.0;
+        let height = 720.0;
+
+        let aspect = width / height;
+        let fov = std::f32::consts::FRAC_PI_3;
+        let znear = 0.125;
+        let camera = Camera::perspective(aspect, fov, znear);
+
+        let mut camera_transform: Transform = Transform::new(
+            Translation3::new(0.0, 0.0, 3.0),
+            UnitQuaternion::identity(),
+            [1.0, 1.0, 1.0].into(),
+        );
+        camera_transform.copy_local_to_global();
+
+        let cursor_pos = Point2::new(width, height);
+        let screen_diag = Vector2::new(width, height);
+        let ray = camera.screen_ray(cursor_pos, screen_diag, &camera_transform);
+
+        let expected_ray = Ray {
+            /// In the znear plane.
+            origin: Point3::new(0.12830007, -0.07216879, 2.875),
+            // Corresponds to 45 degree rotation to the right and 30 degree rotation up. This ray
+            // should pass through the top right corner of the screen.
+            direction: Vector3::new(0.66436374, -0.37370458, -0.6472757),
+        };
+        assert_ulps_eq!(ray.origin, expected_ray.origin);
+        assert_ulps_eq!(ray.direction, expected_ray.direction);
     }
 }


bonsairobo:
Hmm maybe it failed in CI because of floating point differences between CPUs?

Blisto91:
@bonsairobo runs fine on my work laptop. hmm.

bonsairobo:
Honestly I don't understand how ulps works so I'm doing some light reading:

https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html

Blisto91:
It doesn't seem to random because the numbers seems to be be same on each runner no matter if it's linux, Mac with stable, beta, or nightly.
All the numbers I've seen are consistent between runs. 

Maybe windows handles the calculations or methods a bit different

bonsairobo:
@Blisto91 Yea I'm running it on windows, so it could be that.

Blisto91:
👍
If you figure any details out then please share 🙂

Interested in finding out where it happens exactly.

CleanCut:
@bonsairobo After reading a good bit of literature describing what _not_ to do, I hit upon this excellent [blog post](https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/) that also describes _what to do_.  Then I searched around and found the [float-cmp](https://github.com/mikedilger/float-cmp) library which utilizes the method from the article.

My recommendation is to add [float-cmp](https://github.com/mikedilger/float-cmp) as a `dev-dependency` and then use `approx_eq!`.
---
Unrelated question: I understand why we can't use `0.0` -- but it isn't clear to me why the number to use should be `0.5` -- is there a specific reason, or is it an arbitrary choice?  Is there any difference at all between that and `0.1` or `0.9` or any other value?  I'm wondering what tradeoffs are.

CleanCut:
Resolves #2440

CleanCut:
@bonsairobo Rather than put the issue # in the title or even (note to self) comment that it "resolves #xyz", we can make it an actual linked issue: 
image

That has a bunch of benefits, like auto-closing the linked issue, showing up in the UI in various places, etc.

amethyst_rendy/src/camera.rs
```diff
@@ -671,4 +671,36 @@ mod tests {
...
+        let ray = camera.screen_ray(cursor_pos, screen_diag, &camera_transform);
+
+        let expected_ray = Ray {
+            /// In the znear plane.
+            origin: Point3::new(0.12830, -0.07216, 2.875),
+            // Corresponds to 45 degree rotation to the right and 30 degree rotation up. This ray
+            // should pass through the top right corner of the screen.
+            direction: Vector3::new(0.66436, -0.37370, -0.64727),
+        };
+        assert_abs_diff_eq!(ray.origin, expected_ray.origin, epsilon = 0.00001);
```
CleanCut, line 703:
```suggestion
        assert_ulps_eq!(ray.origin, expected_ray.origin, max_ulps = 1);
```

amethyst_rendy/src/camera.rs
```diff
@@ -671,4 +671,36 @@ mod tests {
...
+        assert_abs_diff_eq!(ray.direction, expected_ray.direction, epsilon = 0.00001);
```
CleanCut, line 704:
```suggestion
        assert_ulps_eq!(ray.direction, expected_ray.direction, max_ulps = 1);

```

amethyst_rendy/src/camera.rs
```diff
@@ -181,7 +181,7 @@ impl Camera {
         let matrix = *camera_transform.global_matrix() * self.inverse;
 
         let near = Point3::new(screen_x, screen_y, 1.0);
-        let far = Point3::new(screen_x, screen_y, 0.0);
+        let far = Point3::new(screen_x, screen_y, 0.5);
```
CleanCut, line 184:
```suggestion
        // The constraint on far is: 0.0 < far < 1.0. We arbitrarily chose 0.5 - maybe there is a better value?
        let far = Point3::new(screen_x, screen_y, 0.5);
```

amethyst_rendy/src/camera.rs
```diff
@@ -671,4 +672,37 @@ mod tests {
...
+        let ray = camera.screen_ray(cursor_pos, screen_diag, &camera_transform);
+
+        let expected_ray = Ray {
+            /// In the znear plane.
+            origin: Point3::new(0.12830, -0.07216, 2.875),
+            // Corresponds to 45 degree rotation to the right and 30 degree rotation up. This ray
+            // should pass through the top right corner of the screen.
+            direction: Vector3::new(0.66436, -0.37370, -0.64727),
+        };
+        assert_ulps_eq!(ray.origin, expected_ray.origin, max_ulps = 1);
```
CleanCut, line 704: 🤦‍♂️ Okay, that was _completely_ my fault.  I should have checked to see if the asserts worked with `Vector3` ... they do not.

Even with this newest suggestion the expected ray is failing on my mac because it is missing so much precision that the value is quite a ways off (in terms of steps on the floating point scale).  Where did you get the literal values from?

Here are the values on my mac:
- `ray.origin.coords` - `Matrix { data: [0.12830007, -0.0721688, 2.875] }`
- `ray.direction` - `Matrix { data: [0.6643638, -0.37370467, -0.6472754] }`

```suggestion
        assert_ulps_eq!(ray.origin.coords.x, expected_ray.origin.coords.x, max_ulps = 2);
        assert_ulps_eq!(ray.origin.coords.y, expected_ray.origin.coords.y, max_ulps = 2);
        assert_ulps_eq!(ray.origin.coords.z, expected_ray.origin.coords.z, max_ulps = 2);
```

amethyst_rendy/src/camera.rs
```diff
@@ -671,4 +672,37 @@ mod tests {
...
+        assert_ulps_eq!(ray.direction, expected_ray.direction, max_ulps = 1);
```
CleanCut, line 705:
```suggestion
        assert_ulps_eq!(ray.direction.x, expected_ray.direction.x, max_ulps = 2);
        assert_ulps_eq!(ray.direction.y, expected_ray.direction.y, max_ulps = 2);
        assert_ulps_eq!(ray.direction.z, expected_ray.direction.z, max_ulps = 2);
```

amethyst_rendy/src/camera.rs
bonsairobo: I'm pretty sure `assert_ulps_eq` *does* work with `Vector3`. The issue here is just that the literal values are not very precise. That's why I was using epsilon before.

bonsairobo:
@CleanCut Maybe you can try this on your macbook and see how many ulps are required:

```
        let expected_ray = Ray {
            /// In the znear plane.
            origin: Point3::new(0.12830006, -0.07216878, 2.875),
            // Corresponds to 45 degree rotation to the right and 30 degree rotation up. This ray
            // should pass through the top right corner of the screen.
            direction: Vector3::new(0.66436386, -0.37370467, -0.6472755),
        };
        assert_ulps_eq!(ray.origin, expected_ray.origin, max_ulps = 1);
        assert_ulps_eq!(ray.direction, expected_ray.direction, max_ulps = 1);
```

This works on my Windows machine.

bonsairobo:
@CleanCut I'll try the values you sent me on my Windows machine and figure out how many ulps are required.

bonsairobo:
OK the values you sent me actually worked without an explicit number of ulps (I assume the default is 1). Maybe give that a try in the CI.

PR #2467, opened 2020-08-28
Migrate SystemExt pauseable to Legion

thunderseethe:
## Description

Migrate the Pauseable system wrapper from specs to legion.

I was unable to keep the SystemExt trait for the pauseable method. Legion's `#[system]` macro produces a function that returns an opaque `impl Runnable`. While this is technically fine, Pauseable only relies on Runnables interface, opaque types do not recognize extra trait impls (or I do not know how to make them recognized). Instead of a trait I've exposed a bare method `pauseable` that takes an `impl Runnable` as input and returns a `Pauseable` struct. This struct then also impls Runnable and behaves as expected.

I couldn't find any unit tests for this code in master, I'm assuming the doc acts as a kind of example/unit test combo? While updating the docs I wrote the code snippet as a unit test and confirmed behaves as expected. This can be found embedded in the docs for `pauseable`, I'd be happy to re add it as an explicit unit test if desired.

I ran the given commands but `build` and `test` failed with unrelated errors. I'm assuming that's expected given this branch is mid-migration. Please let me know if that's not the case and I can look harder at resolving things. I included some trivial fixes for what appeared to be code not yet migrated to legion 0.3 under transform/ this was mainly replacing `Universe::new().create_world()` with `World::default()`

## Additions

- amethyst::core::system_ext::pauseable

## Removals

- amethyst::core::system_exit::SystemExt (this was already gone on this branch but I was unable to re-add it)

## PR Checklist

- [X] Added unit tests for new code added in this PR.

- [X] Ran `cargo clippy --workspace --features "empty"` (may require `cargo clean` before)
- [X] Ran `cargo build --features "empty"`
- [X] Ran `cargo test --workspace --features "empty"`

### DIFF
--- a/amethyst_core/src/lib.rs
+++ b/amethyst_core/src/lib.rs
@@ -58,6 +58,5 @@ mod hidden;
 mod named;
 pub mod system_ext;
 
-
 /// A rayon thread pool wrapped in an `Arc`. This should be used as resource in `World`.
 pub type ArcThreadPool = Arc<rayon::ThreadPool>;
--- a/amethyst_core/src/lib.rs
+++ b/amethyst_core/src/lib.rs
@@ -56,6 +56,8 @@ mod event;
 mod hidden;
 //mod hide_system;
 mod named;
+pub mod system_ext;
+
 
 /// A rayon thread pool wrapped in an `Arc`. This should be used as resource in `World`.
 pub type ArcThreadPool = Arc<rayon::ThreadPool>;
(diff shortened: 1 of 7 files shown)


chemicstry:
This looks good in general!

Regarding tests, have you tried running them inside `amethyst_core` and not at the top level dir? I remember successfully running all transform tests that are present in the core. I think it would be better to have an explicit test for this and not just an example snippet.

ezpuzz:
rebased in #2560

PR #2468, opened 2020-08-29
Fix clippy, don't deny warnings

Veetaha:
## Description

Fixed the new warnings introduced in 1.46.0 update.
Also had to remove `--deny warnings` as discussed: https://github.com/amethyst/amethyst/pull/2450#issuecomment-683271243

## PR Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,8 +12,7 @@ on:
       - "**"
 
 env:
-  RUSTFLAGS: --deny warnings -Cdebuginfo=0
-  RUSTDOCFLAGS: --deny warnings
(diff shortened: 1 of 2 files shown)


Veetaha:
cc @Blisto91

Blisto91 (approved):
Good man!

@CleanCut I'm gonna merge this because I think it makes sense in the short term. In the longer term we can still aim for disallowing warnings.

Maybe a temporary compromise could be to create an easy overview of the amount and type of warnings on the repo. 
Could be a separate workflow or a badge counting/linking to the warnings.
Then they could still be in focus without forcing a quick scramble when something new pops up.

PR #2470, opened 2020-08-30
Add correct path to large files in examples

Blisto91:
## Description

The ``.gitattributes`` file wasn't looking in the correct path for the LFS files after the assets were moved into their respective example folders.

I noticed that not all large files are in a folder named large, like ``teapot.obj`` in ``arc_ball_camera``, so it now includes all the files as long as they are in the assets folder.

## PR Checklist


0x6273:
> I noticed that not all large files are in a folder named large, like teapot.obj in arc_ball_camera, so it now includes all the files as long as they are in the assets folder.

That would put smaller files, like the assets for the pong tutorial, into lfs as well. I think we only want the exemples with very large assets to require installing the extension.

Blisto91:
@Blisto91 Ah rgr. Git newbie here! 😁
I guess it should only point to the large assets folder then.

When is a file large enough that it must be tracked with lfs?
``scene.bin`` in ``examples\rendy\assets\large\crystal_stone\`` is only 20KB
While ``teapot.obj`` in  ``examples\arc_ball_camera\assets\mesh\`` is 7.24MB

Blisto91:
I'll squash before merge but i guess this is more correct now.

jojolepro (approved):

PR #2474, opened 2020-08-31
book/src/appendices: Add missing link for Serde

scott-linder:
## Description

Add a missing reference-style link for Serde to Appendix A. This appears to be the first direct reference to Serde in the text, and the original author seems to have intended to make it a link to either the Serde source or documentation; I opted to link to the latter.

## PR Checklist


Blisto91 (approved):
Thank you!

I'm in the middle of going through every link in the book. So the help is appreciated :)

PR #2479, opened 2020-09-05
Fix #2083 - [BUG] FrameLimiter Sleep strategy sleeps too long

### LINKED ISSUE #2083: [BUG] FrameLimiter Sleep strategy sleeps too long

borup3:
## Description

This PR is a proposed fix for issue https://github.com/amethyst/amethyst/issues/2083 that has observed an apparent problem with `FrameLimiter`'s implementation of `FrameRateLimitStrategy::Sleep`. As mentioned in the issue, this is highly likely to over-sleep due to the imprecise nature of operating system sleeps. The proposed fix is a halfway house between sleeping and yielding, not unlike `FrameRateLimitStrategy::SleepAndYield(Duration)` did but with an automatic duration equal to half the requested frame time.

The primary reasoning behind this change is to reduce the complexity of the struct for end-users who are unfamiliar with the undesirable intricacies of operating system sleeps, from a game development point of view. These users may consider `FrameRateLimitStrategy::Sleep` a sensible choice due to its labelling as the more pro battery-life approach and may not understand why their game may suffer from inconsistent timing issues as a result. This fix attempts to strike a balance between `Sleep` and `Yield` without requiring the user to be aware of the intricacies of sleeping, so that `Sleep` becomes a sensible default go-to option.

`FrameRateLimitStrategy::Sleep` has been removed. It is now called `FrameRateLimitStrategy::SleepAndYield`. The prior `FrameRateLimitStrategy::SleepAndYield` is now called `FrameRateLimitStrategy::SleepAndYieldDuration`.

The new `FrameRateLimitStrategy::SleepAndYield` implementation will attempt to sleep for the first 50% of the desired frame time and yield for the remaining 50%. This is done automatically for the end-user of the struct when they invoke `set_rate`. The new `FrameRateLimitStrategy::SleepAndYield` should now offer a sensible default behavior in place of `FrameRateLimitStrategy::Yield`, albeit this change has not been made in this PR.

This demotes `FrameRateLimitStrategy::SleepAndYieldDuration(Duration)` to being nothing more than an alternate version of `FrameRateLimitStrategy::SleepAndYield` that grants the end-user more control over the sleep-to-yield ratio should they need it.

## Additions

This PR adds a new function `pub fn FrameLimiter::get_frame_duration(&self) -> Duration` that returns `self.frame_duration`.

## Removals

`FrameRateLimitStrategy::Sleep` has been removed. It is now called `FrameRateLimitStrategy::SleepAndYield`. The prior `FrameRateLimitStrategy::SleepAndYield` is now called `FrameRateLimitStrategy::SleepAndYieldDuration`.

## Modifications

The `FrameLimiter` struct has a new member `sleep_barrier` that is assigned in `FrameLimiter::set_rate` to half of the desired frame duration. The implementation of `FrameRateLimitStrategy::SleepAndYield` in `FrameLimiter::wait` has been changed to yield once the `sleep_barrier` has been breached.

## PR Checklist

- [X] Ran `cargo clippy --workspace --features "empty"` (may require `cargo clean` before)
- [X] Ran `cargo build --features "empty"`
- [X] Ran `cargo test --workspace --features "empty"`

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -16,6 +16,8 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 
 ### Fixed
 
+[#2083]: https://github.com/amethyst/amethyst/pull/2478
+
 ## [0.15.3] - 2020-08-22
 
 0.15.3 is a corrected version of 0.15.2, which was inadvertently published based off of an incorrect (and broken) commit.
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -16,7 +16,7 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 
 ### Fixed
 
-[#2083]: https://github.com/amethyst/amethyst/pull/2478
+[#2083]: https://github.com/amethyst/amethyst/pull/2479
 
 ## [0.15.3] - 2020-08-22
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -16,7 +16,9 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 
 ### Fixed
 
-[#2083]: https://github.com/amethyst/amethyst/pull/2479
+- Fix `FrameRateLimitStrategy::Sleep` strategy sleeps too long ([#2479])
+
+[#2479]: https://github.com/amethyst/amethyst/pull/2479
 
 ## [0.15.3] - 2020-08-22
 
--- a/amethyst_core/src/frame_limiter.rs
+++ b/amethyst_core/src/frame_limiter.rs
@@ -29,7 +29,7 @@
 //! let assets_dir = "./";
 //! let mut game = Application::build(assets_dir, GameState)?
 //!     .with_frame_limit(
-//!         FrameRateLimitStrategy::SleepAndYield(Duration::from_millis(2)),
+//!         FrameRateLimitStrategy::SleepAndYieldDuration(Duration::from_millis(2)),
 //!         144,
 //!     )
 //!     .build(GameDataBuilder::new())?;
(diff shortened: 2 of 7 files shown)


0x6273:
I think this makes the names a bit confusing. If there is one enum named `SleepAndYield`, I would expect the enum named `Sleep` to just sleep, and not yield.
Why not get rid of `Sleep` and make `SleepAndYield` default to sleeping for half a frame?

borup3:
> I think this makes the names a bit confusing. If there is one enum named `SleepAndYield`, I would expect the enum named `Sleep` to just sleep, and not yield.
> Why not get rid of `Sleep` and make `SleepAndYield` default to sleeping for half a frame?

I think since the dominant behavior of `Sleep` is still sleeping the name still makes sense. Whereas for `SleepAndYield` you can customize the dominant behavior to your needs. For games, you should never limit frames purely with sleeps so it shouldn't even be offered as an option. You don't want to take away the users ability to take full control over their limiter (with `SleepAndYield`) if they want that for whatever reason.

This approach is backwards-compatible and gives any games that are currently built with `Sleep` a better solution without needing any effort or awareness on their end to update their code. Otherwise I would recommend naming `Sleep` as `SleepAndYield` and changing `SleepAndYield` to `SleepAndYieldDuration`. And not offer a `Sleep` behavior at all.

0x6273:
> Otherwise I would recommend naming `Sleep` as `SleepAndYield` and changing `SleepAndYield` to `SleepAndYieldDuration`. And not offer a `Sleep` behavior at all.

That makes a lot more sense to me, since they both sleep and yield, and the only difference is that the former is hard-coded to sleep for half a frame duration. The breakage is probably not a big deal, since most people just use the default, and it's a trivial fix for those that don't.

0x6273:
If you really don't want to break people's code, you could slap a `#[deprecated]` on `Sleep`, and name the new frame limiter `SleepAndYieldHalf`.

borup3:
I've renamed Sleep to SleepAndYield and SleepAndYield to SleepAndYieldDuration.

docs/CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ The format is based on [Keep a Changelog][kc], and this project adheres to
 
 ### Fixed
 
+[#2083]: https://github.com/amethyst/amethyst/pull/2479
```
0x6273, line 19:
```suggestion
[#2479]: https://github.com/amethyst/amethyst/pull/2479
```
Also, you need a line above it like: `- Short summary of the fix ([#2479])`

0x6273 (approved):

jojolepro:
I think it would be preferable to eventually move the sleeping to an external crate. One like this: https://crates.io/crates/spin_sleep
Currently, spin_sleep doesn't support wasm (no access to the `Time` std api). Once it does, delegating the sleeping behavior to a library would allow us to focus on maintaining other parts of the engine instead of dealing over and over with sleep/timing/cpu usage issues. This needs futher discussion with other members so we can agree on a strategy going forward. As such, I opened a forum discussion here: https://community.amethyst.rs/t/sleep-and-cpu-usage-going-forward/1763

jojolepro:
Following conversations, we will close this PR and instead switch to the spin_sleep crate. Thanks for your contribution!

PR #2481, opened 2020-09-05
Cow takeover

Xaeroxe:
## Description

Moo.

image

On a more serious note, this PR removes a bunch of generics from the `amethyst_input` crate, obsoleting the `BindingTypes` trait.

This new approach has a few benefits.

1. Almost all of our users were using constant strings anyways. This is a performance optimization for that use case, while also permitting runtime strings.
2. No more of these generics propagating throughout Amethyst.

## Additions

None

## Removals

- amethyst_input::BindingTypes

## Modifications

- Most of `amethyst_input`. The changes were primarily to substitute `Cow<'static, str>` for generic types.

## PR Checklist

### DIFF
--- a/amethyst_controls/src/bundles.rs
+++ b/amethyst_controls/src/bundles.rs
@@ -1,8 +1,7 @@
-use std::marker::PhantomData;
+use std::borrow::Cow;
 
 use amethyst_core::{ecs::*, math::one, shrev::EventChannel};
 use amethyst_error::Error;
-use amethyst_input::BindingTypes;
 
 use winit::Event;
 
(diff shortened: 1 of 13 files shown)


Xaeroxe:
I'm not sure what the expected state of the `legion_v2` branch is, so the usual routine maintenance work isn't done in this PR.

chemicstry (approved):

jojolepro:
The only thing supposed to get merged into the legion port is.. well the legion port. I'm sorry but this change has to wait for after the port is completed and merged back into master. fletcher's rule...

Xaeroxe:
@fhaynes My thoughts are roughly that 0.16 is going to have a lot of breaking changes already, so it seems like a great place to add some more. These changes are pretty thoroughly unit tested, which significantly lowers the risk.

jlowry:
@Xaeroxe What is the performance benefit?

Xaeroxe:
@jlowry in the event that two static strings are being compared, this version can optimize the equality check to just checking if the references point to the same string. So, in some circumstances you get to compare pointer equality, which is much cheaper than string equality.

jlowry:
When I use Amethyst, I use custom control bindings as described in the [documentation page](https://book.amethyst.rs/stable/input/how_to_define_custom_control_bindings.html). I would like to contine using them.

Is there a way we could have the performance boost and still support CustomBinding?

Xaeroxe:
@jlowry after this PR we will still support that use case, however it won't benefit from the static string optimization.

That's fine, that optimization isn't the primary point of this PR. The point of this PR is to reduce monomorphization.

fhaynes:
This is a reasonable ergonomics improvement, I'm fine with merging it in if we're sure it won't break any existing features/functionality.

PR #2483, opened 2020-09-08
fix examples for legion integration

mockersf:
fixed the three first steps of the pong tutorial, they are now working locally

I'll continue to fix examples, but wanted to get feedback if any

### DIFF
--- a/examples/pong_tutorial_02/pong.rs
+++ b/examples/pong_tutorial_02/pong.rs
@@ -1,7 +1,6 @@
 use amethyst::{
     assets::{AssetStorage, Handle, Loader},
     core::transform::Transform,
-    // ecs::prelude::{Component, DenseVecStorage},
     prelude::*,
     renderer::{Camera, ImageFormat, SpriteRender, SpriteSheet, SpriteSheetFormat, Texture},
 };
--- a/examples/pong_tutorial_02/pong.rs
+++ b/examples/pong_tutorial_02/pong.rs
@@ -1,7 +1,7 @@
 use amethyst::{
     assets::{AssetStorage, Handle, Loader},
     core::transform::Transform,
-    ecs::prelude::{Component, DenseVecStorage},
+    // ecs::prelude::{Component, DenseVecStorage},
     prelude::*,
     renderer::{Camera, ImageFormat, SpriteRender, SpriteSheet, SpriteSheetFormat, Texture},
 };
@@ -15,15 +15,13 @@ const PADDLE_WIDTH: f32 = 4.0;
 pub struct Pong;
 
 impl SimpleState for Pong {
-    fn on_start(&mut self, data: StateData<'_, GameData<'_, '_>>) {
+    fn on_start(&mut self, data: StateData<'_, GameData>) {
         let world = data.world;
 
         // Load the spritesheet necessary to render the graphics.
         // `spritesheet` is the layout of the sprites on the image;
         // `texture` is the pixel data.
-        let sprite_sheet_handle = load_sprite_sheet(world);
-
-        world.register::<Paddle>();
+        let sprite_sheet_handle = load_sprite_sheet(data.resources);
 
         initialise_paddles(world, sprite_sheet_handle);
         initialise_camera(world);
@@ -52,14 +50,10 @@ impl Paddle {
     }
 }
 
-impl Component for Paddle {
-    type Storage = DenseVecStorage<Self>;
-}
-
-fn load_sprite_sheet(world: &mut World) -> Handle<SpriteSheet> {
+fn load_sprite_sheet(resources: &mut Resources) -> Handle<SpriteSheet> {
     let texture_handle = {
-        let loader = world.read_resource::<Loader>();
-        let texture_storage = world.read_resource::<AssetStorage<Texture>>();
+        let loader = resources.get::<Loader>().unwrap();
+        let texture_storage = resources.get::<AssetStorage<Texture>>().unwrap();
         loader.load(
             "texture/pong_spritesheet.png",
             ImageFormat::default(),
@@ -67,8 +61,8 @@ fn load_sprite_sheet(world: &mut World) -> Handle<SpriteSheet> {
             &texture_storage,
         )
     };
-    let loader = world.read_resource::<Loader>();
-    let sprite_sheet_store = world.read_resource::<AssetStorage<SpriteSheet>>();
+    let loader = resources.get::<Loader>().unwrap();
+    let sprite_sheet_store = resources.get::<AssetStorage<SpriteSheet>>().unwrap();
     loader.load(
         "texture/pong_spritesheet.ron",
         SpriteSheetFormat(texture_handle),
@@ -83,11 +77,7 @@ fn initialise_camera(world: &mut World) {
     let mut transform = Transform::default();
     transform.set_translation_xyz(ARENA_WIDTH * 0.5, ARENA_HEIGHT * 0.5, 1.0);
 
-    world
-        .create_entity()
-        .with(Camera::standard_2d(ARENA_WIDTH, ARENA_HEIGHT))
-        .with(transform)
-        .build();
+    world.push((Camera::standard_2d(ARENA_WIDTH, ARENA_HEIGHT), transform));
 }
 
 /// Initialises one paddle on the left, and one paddle on the right.
@@ -104,18 +94,16 @@ fn initialise_paddles(world: &mut World, sprite_sheet_handle: Handle<SpriteSheet
     let sprite_render = SpriteRender::new(sprite_sheet_handle, 0); // paddle is the first sprite in the sprite_sheet
 
     // Create a left plank entity.
-    world
-        .create_entity()
-        .with(sprite_render.clone())
-        .with(Paddle::new(Side::Left))
-        .with(left_transform)
-        .build();
+    world.push((
+        sprite_render.clone(),
+        Paddle::new(Side::Left),
+        left_transform,
+    ));
 
     // Create right plank entity.
-    world
-        .create_entity()
-        .with(sprite_render)
-        .with(Paddle::new(Side::Right))
-        .with(right_transform)
-        .build();
+    world.push((
+        sprite_render.clone(),
+        Paddle::new(Side::Right),
+        right_transform,
+    ));
 }
--- a/examples/pong_tutorial_03/systems/paddle.rs
+++ b/examples/pong_tutorial_03/systems/paddle.rs
@@ -1,29 +1,24 @@
 use crate::pong::{Paddle, Side, ARENA_HEIGHT, PADDLE_HEIGHT};
 use amethyst::{
     core::transform::Transform,
-    derive::SystemDesc,
-    ecs::prelude::{Join, Read, ReadStorage, System, SystemData, WriteStorage},
-    input::{InputHandler, StringBindings},
+    ecs::{Runnable, SystemBuilder},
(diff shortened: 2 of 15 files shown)


mockersf:
for the system in pong tutorial step 3, I couldn't manage to declare the system like this:

```rust
#[system]
#[read_component(Paddle)]
#[write_component(Transform)]
fn paddle(world: &SubWorld, #[resource] input: &InputHandler<StringBindings>) {
    for (paddle, transform) in <(&Paddle, &mut Transform)>::query().iter_mut(world) {
        let movement = match paddle.side {
            Side::Left => get_input_axis_simple(&Some("left_paddle".to_string()), input),
            Side::Right => get_input_axis_simple(&Some("right_paddle".to_string()), input),
        };
        let scaled_amount = 1.2 * movement;
        let paddle_y = transform.translation().y;
        transform.set_translation_y(
            (paddle_y + scaled_amount)
                .min(ARENA_HEIGHT - PADDLE_HEIGHT * 0.5)
                .max(PADDLE_HEIGHT * 0.5),
        );
    }
}
```

it gives me the following errors:
```
error[E0432]: unresolved import `legion`
  --> examples/pong_tutorial_03/systems/paddle.rs:34:1
   |
34 | #[system]
   | ^^^^^^^^^ help: a similar path exists: `amethyst_core::legion`
   |
   = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0433]: failed to resolve: could not find `legion` in `{{root}}`
  --> examples/pong_tutorial_03/systems/paddle.rs:34:1
   |
34 | #[system]
   | ^^^^^^^^^ could not find `legion` in `{{root}}`
   |
   = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0433]: failed to resolve: could not find `legion` in `{{root}}`
  --> examples/pong_tutorial_03/systems/paddle.rs:34:1
   |
34 | #[system]
   | ^^^^^^^^^ not found in `legion::systems`
   |
   = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing one of these items
   |
1  | use amethyst::ecs::SystemBuilder;
   |
1  | use amethyst_core::ecs::SystemBuilder;
   |
1  | use crate::SystemBuilder;
   |
```
and `amethyst::ecs::SystemBuilder` is already imported.

I'm guessing there's something going on with re-export of legion and how the macro use it's imports

mockersf:
after digging more in how the `system` macro works, it's because it uses the format `::legion::...` for all it's uses of `legion` ([see here for example](https://docs.rs/legion_codegen/0.3.0/src/legion_codegen/lib.rs.html#759)) which means that `legion` has to be a direct dependency when using `amethyst`.

cf [Path Clarity in rust docs](https://doc.rust-lang.org/stable/edition-guide/rust-2018/module-system/path-clarity.html):
> Paths starting with :: must reference an external crate.

examples/pong_tutorial_02/pong.rs
```diff
@@ -1,7 +1,7 @@
 use amethyst::{
     assets::{AssetStorage, Handle, Loader},
     core::transform::Transform,
-    ecs::prelude::{Component, DenseVecStorage},
+    // ecs::prelude::{Component, DenseVecStorage},
```
rostyslav52, line 4: Maybe, remove completely?

examples/pong_tutorial_02/pong.rs
```diff
@@ -15,15 +15,13 @@ const PADDLE_WIDTH: f32 = 4.0;
...
-    fn on_start(&mut self, data: StateData<'_, GameData<'_, '_>>) {
+    fn on_start(&mut self, data: StateData<'_, GameData>) {
         let world = data.world;
 
         // Load the spritesheet necessary to render the graphics.
         // `spritesheet` is the layout of the sprites on the image;
         // `texture` is the pixel data.
-        let sprite_sheet_handle = load_sprite_sheet(world);
-
-        world.register::<Paddle>();
```
rostyslav52, line 26: registering Paddle was needed for specs by design, right?

examples/pong_tutorial_03/systems/paddle.rs
```diff
@@ -1,36 +1,30 @@
 use crate::pong::{Paddle, Side, ARENA_HEIGHT, PADDLE_HEIGHT};
 use amethyst::{
     core::transform::Transform,
-    derive::SystemDesc,
-    ecs::prelude::{Join, Read, ReadStorage, System, SystemData, WriteStorage},
-    input::{InputHandler, StringBindings},
+    ecs::{Runnable, SystemBuilder},
```
rostyslav52, line 4: Isn't it part of specs?

examples/pong_tutorial_03/systems/paddle.rs
```diff
@@ -1,36 +1,30 @@
...
-            };
-            if let Some(mv_amount) = movement {
-                let scaled_amount = 1.2 * mv_amount;
+pub fn build() -> impl Runnable {
+    SystemBuilder::new("PaddleSystem")
+        .with_query(<(&Paddle, &mut Transform)>::query())
+        .read_resource::<InputHandler<StringBindings>>()
+        .read_component::<Paddle>()
+        .write_component::<Transform>()
+        .build(move |_commands, world, input, query_paddles| {
```
rostyslav52, line 15: When I was reading this blog post 
https://amethyst.rs/posts/legion-ecs-v0.3
I thought that in Legion systems are just simple Rust functions with attached macros and specific arguments. 

Is SystemBuilder a different way to do the same thing?
And if so, why did you choose this other solution?

examples/pong_tutorial_04/systems/bounce.rs
```diff
@@ -1,62 +1,78 @@
...
+                        let paddle_y = paddle_transform.translation().y - (paddle.height * 0.5);
+                        (
+                            paddle_x,
+                            paddle_y,
+                            paddle.width,
+                            paddle.height,
+                            paddle.side.clone(),
+                        )
+                    })
+                    .collect::<Vec<_>>();
```
rostyslav52, line 34: The previous version used .join and was much shorter.
I understand, that join operations are part of specs design and not Legion, but it would still be awesome if this code could get shorter.

examples/pong_tutorial_04/systems/bounce.rs
```diff
@@ -1,62 +1,78 @@
...
-            if (ball_y <= ball.radius && ball.velocity[1] < 0.0)
-                || (ball_y >= ARENA_HEIGHT - ball.radius && ball.velocity[1] > 0.0)
-            {
-                ball.velocity[1] = -ball.velocity[1];
-            }
-
-            // Bounce at the paddles.
-            for (paddle, paddle_transform) in (&paddles, &transforms).join() {
-                let paddle_x = paddle_transform.translation().x - (paddle.width * 0.5);
-                let paddle_y = paddle_transform.translation().y - (paddle.height * 0.5);
```
rostyslav52, line 38: This is the .join I was talking about in the previous comment

examples/pong_tutorial_04/systems/move_balls.rs
```diff
@@ -1,26 +1,21 @@
...
-            local.prepend_translation_y(ball.velocity[1] * time.delta_seconds());
-        }
-    }
+pub fn build() -> impl Runnable {
+    SystemBuilder::new("MoveBallsSystem")
+        .with_query(<(&Ball, &mut Transform)>::query())
+        .read_resource::<Time>()
+        .read_component::<Ball>()
+        .write_component::<Transform>()
+        .build(move |_commands, world, time, query_balls| {
```
rostyslav52, line 15: Once again, I wonder if using #[system] macro would help as described in 
https://amethyst.rs/posts/legion-ecs-v0.3

rostyslav52 (commented):
Overall, I think it is good. I would like to get answers to my questions though before approving :)

Do I even have rights to approve it?

examples/pong_tutorial_03/systems/paddle.rs
Jazarro: Amethyst re-exports the specs api as amethyst::ecs. It is still specs but you don't import from specs directly.

examples/pong_tutorial_02/pong.rs
mockersf: yes, in an ecstatics like specs, you register your components so that it knows how to store them.
in an ecstatics like legion, based on archetypes, components storage directly depends on how you add them to entities

examples/pong_tutorial_02/pong.rs
mockersf: right, I missed this one

examples/pong_tutorial_03/systems/paddle.rs
mockersf: in the legion_v2 branch, the `ecs` module refers to legion

examples/pong_tutorial_04/systems/bounce.rs
mockersf: the direct translation of that have been would be:
```rust
for (ball, transform) in query_balls.iter_mut(world) {
    ...
    for (paddle, paddle_transform) in query_paddles.iter(world) {
        let paddle_x = paddle_transform.translation().x - (paddle.width * 0.5);
        let paddle_y = paddle_transform.translation().y - (paddle.height * 0.5);
        ...
    }
}
```
but that doesn't compile because `world` is already borrowed as mutable by the first query on `(Ball, Transform)`

The correct way to run nested queries seems to be to split the world like so:
```rust
let (mut ball_world, remaining) = world.split_for_query(query_balls);

for (ball, transform) in query_balls.iter_mut(&mut ball_world) {
    ...
    for (paddle, paddle_transform) in query_paddles.iter(&remaining) {
        ...
    }
}
```
but then this crash during run when the ball is spawned:
```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: ComponentAccessError', /Users/fmockers/.cargo/git/checkouts/legion-6fbc02e8da0bdce7/b93b636/src/internals/query/mod.rs:313:59
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```


A better explanation can be obtained for the crash by splitting the world a little more:
```rust
let (mut ball_world, mut remaining) = world.split_for_query(query_balls);
let (paddle_world, _) = remaining.split_for_query(query_paddles);
```
which then crash at start with:
```
thread '<unnamed>' panicked at 'view accesses components unavailable in this world: world allows only reads: [pong_tutorial_04::pong::Paddle], writes: [], view requires reads: [pong_tutorial_04::pong::Paddle, amethyst_core::transform::components::transform::Transform], writes: []', /Users/fmockers/.cargo/git/checkouts/legion-6fbc02e8da0bdce7/b93b636/src/internals/subworld.rs:119:21
```

So the issue is that both queries are trying to access the `Transform` components, so they can't be nested.

I see 2 workarounds for that:
- run the paddle query outside of the ball query, as it doesn't need to change any components we can just extract the values we need without keeping a reference to the world. this is what I did here
- add a new type wrapping `Transform`, and a system dedicated to applying changes to this new type to the `Transform` component, so that we wouldn't access `Transform` for the paddle query but the new type component. this seems a little too much...

But that's one of the part I'm not happy with, if someone has a better idea how to solve it

rostyslav52: Ok, I see. I don't have any suggestion for now

rostyslav52 (approved):

ezpuzz:
closing in favor of #2531 which continues this work

Issue #2463, opened 2020-08-25
[BUG] Crash with display.ron UTF-8-BOM encoding


Grzybojad:
## Description

A runtime crash when using a `display.ron` file encoded using UTF-8-BOM or other BOM encodings.
Changing the encoding on this file to ANSI or UTF-8 fixes the issue.

## Reproduction Steps

This issue was discovered while following the "Setting up the project" part of the official [Pong tutorial](https://book.amethyst.rs/stable/pong-tutorial/pong-tutorial-01.html) but it should be reproducible for any project utilizing a `display.ron` file.

1. Set up an Amethyst project that utilizes a display config file
2. Create the config file and set its encoding to UTF-8-BOM (using Notepad++ for example)
3. Build the project using `cargo build`
4. Run the project using `cargo run`

## What You Expected to Happen

Expected the program to run or display a clear error suggesting to change the character encoding.

## What Actually Happened

Upon running the build, this error popped up and stopped the program
```
Error: Error { inner: Inner { source: None, backtrace: None, error: Parser(Parser(ExpectedStruct, Position { col: 1, line: 1 })) } }
error: process didn't exit successfully: `target\debug\pong.exe` (exit code: 1)
```

## Screenshots and Logs

![image](https://user-images.githubusercontent.com/18161644/91224373-4de55180-e722-11ea-9792-4aca160199c9.png)

## Your Environment

1. OS: Windows 10 Home
2. Amethyst: 0.15

CleanCut:
For the benefit of folks without access to Windows, would you mind attaching an example file?

Grzybojad:
Here's a UTF-8-BOM encoded display.ron file.
I changed the extension to .txt, because GitHub wouldn't accept a .ron file.

[display.txt](https://github.com/amethyst/amethyst/files/5130480/display.txt)

Blisto91:
So I'm thinking out loud here without really having looked into it.
It should be possible to recode the stream as utf-8 (or whatever is preferred) before the parser tries to make sense of it.
Or even better would be if it didn't care much and were able to interpret all the common file encodings.

Blisto91:
@Grzybojad Can you try pointing to [this branch](https://github.com/amethyst/amethyst/tree/allow-encoding-variants) and see if it fixes it?

Grzybojad:
[...]

PR #2401, opened 2020-08-03
Support for JSON & Binary Config Files

deprilula28:
## Description

Adds support for JSON & Binary config files, under a feature in amethyst_config

## Additions

- Adds the ConfigFormat enum, with Ron, Json & Binary
- Changes the main example on amethyst_config to read & write from all the supported formats
- Adds the "json" and "binary" features for support on those file types

## Modifications

- Deprecates Config.write(AsRef<Path>) in favor of Config.write_format(ConfigFormat, AsRef<Path>)
- Deprecates Config.load_bytes(&[u8]) in favor of Config.write_format(ConfigFormat, &[u8])

## PR Checklist

By placing an x in the boxes I certify that I have:

- [x] Updated the content of the book if this PR would make the book outdated.
- [x] Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.
- [x] Added unit tests for new code added in this PR.
- [x] Acknowledged that by making this pull request I release this code under an MIT/Apache 2.0 dual licensing scheme.

If this modified or created any rs files:

- [x] Ran `cargo +stable fmt --all`
- [x] Ran `cargo clippy --all --features "empty"`
- [x] Ran `cargo test --all --features "empty"` (Also with the binary & json features enabled)

## Other Notes
Output for `cargo run --example main --features json,binary` in `amethyst_config`:
```rust
RON Config Result:
ExampleConfig {
    display: DisplayConfig {
        title: "Amethyst game",
        brightness: 1.0,
        fullscreen: false,
        dimensions: (
            1024,
            768,
        ),
        min_dimensions: None,
        max_dimensions: None,
        vsync: true,
        multisampling: 0,
        visibility: true,
    },
    logging: LoggingConfig {
        file_path: "new_project.log",
        output_level: "warn",
        logging_level: "debug",
    },
}
JSON Config Result:
ExampleConfig {
    display: DisplayConfig {
        title: "Amethyst game",
        brightness: 1.0,
        fullscreen: false,
        dimensions: (
            1024,
            768,
        ),
        min_dimensions: None,
        max_dimensions: None,
        vsync: true,
        multisampling: 0,
        visibility: true,
    },
    logging: LoggingConfig {
        file_path: "new_project.log",
        output_level: "warn",
        logging_level: "debug",
    },
}
Binary Config Result:
ExampleConfig {
    display: DisplayConfig {
        title: "Amethyst game",
        brightness: 1.0,
        fullscreen: false,
        dimensions: (
            1024,
[...]

Issue #2386, opened 2020-07-22
[BUG] Pong Example Sprite Assets Fail to Load on Ubuntu 20.04


0ur0bor0s:
## Description

Running pong code on new amethyst project causes sprites to fail to render followed by error messages.  pong_tutorial_02 in amethyst example directory builds correctly.

## Reproduction Steps

1. Install Amethyst tools with `cargo install amethyst_tools`
2. Create new project with `amethyst new pong`
3. To save time, copy source, asset, and config files from pong_tutorial_02 in amethyst example folder
4. `cargo run`

## What You Expected to Happen

Pong assets are rendered on screen.

## What Actually Happened

Assets are not loaded, followed by error messages

## Screenshots and Logs

Console Output:
```
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
     Running `target/debug/pong`
[INFO][amethyst::app] Initializing Amethyst...
[INFO][amethyst::app] Version: 0.15.0
[INFO][amethyst::app] Platform: x86_64-unknown-linux-gnu
[INFO][amethyst::app] Amethyst git commit: UNKNOWN
[INFO][amethyst::app] Rustc version: 1.45.0 Stable
[INFO][amethyst::app] Rustc git commit: 5c1f21c3b82297671ad3ae1e8c942d2ca92e84f2
[INFO][winit::platform::platform::x11::window] Guessed window DPI factor: 1
[WARN][rendy_factory::factory] Slow safety checks are enabled! Disable them in production by enabling the 'no-slow-safety-checks' feature!
[INFO][rendy_util::wrap] Slow safety checks are enabled! You can disable them in production by enabling the 'no-slow-safety-checks' feature!
[ERROR][amethyst_assets::storage] "renderer::SpriteSheet": Asset "texture/pong_spritesheet.ron" (handle id: Handle { id: 0 }) could not be loaded: Failed to load asset with name "texture/pong_spritesheet.ron"
[ERROR][amethyst_assets::progress] Error loading handle 0, renderer::SpriteSheet, with name texture/pong_spritesheet.ron: Failed to load asset with name "texture/pong_spritesheet.ron"
caused by: Failed to load asset with name "texture/pong_spritesheet.ron"
[...]

PR #2487, opened 2020-09-14
Allow more encodings for text assets and configs

Blisto91:
## Description

This PR allows a few more encoding variants to be used when using text based config files and assets.

Specifically this will try to detect UTF-8-BOM and UTF-16-BOM by BOM sniffing and internally convert them to regular UTF-8 before being deserialized.
If no BOM's are present then the raw bytes are passed through.
Since this is a simple addition that only handles the cases of UTF-8 & UTF-16 with BOM it means that if a file is encoded in UTF-16 without BOM it still won't work, as before. UTF-32 is not supported by this crate at all.

This change should be a non breaking improvement.
It also shouldn't pose a problem for any other formats that might be used in the future since UTF-8 is standard in rust and i except all crates that handle text bytes to support it out of the box.
I have also verified this up against the changes in [PR #2401](https://github.com/amethyst/amethyst/pull/2401).

Further improvements can probably be made for this in the future.
Software that support UTF-16 without BOM usually try and make a best guess with some heuristics and are fairly good at it. But it isn't a guarantee. 
An extra option would be to look for encoding labels if present.

Fixes [#2463 ](https://github.com/amethyst/amethyst/issues/2463) 
Possible fix for [#2386](https://github.com/amethyst/amethyst/issues/2386)

## PR Checklist


CleanCut (approved):
Looks solid.

I suggest adding a test or two.

Blisto91:
Never written a test before but i could probably figure it out.

Unsure how attack it tho since i just pushed it into the existing load function and i guess we want to test that the decoding works and not the config deserialization?
Should this stuff be a separate function? It's quite simple right now but could be expanded.

CleanCut:
We don't need to test the library's decoding functionality, just our usage of it.

I think for `amethyst_assets` it would be more than sufficient to just duplicate [this test](https://github.com/amethyst/amethyst/pull/2487/files#diff-f8945f1ce77d19a1f8eb29b0886acea7R94-R105) and its associated asset file, once each for UTF-8-BOM and UTF-16-BOM.

`amethyst_config` should really have similar three tests, but as it doesn't have any at all yet, which makes it harder.  I'll leave it up to you for whether you think it's worth the effort to figure out a good set of unit tests to write up or not.

CleanCut:
And just to be clear, if writing tests is going to be a big obstacle for you, I'd rather go ahead and merge than leave this hanging.  If you _are_ up to writing tests, that'd be a clear value-add.

Blisto91:
I'll read a bit into it and come back to you :)

Blisto91:
@CleanCut  First attempt!
If there's anything i should do differently or any conventions i'm not following then please tell. :)

CleanCut:
The `amethyst_config` tests look good to me. I verified that the encodings didn't get mangled by Git.  👍

Blisto91:
@CleanCut sorry for the delay here!
Sleep have not been agreeing with me again. 
Will try and include the asset tests this week.

Blisto91:
@CleanCut Added tests for text assets.
More or less the same as with config

CleanCut:
@Blisto91 LGTM. I'm hesitating to merge because I don't want to add any more burden to the legion effort.

jojolepro (approved):

jojolepro:
Please resolve the conflicts and we can merge. Thank! :hearts:

Blisto91:
Rebased, resolved and squashed.
Hope it didn't mess anything up. Still a newb with git in general.

jojolepro:
Thanks!
bors r+

Blisto91:
bors r-

Just noticed i accidently added the logs crate back into amethyst_config cargo.toml

Blisto91:
Fixed.
Guess that is a downside of a rebase. 😁 

bors r+

Blisto91:
Looks like the atelier tests randomly fail on Mac which might be a race condition from my naïve understanding of the the trace and debug info?
[Dunno if this atelier-assets PR will fix it.](https://github.com/amethyst/atelier-assets/pull/67)

ezpuzz:
> 
> 
> Still failed. Tho this time on nightly which is allowed hehe.
> 
> Looks like the atelier tests randomly fail on Mac which might be a race condition from my naïve understanding of the the trace and debug info?
> [Dunno if this atelier-assets PR will fix it.](https://github.com/amethyst/atelier-assets/pull/67)

@kabergstrom has fixed it on another atelier branch, we can ignore those failures for now

PR #2488, opened 2020-09-16
Documenting TransformBundle as a UiBundle dependency

dejaime:
## Description

Added TransformBundle as a dep of UiBundle.
I was trying to use Ui and couldn't even get my code to compile after adding the UiBundle as described in section _13. User Interface_ of the book. Turns out I also needed TransformBundle added before UiBundle, same as InputBundle.

## Modifications

- TransformBundle as a dependency of UiBundle in both /// docs and the book

## PR Checklist


CleanCut (approved):
Nice, thanks!

PR #2489, opened 2020-09-17
Update Rayon to 1.4 to use their new scheduler.

Elabajaba:
## Description

Rayon 1.4 was released a few weeks ago, and includes their new scheduler. This should help with the high idle CPU usage issues people have had with Amethyst.

## PR Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -133,7 +133,7 @@ crossbeam-channel = "0.4.2"
 derivative = "2.1.1"
 fern = { version = "0.6.0", features = ["colored"] }
 log = { version = "0.4.8", features = ["serde"] }
-rayon = "1.3.0"
+rayon = "1.4.0"
 rustc_version_runtime = "0.2.0"
 sentry = { version = "0.18.0", optional = true }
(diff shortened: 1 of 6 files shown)


Elabajaba:
After testing this a bit (using the tiles example and Dwarf Seeks Fortune on Windows 10/AMD 3950x CPU/1080ti GPU and a Linux/Intel i7-4720HQ/Nvidia 960m), I don't notice a difference in CPU usage, whether it's compiled in release mode or in debug mode (without --release) when comparing Rayon 1.3 and 1.4.

Release mode seems to be heavily vsync limited (and there seems to be no way to disable it? I couldn't find any references to vsync in rendy 0.4.1 either to disable it there), but I'm surprised that I didn't see a difference in debug mode.

Could anyone who has run into the high idle CPU usage in release mode test this?

CleanCut (approved):
Very nice. On my 16-core box (32 logical), this brings the `pong_tutorial_06` from saturating ~21 cores on rayon 1.3.0 to saturating ~2.5 cores for a solid order-of-magnitude improvement.  The CPU usage is still way too high, in my opinion, but this is a big step in the right direction.

PR #2492, opened 2020-09-20
Make TextEditingPrefab public

pkrum:
## Description

Make `TextEditingPrefab` public, so an editable `UiWidget::Label` can be build from code and not only parsed from an asset file.

## PR Checklist

- [ ] ~Added unit tests for new code added in this PR.~
- [ ] ~Updated the content of the book if this PR would make the book outdated.~


CleanCut (approved):
Thanks!

PR #2501, opened 2020-10-05
Migrate to rendy 0.5

str4d:
## Description

Migrates `amethyst_rendy` (and therefore the remainder of the `amethyst` crate stack) to use `rendy 0.5`, which includes a migration to `gfx-hal 0.4`.

The most annoying change due to this is the removal of nice HAL wrappers from `gfx-hal`, in particular `ClearColor`'s `From` impls. It might make sense to alter the `RenderToWindow::with_clear` API to use an in-house wrapper.

Full disclosure: I have never used `amethyst`, `rendy`, or `gfx-hal` before, and have no idea what I'm doing. I couldn't see an obvious reason for _not_ migrating in any of the open issues, so hopefully this is somewhat useful 🙂

## Removals

- The `failure` crate was removed from `rendy`, and is no longer used in APIs here.

## Modifications

- APIs that previously returned `failure::Error` now return `rendy` error types, that can bubble back up through the `RenderGroupDesc` trait.

## PR Checklist

### DIFF
--- a/amethyst_rendy/src/bundle.rs
+++ b/amethyst_rendy/src/bundle.rs
@@ -492,7 +492,7 @@ impl TargetImage {
 }
 
 /// Set of options required to create an image node in render graph.
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone)]
 pub struct ImageOptions {
     /// Image kind and size
     pub kind: hal::image::Kind,
@@ -612,6 +612,13 @@ impl<B: Backend> TargetPlan<B> {
             ));
         }
         let mut outputs = self.outputs.unwrap();
+        let suggested_extent = {
+            let metadata = ctx.target_metadata(self.key).unwrap();
+            hal::window::Extent2D {
+                width: metadata.width,
+                height: metadata.height,
+            }
+        };
 
         ctx.mark_evaluating(self.key)?;
 
@@ -648,7 +655,7 @@ impl<B: Backend> TargetPlan<B> {
             match color {
                 OutputColor::Surface(surface, clear) => {
                     subpass.add_color_surface();
-                    pass.add_surface(surface, clear);
+                    pass.add_surface(surface, suggested_extent, clear);
                 }
                 OutputColor::Image(opts) => {
                     let node = ctx.create_image(opts);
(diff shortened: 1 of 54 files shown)


amethyst_rendy/src/bundle.rs
```diff
@@ -956,7 +982,14 @@ mod tests {
                         .with_color_surface()
                         .with_depth_stencil(depth),
                 )
-                .with_surface(surface2, None),
+                .with_surface(
+                    surface2,
+                    hal::window::Extent2D {
+                        width: 1,
+                        height: 1,
+                    },
```
str4d, line 990: No idea how this should actually be set to in the test. This was necessary due to the API change in https://github.com/amethyst/rendy/pull/239.

amethyst_rendy/src/submodules/material.rs
```diff
@@ -84,15 +89,24 @@ impl<B: Backend> SlottedBuffer<B> {
...
+                    },
+                    rendy::memory::Dynamic,
+                )
+                .map_err(|e| match e {
+                    rendy::resource::CreationError::Allocate(
+                        rendy::memory::HeapsError::AllocationError(
+                            hal::device::AllocationError::OutOfMemory(oom),
+                        ),
+                    ) => oom.into(),
+                    _ => CreationError::Other,
```
str4d, line 108: Given that `rendy` now imposes `CreationError` as the error type for `RenderGroupDesc`, all the error types that were previously `failure::Error` now need to be coerced to it. AFAICT `rendy` doesn't provide any simple way to move between its different error types, so I've manually coerced a few where it was obvious.

str4d:
Looks like examples in the book also need updating. I'll hold off doing that until I hear feedback on what to do about `ClearColor`.

ezpuzz:
@str4d it would be best for end users if the api remained compatible.  could you make a version of `with_clear` that takes a vec of f32?

amethyst_rendy/src/bundle.rs
ezpuzz: seems fine, if the test passes it's inconsequential

amethyst_rendy/src/submodules/material.rs
ezpuzz: that should make errors easier to understand in my opinion, nice extra effort to not obscure them!

ezpuzz (changes requested):
I think if we can maintain the API of `with_clear` this looks good to go!

ezpuzz:
fixed by #2554 thanks @str4d

PR #2522, opened 2020-11-16
Autoselect Graphics Backend for Rendy

CleanCut:
## Description

The mandatory requirement to specify `--features=metal` (or `vulkan` or `empty`) is a thorn in the side.  It is a constant obstacle for users, and it causes problems in development as well.

Here's my initial (so far unsuccessful) attempt at having the features auto-detected by platform at the amethyst level (complicated by the fact that it's features at rendy's level).

- I found a way to conditionally specify an entire dependency in Cargo.toml based on platform, _but_ you can't look at `features` in that logic, which means dealing with `empty` has to be done elsewhere.  I had more comprehensive patterns in an earlier (uncommitted) revision.  In this revision things like iOS aren't addressed.
- I tried to alter the macros to accept entire `cfg` patterns (so I could deal with `empty` at the macro level), but macros refuse to take a literal for a `cfg` pattern for whatever reason.  

At the time of writing this, this PR is in a broken state. I need some way to deal with both backend default auto-detection and the ability to override the auto-detection via a feature. Anyone got any ideas?

## Additions

N/A

## Removals

- Remove `metal` and `vulkan` features from everywhere.  Auto-detect them based on operating system.

## Modifications

## PR Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,8 +22,6 @@ travis-ci = { repository = "amethyst/amethyst", branch = "master" }
 # TODO: include animation, ui
 default = ["audio", "locale", "network", "parallel", "renderer"]
 
-vulkan = ["amethyst_rendy/vulkan", "amethyst_rendy/vulkan-x11"]
-metal = ["amethyst_rendy/metal"]
 empty = ["amethyst_rendy/empty"]
 
 ui = [
@@ -340,4 +338,4 @@ path = "examples/tiles/main.rs"
 required-features = [ "tiles" ]
 
 [package.metadata.docs.rs]
-features = ["animation", "audio", "gltf", "tiles", "json", "locale", "network", "sdl_controller", "vulkan"]
+features = ["animation", "audio", "gltf", "tiles", "json", "locale", "network", "sdl_controller"]
--- a/amethyst_animation/Cargo.toml
+++ b/amethyst_animation/Cargo.toml
@@ -35,8 +35,6 @@ thread_profiler = { version = "0.3", optional = true }
 [dev-dependencies]
 
 [features]
-vulkan = ["amethyst_rendy/vulkan", "amethyst_rendy/vulkan-x11"]
-metal = ["amethyst_rendy/metal"]
 empty = ["amethyst_rendy/empty"]
 
 profiler = [ "thread_profiler/thread_profiler" ]
--- a/amethyst_gltf/Cargo.toml
+++ b/amethyst_gltf/Cargo.toml
@@ -35,8 +35,6 @@ image = "0.22.2"
 derivative = "2.1.1"
 
 [features]
-vulkan = ["amethyst_rendy/vulkan", "amethyst_rendy/vulkan-x11"]
-metal = ["amethyst_rendy/metal"]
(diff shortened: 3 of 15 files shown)


CleanCut:
Today:
- Started learning macros
- Rebased the PR against `legion_v2` instead of `master` (oops, habit)
- Uncommented the other conditional rendy dependency variantes in Cargo.toml
- Currently on my mac it tries to pull in the gfx-backend-vulkan dependency, which makes no sense.

jlowry:
This might require using custom [build script](https://doc.rust-lang.org/cargo/reference/build-script-examples.html#conditional-compilation) to create the [rustc-cfg](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-cfg) instructions.

The trade-off is this adds more complexity into the build vs having to specify the features on the invocation of cargo. There might be a better way but performing platform-specific configuration is a example use case given in the docs.

jlowry:
>     * Started learning macros
I'm interested in which resources you use as I'd like to improve my macro skills too. So far I have found David Tolnay's [procedural macros workshop](https://github.com/dtolnay/proc-macro-workshop) to be excellent.

AlveLarsson:
I think this is great if we only make the default auto-detect, but we do need to keep the option to let the user override

CleanCut:
> This might require using custom [build script](https://doc.rust-lang.org/cargo/reference/build-script-examples.html#conditional-compilation) to create the [rustc-cfg](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-cfg) instructions.
> 
> The trade-off is this adds more complexity into the build vs having to specify the features on the invocation of cargo. There might be a better way but performing platform-specific configuration is a example use case given in the docs.

I was super excited about the possibility of using a build script, which can specify just about any run-time flag...except for features. Other than the fact that it won't accept feature output, the build script solution was super straightforward. 😭  This seems pretty consistent throughout every place that the `cfg` system is exposed--everything is supported everywhere...except features. They're only supported if you manually supply them on the command line. It appears that any and every possibility for dynamically specifying them is blocked by design. :table-flip:
 
> > ```
> > * Started learning macros
> > ```
> 
> 
> I'm interested in which resources you use as I'd like to improve my macro skills too. So far I have found David Tolnay's [procedural macros workshop](https://github.com/dtolnay/proc-macro-workshop) to be excellent.

Nothing special.  Quite frankly, the fact that there are multiple forms of macros and they are their each their own little sort-of-language is a huge turn off for me.  Since the macros in `amethyst_rendy/src/types.rs` are all `macro_rules!` macros, I was just reading the book's section on them in https://doc.rust-lang.org/book/ch19-06-macros.html

> I think this is great if we only make the default auto-detect, but we do need to keep the option to let the user override

User-overriding is the easy, supported part. It's the default that's a pain.

CleanCut:
@erlend-sh I don't know how to interpret the 👎 .  Will you elaborate?

erlend-sh:
Oh my, haha, definitely a misclick!

CleanCut:
Closing in favor of #2525, which got a little bit closer.

PR #2525, opened 2020-11-23
Autoselect gfx (only works on nightly)

CleanCut:
A better than #2522 attempt to autoselect the graphics backend BUT only works by using nightly with the command `cargo +nightly build -Zfeatures=itarget` until rust-lang/cargo#7914 becomes stable.

Mostly just making this PR so it's easy to find the attempt later.

~I intend to make yet another attempt via intermediate crates that might work on stable.~ Nope, this really seems like the better approach, once rust-lang/cargo#7914 is stabilized, so I'll just wait.


CleanCut:
After quite a bit of research I am convinced that this fundamental approach is the best one once https://github.com/rust-lang/cargo/issues/7914 is stabilized.

Until then, the only viable path I see for not requiring an explicit feature to always be specified to cargo is to:

- always include all dependencies
- internal to each dependency use `#[cfg(target_os = "<supported os">)]` extensively so that an unsupported dependency _compiles_ but is effectively empty.

If I implement the above type of solution, I'll post a comment with a link (if someone else does it, please do the same 😄).

Blisto91:
If all goes well then the nightly feature might soon move forward with stabilization.
[The RFC is in final comment period.](https://github.com/rust-lang/rfcs/pull/2957)

Blisto91:
RFC merged.
[Cargo PR #8997](https://github.com/rust-lang/cargo/pull/8997) will stabilize these features. 🎉

ezpuzz:
cargo PR is merged but not released yet, so that's cool progress.

I was experimenting with this branch and I didn't seem to automatically get backend features set inside amethyst_rendy itself so it didn't choose a DefaultBackend for me.

Blisto91:
Toyed around with this myself also in the 0.15.3 source code.

Just adding this in amethyst cargo.toml
```
[target.'cfg(target_os = "macos")'.dependencies]
amethyst_rendy = { path = "amethyst_rendy", version = "0.15.3", features = ["metal"], optional = true }

[target.'cfg(target_os = "windows")'.dependencies]
amethyst_rendy = { path = "amethyst_rendy", version = "0.15.3", features = ["vulkan"], optional = true }

[target.'cfg(target_os = "linux")'.dependencies]
amethyst_rendy = { path = "amethyst_rendy", version = "0.15.3", features = ["vulkan", "vulkan-x11"], optional = true }
```

And this in amethyst_rendy if you want it to also autoselect in the lower crate.
```
[target.'cfg(target_os = "macos")'.dependencies]
rendy = { version = "0.4.1", default-features = false, features = ["metal"] }

[target.'cfg(target_os = "windows")'.dependencies]
rendy = { version = "0.4.1", default-features = false, features = ["vulkan"] }

[target.'cfg(target_os = "linux")'.dependencies]
rendy = { version = "0.4.1", default-features = false, features = ["vulkan", "vulkan-x11"] }
```
Seems to make it work.

But if you want to manually select another backend like ``--features=empty`` then it stops working because it'll enable both features and complain about ``DefaultBackend``.
I'm guessing all the extra code here tries to solve some of that?
Bummer that you can't enable these features in the cargo.toml depending on if another feature is set or not.

ezpuzz:
Here is the status as of 16f1add 

- `cargo run -p debug_lines` works without specifying backend
- `cargo test --all-targets --all-features --workspace` passes (when there are conflicting backend features i've defaulted to "empty")
- specifying an override works `cargo build --features empty`

ezpuzz:
with `--all-features` CI is going to give us a bunch more coverage as a bonus

Blisto91:
This look great.

There is a breaking change in that you can't use ``--features=vulkan`` on Mac now if you wanted to use MoltenVK.
Tho i'm not sure if gfx-hal behind the scenes [calls metal anyway?](https://github.com/gfx-rs/gfx/issues/2796#issuecomment-498211878)

ezpuzz:
> 
> 
> This look great.
> 
> There is a breaking change in that you can't use `--features=vulkan` on Mac now if you wanted to use MoltenVK.
> Tho i'm not sure if gfx-hal behind the scenes [calls metal anyway?](https://github.com/gfx-rs/gfx/issues/2796#issuecomment-498211878)
> 
> Edit:
> My tired brain is having trouble digesting all the cfg's right now heh. So not sure if what i said is true. 😁

I didn't know that was supported?  Because of the approach of configuring the rendy flags by platform this is not possible. not to mention it seems like rendy itself only exports the `rendy::vulkan` namespace if on windows :-/

https://github.com/amethyst/rendy/search?q=rendy_with_metal_backend

Blisto91:
Ah rgr. 👍 
I don't actually have a Mac and didn't dig deeper into it so was just from the top of my head .

ezpuzz:
My only point of debate internally was keeping the top-level empty/vulkan/metal features on the amethyst crate.

If we remove them then `--all-features` would enable the real backend instead of `empty` but you would have to specify overrides with `--features amethyst_rendy/empty` to build against empty backend.  I wonder how many people actually override to the empty backend?

ezpuzz (approved):

ezpuzz:
this works on stable btw

bors r+

CleanCut:
You guys rock. I seriously thought it'd be another year before this was unblocked. 🎉

PR #2532, opened 2020-12-08
remove defunct badges section from cargo.toml

ezpuzz:
Removes badges sections that are no longer rendered by crates.io and are defunct because we don't use travis any more.


AlveLarsson (approved):
Seems good. Thanks! :)

PR #2536, opened 2020-12-10
new hidden system

ezpuzz:
port HideHierarchySystem

leaving this as draft because it needs a example to test with


ezpuzz:
closing in favor of #2538

PR #2538, opened 2020-12-10
amethyst_core: Reimplement the hide hierarchy system in legion and add unit tests to it

grzi:
## Description

Re implementation of the hide hierarchy system in legion and unit tests 
I hope it's not a too naïve approch to this. open for reviews. 

## Additions

- Unit tests to validate the system

## Removals

- Spec implementation

## Modifications

- Rename to hide_hierarchy_system 
- Switch to legion

## PR Checklist

- [X] Added unit tests for new code added in this PR.

All of this has been done but due to the legion_v2 state it's actually failing, but not because of my devs


amethyst_core/src/lib.rs
```diff
@@ -13,6 +13,7 @@ pub use self::{
     axis::{Axis2, Axis3},
     event::EventReader,
     hidden::{Hidden, HiddenPropagate},
+    hide_hierarchy_system::build,
```
ezpuzz, line 16: I don't think this is the namespacing we want to use, probably can remove `::build`

amethyst_core/src/hide_hierarchy_system.rs
```diff
@@ -0,0 +1,328 @@
...
+};
+
+use crate::HiddenPropagate;
+use crate::transform::{Children, Parent};
+use std::collections::HashSet;
+
+/// This system adds a [HiddenPropagate](struct.HiddenPropagate.html)-component to all children
+/// of an entity with a [HiddenPropagate](struct.HiddenPropagate.html) and removes it when it is removed
+/// from the parent.
+pub fn build() -> impl Runnable {
```
ezpuzz, line 12: This is awesome! Thanks for adding tests 👍🏻   Way better than my attempt in #2536

ezpuzz (changes requested):
Thanks!

AlveLarsson:
Did you check out the LT repo we have? IDK if it's still valid or not or if this transform PR is better?
https://github.com/amethyst/legion_transform (and the readme is outdated)

ezpuzz (approved):

ezpuzz:
@AlveLarsson probably would be good to review legion_transform as a possible replacement after we port the existing implementations.  also I don't see anything about managing hidden state in legion_transform.

grzi:
@AlveLarsson The current system doesn't belong to the transform part IMHO, it's an additional feature specific to amethyst.

Yeah, legion_transform could replace the actual transform module as the code is nearly the same. Maybe as @ezpuzz said, after all the port is done we could 'merge' what has been done on amethyst (if there are) on the legion_transform project and use it as a replacement

AlveLarsson (approved):
Oh right that wat was what we said, first just a port of the normal transform and after 0.16 the transform repo. Thanks for this! Keep up the great work :)

PR #2541, opened 2020-12-11
[LEGION] system structs instead of build_xyz_system functions

ezpuzz:
# If reviewing use `git diff --minimal legion_v2 -b` for your own good 🥰 

## Description

This creates a new `dispatcher::System` trait that enforces systems have a `build(&mut self)` method to construct a `Runnable` with.  This gets rid of the hard to enforce `build_xyz_system(a, b, c, d,...)` style of system construction and instead passes structs to the dispatcher to be built.  This also allow systems to store their state in their own struct.

## Motivation and Context

This is needed to keep the codebase consistent, to have a standard for system state, and to allow easier refactoring later of system behavior.

One potential downside is that to add a system outside of a bundle you must use `builder.add_system(Box::new(MySystemStruct))` instead of `builder.add_system(build_some_system(x,y,z))` because the system must be allocated to the heap to get a static lifetime and maintain its own state from a `'static` FnMut.  I am open to conversation about other solutions but this is the only way I found to work with legion lifetimes and contain state within the system struct.

## How Has This Been Tested?

This has been tested by porting and running examples that make use of systems.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.


ezpuzz:
The uncommented examples are now working, continuing to test and fix one-by-one.

amethyst_network has not been updated to new systems syntax yet.
ui and tiles are being worked separately to fix logic to legion system format but I can take reigns on converting them to be compat with this after they are merged.  I could wait for them to be merged first as well.

I think a few more examples can be made working before I request merge.

amethyst_core/src/dispatcher.rs
```diff
@@ -50,19 +62,31 @@ impl DispatcherData {
...
+    items: Vec<DispatcherItem<'static>>,
 }
 
-impl DispatcherBuilder {
+impl<'a> DispatcherBuilder {
     /// Adds a system to the schedule.
-    pub fn add_system<T: ParallelRunnable + 'static>(&mut self, system: T) -> &mut Self {
-        self.items.push(DispatcherItem::System(Box::new(system)));
+    pub fn add_system<S: System<'static> + 'static>(&mut self, system: Box<S>) -> &mut Self {
+        let s: &'static mut S = Box::leak(system);
```
grzi, line 71: That's I think the most important change here. I think that, as systems are meant to live during all the lifetime of the app, it make sense.

CleanCut:
@ezpuzz This is a huge diff. Can you please go into more detail about:

- What this change does
- Why you are making the change
- Pros / Cons of this approach

That would help me immensely in my quest to either review or find a reviewer for this PR. 😉

amethyst_core/src/dispatcher.rs
ezpuzz: > 
> 
> @ezpuzz This is a huge diff. Can you please go into more detail about:
> 
>     * What this change does
> 
>     * Why you are making the change
> 
>     * Pros / Cons of this approach
> 
> 
> That would help me immensely in my quest to either review or find a reviewer for this PR. 😉

@CleanCut I've updated the description.

ezpuzz:
> 
> 
> @ezpuzz This is a huge diff. Can you please go into more detail about:
> 
>     * What this change does
> 
>     * Why you are making the change
> 
>     * Pros / Cons of this approach
> 
> 
> That would help me immensely in my quest to either review or find a reviewer for this PR. 😉

I've updated the description.

ezpuzz:
@CleanCut in response to you mentioning that the diff is huge and to expedite review I've added a note at the top of this PR about using `git diff --minimal legion_v2 -b` to make clear that system logic has not been affected by this change.

ezpuzz:
I've reverted the book changes, they were just regex substitutions for the most part anyway, and our CI is not currently validating the book.

Senre (approved):

ezpuzz (commented):
I've reviewed again all the changes.  I think this is safe to merge.  I will get the atelier prefab branch up to date with the rendy changes once merged.

ezpuzz:
I'm basing future work off of this now, so merging to unblock myself.  It's green and there have been no complaints.

valkum:
This does break the usage of the system macros from legion.
While I am open to convert the systems in tiles to the verbose setup form, I really would like a way to register system created using the macros.

PR #2543, opened 2020-12-14
[LEGION] disable failing examples for meaningful CI

ezpuzz:
## Description

This will allow the CI to pass for the legion_v2 branch and prevent build breaking changes from being merged.

It disables the unported crates (animation, gltf, ui, tiles) and fixes most of the examples that do not depend on them.

It fixes all the uncommented (in Cargo.toml) examples to pass testing.  Some of the pong examples have been ported separately in #2531 I will rebase that PR after this is merged.

In development I discovered that CI does not run the `doc` tests at all.  I don't believe that doc passing is _strictly_ necessary for legion_v2 to be merged but I will make a separate PR to track doc fixes.

I've also formatted Cargo.toml with the [taplo formatter](https://taplo.tamasfe.dev/) which I will separately make a PR for incorporating as a linter in to CI.

## Additions

- Detail API additions here if any
- Do not list changes or removals

## Removals

- List API removals here if any

## Modifications

- List changes to existing structures and functions here if any

## PR Checklist

By placing an x in the boxes I certify that I have:

- [x] Added unit tests for new code added in this PR.
- [x] Acknowledged that by making this pull request I release this code under an MIT/Apache 2.0 dual licensing scheme.
- [x] Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.
- [x] Updated the content of the book if this PR would make the book outdated.

If this modified or created any rs files:

- [x] Ran `cargo +stable fmt --all`
- [x] Ran `cargo clippy --workspace --features "empty"` (may require `cargo clean` before)
- [x] Ran `cargo build --features "empty"`
- [x] Ran `cargo test --workspace --features "empty"`

PR #2544, opened 2020-12-14
[LEGION] remove lifetimes from GameData

ezpuzz:
GameData no longer takes lifetimes with Legion

this sort of change will be easier to verify with #2543


ezpuzz:
this makes the diff for #2541 simpler too

ezpuzz:
this does not break build and has primarily affects docs and book so I am merging to make #2541 easier to review

PR #2548, opened 2020-12-17
Fix Plane::intersect_line function that was returning a distance with incorrect sign

Benlitz:
## Description

The function `Plane::intersect_line` is computing a distance to the plane, but this computation was missing a `-` sign (see [formula](https://www.cs.princeton.edu/courses/archive/fall00/cs426/lectures/raycast/sld017.htm)), so the returned value was of the opposite sign that expected. Additionally, the absolute value of the denominator was tested to be greater than `min_value` which always pass since `min_value` is negative. This check has been replaced with `is_positive()` (fails if the value is zero), and the division has been moved inside the if block to avoid dividing by zero.

## Additions

None

## Removals

None

## Modifications

None

## PR Checklist


jojolepro (commented):
Thanks for this pull request! Do you think it would be possible to add a unit test to make sure that now we are calculating the correct values? Thanks!

Benlitz:
> 
> 
> Thanks for this pull request! Do you think it would be possible to add a unit test to make sure that now we are calculating the correct values? Thanks!

Sure. There are tests already for these functions, but I can add an assert on the sign of the distance.

jojolepro (approved):
Thank you and sorry for the delay!

PR #2351, opened 2020-07-05
Legion animation

BlockCat:
## Description

A bit of work on porting specs to legion.

PR #2549, opened 2020-12-17
Port `amethyst_animation` to Legion

ezpuzz:
## Description

Continuing work from #2351 to get amethyst_animation working with legion ~but I've hit an impasse and my legion skills aren't developed enough yet to pick a best path forward.  The AnimationControlSystem wants to pass a mutable world reference through the `process_animation_control` chain but that makes borrow checker unhappy.  Will need to either pass `Vec<(Entity, EntryMut)>` in place of hierarchy or something else.  Any suggestions?~ Edit: just used a world split and seems alright.


ezpuzz:
vertex skinning and examples and book not fixed yet

ezpuzz:
sprite_animation works now in a code-first way and not prefabs.  we are ironing out the last bit of how prefabs should work still.

amethyst_animation/src/material.rs
```diff
@@ -21,6 +22,15 @@ pub enum MaterialPrimitive {
     Offset((f32, f32), (f32, f32)),
 }
 
+use type_uuid::TypeUuid;
+use uuid::Uuid;
+
+impl TypeUuid for Sampler<MaterialPrimitive> {
+    const UUID: type_uuid::Bytes =
```
valkum, line 29: Can we hide those behind a macro?

amethyst_animation/src/bundle.rs
```diff
@@ -1,53 +1,34 @@
...
+        _world: &mut World,
+        _resources: &mut Resources,
+        _builder: &mut DispatcherBuilder,
     ) -> Result<(), Error> {
-        builder.add(
-            VertexSkinningSystemDesc::default().build(world),
-            "vertex_skinning_system",
-            self.dep,
-        );
+        // FIXME: builder.add_system(VertexSkinningSystem);
```
valkum, line 31: What is missing for this?

amethyst_animation/src/resources.rs
```diff
@@ -42,15 +33,11 @@ pub trait AnimationSampling: Send + Sync + 'static + for<'b> ApplyData<'b> {
         &mut self,
         channel: &Self::Channel,
         data: &Self::Primitive,
-        extra: &<Self as ApplyData<'a>>::ApplyData,
+        buffer: &mut CommandBuffer,
```
valkum, line 36: Should we enforce the same naming schema for all CommandBuffers in fn arguments inside of amethyst?

amethyst_animation/src/skinning/systems.rs
```diff
@@ -16,16 +14,12 @@ use super::resources::*;
 /// System for performing vertex skinning.
 ///
 /// Needs to run after global transforms have been updated for the current frame.
-#[derive(Debug, SystemDesc)]
-#[system_desc(name(VertexSkinningSystemDesc))]
+#[derive(Debug)]
 pub struct VertexSkinningSystem {
```
valkum, line 18: This needs migration to legion as well, right? (The system by itself)

amethyst_animation/src/systems/control.rs
```diff
@@ -58,207 +26,186 @@ where
...
-                    if let Some(dur) = self.state_set.get(&deferred_animation.relation.0) {
-                        if *dur < 0. {
-                            (false, 0.)
-                        } else if let DeferStartRelation::Start(start_dur) =
-                            deferred_animation.relation.1
-                        {
-                            let remain_dur = dur - start_dur;
-                            (remain_dur >= 0., remain_dur)
+
+                let mut deferred_start = Vec::default();
```
valkum, line 133: This would allocate each time the system is run, whereas the spec version would only clear the already allocated vec. Should we keep the spec behavior? Does something prevent us from doing it the old way?

amethyst_animation/src/systems/control.rs
```diff
@@ -392,51 +321,55 @@ where
...
-            if check_termination(control.id, hierarchy, &samplers) {
+            if check_termination(control.id, hierarchy, world) {
+                debug!("Animation finished, terminating.");
                 // Do termination
                 for node_entity in hierarchy.nodes.values() {
-                    let empty = samplers
-                        .get_mut(*node_entity)
+                    let empty = world
+                        .entry_mut(entity)
+                        .unwrap()
```
valkum, line 358: I remember we had one time a overhaul to remove all unwraps. Either by replacing them by expects or something more sane. Should we replace this `unwrap()` with a `expect("Failed to resolve entity")` or something along the lines?

amethyst_animation/src/systems/control.rs
```diff
@@ -631,32 +577,37 @@ fn update_animation_rate<T>(
...
     // Check for termination
-    if check_termination(control_id, hierarchy, &samplers) {
+    if check_termination(control_id, hierarchy, world) {
         // Do termination
         for node_entity in hierarchy.nodes.values() {
-            let empty = samplers
-                .get_mut(*node_entity)
+            let empty = world
+                .entry_mut(*node_entity)
+                .unwrap()
```
valkum, line 592: See above.

amethyst_animation/src/systems/control.rs
```diff
@@ -667,14 +618,17 @@ where
...
-        .nodes
-        .iter()
-        .flat_map(|(_, node_entity)| samplers.get(*node_entity))
-        .all(|s| s.check_termination(control_id))
+    hierarchy.nodes.iter().all(|(_, node_entity)| {
+        world
+            .entry_ref(*node_entity)
+            .unwrap()
+            .get_component::<SamplerControlSet<T>>()
+            .unwrap()
```
valkum, line 631: See above

examples/sprite_animation/assets/anims/bat.ron
```diff
@@ -0,0 +1,13 @@
...
+                                            // Time of key frames
+                                            input: [0.0, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1.1, 1.2, 1.3, 1.4],
+                                            // Sprite indexes from SpriteSheet for key frames
+                                            output: [5, 4, 3, 2, 1, 0, 1, 2, 3, 4, 4],
+                                            // Sprites can only ever be animated with Step
+                                        function: Step,
+                                        
+
+)
+}
```
valkum, line 13: newline

examples/sprite_animation/main.rs
```diff
@@ -34,66 +34,99 @@ enum AnimationId {
...
+            anim_set.insert(AnimationId::Fly, anim_handle);
+
+            world.push((SpriteRender::new(sheet, 0), Transform::default(), anim_set));
+        }
         // Creates new entities with components from MyPrefabData
-        world.create_entity().with(bat_prefab).build();
-        world.create_entity().with(arrow_test_prefab).build();
+        // world.create_entity().with(bat_prefab).build();
+        // world.create_entity().with(arrow_test_prefab).build();
+
```
valkum, line 111: I guess this ok for now. Do we want to revert his when asset loading is fully working?

examples/sprite_animation/main.rs
```diff
@@ -102,31 +135,28 @@ impl SimpleState for Example {
...
         Trans::None
     }
 }
 
-fn initialise_camera(world: &mut World) {
+fn initialise_camera(world: &mut World, resources: &mut Resources) {
     let (width, height) = {
-        let dim = world.read_resource::<ScreenDimensions>();
+        let dim = resources.get::<ScreenDimensions>().unwrap();
```
valkum, line 152: Probably use an except here. Helps new users if they copy code from the examples.

examples/sprite_animation/main.rs
```diff
@@ -136,20 +166,11 @@ fn main() -> amethyst::Result<()> {
...
-        ))?
-        .add_bundle(
-            TransformBundle::new()
-                .with_dep(&["sprite_animation_control", "sprite_sampler_interpolation"]),
-        )?
+    let mut game_data = DispatcherBuilder::default();
+    game_data
+        .add_bundle(LoaderBundle)
+        .add_bundle(AnimationBundle::<AnimationId, SpriteRender>::default())
+        .add_bundle(TransformBundle::default())
```
valkum, line 173: As we use the `CommandBuffer` in the animation systems, do we get a 1 frame delay here?
The cmd buffer is not flushed after each system run, right? So we need to manually invoke a flush before the Transform or Rendering Bundle.

valkum (commented):
Overall nice work.

amethyst_animation/src/material.rs
ezpuzz: unfortunately TypeUuid doesn't support generic types yet, so this is a workaround: https://github.com/randomPoison/type-uuid/issues/4

amethyst_animation/src/bundle.rs
ezpuzz: I think i've converted it now, but we won't be able to test until the gltf example is working.

amethyst_animation/src/resources.rs
ezpuzz: I only see one instance of it being named differently, in the removal util.  I think it would be good to stay consistent but I don't know of a linting tool to enforce it.  Do you have something in mind?

amethyst_animation/src/skinning/systems.rs
ezpuzz: Migrated.

amethyst_animation/src/systems/control.rs
ezpuzz: changed back to old behavior

amethyst_animation/src/systems/control.rs
ezpuzz: i've replaced all in animation with expects

examples/sprite_animation/main.rs
ezpuzz: I'm working on prefabs later, plus there is prior conversation throughout the issues of making the examples rely less on prefabs.  The fastest way to get there is to port examples away from prefabs and possibly convert them back later.  We should rely on user feedback as to whether or not they need prefabs for a given example.

This example works in this state to demonstrate an animated sprite so I'm happy with it as is.  I think examples like this one should be focused on demonstrating the programmatic API rather than the magic of prefabs.

I have some work in progress towards a big prefab example that demos a bunch of different prefabs and swapping between them.  With the legion-prefab approach you are essentially just serializing the component directly.

examples/sprite_animation/main.rs
ezpuzz: added a flush for safety, although maybe should be inside AnimationBundle.  let's see what people report.

we can add some slow down functionality in the example in the future to demonstrate features of animation or something like that which will let us debug further if there is an issue.

examples/sprite_animation/assets/anims/bat.ron
ezpuzz: we need a ron formatter, but the existing one doesn't support comments.  I will see if I can make a PR to their project

ezpuzz:
this is blocking gltf, animation example, and my debug-lines example fix so I am going to merge now that i've fixed @valkum's concerns

PR #2550, opened 2020-12-18
[LEGION] fix: Add a component-conditional clause on the parent/children transform system

grzi:
## Description

The transform system was wrongly thinking that a Parent always have a Transform component by unwraping it without any check.
That's a wrong statement as the Parent/Children Hierarchy can be used without Transform or with a UiTransform component.

There may be a better way to do that check but we need that global_matrix value so ..

## Additions

- Condition

## PR Checklist


ezpuzz:
I'm assuming this is blocking your ui porting as well.  I don't see how this would hurt anything.

PR #2559, opened 2020-12-23
feat: Add a track of Modifiers into the Input Handler

grzi:
## Description

Following the recent winit upgrade that introduce the new 'ModifiersChanged' event, I needed it in different places for UI. 
I think it will be useful to keep a track of the current modifiers state. Moreover, it will help to detect key combination with these modifiers easily (without the need to do 'shift left is down OR shift right is down AND my other key is down) 

## Additions

- KeyboarModifiersState
- modifiers on InputHandler

## PR Checklist

### DIFF
--- a/amethyst_input/src/input_handler.rs
+++ b/amethyst_input/src/input_handler.rs
@@ -17,6 +17,37 @@ use super::{
     *,
 };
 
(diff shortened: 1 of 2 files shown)


AlveLarsson (approved):
Looks good as always ;D

PR #2560, opened 2020-12-23
pausable systems

ezpuzz:
- Initial impl of Pausable legion system, does not work with
- Remove SystemExt trait and replace with pauseable method, update docs, minor style/compiler error fixes.
- Formatted code
- Fix clippy warning in system_ext.rs
- Remove SystemExt ref in docs of Pauseable struct
- Fix typo 'pausable' -> 'pauseable' in Pauseable docs
- remove duplicate type


ezpuzz:
Is this useable with DispatcherBuilder?  It seems all the tests are with legion directly.

grzi:
Oh, you're right, forgot about that, I'll see

grzi:
Done.

PR #2565, opened 2020-12-29
Fix bug with frustum culling when using user-defined BoundingSpheres

Valarissa:
## Description

There was an issue with frustum culling where the camera transform was transforming the
points of the center of bounding spheres that should have remained stationary.

This addresses that by only transforming the default point, which uses the origin.

## Additions

- No additions

## Removals

- No removals

## Modifications

- Modifies the frustum culling in `amethyst_rendy::visibility::VisibilitySortingSystem`

## PR Checklist

By placing an x in the boxes I certify that I have:

- [x] Added unit tests for new code added in this PR.
- [x] Acknowledged that by making this pull request I release this code under an MIT/Apache 2.0 dual licensing scheme.
- [x] Added a changelog entry if this will impact users, or modified more than 5 lines of Rust that wasn't a doc comment.
- [x] Updated the content of the book if this PR would make the book outdated.

If this modified or created any rs files:

- [x] Ran `cargo +stable fmt --all`
- [x] Ran `cargo clippy --workspace --features "empty"` (may require `cargo clean` before)
- [x] Ran `cargo build --features "empty"`
- [x] Ran `cargo test --workspace --features "empty"`

PR #2568, opened 2020-12-31
Fix frustum culling for user defined BoundingSpheres

Valarissa:
## Description

There was an issue with frustum culling where the camera transform was transforming the
points of the center of bounding spheres that should have remained stationary.

This addresses that by only transforming the default point, which uses the origin.

Supersedes: #2565 

## Additions

- No additions

## Removals

- No removals

## Modifications

- Modifies the frustum culling in `amethyst_rendy::visibility::VisibilitySortingSystem`

## PR Checklist

### DIFF
--- a/amethyst_rendy/src/visibility.rs
+++ b/amethyst_rendy/src/visibility.rs
@@ -137,12 +137,14 @@ impl System<'static> for VisibilitySortingSystem {
                             entity_query
                                 .iter(world)
                                 .map(|(entity, transform, transparent, sphere)| {
-                                    let pos = sphere.clone().map_or(origin, |s| s.center);
                                     let matrix = transform.global_matrix();
+                                    let pos = sphere
+                                        .clone()
+                                        .map_or(matrix.transform_point(&origin), |s| s.center);
                                     (
                                         *entity,
                                         transparent.is_some(),
-                                        matrix.transform_point(&pos),
+                                        pos,
                                         sphere.map_or(1.0, |s| s.radius)
                                             * matrix[(0, 0)]
                                                 .max(matrix[(1, 1)])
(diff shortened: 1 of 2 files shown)


valkum:
Wouldn't that mean that you need to define your bounding boxes in worldspace? If you apply movement only to the Transform, the position of the bounding box would not by synchronized in that case, or do i understand something wrong?

Valarissa:
Shouldn't the position of the bounding boxes be modified upon the update of the Transform for an entity though? I'm using user defined bounding boxes for chunks in a voxel engine, so they're static. Generally though, I would assume that the bounding boxes would be updated to have the correct center when the entity's Transform itself moves, which is the purview of the user. If you don't update the bounding box's center it can't be used for collision detection or other situations.

The problem is, if you defined the bounding boxes with the correct position as their center, the center for visibility sorting is pushed twice as far from the origin making it so that the frustum culling reacts oddly. If you want to apply that Transform to everything, it might be worthwhile to remove the `center` attribute for the BoundingSpheres, and just assume that the Transform will track where the bounding box is... but that doesn't seem great to me.

valkum:
The bounding sphere center should be set local to the model, so that when the model is not formed uniformly or centered correctly the bounding sphere can be set to include only a minimal part of space that is not part of the model.
There should be no real difference between applying the global_matrix in a system or during sorting. Maybe with some optimizations but in general there should be no difference.

Your usecase should be covered by the current behavior. Maybe the bounding sphere is not large enough? If your model is a cube with the center at (0,0,0) the sphere center should be (0,0,0) as well.

Valarissa:
If that's the case, we should probably make some adjustments to the docs surrounding the BoundingSphere. :)

Issue #2517, opened 2020-11-15
Legion: Final Tracker


AlveLarsson:
Hello! I've been out of the loop for the past months due to some life issues, but now I'm back.

Progress on the Legion port has slowed to a standstill.  I'm creating this issue to track what's been done, what still needs to be done, and what our roadblocks are. This issue will act as a new, clean slate to minimize confusion and get us back on track.

_Edit this comment to change the lists below, or comment below and someone with write access will update the lists for you._

#### Current Progress with Names & Links
- [Project Board](https://github.com/amethyst/amethyst/projects/21)
- [Release Checklist](https://github.com/amethyst/amethyst/issues/2308)
- [x] amethyst parent crate #2320
- [ ] examples #2483
- [ ] book
- [ ] inline documentation (please do it _with_ your subcrate ports!)
- [ ] all subcrates #2400 #2417 #2436
  - [x] amethyst_animation #2334 #2351
  - [x] amethyst_assets @jlowry #2329 #2404 #2520
  - [x] amethyst_audio #2315
  - [x] amethyst_config
  - [x] amethyst_controls #2325 #2378
  - [x] amethyst_core @LechintanTudor #2502 #2560
  - [x] amethyst_derive
  - [x] amethyst_error
  - [x] amethyst_gltf #2335 #2631
  - [x] amethyst_input #2332 #2481
  - [x] amethyst_locale #2331 #2336
  - [x] amethyst_network #2319
  - [x] amethyst_rendy #2318 #2522 
  - [ ] amethyst_test
  - [x] amethyst_tiles @valkum #2555
  - [x] amethyst_ui @grzi  #2330 #2556 
  - [x] amethyst_utils #2350
  - [x] amethyst_window

#### Who is willing to help? What are you currently working on?
- @AlveLarsson - tracking, talking to folks to gather updates
- @CleanCut - help with tracking, try to make `amethyst_rendy` less of a pain, maybe `examples`
- @jlowry - `amethyst_assets`
- @valkum - `amethyst_tiles`
- @LechintanTudor - `amethyst_core`, `amethyst_ui`

Lastly, we need to communicate frequent updates about what's being done or rather the rate of progress that's happening on each crate as we start work.

**Important!** Please discuss the legion port on GitHub issues (like this one, or more specific ones), not on discord or the forum. We need to keep the discussion in one place to keep things sane.

(Special thanks to @CleanCut and @jlowry for catch-ups)

valkum:
[...]

PR #2579, opened 2021-01-06
give examples their own cargo.toml

ezpuzz:
## Description

examples previously shared dev dependencies and paths with the main
amethyst crate.  this lead to bad import paths in examples, additional
joins needed on paths that made examples not easily portable. also it
made it difficult to copy examples as a starter project.

now that each example has a cargo.toml with dependencies we have a best
practice to start a new project packaged alongside each example and
users can copy an example folder in to their own new project with ease
and without editing anything.

## Motivation and Context

This makes it easier for new users to get started.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -134,12 +134,6 @@ legion = { version = "0.3", default-features = false, features = [
 
 [dev-dependencies]
 rusty-hook = { git = "https://git@github.com/swellaby/rusty-hook", branch = "master" }
-derive-new = "0.5.8"
-ron = "0.6.4"
-type-uuid = "0.1"
-serde-diff = "0.4"
-legion-prefab = { git = "https://github.com/amethyst/prefab", rev = "7c30249f106e6177549e223ca2823eec8ab6c70b" }
-prefab-format = { git = "https://github.com/amethyst/prefab", rev = "7c30249f106e6177549e223ca2823eec8ab6c70b" }
 
 [build-dependencies]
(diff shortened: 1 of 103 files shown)


Cargo.toml
```diff
@@ -67,8 +64,35 @@ experimental-spirv-reflection = ["amethyst_rendy/experimental-spirv-reflection"]
...
-members = ["amethyst_*"]
-exclude = ["amethyst_gltf", "amethyst_animation", "amethyst_test"]
+members = ["examples/*", "amethyst_*"]
+exclude = [
+    "amethyst_gltf",
+    "amethyst_animation",
+    "amethyst_test",
+    "examples/Cargo.toml",
+    "examples/_unused_assets",
+    "examples/animation",
```
ezpuzz, line 74: these are the remaining examples to be ported to legion, see #2517

examples/examples.sh
```diff
@@ -0,0 +1,26 @@
+for folder in */; do
```
ezpuzz, line 1: this simple script generates the Cargo.toml file for each example to make it easy to add new examples or modify the template.  we will probably re-export legion-prefab and prefab-format from amethyst in the future but there are macros that have paths to them now and code would get messy to include re-exports in the examples.

this dependencies list is a union of all examples right now.  users can audit it themselves or we could add a `cargo udeps` run to the end.  I think it's alright to have some examples with excess dependencies for now.

PR #2581, opened 2021-01-06
fix tiles example assets

ezpuzz:
## Description

the assets in the tiles example were no longer valid after merging updated atelier code

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.


Benlitz (approved):
I confirm this fixes the issue and this example runs fine now. Thanks

valkum:
I am having trouble getting this to run.
If my current dir is the amethyst root dir, application_root_dir() (because it uses the `CARGO_MANIFEST_DIR` env var) points to the amethyst root dir, and then the display config path is wrong. With a fixed display_path it compiles, but nothing is drawn.
If I change into the tiles example dir, everything gets recompiled (I assume caused by the different set features) but i cannot specify vulkan as a feature to get it to run.

ezpuzz:
> 
> 
> I am having trouble getting this to run.
> If my current dir is the amethyst root dir, application_root_dir() (because it uses the `CARGO_MANIFEST_DIR` env var) points to the amethyst root dir, and then the display config path is wrong. With a fixed display_path it compiles, but nothing is drawn.
> If I change into the tiles example dir, everything gets recompiled (I assume caused by the different set features) but i cannot specify vulkan as a feature to get it to run.
see #2579 docs are now updated to show that examples can be run with `cargo run -p example_name`  in this case `cargo run -p tiles`

PR #2588, opened 2021-01-07
Improve issue and pr templates

ezpuzz:
## Description

This will make the issue and pull request templates more friendly and faster/easier to use.

## Motivation and Context

Previously the issue templates required deleting existing text to enter your own which was a waste of time and resulted in reading the default text frequently when looking at filed issues.

The `status: needs triage` label on every issue served no purpose and issues that had been triaged never had the label removed.

Nobody had ever filed and `type: modification` issue in the history of the project as I could tell.

The checkboxes for PR type were redundant and we should use the GitHub label system as it works in tandem with our release drafter github action to create a nice changelog.

The templates now breathe with whitespace so you can just click and start typing.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,40 +1,43 @@
 ---
 name: Bug Report
 about: Found a bug? Create a report and help us improve!
-title: "[BUG] Brief Description Here"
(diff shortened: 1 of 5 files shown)


ezpuzz:
merging as this purely helps the community and appearance of project on github (no more "needs triage" issues)

PR #2597, opened 2021-01-11
Add test to amethyst audio bundle and fix code

grzi:
## Description

I added a test to guarantee that adding this bundle won't crash. I also had to modify a bit the system as it seems that we can't receive optional resources. 

## Motivation and Context

I was trying to implement an example based on Audio when I saw that there weren't any test on audio. Moreover, adding the Audio bundle was making my projects crash. 

## How Has This Been Tested?
I added an unit test 

## Checklist:

- [x] My code is used in an example. (well it will be soon but I want to validate this first)

### DIFF
--- a/amethyst_audio/src/bundle.rs
+++ b/amethyst_audio/src/bundle.rs
@@ -4,24 +4,46 @@
 use amethyst_core::ecs::*;
 use amethyst_error::Error;
 
-use crate::{output::Output, systems::*};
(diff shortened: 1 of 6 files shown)


grzi:
Mmmmh, should have thought that container don't have audio output. Need to find a solution for this.

ezpuzz (approved):

PR #2601, opened 2021-01-14
Add camera culling for tilemaps

### LINKED ISSUE #2592: Culling tilemaps for better rendering performances
## Feature

Tilemap rendering

## Change

Cull the tilemaps at rendering according to the active camera, using a specific `DrawTiles2DBounds` implementation, so that the render pass does not iterate over all tiles of the map.

## Reason

Tilemaps are currently not culled when being rendered, making them very slow for large maps (say, 1000x1000 tiles). Culling allows to mitigate the problem.

## Additional Notes

I already have an implementation of this [in this branch](https://github.com/amethyst/amethyst/compare/master...Benlitz:tilemap-camera-culling). I can clean it up and open a PR if this idea is generally approved in this ticket (the meaningful part is just the large chunk of code in `pass.rs` if anyone wants to pre-review). There are a few caveats/things to know:
- this implementation is general, casting 4 rays at the four corners of the camera to find which tiles are going to be rendered there, and uses these tiles to build a bounding `Region`. It works for all cameras (ortho/persp) with all transforms (at least I think so :) I tested quite intensively). More efficient culling can be achieved with more constraints on the camera itself, but this approach already drastically improves performances
- it requires an `ActiveCamera` to be set. From what I've seen, this is not automatically set when the first camera is created and requires to do something like `world.fetch_mut::<ActiveCamera>().entity = Some(camera);` in the user code. Because of this, this `DrawTiles2DBounds` isn't the default one. Automatically assigning the active camera on the engine side could be considered.
- if there is no default `ActiveCamera`, the returned region is empty. This makes it easier to track a "missing active camera" problem, rather to falling back to the default method of rendering everything which would hide the problem but still have the same poor performances on large map. Other options would be panicking or at least outputing log, but I think disabling the active camera on purpose is something normal one user might want to do.
- the culling itself isn't ideal because the render pass is using `Region` which are axis-aligned (in tile space) rectangles (for example in case of a 45° rotated square tilemap, 2x more tiles than needed will be still rendered). Something to keep in mind for future improvement maybe, but this implem is still "good enough" I believe.
- there's still an issue with the fact that tiles are arranged using the Morton encoder (by default), while they are iterated linearly in the render pass (because iterating a region [is linear](https://github.com/amethyst/amethyst/blob/6d8d4c495fbb78bd3ec4b132dd38d899f19d3d3c/amethyst_tiles/src/iters.rs#L181-L203)), which defeat the purpose of using something like morton for tiles - I actually have better performances using a linear encoder instead, although I didn't precisely measure that. Initializing and filling the storage with morton is particularly slow to my observation. This last point is slightly off-topic but it might be worth re-evaluating the value of the morton encoder in a different ticket maybe.

ezpuzz:
## Description

fixes #2592 

## Checklist:

### DIFF
--- a/amethyst_assets/src/prefab/component_registry.rs
+++ b/amethyst_assets/src/prefab/component_registry.rs
@@ -63,13 +63,13 @@ impl ComponentRegistryBuilder {
         FromT: Component,
         IntoT: Component,
         F: Fn(
-            &Resources,                                 // resources
-            Range<usize>,                               // src_entity_range
-            &Archetype,                                 // src_arch
-            &Components,                                // src_components
-            &mut ComponentWriter<'_, IntoT>,            // dst
-            fn(&mut ComponentWriter<'_, IntoT>, IntoT), // push_fn
-        ) + Send
+                &Resources,                                 // resources
+                Range<usize>,                               // src_entity_range
+                &Archetype,                                 // src_arch
+                &Components,                                // src_components
+                &mut ComponentWriter<'_, IntoT>,            // dst
+                fn(&mut ComponentWriter<'_, IntoT>, IntoT), // push_fn
+            ) + Send
             + Sync
             + 'static,
     {
--- a/amethyst_tiles/src/iters.rs
+++ b/amethyst_tiles/src/iters.rs
@@ -76,8 +76,8 @@ impl<'a> From<&'a Region> for MortonRegion {
 }
 
 /// Axis aligned quantized region of space represented in tile coordinates of `u32`. This behaves
-/// like a bounding box volume with `min` and `max` coordinates for iteration. This regions limits are *inclusive*,
-/// in that it considers both min and max values as being inside the region.
+/// like a bounding box volume with `min` and `max` coordinates for iteration. The lower (min) coordinates
+/// are inclusive and the upper (max) coordinates are exclusive.
 #[derive(Copy, Clone, Debug, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
 pub struct Region {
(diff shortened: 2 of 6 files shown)


Benlitz:
This is more or less ready for review, but I cannot edit the PR description because I don't own it.

Description/motivation/context is described in the linked issue (https://github.com/amethyst/amethyst/issues/2592)

**How has this been tested?**

I tested a lot of cases with the both the tilemap and the camera at various locations and orientations. I made sure it works fine even with a perspective camera. I did all my tests in the previous specs implementation, I just migrated to legion but everything still seems to work fine, perf improvement is still there and the maths were not affected by the change anyway
I haven't updated the book yet (depends if there's a decision to make this the default mode or not...). I haven't written tests specifically for this, it's not trivial to write test for this (need to build a mock GraphAuxData)
A good way to see the performance increase is to modify `examples\tiles\main.rs` to create a tilemap with a size of eg. `Vector3::new(1000, 1000, 1),` and run this example. Without this change, the tilemap will be extremely slow, but it works at normal speed with it (note: initialization is still slow due to Morton encoder, but that's not related to this PR).

Benlitz:
Actually hold on for now, I'm trying to implement tests and I found some issue with the `Region` struct that I might want to fix first

Benlitz:
Ok, this should be good now!
So additionally, I modified `Region` to exclude the max value because otherwise it's impossible to have an empty region (eg. `Region::empty()` doesn't return an actual empty region). When culling this is annoying because if a tilemap is out of sight it may still render a tile or a row of tiles.
I also included tests with various transformations

ezpuzz:
Great work @Benlitz !

PR #2609, opened 2021-01-19
replace thread_profiler

### LINKED ISSUE #2593: Flamegraph Profiling
## Description

A way to generate flamegraphs for my game.

## What problem does this solve? What need does it fill?

When I'm making a game I want to know why performance is bad so I can fix it.

## Describe the solution you'd like

I would like to be able to enable a feature flag on amethyst and immediately get flamegraphs of all internal systems and all of my systems that implement the `System` trait to see which are slow.  Also my state update and handle_event calls.

## Describe alternatives you've considered

Using the Stopwatch from amethyst_core to print timings.

## Additional context

https://crates.io/crates/flame
https://crates.io/crates/flamer
https://github.com/jonhoo/inferno
https://github.com/flamegraph-rs/flamegraph

ezpuzz:
## Description

This is a start at using a different profiling mechanism.  thread_profiler is not really suitable for legion and takes a lot of naming and macro fidgeting.  I will be trying out different solutions on this branch.

closes #2593 hopefully

## Checklist:


ezpuzz:
early investigation notes:

flame/flamer has some potential, ideally we can surround the system closure somehow and get the name from the systembuilder or the struct.  it does make me regret a little going all in on the SystemBuilder approach rather than legion macro because flaming an fn is easier, but maybe there's some way around it.

jojolepro:
Hi! I'm super interested by the work you are doing here. Any news?

PR #2610, opened 2021-01-19
Prefabs with modified dependencies

jlowry:
## Checklist:


ezpuzz:
I noticed this bug too, I was having issues in the prefab example with the root entity not being preserved.  I believe I fixed it with similar code to this in #2611 can you review that PR?

ezpuzz:
implemented your fix in #2611

PR #2611, opened 2021-01-19
Fix `auto_fov` example

ezpuzz:
## Description

This uses working UI prefabs to get the auto_fov example working again.

I did not use prefabs for the graphics in the example because it would have made the diff much larger.

I used prefabs for UI because it would've been repetitive code otherwise.

I had to fix the prefab spawning because I didn't implement it right in #2567 

## Checklist:


ezpuzz:
closes #2610

PR #2614, opened 2021-01-20
test inline documentation in CI

ezpuzz:
## Description

inline docs are passing with this pr, this will enable `cargo test --doc` in CI and enable mdbook and linkcheck as well.

fixes docs highlighting in vscode as well and formats inline examples.

takes a first stab at updating some of the book copy as well, but that will be a page-by-page endeavor.

have a working local method of testing book, but checking how CI runs with this PR.  may disable the mdbook check if failing and do separate PR for it.

## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,5 +78,5 @@ jobs:
       - run: mdbook build book
         if: matrix.toolchain == 'stable' && matrix.os == 'ubuntu-latest'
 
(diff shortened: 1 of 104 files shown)


ezpuzz:
yep, mdbook test doesn't even work without the extern crate statements and those are a mess.  I found a way to run rustdoc directly but need to figure out how to get that in CI.  Will open a separate PR because inline docs are just as important.

PR #2620, opened 2021-01-27
Fix FpsCounterBundle not inserting FpsCounter

LokiVKlokeNaAndoke:
## Description

Insert `FpsCounter` resource in `FpsCounterBundle`.

## Motivation and Context

The docs on `FpsCounterBundle` say:
> Automatically adds a FpsCounterSystem __and a FpsCounter resource with the specified sample size__.

(emphasis mine)

This pull request makes the implementation comply with the docs.

## How Has This Been Tested?

```
cargo test --workspace --all-features
```

## Checklist:


ezpuzz:
@LokiVKlokeNaAndoke could you make the rustdoc comment more clear as to what are reasonable values for the samplesize parameter?  I'm not sure what it should be.

ezpuzz (changes requested):
needs more docs

LokiVKlokeNaAndoke:
Neither do I. I just added `new` and `default` to `FpsCounterBundle` just like in `FpsCounter`, and they basically do the same thing. I thought about injecting `FpsCounter` into `FpsCounterBundle`, but it would make the interface too complex.

What about using the builder pattern for this? It would probably make things less confusing.
```rust
///Automatically adds a `FpsCounterSystem` and a [`FpsCounter`] resource with the specified sample size.
#[derive(Default, Debug)]
pub struct FpsCounterBundle {
    samplesize: Option<usize>,
}

impl FpsCounterBundle {
    /// Set the sample size the [`FpsCounter`] uses. The default is 20.
    pub fn sample_size(self, samplesize: usize) -> Self {
        Self {
            samplesize: Some(samplesize),
        }
    }
}
```

instead of

```rust
///Automatically adds a `FpsCounterSystem` and a [`FpsCounter`] resource with the specified sample size.
#[derive(Default, Debug)]
pub struct FpsCounterBundle {
    samplesize: Option<usize>,
}

impl FpsCounterBundle {
    /// Create a new [`FpsCounterBundle`] with the specified sample size. The default is 20, see [`FpsCounter`].
    pub fn new(samplesize: usize) -> Self {
        Self {
            samplesize: Some(samplesize),
        }
    }
}
```

ezpuzz (approved):

PR #2638, opened 2021-02-23
Fix CI for audio 

### LINKED ISSUE #2637: Audio Unit tests don't seem to work anymore on CI
## Description

Audio unit tests are failing since runner are on ubuntu 20.04 instead of 18.04
## Reproduction Steps

1. Launch the CI to reproduce

grzi:
A quick fix to enable CI to work against main branch

to fix #2637

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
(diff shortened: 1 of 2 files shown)


grzi:
The fix is actually to set 18.04 but I want to reproduce in 20.04 first and then put it to 18.04 to validate it's ok

PR #2650, opened 2021-03-02
feat: Add shape prefab with distill / legion

jerthz:
## Description

In order to be able to add Shape prefab (as before), it was needed to implement this. 

So with this PR, we can now add a component "ShapePrefab" to en entity in a prefab. 
A thread local fn will then read this and translate this into an Handle<Mesh> using the default loader and the data provided in the prefab.

## Motivation and Context
Shape Prefabs were a thing in 0.15.3, they must be back in 0.16.

The plan is to do the same kind of thing for Material (and I would also like Obj) so we can have prefabs with complete Meshes. 

## How Has This Been Tested?
In the main prefab example.

## Checklist:


dawnlarsson (approved):

amethyst_rendy/src/prefab.rs
```diff
@@ -0,0 +1,112 @@
...
+    where
+        A: de::SeqAccess<'de>,
+    {
+        unimplemented!()
+    }
+}
+
+impl Default for ShapePrefab {
+    fn default() -> Self {
+        unimplemented!()
```
jojolepro, line 63: Should this and the other unimplemented be implemented? This looks like it will cause quite a few errors to users thinking that the type can be created with ::default.

AnneKitsune (changes requested):

PR #2659, opened 2021-03-14
Send `ValueChange` UI event when text is deleted

hampuslidin:
## Description

This PR adds functionality for firing UI events with the type `ValueChange` whenever text is modified by deletion of pressing keys Back or Delete.

## Motivation and Context

I found that the editable text lacks this functionality, since deletion of text should also be considered a change in value.

## How Has This Been Tested?

I have tested the event firing against a private project, to see that the event is fired when text is deleted. I am not sure if there is more extensible testing needed in this regard, since I couldn't find any prior tests for this kind of functionality.

## Screenshots (if appropriate):
N/A

## Checklist:


AnneKitsune (approved):

jojolepro:
Thanks! :hearts:

PR #2669, opened 2021-04-14
Fix "panic message is not a string literal"

InterruptSpeed:
## Description

cargo build was warning "this is no longer accepted in Rust 2021" on the format! used within the debug_assert!

modified lines to use a format string and arguments,without needing to use format!

## Motivation and Context

this very minor pull request prevents 2 warnings on build. it was the most obvious way that I could try and be involved.

## How Has This Been Tested?

cargo build before the change output 2 warnings
cargo build after the change does not output any warnings.

## Checklist:


AlveLarsson:
Thanks for the PR! this issue fix got merged in #2676

PR #2675, opened 2021-05-02
Implement "spin-sleep" strategy for the frame rate limiter sleep

### LINKED ISSUE #2083: [BUG] FrameLimiter Sleep strategy sleeps too long

saveriomiroddi:
## Description

Replace the thread::sleep() invocation with the `spin-sleep` crate API, which, on Windows, yields very good accuracy at virtually no cost.

This is though, a complex subject, so, for more information, see the references added to the documentation.

Adds the [`spin-sleep`](https://github.com/alexheretic/spin-sleep) dependency.

## Motivation and Context

The current frame limiter sleep strategy is inaccurate, while the yield one is excessively resource-intensive.

Closes #2083.

## How Has This Been Tested?

Tested by running the `fly_camera` example, although, for the experimental data, the new strategy has been tested on multiple O/Ss and projects. See the referenced issue.

## Checklist:

- [x] My code follows the code style of this project. Hopefully!
- [x] ~If my change required a change to the documentation I have updated the documentation accordingly.~ Doesn't apply
- [x] ~I have updated the content of the book if this PR would make the book outdated.~ Doesn't apply.
- [ ] I have added tests to cover my changes. **This point requires review** - I'm not aware of any realistic test strategy that can distinguish the change from the original strategy; also note that the frame limiter does not have a test suite.
- [x] ~My code is used in an example.~ Doesn't apply.

## Further considerations

I've done the change at the best of my capacity, however, this area as a whole should be at the very least tested on other machines, in order to be considered solid (IMO).  
In particular, I've tested multiple projects on Linux, and the performance on such O/S was terrible. With my setup only, I can't prove whether Amethyst just performs poorly on Linux, or if it's a problem with my setup.  
Additionally, the setup I've used for experimentation may not be representative of the typical one (see referenced issue).

Having said that, the PR has been very carefully researched an thought out. I only think that this area (the frame limiter) should receive attention to be very solid.

In particular, the new strategy makes the Yield and SleepYield strategy likely obsolete (but again, I can't prove this using my setup only).

Coffee for the masses! S.

### DIFF
--- a/amethyst_core/Cargo.toml
+++ b/amethyst_core/Cargo.toml
@@ -36,6 +36,7 @@ rayon = "1.5"
 shrev = "1.1.1"
 simba = { version = "0.3" }
 smallvec = "1.4"
+spin_sleep = "1.0.0"
 thread_profiler = { version = "0.3", optional = true }
 serde-diff = "0.4"
 
--- a/amethyst_core/src/frame_limiter.rs
+++ b/amethyst_core/src/frame_limiter.rs
@@ -44,8 +44,8 @@
 //!   passed. This will result in the most accurate frame timings, but effectively guarantees
 //!   that one CPU core will be fully utilized during the frame's idle time.
 //! * `Sleep` will sleep for the approximate remainder of the frame duration. This will result in
-//!   lower CPU usage while the game is idle, but risks fluctuations in frame timing if the
-//!   operating system doesn't wake the game until after the frame should have started.
+//!   lower CPU usage while the game is idle; this is the best compromise between resources
+//!   occupation and accuracy.
 //! * `SleepAndYield` will sleep until there's only a small amount of time left in the frame,
 //!   and then will yield until the next frame starts. This approach attempts to get the
 //!   consistent frame timings of yielding, while reducing CPU usage compared to the yield-only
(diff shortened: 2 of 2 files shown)


AlveLarsson:
Thanks for the PR! Sorry for the late response!
we are going to discuss this more in the future but this is a good fix, for now, thanks!

PR #2676, opened 2021-05-16
removed unnecessary format! in debug_assert!

Tobias418:
damn warnings kept spamming my output so i fixed it


AlveLarsson (approved):
Thanks for the PR!

PR #2677, opened 2021-05-18
Fixed missed line in Book, beginner pong example forgot to include a 'pub use'

cjglo:
In the "pong tutorial" section of the Amethyst Book, there is a part where two systems are added to project, but the book forgets to mention to expose these systems with a `pub use` inside the mod.rs

I expanded on the section that says to include `mod move_balls` and `mod bounce` and included the use statements of the two systems.  I expanded it to be a code snippet in an effort to make it readable, now that it includes adding several things.  

Removing two lines that stated to add the mod statements.

Added a similar line but including a code snippet of the new mod.rs file.

## Motivation and Context

Fixing small documentation inconsistencies.  If these lines I added to the book are not 
added to the pong project, the code won't compile.

## How Has This Been Tested?

I have tested the code that the book now recommends, it runs the pong game as it should to that point.
(cargo run)

## Checklist:

- [x ] My code follows the code style of this project.

### DIFF
--- a/book/src/pong-tutorial/pong-tutorial-04.md
+++ b/book/src/pong-tutorial/pong-tutorial-04.md
@@ -266,8 +266,18 @@ The following image illustrates how collisions with paddles are checked.
 
 ![Collision explanotary drawing](../images/pong_tutorial/pong_paddle_collision.png)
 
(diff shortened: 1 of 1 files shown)


AlveLarsson (approved):
Thanks!

PR #2684, opened 2021-05-26
Fix a path in the tiles example

mxgrey:
Signed-off-by: Michael X. Grey <grey@openrobotics.org>

## Description

This simply fixes a hardcoded file path in the `tiles` example. It seems the path is a hold over from when examples were compiled and run as part of the overall `amethyst` crate whereas now they seem to be treated independently.

## Motivation and Context

Without this change, the path is wrong and the program panics.

## How Has This Been Tested?

`cargo run` inside of the `examples/tiles` directory.

## Checklist:

- [ ] ~~If my change required a change to the documentation I have updated the documentation accordingly.~~
- [ ] ~~I have updated the content of the book if this PR would make the book outdated.~~
- [ ] ~~I have added tests to cover my changes.~~
- [ ] ~~My code is used in an example.~~

### DIFF
--- a/examples/tiles/main.rs
+++ b/examples/tiles/main.rs
@@ -156,7 +156,7 @@ fn main() -> amethyst::Result<()> {
     dispatcher.add_bundle(LoaderBundle);
     dispatcher.add_bundle(TransformBundle);
     dispatcher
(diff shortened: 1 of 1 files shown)


jojolepro (approved):
Thanks and sorry for the delay! :hearts:

PR #2699, opened 2021-07-18
Port SamplerInterpolationSystem to legion macros (0.4) syntax

luckystreak63:
## Description
Modifications:
- SamplerInterpolationSystem now uses legion macros (0.4) syntax.
  - The `#[state]` is used to keep what looked like a previous optimization of re-using the same `Vec`s to avoid re-allocating every frame. Alternative ways to conserve this optimization without using `#[state]` could be better.
- Fixed compile errors in `animation` example.

## Motivation and Context
- Legion macros are a more concise way of writing systems
- Using legion macros internally sets an example of how they can be used
- This change could be used as an example of how to port systems to legion macros.

## How Has This Been Tested?
Compiled and ran `animation` example.

## Checklist:
- [x] I have added tests to cover my changes. (Fixed + ran animation example)
- [x] My code is used in an example. (animation)

### DIFF
--- a/amethyst_animation/Cargo.toml
+++ b/amethyst_animation/Cargo.toml
@@ -32,6 +32,7 @@ alga = "0.9.3"
 type-uuid = "0.1.2"
 uuid = "0.8.2"
 legion-prefab = { version = "0.1", git = "https://github.com/amethyst/prefab", rev = "8565883493aaf6044e210ffe2e72eec519d46369" }
+legion = "0.4"
 
 [dev-dependencies]
 amethyst = { path = "../", version = "0.16.0", features = ["renderer"] }
--- a/amethyst_animation/src/bundle.rs
+++ b/amethyst_animation/src/bundle.rs
@@ -4,7 +4,10 @@ use amethyst_core::ecs::{DispatcherBuilder, Resources, SystemBundle, World};
 use derivative::Derivative;
 use marker::PhantomData;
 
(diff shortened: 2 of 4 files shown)


AlveLarsson:
Thanks for the PR.

Okay, so using the macros is definitively the way we want to do things, but atm legion code-gen can't re-export due to the macros expecting to import Legion from the super crate itself, I've been looking into fixing this as a part of a bigger usability improvement PR

but this can certainly be the temporary solution as it would be quite trivial to change so the examples use the re-exported macros in the amethyst::ecs namespace.

luckystreak63:
Do you think we should wait until you complete your usability improvement PR before merging this?

I think the only debt this causes is the addition of `legion = "0.4"` to amethyst_animation's Cargo.toml, but if we merge it now we might forget to remove it later?

AlveLarsson:
No i think this should be fine, it's something I would ensure to go over in that PR, I can't say any time frame for when I'm able to get this PR done, that's why I'm ok with this solution :)

luckystreak63:
Does the CI only run if someone else approves it?

AlveLarsson:
> First-time contributors need a maintainer to approve running workflows.
Yes :)

AlveLarsson:
but the CI is somewhat broken, or rather we have some issues with clippy

AlveLarsson (approved):
Thanks!

PR #2704, opened 2021-08-01
Port Amethyst's audio to rodio v0.14.0 and refactor its API

kirusfg:
## Description

In version 0.12.0, breaking changes were made in rodio's API. The major one is the introduction of Streams, which replaced Devices as the main way to play audio.

More precisely, there is an OutputStream, which can be constructed from a default device or a Device that the user provides. This struct is !Sync and !Send, i.e. it cannot be send across threads. There is, however, an OutputStreamHandle which provides thread-safe access to an audio device. Sinks are now created from those handles instead of Devices.

Here are the adjustments made to this crate:
- Output now contains an OutputStreamHandle and the name of the device it was created with. It also provides a method to construct a Sink from the handle.
- AudioSink has been renamed to Sink.
- AudioSystem and DjSystem have been reworked according to the above changes.
- The bundles for the systems now insert at least two resources: the OutputStream resource, which must not be accessed by the systems, and the Output that is used by those systems.
- All audio functions now return results. That should make identifying the problems with audio easier.

## Motivation and Context

These changes are required to support [Android's AAudio API, which is not completely thread-safe][aaudio].

[aaudio]: https://developer.android.com/ndk/guides/audio/aaudio/aaudio#thread-safety

## How Has This Been Tested?

This has been tested using the module's unit tests and pong_tutorial_06.

## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ homepage = "https://amethyst.rs/"
 repository = "https://github.com/amethyst/amethyst"
 
 readme = "README.md"
-license = "MIT/Apache-2.0"
+license = "MIT OR Apache-2.0"
 
 autoexamples = false # Our examples come with their own Cargo.toml and are in the [workspace] section
 
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ in
       xlibs.libX11
     ];
 
-    APPEND_LIBRARY_PATH = stdenv.lib.makeLibraryPath [
+    APPEND_LIBRARY_PATH = lib.makeLibraryPath [
       vulkan-loader
       xlibs.libXcursor
       xlibs.libXi
--- a/amethyst_animation/Cargo.toml
+++ b/amethyst_animation/Cargo.toml
@@ -32,6 +32,7 @@ alga = "0.9.3"
 type-uuid = "0.1.2"
 uuid = "0.8.2"
 legion-prefab = { version = "0.1", git = "https://github.com/amethyst/prefab", rev = "8565883493aaf6044e210ffe2e72eec519d46369" }
+legion = "0.4"
 
 [dev-dependencies]
 amethyst = { path = "../", version = "0.16.0", features = ["renderer"] }
--- a/amethyst_animation/Cargo.toml
+++ b/amethyst_animation/Cargo.toml
@@ -14,7 +14,7 @@ homepage = "https://amethyst.rs/"
 repository = "https://github.com/amethyst/amethyst"
 
 readme = "README.md"
-license = "MIT/Apache-2.0"
+license = "MIT OR Apache-2.0"
 
 [dependencies]
 amethyst_assets = { path = "../amethyst_assets", version = "0.16.0" }
--- a/amethyst_animation/src/bundle.rs
+++ b/amethyst_animation/src/bundle.rs
@@ -4,7 +4,10 @@ use amethyst_core::ecs::{DispatcherBuilder, Resources, SystemBundle, World};
 use derivative::Derivative;
 use marker::PhantomData;
 
-use crate::{bundle, resources::AnimationSampling, skinning::VertexSkinningSystem};
+use crate::{
+    bundle, resources::AnimationSampling, skinning::VertexSkinningSystem,
+    systems::sampling::sampler_interpolation_system,
+};
 
 /// Bundle for vertex skinning
 ///
@@ -56,7 +59,7 @@ where
         _resources: &mut Resources,
         builder: &mut DispatcherBuilder,
     ) -> amethyst_core::Result<()> {
-        builder.add_system(crate::systems::sampling::SamplerInterpolationSystem::<T>::default());
+        builder.add_system(|| sampler_interpolation_system::<T>(Vec::new(), Vec::new()));
 
         Ok(())
     }
--- a/amethyst_animation/src/systems/sampling.rs
+++ b/amethyst_animation/src/systems/sampling.rs
@@ -65,7 +65,7 @@ pub(crate) fn sampler_interpolation<T: AnimationSampling + std::fmt::Debug>(
                 }
 
                 Some(BlendMethod::Linear) => {
-                    if let Some(p) = linear_blend::<T>(channel, &inner) {
+                    if let Some(p) = linear_blend::<T>(channel, inner) {
                         comp.apply_sample(channel, &p, commands);
                     }
                 }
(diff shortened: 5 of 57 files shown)


kirusfg:
Please suggest improvements and point out problems. This will be ready for merge when I write an example dedicated to audio and update the book. Before that, I need to figure out a bug with distill that does not allow me load big audio files. Help would be appreciated!!!

kirusfg:
Here is a summary of what I did in an attempt to solve the problem mentioned above.

When running `pong_tutorial_06` both on the main branch and my branch, it panics here:

```
thread 'main' panicked at 'refcount underflow for indirect load 9223372036854775821:IndirectLoad { id: PathWithType("audio/Computer_Music_All-Stars_-_Wheres_My_Jetpack.ogg", AssetTypeId([91, 166, 57, 7, 56, 131, 69, 62, 165, 89, 155, 119, 130, 136, 245, 210])), state: RequestingMetadata, resolved_uuid: None, refs: 18446744073709551614, pending_reresolve: false }', /home/kirusfg/Work/projects/distill/loader/src/loader.rs:308:17
```

Here is the part of stack backtrace that interests us:

```
14:     0x55e726b35dfb - std::panicking::begin_panic_fmt::h5479575e112541f8
                               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/panicking.rs:457:5
15:     0x55e7285b9a43 - distill_loader::loader::LoaderState::remove_refs::h30d18a56f32ad684
                               at /home/kirusfg/Work/projects/distill/loader/src/loader.rs:308:17
16:     0x55e7285c27dc - distill_loader::loader::Loader::remove_ref::hdcc302b876383932
                               at /home/kirusfg/Work/projects/distill/loader/src/loader.rs:1143:9
17:     0x55e72829f06d - <amethyst_assets::loader::DefaultLoader as amethyst_assets::loader::Loader>::process::h35737f1bb5455bbd
                               at /home/kirusfg/Work/projects/amethyst/amethyst_assets/src/loader.rs:253:21
18:     0x55e7282089bb - amethyst_assets::bundle::asset_loading_tick::hb28f5357ed238b61
                               at /home/kirusfg/Work/projects/amethyst/amethyst_assets/src/bundle.rs:13:5
19:     0x55e72821dba0 - core::ops::function::FnMut::call_mut::hd98eb8a6c37af957
                               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/ops/function.rs:150:5
20:     0x55e728cab1bb - <alloc::boxed::Box<F,A> as core::ops::function::FnMut<Args>>::call_mut::h62f5b8ea42001a5d
                               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/alloc/src/boxed.rs:1582:9
21:     0x55e728cdc551 - legion::internals::systems::schedule::Schedule::execute_internal::ha03b3131702dc4e6
                               at /home/kirusfg/.cargo/registry/src/github.com-1ecc6299db9ec823/legion-0.4.0/src/internals/systems/schedule.rs:599:50
22:     0x55e728cdc2b7 - legion::internals::systems::schedule::Schedule::execute::h9b310cc34a306c89
                               at /home/kirusfg/.cargo/registry/src/github.com-1ecc6299db9ec823/legion-0.4.0/src/internals/systems/schedule.rs:552:9
23:     0x55e728b0f42c - amethyst_core::dispatcher::Dispatcher::execute::h6af083409e20fe2f
                               at /home/kirusfg/Work/projects/amethyst/amethyst_core/src/dispatcher.rs:211:9
24:     0x55e726f95a34 - amethyst::game_data::GameData::update::hbe92c46014c4bde7
                               at /home/kirusfg/Work/projects/amethyst/src/game_data.rs:39:13
25:     0x55e726b538a9 - <T as amethyst::state::State<amethyst::game_data::GameData,amethyst::state_event::StateEvent>>::update::h4ac441ac831a35d2
                               at /home/kirusfg/Work/projects/amethyst/src/state.rs:345:9
26:     0x55e726d18f97 - amethyst::state::StateMachine<T,E>::update::h4ed905f014cf8ff3
                               at /home/kirusfg/Work/projects/amethyst/src/state.rs:494:21
27:     0x55e726eafb06 - amethyst::app::CoreApplication<T,E,R>::advance_frame::hdbaf9fb62015c7dd
```

And here is the code that panics:

```
assert!(
    state.refs.fetch_sub(num_refs, Ordering::Relaxed) < usize::MAX - num_refs,
    "refcount underflow for indirect load {:?}:{:?}",
    load.0,
    *state
);
```

The number of references for an asset is of type `AtomicUsize`, which wraps around on overflow. The assertion above is needed to check whether the number of refs has reached 0, in which case we should not be subtracting from it, and `distill` must unload the asset instead. However, that does not happen for some reason.

This code is from the `remove_refs` function, which decreases the number of references to an asset. Its wrapper, `remove_ref`, which only subtracts 1 from the number of references, is called by distill's `loader` when it receives a `RefOp::Decrease`. When does that happen? Whenever the asset is dropped (technically, a handle to an asset's loading state):

```
impl Drop for HandleRef {
    fn drop(&mut self) {
        use HandleRefType::*;
        self.ref_type = match std::mem::replace(&mut self.ref_type, None) {
            Strong(sender) => {
                let _ = sender.send(RefOp::Decrease(self.id));
                Weak(sender)
            }
            r => r,
        };
    }
}
```

I don't think that the bug originates from `distill` or `amethyst_assets` since the `AudioSystem` works just fine (loads and plays assets without errors) while `DjSystem` does not. My guess is that the `DjSystem` or its builder are holding on to the resources incorrectly. However, there is nothing indicating that in the stack backtrace.

My understanding of `legion` is very limited, but my intuition tells me that the `System` implementation for `DjSystem` is incorrect. Any thoughts?

luckystreak63:
See https://github.com/amethyst/amethyst/pull/2706 for a fix to the `refcount underflow for indirect load` error

AlveLarsson:
This is fantastic! @kirusfg ❤️
I'm going to enable this PR to run the workflows

Cargo.toml
```diff
@@ -16,7 +16,7 @@ homepage = "https://amethyst.rs/"
 repository = "https://github.com/amethyst/amethyst"
 
 readme = "README.md"
-license = "MIT/Apache-2.0"
+license = "MIT OR Apache-2.0"
```
AlveLarsson, line 19: @kirusfg
This has to be rejected, we can't have 3rd party changes to the license
Even tho it's just formating.

kirusfg: This change was committed on Aug 15 and not by me.

Also, I barely use git and GitHub, so I need to ask whether I did this right. Whenever the main branch is updated, how does one get all those updates merged into the branch they are working on? I did a rebase off of the main branch, it added all those commits (one of which introduced that change to the license you mentioned) and duplicated my work onto the head of the branch. Is this the correct way to do it?

AlveLarsson: Oh, see if you can squash / update it

Github itself should have buttons for it if you go into your branch

PR #2706, opened 2021-08-08
Fix refcount underflow error in tutorial pong06

luckystreak63:
Indirect `LoadHandle`s are not always immediately available from the
indirection table. A cloned indirect handle would thus have their
refcount decreased on Drop, but not increased on Clone. Remove the need
for indirection table resolution to fix this bug.

## Motivation and Context
The tutorial Pong06 is crashing due to a negative refcount.
Using the DJSystem in any other project in a similar manner causes the same crash.

## How Has This Been Tested?
Tested by running the following examples:
- pong01
- pong06
- sphere
Also spam-cloned / dropped indirect handles while logging the refcount.

## Checklist:

### DIFF
--- a/amethyst_assets/src/loader.rs
+++ b/amethyst_assets/src/loader.rs
@@ -253,9 +253,7 @@ impl Loader for DefaultLoader {
                     self.loader.remove_ref(handle);
                 }
                 Ok(RefOp::Increase(handle)) => {
-                    self.loader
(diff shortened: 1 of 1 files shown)


jojolepro (approved):
looks good to me!

AlveLarsson (approved):
Thanks!

Issue #2703, opened 2021-07-29
RenderToWindow::from_config_path - Parse Errors include filename that couldn't be parsed


TimVanDyke:
## Feature
from_config_path

## Change
I would like to see the file path or file name printed to the console in parsing errors.

Ex: 
Current: 
`Error: Error { inner: Inner { source: None, backtrace: None, error: Parser(Parser(ExpectedMapColon, Position { col: 13, line: 3 })) } }`

Preferred: 
`Error: Error { inner: Inner { source: None, backtrace: None, error: Parser(Parser(*FILE_PATH:* ExpectedMapColon, Position { col: 13, line: 3 })) } }`

## Reason
I'm a Rust and Amethyst noob and seeing more detailed errors will help me better find my errors. I am also not the only one who has had issues with not knowing where my error is. 
 - https://github.com/amethyst/amethyst/issues/2460

kirusfg:
This would require refactoring `amethyst_error` crate's `ConfigError` and everything that uses it. Not a difficult, but rather tedious task. I would label it as a good first issue, @AlveLarsson.

PR #2708, opened 2021-08-14
#2703: Include a path of the file in ron parse errors

podusowski:
## Description
Added additional type to `ConfigError` which holds both ron error and a path to the file being parsed. Alternative was to add a path to the existing `Parser` variant, but it would require to fill it for cases that are not really appropriate like reading directly from bytes in memory.

## Motivation and Context
https://github.com/amethyst/amethyst/issues/2703

## How Has This Been Tested?
Written an unit test for it.

## Checklist:


AlveLarsson (approved):
Thanks!

PR #2711, opened 2021-08-26
Fix deserialize bug for material example

luckystreak63:
## Description
Material example panicked and threw this error on start:
```
Error: Error { inner: Inner { source: None, backtrace: None, error: FileParser(Error { code: TrailingCharacters, position: Position { line: 6, col: 1 } }, ".../amethyst/examples/material/config/display.ron") }
```
This is due to a recent change. This commit reverts the code back to a working state.

## Motivation and Context

Examples that crash is confusing for newcomers.

## How Has This Been Tested?
Running the `material` example now works.
`cargo run -p material`

## Checklist:


AlveLarsson (approved):
Thanks!

PR #2712, opened 2021-08-29
fix: remove documented tab-order reference

RossMurr4y:
## Description

So far as I can tell as I follow along, the referenced configuration to `tab-order` is not actually set.

> Next, we create a transform for each of our two scores by giving them a unique id (P1 and P2), a UI Anchor at the top middle of our window, and then adjust their global x, y, and z coordinates, width, height, and tab-order.

```rust
    let p1_transform = UiTransform::new(
        "P1".to_string(), Anchor::TopMiddle, Anchor::TopMiddle,
        -50.0, -50.0, 1.0, 200.0, 50.0,
    );
```

If it is set, its not immediately clear where.

## Motivation and Context

Align documentation with changes made.

## How Has This Been Tested?

Locally built with mdbook

## Checklist:

- n/a - I have added tests to cover my changes.
- n/a - My code is used in an example.

### DIFF
--- a/book/src/pong-tutorial/pong-tutorial-05.md
+++ b/book/src/pong-tutorial/pong-tutorial-05.md
@@ -314,8 +314,7 @@ our `UiText` components).
 
 Next, we create a transform for each of our two scores by giving them a unique
 id (`P1` and `P2`), a UI `Anchor` at the top middle of our window, and then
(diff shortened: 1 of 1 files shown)


AlveLarsson (approved):
Thanks!

PR #2714, opened 2021-09-17
add resources to Tile.sprite and Tile.tint params

SET001:
## Description
This PR adds `Resources` in methods `sprite` and `tint` of `Tile` trait

## Motivation and Context
It may be useful to generate/modify map in separate resource and then just read it from there inside `Tile.sprite`. Specs `world` had `read_resource` method. But Legion stores resources separately from world so we should pass them in here.

## Checklist:


luckystreak63 (approved):
Look good to me!

AlveLarsson (approved):
Thanks! 👍

PR #2720, opened 2021-10-03
fix: clippy error for amethyst_error crate

grzi:
Fix the clippy warning in amethyst_error doc
Fixed a clippy recommandation in amethyst_derive
Fixed a derivable impl in amethyst_network


AlveLarsson (approved):
Thanks!

